Issue #159, opened 2015-01-01
Affix sidebar on module page breaks when navigating from a module page to another module page


yangshun:
Steps to reproduce:
1. Go to a module page, [/modules/ACC2002](/modules/ACC2002) for example.
2. Scroll to the bottom of the page.
3. Click on the search bar and navigate to another page.

The sidebar loses its affix functionality.

yangshun:
@xbili if you are interested you can try tackling this too :smile:

xbili:
I'll look into this!

yangshun:
Fixed via #160

PR #160, opened 2015-01-02
CSS fix for affix sidebar positioning

xbili:
For issue [#159]

The problem was caused by jQuery's ```.offset()``` function. More information can be found [here].

```.affix-bottom``` has to be added to ensure that jQuery doesn't set ```position: relative;``` inline every time we navigate to a new module.

[#159]: https://github.com/nusmodifications/nusmods/issues/159
[here]: https://github.com/twbs/bootstrap/issues/9342

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -715,6 +715,9 @@ colgroup.hover {
     &.affix {
       top: 85px;
     }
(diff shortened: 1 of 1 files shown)


yangshun:
Awesome! Affix is really tricky to use..

Issue #156, opened 2014-12-30
Optimize moduleInformation.json


yangshun:
`moduleInformation.json` is HUGE at ~4.2mb and the page may appear to have hung if it takes a long time to load. 

A temporary fix will be to add a loader/spinner or some message to notify of load times. Related to #95 

Reported by @yjwong

yangshun:
After gzipping by @xinan it has been reduced to 1.2mb. There's no real optimization we can do if we are to stick with client side filtering.

Issue #95, opened 2014-07-30
Add loading indicators


ahbeng:
Currently there's no indication, only waiting till data loads before showing a view.

yangshun:
Closed via #172

PR #172, opened 2015-01-11
Added loading indicator to modules page

xbili:
For [#95]. A temporary fix for [#156] as well.

I don't think this is the best way to add in a spinner, apparently people have been pushing for Marionette's CollectionView to come with a LoadingView [here] as well.

Any better way to do this? 

[#95]: https://github.com/nusmodifications/nusmods/issues/95
[#156]: https://github.com/nusmodifications/nusmods/issues/156
[here]: https://github.com/marionettejs/backbone.marionette/issues/510


app/scripts/common/templates/loading.hbs
```diff
@@ -0,0 +1 @@
+<div class="loading"></div>
```
yangshun: Please add newline at the end of file

yangshun: It'd be good if we can namespace all our own classes by prefixing with `nm-`. Sorry I should come up with a style guide soon >_<

yangshun: Could you make it:
```
<div class="js-nm-loading nm-loading"></div>
```
instead? and use the `js-nm-loading` class in `LoadingView.js` and `nm-loading` in the CSS file.

xbili:
I see, I'll namespace it first then!

app/scripts/common/templates/loading.hbs
xbili: I see, I'll namespace it first then!

yangshun: This is a good read recommended for front end devs: http://philipwalton.com/articles/css-architecture/

xbili: Great article! Perfect for future reference as well.

So in this case we're using js-nm-loading for js manipulations and the other for styling only? That's really nice and neat :D

Made the changes!

yangshun: That's right, and `js-` prefixed selectors should not appear in the CSS file at all (although there are some existing there, which I should clean up). Awesome, lemme test it and I'll merge (:

yangshun:
@xbili it works well! I'd want to modify its appearance though. I'll merge it and take over from here. Thanks! (:

xbili:
Sure no problem! :)

Issue #103, opened 2014-07-31
Module code hover brief info


AHaliq:
In the module details page, in certain texts there are module code hyperlinks such as pre-requisite modules. Itd be great if hovering over these texts will display the module name.

yangshun:
This behaviour can also be supported on module finder page. It's okay not to have that functionality for touch screen devices.

PR #173, opened 2015-01-13
Added qtips for hyperlinks on Module pages

xbili:
For issue [#103].

I'm not very sure if it's bad practice to instantiate a new Model inside a View. There might be a better way to do this.

Also, I included in onBeforeDestroy to deal with qtips not destroying themselves after view changes. 

[#103]: https://github.com/nusmodifications/nusmods/issues/103

### DIFF
--- a/app/scripts/common/behaviors/ModuleHoverBehavior.js
+++ b/app/scripts/common/behaviors/ModuleHoverBehavior.js
@@ -6,7 +6,7 @@ var Marionette = require('backbone.marionette');
 
 module.exports = Marionette.Behavior.extend({
   onShow: function () {
-    // Module code hover brief info
+    // Module hover brief info
     $("a[href^='/modules/'").qtip({
       content: {
         text: function(event, api) {
@@ -38,6 +38,14 @@ module.exports = Marionette.Behavior.extend({
       show: {
         effect: function() {
           $(this).fadeTo(200, 0.85);
+        }
+      },
+      events: {
+        show: function(event, api) {
+          // Prevents qtip from appearing when hovering over main module link
+          if ($(event.originalEvent.currentTarget).parent().is('h2')) {
+            event.preventDefault();
+          }
         }
       }
     });
--- a/app/scripts/common/behaviors/ModuleHoverBehavior.js
+++ b/app/scripts/common/behaviors/ModuleHoverBehavior.js
@@ -51,3 +51,4 @@ module.exports = Marionette.Behavior.extend({
     });
   }
 });
+
--- a/app/scripts/common/behaviors/ModuleHoverBehavior.js
+++ b/app/scripts/common/behaviors/ModuleHoverBehavior.js
@@ -6,7 +6,8 @@ var Marionette = require('backbone.marionette');
 
 module.exports = Marionette.Behavior.extend({
   events: {
-    "mouseenter a[href^='/modules/']" : 'showDetails'
+    "mouseenter a[href^='/modules/']" : 'showDetails',
+    "click a[href^='/modules/']": 'destroyDetails'
   },
 
   showDetails: function(event) {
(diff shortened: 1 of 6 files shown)


yangshun:
@ahbeng could you review this? I'm not sure about @xbili 's concerns either. One thing to note that the modules finder page (`ModulesView.js`) can and should also have this qtip behaviour (and perhaps even more pages in future). 

I think it'll be suitable to use the [Event Delegation Pattern](http://davidwalsh.name/event-delegate) in this case. Perhaps we can add a class like `.js-nm-module-hover-info` and the attribute `data-nm-mod-code=<module code>` to any links that we want to have this behaviour. The listener will look like: 
```
$('body').on('hover', '.js-nm-module-hover-info', function () {
    // Retrieve hovered DOM element
    // Get module code from data attribute 
    // Bind the qtip and then show it
});
```

Benefits of this is that there isn't a need to add listeners when creating the view.

As for `modulify.js`, to keep things flexible, it can take in an additional flag parameter, to indicate whether to add in `.js-nm-module-hover-info` and  `data-nm-mod-code=<module code>` into the `<a>`, so as to provide the option of not showing the module info on hover when we use `linkifyModules`.

ahbeng:
@xbili It's a bit confusing, but a View in Backbone acts somewhat like a controller, and a Controller in Marionette isn't a Controller in the traditional MVC sense but is meant to be used with routers. So that usage is fine.

@yangshun You're right, that pattern could be used, but doing it on `body` as in that code fragment is rather globalish for my liking. Using a [Marionette.Behavior](http://marionettejs.com/docs/v2.3.0/marionette.behavior.html) would be using event delegation but in a cleaner way, e.g. like in [GoToTopBehavior](https://github.com/nusmodifications/nusmods/blob/master/app/scripts/common/behaviors/GoToTopBehavior.js).

ahbeng:
Since this behavior is probably something you want almost every time you link to a module, maybe an attribute starts with selector could be used instead of adding a new class? Something like `a[href^="/modules/"]`, but with additional restrictions to not select the main module links.

xbili:
Right, I'll look into this again!

xbili:
Yep I've shifted them into a Behavior instead. Thanks for the help! :)

The current code works now, but I don't think adding all of it in ```onShow``` would be good. I'll clean it up a bit more first.

xbili:
It's done! Do let me know if there are any issues with this :)

yangshun:
Hey @xbili i will look at it over the weekend, cheers! (:

app/scripts/common/behaviors/ModuleHoverBehavior.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+      position: {
+        effect: 'false',
+        my: 'bottom center',
+        at: 'top center'
+      },
+      events: {
+        show: function(event, api) {
+          // Prevents qtip from loading if main module link is selected
+          if ($(curr).parent().is('h2')) {
```
yangshun: This may change in the future. We might change to other HTML elements instead of `h2` and this will break. A better way would be identify such cases would be to add add an attribute `data-no-module-qtip` to any links that you don't want this behaviour. In future there might be other cases (not only these main module links) where we don't want the hover behaviour.

yangshun:
The module qtip does not disappear after clicking. Steps to reproduce:
1. Go to module finder page: `/modules`.
2. Click on **ACC1002** under the preclusions section of the module **ACC1002**.
3. The qtip stays there even though the page has transitioned to the ACC1002's module page. Refer to attached screenshot:

screen shot 2015-01-19 at 7 21 40 am

xbili:
Oops, sorry about that. I've added the changes!

yangshun:
Thanks @xbili . You rock :+1:

xbili:
Hahaha thanks, and welcome! :)

PR #176, opened 2015-01-22
Renamed size options for image export

xinan:
* Changed the size options to fa icons. 
* Added link titles.
* The normal-size icon is a bit special because it is alongside the description. So I used a bit of magic number (left: 12px;) to adjust the distance between it and the next icon. Tell me if there is any better ways.

### DIFF
--- a/app/scripts/timetable/templates/export.hbs
+++ b/app/scripts/timetable/templates/export.hbs
@@ -6,35 +6,35 @@
   </button>
   <ul class="dropdown-menu">
     <li>
-      <a download="My NUSMods.com Timetable.xls" href="#" id="xls-file">
+      <a download="My NUSMods.com Timetable.xls" href="#" data-export-type="xls-file">
         <i class="fa fa-file-excel-o fa-fw"></i>&nbsp;Excel File
       </a>
     </li>
     <li>
-      <a download="My NUSMods.com Timetable.html" href="#" id="html-file">
+      <a download="My NUSMods.com Timetable.html" href="#" data-export-type="html-file">
         <i class="fa fa-file-code-o fa-fw"></i>&nbsp;HTML File
       </a>
     </li>
     <li{{#if isSpecialTerm}} class="disabled"{{/if}}>
-      <a download="My NUSMods.com Timetable.ics" href="#" id="ical-file">
+      <a download="My NUSMods.com Timetable.ics" href="#" data-export-type="ical-file">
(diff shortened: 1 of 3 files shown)


app/scripts/timetable/templates/export.hbs
```diff
@@ -22,11 +22,16 @@
       </a>
     </li>
     <li class="horiz-menu">
-      <a href="#" class="col-xs-6" id="jpg-file" data-imagesize="1">
-        <i class="fa fa-file-image-o fa-fw"></i>&nbsp;Image (.jpg)
+      <a href="#" class="col-xs-8" id="jpg-file" data-imagesize="1" title="Normal">
```
yangshun: `id`s should be unique!

app/styles/_timetable.scss
```diff
@@ -52,6 +52,17 @@
     padding-right: 0;
     text-align: center;
   }
+  .image-normal {
+    position: relative;
```
yangshun: Sort CSS properties alphabetically please

app/styles/_timetable.scss
```diff
@@ -52,6 +52,17 @@
...
+    font-size: 0.8em;
+    left: 12px;
+  }
```
yangshun: Sort CSS properties alphabetically please

app/scripts/timetable/templates/export.hbs
xinan: Using unique `id` requires duplication of js code o.O Because it uses a switch statement.

yangshun: Ok, then change all to class, and use data attribute to differentiate between the `jpg-file` buttons being clicked

xinan: I think changing all `id`s to `data-attribute`s will be better because some elements might have multiple classes which will be difficult to `switch`. Then for the image links, use another `data-attribute` just like what I have done.

yangshun: Sounds reasonable. Go ahead

app/styles/_timetable.scss
```diff
@@ -52,6 +52,17 @@
     padding-right: 0;
     text-align: center;
   }
+  .image-normal {
```
yangshun: Prefix all these classes with `icon-` please. These sound like the class names for ``.

xinan: Sure :smile_cat:

PR #177, opened 2015-02-02
Improve language used.

tgxworld:
`Correct as of` seems to be the correct usage here.


tgxworld:
Discussed with @yangshun. `Correct as at` is appropriate too :smile_cat:

PR #180, opened 2015-02-03
Show contributor's commits instead of GitHub profile.

tgxworld:
@yangshun Not sure if you'll prefer this but I think it makes more sense to showcase all the contributions made to NUSMods instead of just the contributor's GitHub profile.

### DIFF
--- a/app/scripts/contribute/templates/developers_item.hbs
+++ b/app/scripts/contribute/templates/developers_item.hbs
@@ -1,13 +1,16 @@
-<a href="{{html_url}}" target="_blank">
-  <div class="nm-ct-dev-head">
-    <img class="img-rounded img-responsive" src="{{avatar_url}}" alt="{{login}}">
-    <div class="nm-ct-dev-stat img-rounded">
-      <p>
+<div class="nm-ct-dev-head">
+  <img class="img-rounded img-responsive" src="{{avatar_url}}" alt="{{login}}">
+  <div class="nm-ct-dev-stat img-rounded">
+    <p>
+      <a href="{{html_url}}" target="_blank" class="nm-ct-dev-profile">
+        <i class="fa fa-github fa-lg"></i>
+      </a>
+      <a href="https://github.com/nusmodifications/nusmods/commits?author={{login}}" target="_blank">
         <span class="nm-text-xlarge">{{contributions}}</span>
         <span class="nm-text-medium hidden-xs hidden-sm"><br></span>
         commit{{#notequals contributions 1}}s{{/notequals}}
-      </p>
-    </div>
+      </a>
+    </p>
   </div>
-  <h4 class="nm-ct-dev-id"><i class="fa fa-github"></i> {{login}}</h4>
-</a>
+</div>
+<h4 class="nm-ct-dev-id">{{login}}</h4>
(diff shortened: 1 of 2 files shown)


app/scripts/contribute/templates/developers_item.hbs
```diff
@@ -1,4 +1,4 @@
-<a href="{{html_url}}" target="_blank">
+<a href="https://github.com/nusmodifications/nusmods/commits?author={{login}}" target="_blank">
```
tgxworld: @yangshun Haha is there a way I can pass the URL as a template variable/argument (what is the proper term?)?

yangshun: I actually considered this before, but in the end, I decided that the purpose of this page will be to showcase our developers' portfolio, rather than their contributions towards NUSMods. Looking at someone's Github profile is more meaningful than rows of commits by them (who is going to read them anyways?).

yangshun: I think your way of adding that url there is fine, but as mentioned earlier, I intentionally put a link to their Github profile rather than the list of commits they have made.

tgxworld: >  Looking at someone's Github profile is more meaningful than rows of commits by them (who is going to read them anyways?).

Haha I was more interested in what commits people have made :smiley_cat: But I think I'll looking at it from a developer's perspective.

ahbeng: Why not both? How about linkifying "commits" to the actual commits and the rest of the box still leads to the profile?

tgxworld: I would go for two separate links "commits" and "profile". Having the rest of the box linking to profile isn't immediately obvious to someone. Thoughts?

yangshun: I agree with @ahbeng 's suggestion. How about something like this?

screen shot 2015-02-07 at 7 41 30 pm

The Github icon leads to the developer's profile while the commit number leads to the commit page made by the developer.

tgxworld: :+1:

app/styles/_contribute.scss
```diff
@@ -42,8 +49,21 @@
...
+          color: black;
+        }
+      }
+      @include dev-github-profile(5em);
+      @media (max-width: 1250px) {
+        @include dev-github-profile(3em);
       }
       @media (max-width: 1024px) {
+        .nm-ct-dev-profile {
+          display: none;
```
tgxworld: @yangshun I decided to hide on the profile icon on smaller devices in order to fit the current design.

tgxworld:
@yangshun Could you take a look at this? (especially the CSS portion)

app/styles/_contribute.scss
```diff
@@ -114,4 +134,4 @@
       }
     }
   }
-}
```
yangshun: thanks. i love newlines at the end of files. lol

tgxworld: haha just have to enabled it on Sublime.

xinan: There is even a option to strip trailing whitespaces when saving automatically. Haha!

yangshun: wah cool. i have enabled both (:

Issue #129, opened 2014-08-08
Timetable tooltip remains after navigating via keyboard shortcut


ahbeng:
Reported by @yangshun.

yangshun:
Fixed vis #157

PR #184, opened 2015-02-07
Removes qtips on module page before view changes.

xbili:
Similar bug to issue [#129]. 

To reproduce the bug:
1. Head to modules page
2. Hover over one of the module pre-requisites
3. Hit '?' to access the help page
4. qtip does not disappear

Did not take this into consideration when implementing the qtips on the module page, apologies! 

Is there a better way to implement this though, instead of including it into ```onBeforeDestroy``` for every ```View``` that we have?

### DIFF
--- a/app/scripts/modules/views/ModuleItemView.js
+++ b/app/scripts/modules/views/ModuleItemView.js
@@ -66,5 +66,9 @@ module.exports = Marionette.ItemView.extend({
 
   onShow: function() {
     this.$('.js-nm-mi-filter').qtip();
+  },
+
+  onBeforeDestroy: function() {
+    $("a[href^='/modules/']").qtip('destroy', true);
   }
 });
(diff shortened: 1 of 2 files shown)


app/scripts/modules/views/ModuleItemView.js
```diff
@@ -66,5 +66,9 @@ module.exports = Marionette.ItemView.extend({
 
   onShow: function() {
     this.$('.js-nm-mi-filter').qtip();
+  },
+
+  onBeforeDestroy: function() {
+    $("a[href^='/modules/']").qtip('destroy', true);
```
yangshun: single quotes for strings please. use double quotes for the attribute

yangshun:
> Is there a better way to implement this though, instead of including it into onBeforeDestroy for every View that we have?

I'm not exactly sure. Maybe can destroy when there's a navigation event? @ahbeng any suggestions?

ahbeng:
I think this is fine, it's just a line which isn't in many places and don't think it's worth it to abstract out further since it has to target different elements for different views anyway?

xbili:
@ahbeng I see, yes it is actually quite specific as for which element to select with jQuery. Thanks! :+1:

Issue #162, opened 2015-01-02
Display current total number of MCs in timetable


yangshun:
Have to take into account that FYP modules are 12 MCs on paper but only 6 MCs per semester. Can possibly maintain a list of FYP module codes and check against it when calculating.

PR #185, opened 2015-02-08
[WIP] Display total module credits.

tgxworld:
Closes: https://github.com/nusmodifications/nusmods/issues/162.

### DIFF
--- a/app/scripts/timetable/collections/ExamCollection.js
+++ b/app/scripts/timetable/collections/ExamCollection.js
@@ -9,6 +9,7 @@ module.exports = Backbone.Collection.extend({
 
   initialize: function () {
     this.clashCount = 0;
+    this.totalModuleCredits = 0;
     this.listenTo(this, {add: this.onAdd, remove: this.onRemove});
   },
 
@@ -19,11 +20,13 @@ module.exports = Backbone.Collection.extend({
       examStr: module.get('examStr'),
       id: module.id,
       ModuleTitle: module.get('ModuleTitle'),
-      Semester: module.get('Semester')
+      Semester: module.get('Semester'),
+      moduleCredit: module.get('ModuleCredit')
     });
   },
 
   onAdd: function (exam) {
+    this.totalModuleCredits += exam.get('moduleCredit');
     // Compute clashes based on keys with clustered hours.
     var clashes = this.where({key: exam.get('key')});
     if (clashes.length > 1) {
@@ -36,6 +39,8 @@ module.exports = Backbone.Collection.extend({
   },
 
   onRemove: function(exam) {
+    this.totalModuleCredits -= exam.get('moduleCredit');
+
     if (exam.get('clash')) {
       var clashes = this.where({key: exam.get('key')});
       if (clashes.length === 1) {
--- a/app/scripts/timetable/templates/exams.hbs
+++ b/app/scripts/timetable/templates/exams.hbs
@@ -2,8 +2,17 @@
   <tr>
     <th>Code</th>
     <th>Title</th>
+    <th>Module Credits</th>
     <th class="exam-time">Exam Time</th>
     <th>Actions</th>
   </tr>
 </thead>
 <tbody></tbody>
+<tfoot>
+  <tr>
+    <td colspan="5" class="total-module-credits">
+      Total Module Credits:
+      <span id="js-nm-total-mc"></span>
+    </td>
+  </tr>
+</tfoot>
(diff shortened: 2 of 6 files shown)


tgxworld:
@yangshun :smile:

app/scripts/timetable/collections/ExamCollection.js
```diff
@@ -51,5 +58,11 @@ module.exports = Backbone.Collection.extend({
   // as intended.
   comparator: function(exam) {
     return exam.get('key');
+  },
+
+  calculateModuleCredit: function(exam) {
+    var moduleCredit = exam.get('ModuleCredit');
+    moduleCredit = exam.get('ModuleTitle').match(/dissertation/i) ? moduleCredit / 2 : moduleCredit
```
tgxworld: I'll update this soon once I find the time to go through the list of all FYPs

app/scripts/timetable/collections/ExamCollection.js
```diff
@@ -9,21 +9,26 @@ module.exports = Backbone.Collection.extend({
...
+
     this.add({
       color: module.get('color'),
       ExamDate: module.get('ExamDate'),
       examStr: module.get('examStr'),
       id: module.id,
       ModuleTitle: module.get('ModuleTitle'),
-      Semester: module.get('Semester')
+      Semester: module.get('Semester'),
+      ModuleCredit: moduleCredit
```
yangshun: we start with lowerCase for our own properties. The properties that start with UpperCase are because they are from the NUSMods API.

yangshun:
The `colspan` attribute of `EmptyView` in `ExamsView` will have to be updated because you added a new column. Currently it looks like this: 

screen shot 2015-02-08 at 7 13 09 pm

app/scripts/timetable/templates/exams.hbs
```diff
@@ -2,8 +2,16 @@
...
     <th>Code</th>
     <th>Title</th>
+    <th>Module Credits</th>
     <th class="exam-time">Exam Time</th>
     <th>Actions</th>
   </tr>
 </thead>
 <tbody></tbody>
+<tfoot>
+  <td/>
```
yangshun: You need a `<tr>` here also

app/scripts/timetable/templates/exams.hbs
```diff
@@ -2,8 +2,16 @@
...
+  <td/>
+  <td id="total-module-credit">Total: {{totalModuleCredits}}</td>
```
yangshun: Make it `colspan="5"`. Nicer to take up the full width.

yangshun: if a selector needs to be accessed via js, we prefix with js. NUSMods also namespaces our own selectors using nm. Name it this instead: `js-nm-total-mc`.

app/scripts/timetable/collections/ExamCollection.js
```diff
@@ -9,21 +9,26 @@ module.exports = Backbone.Collection.extend({
...
     });
   },
 
   onAdd: function (exam) {
+    this.totalModuleCredits += parseInt(exam.get("ModuleCredit"));
```
yangshun: Putting different logic to update `totalModuleCredits` in both `onAdd` and `onRemove` is clunky. It's cleaner to re-calculate the total MCs and just re-render the relevant row by watching the `totalModuleCredits` property. This way, you won't need a selector to access the relevant `<td>` element too. You will need to create a new `ItemView` for this partial re-render though.

It feels unnecessary to have two ways (template rendering and jquery manipulation) to update the total MC count, and it's a cause for bugs. In fact, the way you are doing now, rendering it via the template doesn't serve any purpose because when it is first rendered, the collection is empty and you'll get 0 MCs.

Simplest way I'd recommend: Remove template rendering, and write a function to recalculate the total MCs everytime the modules change, and set the value via jquery.

app/scripts/timetable/templates/exams.hbs
yangshun: Just use a `<span id="js-nm-total-mc">` around the value. The words will not change and can be left as part of the template.

app/scripts/timetable/views/ExamsView.js
```diff
@@ -18,9 +18,13 @@ module.exports = Marionette.CompositeView.extend({
   childViewContainer: 'tbody',
   emptyView: EmptyView,
   template: template,
+  templateHelpers: function() {
+    return { totalModuleCredits: this.collection.totalModuleCredits };
+  },
   collectionEvents: {
     'add remove': function() {
       $('#clash').toggleClass('hidden', !this.collection.clashCount);
+      $('#total-module-credit').text("Total: " + this.collection.totalModuleCredits);
```
yangshun: Since the words will be part of the template, `"Total: "` can be left out from here.

yangshun:
Refer to #185

Issue #183, opened 2015-02-05
Remove filter from 'Active Filters' section


yangshun:
Currently there is no convenient way to remove the active filters except clicking on the respective checkboxes from the *Search Options* section. It will be good if there is a **x** symbol in the labels of the **Active Filter** section that can be used to remove that particular filter.

![screen shot 2015-02-06 at 3 13 59 am](https://cloud.githubusercontent.com/assets/1315101/6067262/3e6ee4e0-adae-11e4-969d-1dd8093b3a0e.png)

PR #186, opened 2015-03-25
#183 Remove filter from 'Active Filters' section

chunqi:
Clicking on a filter in the meta list ‘clicks’ on the corresponding
filter’s checkbox, removing it from the current selection.

### DIFF
--- a/app/scripts/modules/views/ModulesFilterMetaView.js
+++ b/app/scripts/modules/views/ModulesFilterMetaView.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var $ = require('jquery');
 var Marionette = require('backbone.marionette');
 var _ = require('underscore');
 var template = require('../templates/modules_filter_meta.hbs');
@@ -9,13 +8,6 @@ module.exports = Marionette.CompositeView.extend({
   template: template,
   initialize: function () {
     this.parseSelectedFilters();
-  },
-  events: {
-    'click button': function(event) {
-      var deselectedFilter = event.target.getAttribute('data-filter');
-      var selector = "input[data-filter='" + deselectedFilter + "']";
-      $(selector).click();
-    }
   },
   parseSelectedFilters: function () {
     var selectedFilters = this.model.get('selectedFilters');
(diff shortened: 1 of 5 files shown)


yangshun:
Hi @chunqi , nice work! However, I think it's kinda hackish to simulate a click on the `<input>` checkbox. It would be better if you modified the model like how `FilterView` does it.

chunqi:
Hello @yangshun thanks for the suggestion. I have changed the method of deselecting a filter to be less hackish. It now grabs the label of the filter being deselected and toggles it directly from its model. As a result of the need to access the filters collection the constructor for the model of ModulesFilterMetaView is passed a copy.

screenshot
The UI has been changed from a span label to a button for the hover cursor effect.

app/scripts/modules/views/ModulesFilterMetaView.js
```diff
@@ -9,6 +10,22 @@ module.exports = Marionette.CompositeView.extend({
...
     this.parseSelectedFilters();
   },
+  events: {
+    'click button': function(event) {
+      //Get the label of the filter that the user deselected
+      var deselectedFilter = event.currentTarget.getAttribute("data-filter");
+      //Find and deselect the filter
+      _.each(this.model.get('collection').models, function (facet) {
+        _.each(facet.get('filters').selected, function (filter) {
+          if(filter.get('label') == deselectedFilter) {
```
yangshun: use `===`

yangshun:
This looks great! One last small request before I merge it. Could you add in small crosses beside the text within the label? Something like: http://staging.nusmods.com/friends.

screen shot 2015-03-28 at 12 57 51 am

xinan:
Clicking on the label perform a "remove" action. It is not very intuitive. I would sometimes click on the text or select them while I read. It is better to have a cross beside the text in the label.

chunqi:
@yangshun @xinan thanks for the input, it now looks like this

screenshot

yangshun:
Looks great!!!

chunqi:
Changed it, @yangshun thanks for the guidance!

yangshun:
Thanks! Good work! :+1:

xinan:
Deployed. Thanks! :+1:

yangshun:
Closes #183

PR #187, opened 2015-03-28
Show total mc count

### LINKED ISSUE #162: Display current total number of MCs in timetable

yangshun:
Added on to @tgxworld 's #185 . @xinan @xbili please review (:
Closes #162 

screen shot 2015-03-29 at 12 40 42 am

### DIFF
--- a/app/scripts/timetable/collections/ExamCollection.js
+++ b/app/scripts/timetable/collections/ExamCollection.js
@@ -51,7 +51,7 @@ module.exports = Backbone.Collection.extend({
 
   getTotalSemesterModuleCredits: function () {
     var totalCount = this.reduce(function (memo, module) {
-      return memo + module.getSemesterModuleCredit();
+      return memo + module.get('semesterModuleCredit');
     }, 0);
     return totalCount;
   },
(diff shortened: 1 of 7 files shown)


xinan:
Looks quite good. Shall we change the colour of the last row on dark theme? Currently it looks like:

total-mc-screenshot-dark

tgxworld:
I think the dissertation module should show `6` instead of `12` Otherwise, it'll really be confusing when people add up the column and see something different,

yangshun:
@tgxworld true, but in the transcript or MyISIS, it shows the total MCs for that module regardless of whether it spans across multiple semesters. I think we should follow that for consistency.

tgxworld:
Do we have a disclaimer that we're not liable for any mistakes in the counting of MCs? Just in case.

yangshun:
We'll come up with a Terms of Service soon. Put our IS1103FC knowledge to good use @xbili  lol

xbili:
@yangshun Yup that sounds good. I think a bit more effort can be put to make it more readable as well. We care about our users 👍

PR #194, opened 2015-04-05
Updates + Cleanup

ahbeng:
Everything has been updated to latest except:
- Select2, which remains at 3.x since 4.x is still RC and has breaking changes
- localForage, which remains at forked version which has fixes for Firefox/Safari private modes


xinan:
Looks good to me. 

I'm not so sure about whether those dependencies updates will have any side effect or not. 

Btw, we can consider integrating ImageOptim into our build process. Something like [grunt-imageoptim](https://github.com/JamieMason/grunt-imageoptim).

ahbeng:
We're already using [grunt-contrib-imagemin](https://github.com/gruntjs/grunt-contrib-imagemin) in the build process, which is cross-platform. ImageOptim does seem to do a slightly better job but it's only on Mac, as is grunt-imageoptim.

I was actually considering whether optimizing images as part of the build process makes sense, as we have very few images, and it's a significant contributor in slowing down the build process, especially since our images are already optimized. May be better to just make it generally understood that you should pre-optimize any images you add.

I checked for breaking changes for the modules with version number jumps, and tested everything I could think of. Might have missed something, but we'd find that out pretty soon enough. I think we're pretty out of date on our dependencies, and this would pave the way for further updates, like to Select2 4.x.

Issue #108, opened 2014-07-31
Workload filters


ahbeng:
Suggested by Zeus Khambatta. In particular, 'no project work' might be helpful.

yangshun:
Closed via #197

PR #197, opened 2015-04-13
#108 Add workload filter

xming13:
In the case of invalid workload data, I categorized them under 'Others'.
e.g. mod.Workload is undefined or mod.Workload does not follow the format x-x-x-x-x

### DIFF
--- a/app/scripts/modules/views/ModulesView.js
+++ b/app/scripts/modules/views/ModulesView.js
@@ -110,12 +110,34 @@ module.exports = Marionette.LayoutView.extend({
         mod.Types = ['Not in CORS'];
       }
       mod.semesterNames = [];
+
       for (var i = 0; i < mod.History.length; i++) {
         var history = mod.History[i];
         var sem = history.Semester;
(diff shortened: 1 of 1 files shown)


xbili:
Looks good to me, anybody have issues with this?

yangshun:
Haven't got the time to test >_< @xinan said it was ok in Chrome but laggy in Safari. But that might not be caused by this PR.

xinan:
Haha it's not caused by this PR. The filter was laggy in Safari. I have one suggestion for this PR, instead of the super long list of checkboxes, how about using a select?

yangshun:
What if want to check a few? `<select>` doesn't offer that

xinan:
Oh ya. Then it's ok for now. We can find something like "MultiSelect" in the future.

xming13:
Some kind of range picker will be good too. 
e.g. 0 - 5 hours

The original motivation for this issue was to allow user to select 0 for the hours. Useful for cases like no project work or no tutorial.

Issue #188, opened 2015-03-28
Change to FB Page Plugin


xinan:
Facebook Like Box is deprecated and will stop working soon. 

https://developers.facebook.com/docs/plugins/like-box-for-pages

> With the release of Graph API v2.3, the Like Box plugin is deprecated and will stop working on **June 23rd 2015**. Use the new [Page Plugin](https://developers.facebook.com/docs/plugins/page-plugin) instead. The Page Plugin allows you to embed a simple feed of content from a Page into your websites.

yangshun:
It looks great, let's use that!

yangshun:
Closed via #204

PR #204, opened 2015-05-10
Change to FB Page Plugin #188

MrFool:
Change to FB Page Plugin for #188 

Demo:

screen shot 2015-05-10 at 9 28 57 pm
screen shot 2015-05-10 at 9 28 52 pm

Discussion:

1. Page plugin requires JS-SDK as per FB Page Plugin documentation
2. Because of 1, we need to insert it somewhere where it'll be loaded upon refresh, appending it under fb-root in index.html causes the FB Page Plugin to disappear upon refresh ->

screen shot 2015-05-10 at 9 41 53 pm
screen shot 2015-05-10 at 9 41 48 pm

3. Currently, I'm not very sure where to place such a JS and have put it within one of the view's JS (added a TODO to move it somewhere else) so what would be a good place to put this SDK? (Tried to put it beside Google Analytics but that doesn't work either, causes it to disappear upon refresh as in 2)

Let me know if anything is gravely wrong, still figuring out how nusmods work and how the app is structured :)

### DIFF
--- a/app/scripts/timetable/templates/timetable.hbs
+++ b/app/scripts/timetable/templates/timetable.hbs
@@ -22,7 +22,7 @@
           <div class="url-sharing-region"></div>
         </div>
         <div class="hidden-xs">
-          <iframe src="//www.facebook.com/plugins/likebox.php?href=https%3A%2F%2Fwww.facebook.com%2FNUSMods&amp;width&amp;height=558&amp;colorscheme=light&amp;show_faces=true&amp;header=false&amp;stream=true&amp;show_border=false&amp;appId=491857110934417" scrolling="no" frameborder="0" style="border:none; overflow:hidden; height:448px; width: 100%" allowTransparency="true"></iframe>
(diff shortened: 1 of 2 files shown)


yangshun:
I think you don't need refresh, you mean navigating away and navigating back. There's no need to load the JS SDK again upon navigating back to the timetable view. Upon navigating back, simply do `      FB.XFBML.parse();` to bootstrap the loaded template again (:

I ended up doing this in the end:

```
    if (window.FB) {
      FB.XFBML.parse();
    } else {
      (function(d, s, id) {
        var js, fjs = d.getElementsByTagName(s)[0];
        if (d.getElementById(id)) return;
        js = d.createElement(s); js.id = id;
        js.src = "//connect.facebook.net/en_GB/sdk.js#xfbml=1&version=v2.3&appId=1524196174461544";
        fjs.parentNode.insertBefore(js, fjs);
      }(document, 'script', 'facebook-jssdk'));
    }
```

MrFool:
Nice, what does xfbml stand for?

Btw the block quote thing is provided by Facebook when we get the code from them, believe it's for when JS is disabled in the browser.

yangshun:
xfbml is fb's markup document syntax. haha i'd sacrifice the momentary flash of `blockquote` placeholder for aesthetic purposes. If JS was disabled, the rest of the site won't render anyways, and it's quite pointless putting a link to NUSMods haha.

MrFool:
Oh, xml with fb inserted in between LOL.

Yeah most websites won't even render without JS xD

Issue #206, opened 2015-05-12
Use nusmoderator library instead of academicCalendar.js


yangshun:
An academic calendar code is currently still using our local `academicCalendar.js` code. @zhouyichen has ported  `academicCalendar.js` into an NPM package and can be found at: https://www.npmjs.com/package/nusmoderator. 

We should update our code to make use of the **nusmoderator** package instead.

yangshun:
Closed via 0b0c36a84649916ae6f006433affe436502f32bc

PR #209, opened 2015-05-13
Refactor academicCalendar.js #206

zhouyichen:
Changed the return format to JSON.
Included support for special semester and orientation week.


yangshun:
Thanks @zhouyichen ! (: Fixes #206 .

Issue #208, opened 2015-05-13
Migrate to grunt-sass from grunt-contrib-compass, removing dependency on Ruby


ahbeng:
Bootstrap-sass doesn't require Compass, and we don't use any Compass-specific features, so [grunt-sass](https://github.com/sindresorhus/grunt-sass) should be faster, and remove the need for Ruby. Should also remove Ruby-related provisioning.

PR #217, opened 2015-06-02
Remove ruby dependencies for SASS

yangshun:
Referring to #208 , I have migrated the project from `grunt-contrib-compass` to `grunt-sass`. A few things to take note:
- Version 3.1.3 of jasny-bootstrap does not work with the new sass task because of a bug. I have modified `bower.json` to point to a stable commit on master. Looks like we'll have to use this for a while.
- The `sass:server` task has many CSS files included there. Not sure if there is a better way to do it.
- Removed `Gemfile` and  `Gemfile.lock` because bundle is no longer needed.

Please review this @xinan @xbili @ahbeng . Try to clone the project from scratch and run `grunt serve` and `grunt build`.


xbili:
Works fine, everything still works for me. :+1: 

The styling of the page seems a bit off as compared to the current live version of the page. Try switching between the two and notice the bookmark button's difference in margin. Feels kinda negligible, any concern for that? 

Also, `compass:server` took pretty long as well compared to `sass:server`. Not sure what else can be done :\

yangshun:
Hey @xbili , thanks! Do you mean the other way round that `sass:server` is taking longer than `compass:server`?

xinan:
Hmm I timed both version. `compass:server` takes more than 6 seconds whereas `sass:server` takes less than 2 seconds. So I think Bili doesn't mean that.

Gruntfile.js
```diff
@@ -149,34 +149,21 @@ module.exports = function (grunt) {
...
-                cssDir: '.tmp/styles',
-                generatedImagesDir: '.tmp/images/generated',
-                imagesDir: '<%= yeoman.app %>/images',
-                javascriptsDir: '<%= yeoman.app %>/scripts',
-                fontsDir: '<%= yeoman.app %>/styles/fonts',
-                importPath: '<%= yeoman.app %>/bower_components',
-                httpImagesPath: '/images',
-                httpGeneratedImagesPath: '/images/generated',
-                httpFontsPath: '/styles/fonts',
-                sourcemap: true,
```
xinan: The little off in the style is because of a problem in the `compass` setting. In order to work with Bootstrap correct, we should set the SASS number precision to 10 by adding `raw: 'Sass::Script::Number.precision = 10\n'` into the options. 
`grunt-sass` by default sets it to 10 for us. So actually the new one is more accurate.

PR #220, opened 2015-06-08
More friendly setup for Windows Developers

Yunheng:
Windows developers only have to use "vargrant up" to work on NUSMods locally. No need for "pip install ansible" :)


xinan:
Then where does it get `ansible`? The windows version of `vagrant` comes with it?

Yunheng:
The provisioning script sets up ansible within the VM.

xinan:
Oh I see. :sweat_smile: Sorry didn't read the script. Thanks for that! :+1:

PR #229, opened 2015-07-11
Refactor academicCalendar.js, added more comments

zhouyichen:
Sorry there was a bug in the previous commit. The week number was wrong for Special Terms.


xinan:
For constants it would be better if they are in ALL_CAPS. :smile:

zhouyichen:
Okay, have changed the constants to capital letters

xinan:
:+1: Merged with thanks :smile:

PR #232, opened 2015-07-17
Display modules without exams before those with exams

zhouyichen:
Modules without exams usually end during week 13 or reading week, which is before the starting of final exams of other modules. Therefore it is more intuitive to have modules without exams displayed before those with exams.


xbili:
I think the purpose of the exam timetable is for students to check their exam date and timing. So it should be more intuitive to leave those modules without exams at the bottom. What do the rest think about this?

yangshun:
I think this ordering makes sense. Modules that have no exam usually have huge project components that are due like in recess week, and these can be viewed as exam equivalents. This new ordering is a better indicator of the looming exam dates for each module even for those without exams. Thanks @zhouyichen !

PR #243, opened 2015-10-28
Use our very own Nusmoderator library for academic year calculation (:

yangshun:
Install [nusmoderator](https://github.com/nusmodifications/nusmoderator) via bower and use it in NUSMods for our academic year calculation (:

Much thanks to @zhouyichen and @karenang for their work on extracting out nusmoderator!


xbili:
:+1:

PR #244, opened 2015-10-28
Improve module page section scrolling behaviour

yangshun:
Not sure since when the scrolling to section behaviour on a module page is jerky (double jumping to a section is seen). Fixed this by only setting the hash after the scrolling animation is over.


xbili:
:+1:

PR #249, opened 2016-01-27
Current Day of Week + Time of Day indicator

mauris:
Human has no time to find the current day / time on the timetable. Human needs help from machine. So machine decides to show when is the day of week and time of day on the timetable with an orange background.

I have no background on Backbone + Marionette, but looking at the code I roughly know what's going on. I've added the timer initialization into the onRender event of TableView.js, using jQuery to update and reset background colours of the `td` on the timetable. 

Not sure if this is the best way to write the code, but I am willing to listen to changes. 

Hope this feature would be very useful to all.


app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +27,49 @@ module.exports = Marionette.CompositeView.extend({
     colgroups: 'colgroup'
   },
 
+  onRender: function() {
+    var self = this;
+
+    // updates every minute thereafter
+    window.setTimeout(function() {
```
yangshun: Is this periodic really needed? It's a nice idea but if it takes up too much memory then it might not be so worth it. :joy:

Anws it's not ideal to run a check every minute. Some proposals I have:
1. Check once every 60/30/15 minutes (I'm undecided)
2. Check on window `focus` event.

app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +27,49 @@ module.exports = Marionette.CompositeView.extend({
...
+  },
+
+  updateDayTimeIndicator: function() {
+    var nowDate = new Date();
+    var day = nowDate.getDay();
+    var hour = nowDate.getHours();
+    var minutes = nowDate.getMinutes();
+
+    // leading zero check
+    if (hour < 10) {
```
yangshun: We have a `padTwo` util function. Check out `common/utils/padTwo.js`

app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +27,49 @@ module.exports = Marionette.CompositeView.extend({
...
+      'mon',
+      'tue',
+      'wed',
+      'thu',
+      'fri',
+      'sat'
+    ];
+
+    this.$('#timetable td').css('background', 'none');
+    this.$('#' + dayMapping[day] + ' .h' + hour + '.m' + minutes).css('background', 'rgba(255, 140, 20, 0.5)');
```
yangshun: Some comments
- Can we add a class instead of adding inline CSS?
- Possible to apply a class to the `tr` instead of `td`?

app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +27,49 @@ module.exports = Marionette.CompositeView.extend({
...
+    }
+
+    if (minutes < 30) {
+      minutes = '00';
+    } else {
+      minutes = '30';
+    }
+
+    var dayMapping = [
+      'sun',
```
yangshun: Don't think we have a sunday row.

yangshun: Ok looks like this is necessary due to how the native JS Date works.

app/scripts/timetable/views/TableView.js
yangshun: Sorry re-read the code and realized it shows hour interval too. Now I understand why you chose to do such frequent updates. Hmm maybe 5 minutes interval then?

app/scripts/timetable/views/TableView.js
yangshun: Use a class to apply to the td instead of setting inline style.

yangshun:
Thanks @mauris , I'll look into it more in-depth tonight. Not sure how necessary this is but if some people find it helpful and it doesn't impact performance much we'll take it.

mauris:
alright! Thanks @yangshun for the comments. I'll work on the comments shortly after class.

app/scripts/timetable/views/TableView.js
yangshun: I think you meant `setInterval`. `setTimeout` only runs once.

mauris: ah yes, I needed `setInterval`. will change that.

app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +27,49 @@ module.exports = Marionette.CompositeView.extend({
...
+      self.updateDayTimeIndicator.call(self);
+    }, 60000);
+
+    // initial updating
+    this.updateDayTimeIndicator();
+  },
+
+  updateDayTimeIndicator: function() {
+    var nowDate = new Date();
+    var day = nowDate.getDay();
```
yangshun: Can use `moment` instead? We already have it in the code. Moment.js is more awesome (:

app/scripts/timetable/views/TableView.js
yangshun: Calling this inside `onRender` is not right. Everytime the timetable is rendered, a timer will be set. If I go to timetable -> other page -> timetable, two timers will be set. Two ways to do it:

1. Set the interval in some global script that runs on app initialization like `app.js` to notify the timetable to update itself, 2. Set the timer on render but clear the timer whenever the view is destroyed in the `onBeforeDestroy` event.

app/scripts/timetable/views/TableView.js
yangshun: It's weird to have half a block being highlighted. Can we do every hour interval instead?

app/scripts/timetable/views/TableView.js
tgxworld: :+1:

app/scripts/timetable/views/TableView.js
yangshun: Actually I'm not exactly sure what's the Marionette way of doing this. @ahbeng any suggestions?

app/scripts/timetable/views/TableView.js
mauris: sure thing. My initial code does hourly highlighting, then I figured since m00 and m30 is available, why not be more precise

app/scripts/timetable/views/TableView.js
mauris: will switch to moment. wasn't aware moment's around!

mauris:
in the mean time, I've added `clearInterval` to clear the timer in the `onBeforeDestroy` event.

mauris:
@yangshun any update on this? 😃

app/scripts/timetable/views/TableView.js
```diff
@@ -27,6 +30,30 @@ module.exports = Marionette.CompositeView.extend({
     colgroups: 'colgroup'
   },
 
+  onRender: function() {
+    var self = this;
+
+    // updates every minute thereafter
+    this.timerUpdateDayTime = window.setInterval(function() {
+      self.updateDayTimeIndicator.call(self);
```
yangshun: I don't think you need `.call(self)` here but I might be wrong. Can try to remove it and see if it works?

app/styles/_timetable.scss
```diff
@@ -13,6 +13,9 @@
   th {
     background-color: #fff;
   }
+  .currentDayTime {
```
yangshun: Sorry for the nit pick, but please use kebab case for the class name.

mauris:
@yangshun addressed what you wrote! Thanks for going through and pointing them out.

yangshun:
It just occurred to me... What if the time is outside the range of the timetable? Would it error because the table cell for that time cannot be found?

yangshun:
@mauris I deployed latest master to staging: http://staging.nusmods.com/timetable/2016-2017/sem1?AR5951C=&CS3210[LEC]=1&CS3210[TUT]=3. We'll find out after 12mn 😂

yangshun:
It seems fine (:

mauris:
yeah if the hour doesn't exist it'll just do nothing

PR #250, opened 2016-02-02
Fix ansible provision on windows

ngzhian:
1. in install_dependencies.yml, we use the command module and to
   run bower, in that case we need to sepcify --allow-root, otherwise
   bower will crap out saying that we cannot run with root
2. in windows.sh (the shim to do ansible provisioning for us when
   the host machine is a Windows), the variable `ansible_ssh_user` is
   missing, so the other config files that rely on that variable, and
   also on `user` variable gets a blank string, which is not good.
   To fix, we pass in the `ansible_ssh_user` via the --extra-vars
   command line param, set to the value of `vagrant`, which is what
   it is if we use vagrant to provision the box.


yangshun:
Great, thanks for this @ngzhian ! (:

ngzhian:
No probs, There's some issue with provisioning, I'll make another issue about it

Issue #40, opened 2014-07-09
Migrate build system from grunt to gulp


ahbeng:

ngzhian:
In lieu of the ever-changing landscape of nodejs build system, is gulp still our target? Or should we aim for webpack/<any other build framework>?
(sorry I'm a nodejs noob so I'm not too familiar with what's *really* good, I just know some buzz-framework-names)
cc @yangshun

yangshun:
Gulp is still relevant. If we use webpack then we can do away with browserify and it'll involve a major overhaul. We can afford to migrate to gulp first, for faster build times

yangshun:
Closed via #258

PR #258, opened 2016-02-04
Migrate to Gulp

ngzhian:
So this is the start of #40, I tried my best to ensure that this change doesn't break anything by:

1. looking at the files in the `dist` folder, making sure all the files that used to be there are there
2. checking sizes of important files to make sure they match
3. visually examining the debug server and trying out things like livereload

If anyone has any clue how to better test this, or is willing to test it, lmk!

Otherwise there's a bit more work to to with regards to documentation, simplifying the rules (I feel like there's some redundancy and inconsistencies), which can be done in this PR or another.

Important thing to note, we can't serve versioned logo images because gulp-usemin doesn't support it (yet). I've checked and it seems like we don't do crazy things with that (and logos don't change often), so I think it's acceptable.

TODO: in grunt we used this 'newer' thing, gulp has it's own way of building changed files


gulpfile.js
```diff
@@ -0,0 +1,287 @@
...
+
+var imageminOptions = {
+  interlaced: true,
+  optimizationLevel: 3,
+  progressive: true
+}
+
+gulp.task('imagemin', function() {
+  // we cannot revision the logos until gulp-usemin is able to handle rev images
+  var images = gulp.src('app/images/{,*/}*.{gif.jpeg,jpg,png}')
```
yangshun: `{gif.jpeg,jpg,png}` <- is the `.` a typo?

gulpfile.js
```diff
@@ -0,0 +1,287 @@
...
+});
+
+gulp.task('clean:server', function() {
+  del.sync(['.tmp/**']);
+});
+
+// Make sure code styles are up to par and there are no obvious mistakes
+gulp.task('jshint', function(cb) {
+  return gulp.src([
+      'Gruntfile.js',
```
yangshun: Hmm isn't `Gruntfile.js` deleted?

gulpfile.js
ngzhian: Good catch, yes it is!

yangshun:
Looks good, thanks for this @ngzhian ! This is awesome, one of our more significant PRs yet haha.

I haven't patched it and tried it, but I will do so by this weekend (:

gulpfile.js
ngzhian: Good catch, fixed!

yangshun:
Hi @ngzhian , very nice work. Except a few small issues:

1. `gulp-sourcemaps` is missing from `package.json`. I had to install it.
2. Livereload for styles work but doesn't work for the JS and `hbs`. This is really important for development as not only is the page not reloaded, the app is not being recompiled. After some investigation, I think it should be `gulp.dest('.tmp/scripts/')` instead of `gulp.dest('./tmp/scripts/')` in `browserify:watch`?
3. `gulp build` fails for me. The image versioning is not important at all because we don't have many (one?) images in NUSMods at all. Feel free to do a direct copy of the image file when building.
4. Could you update the `README.md` as well for the new `gulp` commands to develop?

Really excited to get this in! Lemme know if you need help with it. (:

ngzhian:
@yangshun thanks for your meticulous checking.
I fixed all the issues above except 3, may I have the error log for your `gulp build` command?
It works fine for me (and travis) so idk what's wrong.
There's still one big piece of the puzzle missing here: incremental browserify bundling. Right now I set it such that everytime a `.js` or `.hbs` file changes, we bundle the whole thing again -> this is slow. I tried using `watchify` to set up fast builds, but somehow it isn't working.
If you have time I would love another pair of eyes to look at that.

ngzhian:
also, think there's an issue with my host machine being a windows machine, it doesn't pick up file changes due to the synced folder. lmk if you face this issue, a quick way to test is to only run watchify as such (from root of this repo):
`./node_modules/watchify/bin/cmd.js -v -o tmp/output.js app/scripts/main.js`
and then try to edit `main.js` and see if watchify builds it.

ahbeng:
I might have gotten watchify to work before in an experimental branch - will dig up to compare and review tomorrow.

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+var rev = require('gulp-rev');
+var connect = require('gulp-connect');
+var modRewrite = require('connect-modrewrite');
+var merge = require('merge-stream');
+var usemin = require('gulp-usemin');
+var uglify = require('gulp-uglify');
+var minifyHtml = require('gulp-htmlmin');
+var minifyCss = require('gulp-cssnano');
+var rsync = require('gulp-rsync');
+var mocha = require('gulp-mocha');
```
ahbeng: Could consider using [gulp-load-plugins](https://www.npmjs.com/package/gulp-load-plugins) to avoid manually requiring all these gulp plugins.

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+gulp.task('browserify', function() {
+  var b = browserify({
+    entries: ['app/scripts/main.js'],
+  });
+  return b.bundle()
+    .pipe(source('main.js'))
+    .pipe(buffer())
+    .pipe(uglify())
+    .pipe(gulp.dest('.tmp/scripts/'));
+});
```
ahbeng: Looks like the API changed a bit, this worked for me previously:
```js
var dist = false;

gulp.task('setDist', function() {
  dist = true;
});

gulp.task('browserify', function () {
  var bundleMethod = dist ? browserify : watchify;
  var bundler = bundleMethod('./app/scripts/main.js');

  var bundle = function() {
    return bundler
      .bundle({debug: true})
      .on('log', console.log.bind(console))
      .on('error', console.error.bind(console))
      .pipe(source('main.js'))
      .pipe(gulp.dest((dist ? 'dist' : '.tmp') + '/scripts'));
  };

  if (!dist) {
    bundler.on('update', bundle);
  }

  return bundle();
});
```

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+    'app/vendor/facebook/php-sdk-v4/autoload.php',
+    'app/vendor/facebook/php-sdk-v4/src/Facebook/**/*',
+    'app/bower_components/font-awesome/fonts/*.*',
+  ], { base: 'app' })
+    .pipe(gulp.dest('dist/'));
+
+  var zeroclipboard = gulp.src('node_modules/zeroclipboard/dist/ZeroClipboard.swf')
+    .pipe(gulp.dest('dist/'));
+
+  return merge(apps, zeroclipboard);
```
ahbeng: I think the merge isn't needed? Both gulp.src calls could be combined if `base: 'app'` is removed, since all the files there are under `app/` anyway.

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+    .pipe(jshint.reporter(stylish));
+});
+
+gulp.task('livereload', function() {
+  gulp.src([
+      'app/{,*/}*.html',
+      '.tmp/scripts/main.js',
+      '.tmp/styles/{,*/}*.css',
+      'app/images/{,*/}*./{gif,jpeg,jpg,png,svg,webp}'])
+   .pipe(connect.reload());
```
ahbeng: I feel like [Browsersync](https://www.browsersync.io/docs/gulp/) has more features than LiveReload, but would require a bit more setup. This was what I had previously:

```js
gulp.task('serve', ['styles', 'browserify'], function () {
  browserSync({
    notify: false,
    server: {
      baseDir: ['.tmp', 'app', 'api/app', 'node_modules/zeroclipboard/dist'],
      middleware: [
        // Rewrite everything that does not contain a '.' to support pushState
        modRewrite(['^[^\\.]*$ /index.html [L]'])
      ]
    }
  });

  gulp.watch(['app/**/*.html'], reload);
  gulp.watch(['app/styles/**/*.{scss,css}'], ['styles', reload]);
  gulp.watch(['app/scripts/**/*.js'], ['jshint']);
  gulp.watch(['app/images/**/*'], reload);
});
```

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+  gulp.watch('app/styles/{,*/}*.css', function() {
+    runSequence('copy:styles', 'livereload');
+  });
+  gulp.watch('app/{,*/}*.html', ['livereload']);
+  gulp.watch('.tmp/scripts/main.js', ['livereload']);
+  gulp.watch('app/images/{,*/}*./{gif,jpeg,jpg,png,svg,webp}', ['livereload']);
+  gulp.watch('package.json', ['browserify']);
+  gulp.watch('app/scripts/**/*.{js,hbs}', function() {
+    runSequence('browserify', 'livereload');
+  });
```
ahbeng: The globs with the same tasks to run could be combined into an array passed to one `gulp.watch` call.

gulpfile.js
```diff
@@ -0,0 +1,289 @@
...
+      'hostname': '0.0.0.0',
+      'middleware': connectMiddleware,
+      'root': ['.tmp', 'test', 'app', 'api/app']
+  });
+});
+
+gulp.task('serve:dist', ['build', 'connect:dist']);
+gulp.task('serve', function() {
+  runSequence(
+    'clean:server', ['sass', 'copy:styles'],
```
ahbeng: Maybe clearer with 'clean:server' as a gulp.task dependency? Here and below.

ahbeng:
Very nice @ngzhian! Some of my comments are more for future reference as to possible improvements, feel free to ignore for this PR.

ngzhian:
oo @ahbeng such details! Thanks for the feedback, I'll have time to patch this up this weekend and hopefully be able to push this out then! For now, back to ruby-land :sob:

ngzhian:
Took care of almost all of your comments except:
  - [ ] use browsersync

Interestingly, watchify doesn't work that great when I across the mounted filesystem, I think the package that it uses to detect file changes doesn't work for this, whereas the one that `gulp.watch` uses does. I tested this by running `gulp serve` on an identical NUSMods repo in a different directory (that is not shared) in vagrant, and it works! Will be cool if any of you can try to repro this problem.

ngzhian:
So, as the last commit shows, I hacked it such that watchify uses gulp.watch instead of chokidar, apparently gulp.watch works fine for our synced folders, but chokidar doesn't... I *might* spend some time investigating, but right now this works for me, would love some help making sure it works for your setup too! Thanks :)
@yangshun @ahbeng

yangshun:
Sure thing. I'll do so asap. @ngzhian btw the tests are failing >_<

@xinan If you are free could you help try too? This is a very important PR :sweat_smile:

xinan:
:sweat_smile: Okay I will try after finishing all my homework and mid-terms

ngzhian:
ping :)

ahbeng:
Sorry didn't realize this hadn't been looked at yet! I've an actual Windows machine now, will try during weekend.

yangshun:
Thanks @ngzhian (:

ngzhian:
✨ 🎆

ngzhian:
time to work on my PR to migrate to webpack! *jk*

PR #261, opened 2016-02-10
Remove fb btn from each mod at /modules

### LINKED ISSUE #260: Do we really need a "like" button for each module at /modules ?
I find usage really low, and (not tested experimentally) feel that so many iframes slow down the page slightly :P 
See picture:

ngzhian:
Fixes #260


yangshun:
Could you remove the corresponding style in `main.scss` also? Then just flatten and merge. Thanks!

yangshun:
:+1: thanks!

PR #269, opened 2016-07-09
Fix ansible dep

### LINKED ISSUE #268: Ansible deprecation warnings
Running on ansible 2.1.0.0 gives us a bunch of deprecation warnings

```
[DEPRECATION WARNING]: Instead of sudo/sudo_user, use become/become_user and 
make sure become_method is 'sudo' (default).
This feature will be removed in a 
future release. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

```
TASK [app_server : template] ***************************************************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your 
playbooks so that the environment value uses the full variable syntax 
('{{domains}}').
This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```
```
TASK [app_server : file] *******************************************************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your 
playbooks so that the environment value uses the full variable syntax 
('{{domains}}').
This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

ngzhian:
Fixes #268 
- remove bare variables in `with-items` clauses
- use `become` and `become_user` instead of `sudo`

review? @ahbeng

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 On Unix/Mac:
 
 ```bash
-$ pip install ansible
+$ pip install ansible==2.1.0.0
 $ vagrant up
 ```
 
--- a/provisioning/production.yml
+++ b/provisioning/production.yml
@@ -1,7 +1,7 @@
 ---
 - hosts: all
   roles:
-    - { role: git, sudo: yes }
+    - { role: git, become: yes }
   tasks:
     - git: repo=https://github.com/nusmodifications/nusmods.git dest={{ project_dir }}
         accept_hostkey=yes
--- a/provisioning/roles/app_server/tasks/main.yml
+++ b/provisioning/roles/app_server/tasks/main.yml
@@ -15,11 +15,11 @@
 - template:
     src=etc/nginx/sites-available/{{ item }}.j2
     dest=/etc/nginx/sites-available/{{ item }}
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -18,7 +18,7 @@
 On Unix/Mac:
 
 ```bash
-$ pip install ansible
+$ pip install ansible==2.1.0.0
```
ahbeng: These breakages don't happen all too often. My concern is that this would mean we have to update this version in the instructions periodically also.

While I assume anyone who does this step can either go back to a working version themselves, file an issue or better still, fix it?

ngzhian: True that this doesn't happen too often, but not specifying the version causes random warnings, and eventually breakages when ansible upgrades.
By specifying the version, a new developer won't be faced with random breakages when they get a newer version of ansible (that we have not updated the yaml files to support)

ahbeng: Makes sense to ease any potential kinks in initial dev setup. They wouldn't need to necessarily follow that after they're more familiar, and could then find issues with more recent ansible versions.

PR #270, opened 2016-07-14
Implement tab to search on chrome omnibar

ngzhian:
Fix 262

### DIFF
--- a/app/opensearch.xml
+++ b/app/opensearch.xml
@@ -2,5 +2,5 @@
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
   <ShortName>NUSMods Module Search</ShortName>
   <Description>Search for modules on NUSMods</Description>
-  <Url type="text/html" method="get" template="/modules/{searchTerms}"/>
+  <Url type="text/html" method="get" template="https://nusmods.com/modules/{searchTerms}"/>
 </OpenSearchDescription>
--- a/app/opensearch.xml
+++ b/app/opensearch.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
- <ShortName>NUSMods Module Search</ShortName>
- <Description>Search for modules on NUSMods</Description>
- <Url type="text/html" method="get" template="https://nusmods.com/modules/{searchTerms}"/>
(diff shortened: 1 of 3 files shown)


app/opensearch.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+ <ShortName>NUSMods Module Search</ShortName>
```
yangshun: Indent by 2 or 1?

app/opensearch.xml
```diff
@@ -0,0 +1,6 @@
...
+ <Description>Search for modules on NUSMods</Description>
+ <Url type="text/html" method="get" template="https://nusmods.com/modules/{searchTerms}"/>
```
yangshun: Is it possible to use relative path here?

ngzhian: I've just changed it to use relative paths, can you pull this branch and try it to see if it works? I tried it but it might be because chrome cached the older version.

yangshun:
Why assign to me? :joy:

ngzhian:
cos you da best!!

yangshun:
@ngzhian I deployed to staging: staging.nusmods.com. But I'm not sure how to trigger the search. Doesn't seem to trigger for me :S

ngzhian:
hm, if you type staging.nusmods.com into your bar, it should show up. Otherwise visit staging.nusmods.com once, and the next time you type staging.nusmods.com it should show up.

app/opensearch.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>NUSMods Module Search</ShortName>
+  <Description>Search for modules on NUSMods</Description>
```
yangshun: More accurately: "Enter a module code" since we don't support other forms of search

ngzhian:
I got an email with a comment
> More accurately: "Enter a module code" since we don't support other forms of search

You think we should change the description? The problem with using "Enter a module code" is that it doesn't read well - "Search enter a module code".

yangshun:
@ngzhian I realized that, that’s why I deleted the comment 😂

ngzhian:
@yangshun okay :) merge? :D

yangshun:
@ngzhian ship it

PR #276, opened 2016-07-24
Add Faculty facet to modules

### LINKED ISSUE #25: [Design] Group by Faculty
Why the courses are no longer grouped in the new release?

The issue with this design is that I couldn't browse for the courses under a certain school.
For example: I wanted to browse for all module under "School of Business", or "School of Computing", I have to find out what are the courses from school of computing and school of business. Such as:
1. Information System
2. Computer Science
3. Financial Accounting
4. Strategy & Policy
5. etc..

screenshot from 2014-07-07 14 31 42

ngzhian:
Fixes #25

### DIFF
--- a/app/scripts/modules/views/ModulesView.js
+++ b/app/scripts/modules/views/ModulesView.js
@@ -40,12 +40,12 @@ module.exports = Marionette.LayoutView.extend({
 
   initialize: function (options) {
     this.mods = options.mods;
-    const departmentToFaculty = {}
-    Object.keys(options.facultyDepartments).map((fac) => {
-      options.facultyDepartments[fac].map((dept) => {
-        departmentToFaculty[dept] = fac
-      })
-    })
+    var departmentToFaculty = {}
+    Object.keys(options.facultyDepartments).map(function(fac) {
+      options.facultyDepartments[fac].map(function(dept) {
+        departmentToFaculty[dept] = fac;
+      });
+    });
     this.departmentToFaculty = departmentToFaculty;
   },
 
--- a/app/scripts/modules/views/ModulesView.js
+++ b/app/scripts/modules/views/ModulesView.js
@@ -172,8 +172,14 @@ module.exports = Marionette.LayoutView.extend({
       label: 'Types',
       slug: 'types'
     }));
+    facets.add({
+      filteredCollection: mods,
+      key: 'faculty',
+      label: 'Faculty',
+      slug: 'faculty'
+    })
     facets.add(_.map({
-      Department: 'Faculty / Department',
+      Department: 'Department',
       level: 'Level'
     }, function(label, key) {
       return {
@@ -183,12 +189,6 @@ module.exports = Marionette.LayoutView.extend({
         slug: slugify(label)
       };
     }));
-    facets.add({
-      filteredCollection: mods,
-      key: 'faculty',
-      label: 'Faculty',
-      slug: 'faculty'
-    })
     facets.add({
       filteredCollection: mods,
       key: 'ModuleCredit',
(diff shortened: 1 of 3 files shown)


ngzhian:
@ahbeng do you think we should remove the existing "Faculty / Department" facet, or rename it to "Department"?

ngzhian:
@yangshun @9gix any thoughts? :)

yangshun:
@ngzhian I’ll look at it tonight (:

ahbeng:
I think "Faculty / Department" may have been since some smaller faculties have only one department so they are named the same? Renaming to "Department" sounds fine since there's now "Faculty".

Why do you think "Department" should  be replaced, they still filter separate things, as far as I understand?

yangshun:
> My thought is that we should replace the existing "Faculty / Department" with this new "Faculty" one

We shouldn't do this. Filtering by course / dept is still a valuable feature to have. Arts students will appreciate this especially. FASS has over 1.3k modules and filtering out just FASS modules isn't very useful.

I think renaming "Faculty / Department" -> "Department" makes sense.

Could you also help to reorder the categories in the following order (if you think there is another order that makes more sense, feel free to comment)

- Semesters
- Types
- Faculty
- Department
- Level
- MCs
- ... (the rest remains)

app/scripts/modules/views/ModulesView.js
```diff
@@ -40,6 +40,13 @@ module.exports = Marionette.LayoutView.extend({
 
   initialize: function (options) {
     this.mods = options.mods;
+    const departmentToFaculty = {}
```
yangshun: In Safari you'll get an error: `SyntaxError: Unexpected keyword 'const'. Const declarations are not supported in strict mode.`

app/scripts/modules/views/ModulesView.js
```diff
@@ -40,6 +40,13 @@ module.exports = Marionette.LayoutView.extend({
...
+    Object.keys(options.facultyDepartments).map((fac) => {
```
yangshun: AFAIK not all browsers support es6 at the moment, so it'll break in Safari. Please refrain from using es6 syntax at the moment. We'll get there eventually >_<

app/scripts/modules/views/ModulesView.js
```diff
@@ -40,6 +40,13 @@ module.exports = Marionette.LayoutView.extend({
...
+      options.facultyDepartments[fac].map((dept) => {
+        departmentToFaculty[dept] = fac
+      })
```
yangshun: Missing `;`

app/scripts/modules/views/ModulesView.js
```diff
@@ -40,6 +40,13 @@ module.exports = Marionette.LayoutView.extend({
 
   initialize: function (options) {
     this.mods = options.mods;
+    const departmentToFaculty = {}
+    Object.keys(options.facultyDepartments).map((fac) => {
+      options.facultyDepartments[fac].map((dept) => {
+        departmentToFaculty[dept] = fac
```
yangshun: Missing `;`

app/scripts/modules/views/ModulesView.js
```diff
@@ -40,6 +40,13 @@ module.exports = Marionette.LayoutView.extend({
...
+      })
+    })
```
yangshun: Missing `;`

yangshun:
LGTM, except for the es6 syntax and semicolons. Or should we not use semicolons? Up for debate!

ngzhian:
Ah makes sense, sorry I was being selfish and thinking that department helps (cos that's what i care about :P)
So I'll:
1. Rename "Faculty / Department" -> "Department"
2. Reorder the facets, I think there's no correct order, and I like the one suggested, so will update it to that :)

app/scripts/modules/views/ModulesView.js
ngzhian: ahh shit okay, i keep assuming that browserify will compile it to es5... we need babel :P

ngzhian:
Updated, ptal :)

PR #277, opened 2016-07-25
Fix eslint and plugin versions

ngzhian:
I somehow messed up the versions so that npm install on a clean nusmod clone will fail, this fixes the versions.


ngzhian:
Okay i tried npm install locally and this works, so just gonna go ahead and merge this :)

PR #280, opened 2016-07-25
Move to npm run commands

yangshun:
Moved the common commands to npm run scripts so that we can use the same commands regardless of underlying implementation (task runner choice),

Ref: https://docs.npmjs.com/cli/run-script

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,6 @@ Install the necessary packages.
 $ npm install -g npm@3.6.0
 $ npm install -g bower@1.7.9 gulp-cli@1.2.2
 $ npm install
-$ cd api && npm install && cd .. # install npm packags for api
 $ bower install
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,7 +43,6 @@ Install the necessary packages.
 $ npm install -g npm@3.6.0
 $ npm install -g bower@1.7.9 gulp-cli@1.2.2
 $ npm install
-$ cd api && npm install && cd .. # install npm packags for api
```
yangshun: @ngzhian I removed this because it's being covered in the "Working with the NUSMods API" section.

README.md
```diff
@@ -97,12 +90,11 @@ default. To work with a local copy of the API:
 $ git submodule update --init
 $ cd api
 $ npm install
-$ gulp
```
yangshun: I believe this was mistakenly changed to `gulp` when we did #258 . This should be the command to run the scraping process and is still in `grunt`.

yangshun:
Please review @ngzhian @ahbeng

package.json
```diff
@@ -18,6 +18,10 @@
     }
   ],
   "scripts": {
+    "start": "gulp serve",
```
ngzhian: let's call this `serve`? consistent with `build` and `rsync` :)

ngzhian:
1 nit, otherwise lgtm :shipit:

package.json
yangshun: `start` and `test` are actually special commands supported by `npm` where you can do `npm start`, `npm test` rather than the rest where you have to do `npm run build`, `npm run rsync` etc. It's the convention to use `start` (:

ngzhian: OHHH, makes sense. :shipit:  then :) thanks for your explanation!

yangshun:
Thanks @ngzhian ! 😄

PR #286, opened 2016-07-27
Add babel

### LINKED ISSUE #281: Babel?
So we can write ES2015 code :)

Out of curiousity, found out that facebook/react uses browserify, grunt, and gulp. They do this thing where they babel-fy all their source code, then use browserify to bundle them up, we can probably do something like this? But I'm not sure how well that will work with watchify.

ngzhian:
Fixes #281 .

So seems like it isn't that hard, the thing I had to do was to use `gulp-babel` to copy the files into `.tmp/modules`, and then change the `browserify` tasks to bundle starting from `.tmp/modules/main.js` instead of the original `app/scripts/main.js`.
Also had to copy some extra `.json` and `.hbs` files into `.tmp` so that browserify can find them.

I also looked at the generated `main.js` file (was worried that it would cause sizes of file to increase), seems like it doesn't, for now!

```
old: 2273158
new: 2272611
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -121,16 +121,16 @@ function bundle() {
 
 gulp.task('bundle', bundle);
 
-gulp.task('browserify', function() {
+gulp.task('browserify', ['copyscripts'], function() {
   bundler = browserify({
-    entries: ['app/scripts/main.js'],
+    entries: ['.tmp/modules/main.js'],
   });
   return bundle();
 });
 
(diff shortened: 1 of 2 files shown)


ngzhian:
Ooops this doesn't work with browsersync hm.. need to fix it

ngzhian:
Fixed by previous commit, ready for review!

ngzhian:
Closed in favour of webpack #287

PR #287, opened 2016-07-28
Webpack!

ngzhian:
`npm start` runs the `webpack-dev-server` for development
`npm run build` builds the app into `build`
`npm test` runs `jshint` and `tests`


yangshun:
How about using Webpack 2? We have been yolo-ing thus far, might as well continue the trend :trollface:

ngzhian:
LOL that sounds good, let's move once it's release!

ngzhian:
The commits are really messy sorry :(
But now i'm at the stage where by we don't need bower anymore, and running in dev mode works. Dist doesn't work cos I couldn't get `usemin` to substitute the hashed file names properly, working on that next.

app/index.ejs
```diff
@@ -37,19 +37,9 @@
...
-  <!-- endbower -->
-  <!-- endbuild -->
-  <!-- build:css(.tmp) /styles/main.css -->
-  <link rel="stylesheet" href="/styles/main.css">
-  <link rel="stylesheet" href="/styles/modes.css">
-  <link rel="stylesheet" href="/styles/themes.css">
-  <!-- endbuild -->
-  <link rel="stylesheet" id="mode">
+  <% for (var css in htmlWebpackPlugin.files.css) { %>
+  <link href="/<%= htmlWebpackPlugin.files.css[css] %>" rel="stylesheet">
```
ahbeng: Output would be neater with rel="stylesheet" first?

app/scripts/common/views/AppView.js
```diff
@@ -176,8 +176,7 @@ module.exports = Backbone.View.extend({
       $('body').removeClass('js-toggled');
     });
 
-    attachFastClick(document.body);
-
+    attachFastClick.attach(document.body);
```
ahbeng: Rename attachFastClick to FastClick since it's no longer FastClick.attach which is exported?

package.json
```diff
@@ -18,46 +18,68 @@
...
     "es6-promise": "^3.0.2",
     "eslint": "^2.10.0",
     "eslint-config-airbnb": "^9.0.1",
     "eslint-plugin-import": "^1.11.1",
     "eslint-plugin-jsx-a11y": "^1.2.2",
     "eslint-plugin-react": "^5.2.2",
+    "extract-text-webpack-plugin": "^1.0.1",
+    "file-loader": "^0.9.0",
     "gulp": "^3.9.0",
     "gulp-autoprefixer": "^3.1.0",
```
ahbeng: Some of the gulp-* modules aren't required anymore?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
+var webpack = require('webpack');
```
ahbeng: Should these be const too?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+var path = require('path');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+const merge = require('webpack-merge');
+const validate = require('webpack-validator');
+const pkg = require('./package.json');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
```
ahbeng: NIt: alphabetical order?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+        test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=application/octet-stream&name=[name].[ext]"
+      }, {
+        test: /\.eot(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "file?name=[name].[ext]"
+      }, {
+        test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=image/svg+xml"
+      },
+      { test: /\.otf/, loader: "file"},
```
ahbeng: $/ for consistency?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+  },
+  output: {
+    path: PATHS.build,
+    filename: "[name].js"
+  },
+  module: {
+    loaders: [
+      {
+        test: /\.woff(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=application/font-woff&name=[name].[ext]"
```
ahbeng: If it isn't easy to get [hash] in the name to work, probably not worth it until far future expires headers on static resources are set up in nginx.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+      },
+      { test: /\.otf/, loader: "file"},
+      { test: /\.swf$/, loader: "file"},
+      {
+        test: /\.jpe?g$|\.gif$|\.png$|\.svg$/,
+        loader: 'file?name=[path][name].[ext]'
+      },
+      // copy favicons to root
+      {
+        test: /favicon-\d+.png$|favicon.ico$/,
```
ahbeng: Regexes could be merged with everything else using the same loader?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+      $: "jquery",
+      jQuery: "jquery",
+      "window.jQuery": "jquery",
+      FastClick: "FastClick"
+    })
+  ],
+  resolve: {
+    alias: {
+      bootstrap: 'bootstrap-sass/assets/javascripts/bootstrap'
+      // localforage: 'localforage/src/localforage'
```
ahbeng: Would it work if aliased to localforage/dist/localforage.nopromises? The promise polyfill which comes with the regular version isn't needed since we're using Bluebird.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+      //   name: 'vendor',
+      //   entries: Object.keys(pkg.dependencies)
+      // })
+    );
+    break;
+  default:
+    config = merge(
+      common,
+      {
+        devtool: 'eval-source-map'
```
ahbeng: Chrome doesn't seem to detect that the source map, even though there are webpack:///./app/* files. 'source-map' worked, but might be slowish? The source mapping works with the gulp version though.

ahbeng:
Nice! Added some minor line comments. Everything seems working fine as far as I tested. As long as the production build works fully, feel free to merge yourself, then can iterate more later on?

webpack.config.js
ngzhian: Separating it makes it clearer how we are handling the different kinds of files we have, okay with merging it too.

webpack.config.js
ngzhian: What's not worth it? Inlining?
We do not need hash for proper caching of fonts to work, because font awesome is retrieved via URL query param `?` that has the version.

webpack.config.js
ngzhian: @ahbeng How does it not work for you?
For me it does what I expect it to:
I went in to add a `console.log` in `TimetableView.js`, and when I click the link to the file I jump to the right place in the source file.

webpack.config.js
yangshun: I use this at work and it has worked for me so far:

```
     {
        test: /fonts\/.*\.woff(2)?(\?v=[0-9]\.[0-9]\.[0-9])?$/,
        loader: 'file?name=fonts/[name].[ext]&limit=10000&mimetype=application/font-woff'
      },
      {
        test: /fonts\/.*\.(ttf|eot|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
        loader: 'file?name=fonts/[name].[ext]'
      },
```

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+        loader: 'file?name=[name].[ext]'
+      },
+      { test: /\.hbs$/, loader: "handlebars-loader" },
+      { test: /\.json$/, loader: "json-loader" },
+      {
+        test: /\.js$/,
+        exclude: /node_modules/,
+        loader: "babel",
+        query: {
+          presets: ['es2015', 'react'],
```
yangshun: I think these would be better in `.babelrc`.

yangshun: Do we have react yet? 😂

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+          new webpack.optimize.UglifyJsPlugin({
+            compress: {
+              warnings: false
+            }
+          })
+        ]
+      }
+      // extractBundle({
+      //   name: 'vendor',
+      //   entries: Object.keys(pkg.dependencies)
```
yangshun: Good idea to not use this approach. Automagically reading from `pkg.dependencies` is not a very good idea from the survivejs tutorial, because the bundle order is alphabetical and some packages depend on others that might not have been loaded alphabetically yet. We should keep an array of vendor dependenices and manually update it.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+        test: /\.eot(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "file?name=[name].[ext]"
+      }, {
+        test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=image/svg+xml"
+      },
+      { test: /\.otf$/, loader: "file"},
+      { test: /\.swf$/, loader: "file"},
+      {
+        test: /\.jpe?g$|\.gif$|\.png$|\.svg$/,
```
yangshun: Can be shortened to: `/\.(jpe?g|gif|png|svg)$/`.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+        test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=application/font-woff&name=[name].[ext]"
+      }, {
+        test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "url?limit=10000&mimetype=application/octet-stream&name=[name].[ext]"
+      }, {
+        test: /\.eot(\?v=\d+\.\d+\.\d+)?$/,
+        loader: "file?name=[name].[ext]"
+      }, {
+        test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
```
yangshun: What's the use case for this `.svg` format? Is it not sufficient to just have the one in the general image format one below?

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+        loader: "url?limit=10000&mimetype=image/svg+xml"
+      },
+      { test: /\.otf$/, loader: "file"},
```
yangshun: Perhaps can shift `.otf` up together to the fonts file format section.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+    app: PATHS.app,
+    style: path.join(PATHS.style, 'main.scss')
+  },
+  output: {
+    path: PATHS.build,
+    filename: "[name].js"
+  },
+  module: {
+    loaders: [
+      {
```
yangshun: Can leave a comment about this section being about fonts.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+}
+
+const extractCSS = function() {
+  return {
+    module: {
+      loaders: [{
+        test: /\.scss$/,
+        loader: ExtractTextPlugin.extract(['css', 'sass'])
+      }, {
```
yangshun: I don't think we have CSS outside the `styles` dir, so to make it faster you can add `include: PATHS.style` here.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+  plugins: [
+    new HtmlWebpackPlugin({
+      template: 'app/index.ejs',
+      inject: false
+    }),
+    new webpack.ProvidePlugin({
+      $: "jquery",
+      jQuery: "jquery",
+      "window.jQuery": "jquery",
+      FastClick: "FastClick"
```
yangshun: Any reason why FastClick is in provide plugin? FastClick is only needed in the app bootstrapping and is only used one place in the app. 

Actually I'm not even sure if FastClick is relevant these days...

package.json
```diff
@@ -18,148 +18,91 @@
...
     "node": ">=0.8.0"
   },
-  "private": true,
-  "browserify": {
-    "transform": [
-      "browserify-shim",
-      "hbsfy"
-    ]
-  },
-  "browser": {
```
yangshun: Wow.. the amount of config we saved by moving to webpack 👍

package.json
```diff
@@ -18,148 +18,91 @@
...
+    "url-loader": "^0.5.7",
+    "vinyl-source-stream": "^1.1.0",
+    "webpack": "^1.13.1",
+    "webpack-dev-server": "^1.14.1",
+    "webpack-merge": "^0.14.1",
+    "webpack-stream": "^3.2.0",
+    "webpack-validator": "^2.2.7"
   },
   "engines": {
     "node": ">=0.8.0"
```
yangshun: I think we can change this to node `>=5.11.0`

app/styles/main.scss
```diff
@@ -1,62 +1,67 @@
...
+@import "bootstrap-sass/assets/stylesheets/bootstrap/utilities";
+@import "bootstrap-sass/assets/stylesheets/bootstrap/responsive-utilities";
 
 @import "jasny-bootstrap/scss/jasny-bootstrap";
 
-$fa-font-path: "/bower_components/font-awesome/fonts";
+$fa-font-path: "";
 @import "font-awesome/scss/font-awesome";
 
+@import "../../node_modules/qtip2/dist/jquery.qtip.css";
```
yangshun: I think you can do `@import "~qtip2/dist/jquery.qtip.css";`. `~` resolves the path like a module import. Same goes for the rest with `../../node_modules/...`

webpack.config.js
yangshun: Might be worth moving other app-widely use libraries like `backbone` and `underscore` into this too.

webpack.config.js
```diff
@@ -0,0 +1,207 @@
...
+    })
+  ],
+  resolve: {
+    alias: {
+      bootstrap: 'bootstrap-sass/assets/javascripts/bootstrap',
+      localforage: 'localforage/dist/localforage.nopromises.js',
+    }
+  },
+  sassLoader: {
```
yangshun: Do we need this?

yangshun:
@ngzhian Just some small comments. I think we might have to revert to Beng's fork of localforage to cater for incognito Safari. With those out of the way, it's good to merge!

webpack.config.js
ngzhian: I think it has to do with the way svgs are loading in some css library where they use a query to load different versions.

webpack.config.js
ngzhian: what happens for css inside `node_modules`?

webpack.config.js
ngzhian: think added this here to fix fastclick deps previously.
backbone and underscore seems to be fine without being here, so i'll leave them out.

webpack.config.js
yangshun: If they were `@include`-ed from the CSS in the styles directory, it should be fine

PR #292, opened 2016-08-15
[FIX] Module not found, redirect back to timetable

mauris:
Currently if a module code is not found, the spinner will spin until the laptop battery goes flat. So I added a catch to the `getMod()` so that it will redirect the user to their timetable if the module is not found.

I cross-matched the code against other code for the practices used by other developers.

Preferably, if possible, after redirect we should tell user that the module is not found.


yangshun:
This is great! Thanks @mauris

PR #293, opened 2016-08-17
[fix] for day time indicator not being removed.

mauris:
for some reason within the scope of "onRender", jQuery variable "$" is not available (probably because the visual is not rendered yet). so the function needs to use "this.$" instead (which i think is wrong). Hence by putting a window.setTimeout() for the update function, I will then be able to use "$" because the timeout event executes after the render event loop completes.

### DIFF
--- a/app/scripts/timetable/views/TableView.js
+++ b/app/scripts/timetable/views/TableView.js
@@ -33,13 +33,17 @@ module.exports = Marionette.CompositeView.extend({
   onRender: function() {
     var self = this;
 
-    // updates every minute thereafter
-    this.timerUpdateDayTime = window.setInterval(function() {
+    var updateDayTimeIndicatorWrapper = function() {
       self.updateDayTimeIndicator();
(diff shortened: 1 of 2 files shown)


ngzhian:
Hi @mauris , since this is a frontend change, a screenshot will be really helpful :)

yangshun:
Did you manage to test whether the bug stops occurring with this fix?

mauris:
The original problem:
screenshot 2016-08-17 16 08 35

The changes made to the code from the last commit of this PR so that we can test it faster without having to wait:
screenshot 2016-08-17 18 33 37

The result:
nusmods

ahbeng:
@mauris I put a todo to review this one again. Fix works but would like to understand what's happening

mauris:
@ahbeng sure thing. I just tested the 9pm tick, fixes the bug reported.

yangshun:
@paradite reported that if he's on the page the cell won't update

mauris:
simply lose focus and come back to the page, instant update

PR #295, opened 2016-08-17
Upgrade localforage to 1.4.2 to fix indexdb version

ngzhian:
So apparently, indexdb in latest chrome is version 2, but our existing version of localforage tries to request in a version 1 format. Upgrade our version to fix this.

ref: http://stackoverflow.com/questions/33660849/domerror-localforage-indexeddb-cases

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -9,7 +9,7 @@
     "jquery-file-download": "~1.4.3",
     "jquery.ui": "~1.11.4",
     "jqueryui-touch-punch": "#4bc009145202d9c7483ba85f3a236a8f3470354d",
-    "localforage": "ahbeng/localForage#2da72b902fb55866f6a4c0415970931d94d98592",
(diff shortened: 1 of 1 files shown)


ahbeng:
@ngzhian I forked it to fix localForage not working in some incognito/private modes. Had some other fixes that did get merged upstream but not that one, not sure if it has already been fixed.

Key difference over localStorage that I thought was useful at the time was having a nice API over other storage methods that could be accessed async where possible - localStorage is sync.

ahbeng:
@ahbeng Clarification: any local storage method obviously wouldn't be useful in incognito mode. But localForage used to break. IIRC my fork just makes sure it doesn't break but doesn't otherwise use local storage.

ngzhian:
Verified that 1.4.2 works in incognito mode: no error messages, and data persists across multiple tabs in the same incognito window.

PR #318, opened 2016-09-14
Change from modules to browse

yangshun:
@li-kai I retained the routes path as `/modules`.


li-kai:
@yangshun Nice, it's exactly what I had in mind. Do I merge this now?

yangshun:
@li-kai go ahead

PR #322, opened 2016-09-17
Use AVA for tests

li-kai:
### AVA
Simple enough, just removed mocha and chai and changed the tests to run with AVA.

### Problems getting flow working
No problems with getting flow annotations, they'll get stripped out automatically
Thing is, `.flowconfig` is the only way to specify where the flow parsing server runs, and there aren't any runtime options or flags to do so. So, there is no way to stop flow from checking files in `/test/` when running `npm start`, so you end up with a bunch of import errors (as webpack doesn't resolve /test/).

Similarly, since there's no way to stop flow from checking files in `src`, errors would pop up there too. I think flow still needs some more time maturing haha. Having multiple `.flowconfig` at different places isn't a really good solution.


v3/.eslintrc.js
```diff
@@ -14,12 +14,6 @@ module.exports = {
     "import",
     "flowtype"
   ],
-  "parserOptions": {
```
yangshun: how come these are not needed anymore?

v3/test/actions/helpers.js
```diff
@@ -0,0 +1,48 @@
+import test from 'ava';
+import * as actions from 'actions/helpers';
+
+function testShouldReturnArr(t, type, action) {
+  const payload = [];
+  const expectedResult = {
+    type,
+    payload: payload,
```
yangshun: can use es6 object literal shorthand.

v3/test/actions/helpers.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('resetRequestState should convert item to an array', (t) => {
+  testShouldReturnArr(t, actions.RESET_REQUEST_STATE, actions.resetRequestState);
+});
+
+test('resetErrorState should convert item to an array', (t) => {
+  testShouldReturnArr(t, actions.RESET_ERROR_STATE, actions.resetErrorState);
+});
+
+
```
yangshun: Extra new line here. Hmm seems like the linting doesn't happen in the tests?

If the goal is to categorize the tests, you can leave a comment instead.

v3/test/actions/moduleBank.js
```diff
@@ -0,0 +1,14 @@
+import test from 'ava';
+import * as actions from 'actions/moduleBank';
+
+test('fetchModuleList should dispatch a request', (t) => {
+  t.true(typeof actions.fetchModuleList() === "function");
```
yangshun: singlequotes for strings.

v3/test/example.js
```diff
@@ -0,0 +1,5 @@
+import test from 'ava';
+
+test('should always pass', (t) => {
+    t.pass();
```
yangshun: 2 spaces

v3/.eslintrc.js
li-kai: [SurviveJS says to remove them](http://survivejs.com/webpack/advanced-techniques/linting/#connecting-eslint-with-babel). I also found no problems running the build so I reckoned it is fine to do so.

PR #326, opened 2016-09-22
Refactor Settings page

li-kai:
Moved the loop to its own component hoping to get the eslint error resolved...

It didn't work.

### DIFF
--- a/v3/src/js/views/settings/SettingsContainer.jsx
+++ b/v3/src/js/views/settings/SettingsContainer.jsx
@@ -1,14 +1,22 @@
 // @flow
 
 import React from 'react';
+import { connect } from 'react-redux';
 import DocumentTitle from 'react-document-title';
 import config from 'config';
 
+import { changeTheme } from 'actions/theme';
 import availableThemes from 'data/themes.json';
 
-import SettingsThemeSelect from './SettingsThemeSelect';
+import ThemeOption from './ThemeOption';
 
-function SettingsContainer() {
+type Props = {
+  currentThemeId: string,
+
+  changeTheme: Function,
+};
+
+function SettingsContainer(props: Props) {
   return (
     <DocumentTitle title={`Settings - ${config.brandName}`}>
       <div className="settings-page-container page-container">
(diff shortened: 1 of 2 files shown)


v3/src/js/views/settings/SettingsThemeSelect.jsx
```diff
@@ -0,0 +1,54 @@
+// @flow
+
+import React from 'react';
+import { connect } from 'react-redux';
```
yangshun: This component is repeatedly rendered in a map and we try to make it a dumb component. Dumb components simply have props and do not connect to redux stores. The typical redux+react approach is that only the container connects to the redux stores and pass state and actions into child components via props. More reading here: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

Another reason why it's a bad idea to `connect` repeatedly rendered items to the store is because there will be each component listens to the store. That's not ideal for performance.

You can pass the `changeTheme` action into the component as a prop.

v3/src/js/views/settings/SettingsThemeSelect.jsx
```diff
@@ -0,0 +1,54 @@
...
+import type { Theme } from 'types/views';
+
+type Props = {
+  theme: Theme,
+  currentThemeId: string,
+
+  changeTheme: Function,
+};
+
+function SettingsThemeSelect(props: Props) {
```
yangshun: Just call it `ThemeSelect` or `ThemeOption`

v3/src/js/views/settings/SettingsThemeSelect.jsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <div className={classnames('col-sm-4 theme-item', {
+      [`theme-${props.theme.id}`]: true,
+      'is-selected': props.currentThemeId === props.theme.id,
```
yangshun: If we factor this out into a component, from the component point of view, giving it a `currentThemeId` doesn't make much sense to it. The component should only be concerned about whether it is selected, it shouldn't be concerned about why it is selected (that's the job of the container). Hence, you can add a boolean prop called `isSelected` or `selected` and the condition for it to be `true` is `props.currentThemeId === props.theme.id` (evaluated in the container, and result passed to this child component.

v3/src/js/views/settings/SettingsThemeSelect.jsx
```diff
@@ -0,0 +1,54 @@
...
+    })}
+      onClick={() => {
+        props.changeTheme(props.theme.id);
```
yangshun: The awesome thing about using `props` is that even though you will change this prop to be passed down from the parent component rather than `connect`, you don't have to change this part :joy:

PR #327, opened 2016-09-22
Flow jsx

li-kai:
High level changes
`type Lesson` is now `type RawLesson`
and `type TimetableLesson` is now `type Lesson`.

Because there are only 2 places that used the first type.


v3/src/js/types/reducers.js
```diff
@@ -1,7 +1,38 @@
...
+  ModuleCondensed,
 } from 'types/modules';
 
+/* app.js */
 export type AppState = {
   activeLesson: ?Lesson,
 };
+
+/* requests.js */
+export type FetchRequest = {
```
yangshun: I thought it would be better to co-locate the types in the corresponding reducer files

v3/src/js/types/timetables.js
```diff
@@ -9,17 +9,19 @@ import type {
   Semester,
 } from 'types/modules';
 
+export type Lessons = Array<Lesson>;
```
yangshun: I contemplated doing this but I decided that there wasn't much value in it creating a new type just for an array of types. `Array<Lesson>` is more explicit and immediately obvious what it is. I'll have to look up some flow examples to see if this is an anti-pattern.

v3/src/js/utils/modules-search.js
```diff
@@ -22,7 +22,7 @@ export class ModulesSearchIndex {
     const query: string = tokens[0];
     let uidToDocumentMap: Object = this.tokenToUidToDocumentMap[query] || {};
 
-    for (let i = 1, numTokens = tokens.length; i < numTokens; i++) {
+    for (let i = 1, numTokens = tokens.length; i < numTokens; i += 1) {
```
yangshun: Actually why can't this be changed to `tokens.forEach(...)`?

v3/src/js/utils/modules.js
```diff
@@ -2,12 +2,12 @@
 
 import _ from 'lodash';
 import type {
+  RawLesson,
```
yangshun: Sort by alphabetical (:

v3/src/js/views/browse/ModulePageContainer.jsx
```diff
@@ -1,25 +1,46 @@
...
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps(nextProps: Props) {
     if (nextProps.routeParams.moduleCode !== this.props.routeParams.moduleCode) {
-      loadModuleInformation(nextProps);
+      this.loadModuleInformation(nextProps);
     }
   }
 
+  loadModuleInformation(props: Props) {
```
yangshun: It doesn't really matter in this case, but in general it's better to take functions out because it won't create an instance method for each instance of the component.

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -118,7 +145,7 @@ export class TimetableContainer extends Component {
               <br/>
               <TimetableModulesTable modules={
                 Object.keys(this.props.semTimetable).sort((a, b) => {
-                  return a > b;
+                  return a.localeCompare(b);
```
yangshun: Wow I nvr knew such a function exists.

v3/src/js/views/timetable/TimetableRow.jsx
```diff
@@ -1,24 +1,30 @@
...
 
-function generateCells(lessons, onModifyCell) {
-  const lessonsGroupedByStartTime = _.mapValues(_.groupBy(lessons, lesson => lesson.StartTime), value => value[0]);
+function generateCells(lessons: ?Lessons, onModifyCell: ?Function) {
+  const lessonToStartTimeMap: {[time: LessonTime]: Lesson} = _.mapValues(
+    _.groupBy(lessons, lesson => lesson.StartTime),
+    value => value[0]
+  );
 
   const cells = [];
```
yangshun: Missing type annotation for this?

v3/test/utils/timetable.js
```diff
@@ -41,6 +43,8 @@ import lessonsArray from '../mocks/lessons-array.json';
 function createGenericLesson(dayText: DayText, startTime: LessonTime,
                             endTime: LessonTime, lessonType?: LessonType, classNo?: ClassNo): Lesson {
   return {
+    ModuleCode: 'GenericCode',
```
yangshun: Make it realistic, `GC1101` 😋

v3/src/js/utils/modules-search.js
li-kai: Index is 1!

v3/src/js/views/timetable/TimetableRow.jsx
li-kai: Nope...if I add them, compiler throws error about `unexpected token`.

v3/src/js/types/reducers.js
li-kai: Then type folder will have as many layers as we have in `src/js`. I also think putting it in one file makes it clear which reducers do what.

v3/src/js/views/timetable/TimetableRow.jsx
yangshun: So it's a flow/babel bug?

v3/src/js/utils/modules-search.js
yangshun: oic. my bad

v3/src/js/types/reducers.js
yangshun: ok let's do that

v3/src/js/views/timetable/TimetableRow.jsx
li-kai: seems to me that you can't add flow type to single line array functions yet

PR #328, opened 2016-09-25
Timetable revamp

yangshun:
- Rewrite in (almost) non-flex approach. 
- Add a toggle orientation button.

### DIFF
--- a/v3/src/js/reducers/theme.js
+++ b/v3/src/js/reducers/theme.js
@@ -2,16 +2,20 @@
 
 import _ from 'lodash';
 import { ADD_MODULE, REMOVE_MODULE } from 'actions/timetables';
-import { CHANGE_THEME } from 'actions/theme';
+import { CHANGE_THEME, TOGGLE_TIMETABLE_ORIENTATION } from 'actions/theme';
 
 import type { FSA } from 'redux';
 import type { ColorIndex, ColorMapping, ThemeState } from 'types/reducers';
+
+export const VERTICAL: string = 'vertical';
+export const HORIZONTAL: string = 'horizontal';
 
 const defaultColorsState: ColorMapping = {};
 const defaultThemeState: ThemeState = {
   // Available themes are defined in `themes.scss`
   id: 'ocean',
   colors: defaultColorsState,
+  timetableOrientation: HORIZONTAL,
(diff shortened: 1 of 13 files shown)


v3/src/js/reducers/theme.js
```diff
@@ -2,16 +2,26 @@
...
+/* eslint-disable no-duplicate-imports */
+import type {
+  ColorIndex,
+  ColorMapping,
+  ThemeState,
+} from 'types/reducers';
+import {
+  VERTICAL,
+  HORIZONTAL,
+} from 'types/reducers';
```
li-kai: Just merge with the above? Why import twice? The constants should be ranked higher up too in any case.

v3/src/js/views/timetable/Timetable.jsx
```diff
@@ -1,43 +1,113 @@
 // @flow
 
+import type { DayText, LessonTime, RawLesson } from 'types/modules';
+import type { TimetableArrangement } from 'types/timetables';
+
 import React, { Component } from 'react';
```
li-kai: React and module dependencies should be imported first, then followed by types.

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -1,11 +1,28 @@
 // @flow
+/* eslint-disable no-duplicate-imports */
+import type {
```
li-kai: Types should really be closer to the code as it makes it easier to reference where the types come from.

v3/src/js/views/timetable/TimetableRow.jsx
```diff
@@ -40,16 +39,25 @@ function generateCells(lessons?: Array<Lesson>, onModifyCell?: Function) {
 }
 
 type Props = {
-  day: string,
+  cellSize: number,
+  cellOrientationStyleProp: string,
+  horizontalOrientation?: boolean,
+  proportion?: number,
```
li-kai: Would `scale` be better? `props.proportion` sounds like it could be confusing.

yangshun: Ok `scale` it is.

PR #329, opened 2016-09-26
Webpack it right

li-kai:
## Big changes
12.0.0 of airbnb-eslint-config dropped, so we could drop a few custom rules. One thing led to another, and I discovered we're weren't using `eslint-plugin-import` the right way.
Since we're doing absolute imports, we needed `eslint-import-resolver-webpack` too.

### Smaller ones

- `transform-es2015-destructuring` isn't needed because it's included in the [es2015 preset](https://github.com/babel/babel/blob/master/packages/babel-preset-es2015/src/index.js#L25).
- `js-search` is part of dependencies since virtualized-select need them, and I'll use them for search in the future too.
- `babel-polyfill` only needs to be injected either at `main.js` or in webpack. I removed the duplicate.
- `history` cannot be listed as dependency because current version is >4 and `react-router` is still at 2.11 something. Updating causes issues.


v3/webpack.config.js
```diff
@@ -14,24 +14,26 @@ const PATHS = {
   scripts: path.join(__dirname, SRC, 'js'),
   styles: path.join(__dirname, SRC, 'styles'),
   images: path.join(__dirname, SRC, 'img'),
-  build: path.join(__dirname, BUILD)
+  build: path.join(__dirname, BUILD),
 };
 const vendor = [
```
yangshun: Heh so you kept this in the end 😝

PR #335, opened 2016-09-29
Timetable improvement

yangshun:
- Some styling fixes
- Render timetable to earliest and latest hour bounds


v3/src/js/utils/timify.js
```diff
@@ -17,3 +20,20 @@ export function convertIndexToTime(index: number): LessonTime {
...
+
+export function calculateBorderTimings(lessons: Array<RawLesson>): { startingIndex: number, endingIndex: number } {
+  let earliestTime: number = convertTimeToIndex(DEFAULT_EARLIEST_TIME);
+  let latestTime: number = convertTimeToIndex(DEFAULT_LATEST_TIME);
+  lessons.forEach((lesson) => {
+    earliestTime = Math.min(earliestTime, convertTimeToIndex(lesson.StartTime));
+    latestTime = Math.max(latestTime, convertTimeToIndex(lesson.EndTime));
+  });
+  return {
+    startingIndex: earliestTime % 2 === 0 ? earliestTime : earliestTime - 1, // start at earliest hour
```
li-kai: It would be clearer with `// floor to earliest hour` and `// ceil to latest hour` instead.

v3/src/js/views/timetable/Timetable.jsx
```diff
@@ -1,41 +1,27 @@
 // @flow
 
-import type { DayText, LessonTime, RawLesson } from 'types/modules';
-import type { TimetableArrangement } from 'types/timetables';
+import type { RawLesson } from 'types/modules';
+import type { TimetableArrangement } from 'types/timetable';
 
-import React, { Component } from 'react';
 import _ from 'lodash';
+import React, { Component } from 'react';
```
li-kai: React should be imported before lodash.

v3/src/js/utils/timify.js
```diff
@@ -17,3 +20,20 @@ export function convertIndexToTime(index: number): LessonTime {
   const minute: string = (index % 2) === 0 ? '00' : '30';
   return (hour < 10 ? `0${hour}` : hour.toString()) + minute;
 }
+
+export const SCHOOLDAYS: Array<DayText> = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+export const DEFAULT_EARLIEST_TIME: LessonTime = '1000';
```
li-kai: This should be `'0800'` right?

v3/src/js/utils/timify.js
```diff
@@ -1,13 +1,16 @@
...
 // Converts a 24-hour format time string to an index.
 // Each index corresponds to one cell of each timetable row.
 // Each row may not start from index 0, it depends on the config's starting time.
 // 0000 -> 0, 0030 -> 1, 0100 -> 2, ...
 export function convertTimeToIndex(time: LessonTime): number {
-  return (parseInt(time.substring(0, 2), 10) * 2) + (time.substring(2) === '00' ? 0 : 1);
+  const hour: number = parseInt(time.substring(0, 2), 10);
+  const minute: string = time.substring(2);
+  /* eslint-disable quote-props */
+  return (hour * 2) + { '00': 0, '30': 1, '59': 2 }[minute];
```
li-kai: We should write a more general version. Better to do it now than when lessons with other timings get introduced.

v3/src/js/views/timetable/Timetable.jsx
li-kai: React should be imported before lodash.

v3/src/js/views/timetable/Timetable.jsx
```diff
@@ -47,9 +33,15 @@ class Timetable extends Component {
   props: Props;
 
   render() {
-    const { startingIndex, endingIndex } = calculateBorderTimings(this.props.lessons);
+    const lessons: Array<RawLesson> = [];
+    SCHOOLDAYS.forEach((day) => {
+      const lessonsArray: Array<RawLesson> = _.flatten(this.props.lessons[day]);
+      lessons.push(...lessonsArray);
```
li-kai: `lessons.concat(lessonsArray)`

v3/src/js/views/timetable/Timetable.jsx
```diff
@@ -74,11 +66,12 @@ class Timetable extends Component {
         'horizontal-mode': this.props.horizontalOrientation,
         'vertical-mode': !this.props.horizontalOrientation,
       })}>
+        <style>{`
+          .vertical-mode .timetable-inner-container { height: ${timetableHeight}px; }
```
li-kai: Can we use `em` or `rem` units instead? It'll help with zooming in and high density screens.

v3/src/js/utils/timify.js
yangshun: The more general version would mean every minute is 1 index :joy: How would you propose to generalize this?

v3/src/js/utils/timify.js
yangshun: As discussed, we'll change it to `1000` since we increase as needed. Starting at `1000` still covers close to 90% of lessons.

v3/src/js/utils/timify.js
yangshun: Great suggestion. 👌

v3/src/js/views/timetable/Timetable.jsx
yangshun: `concat` does not modify itself.

li-kai: Ah, right. I'm still surprised there's no method to append to an array in JS lol. Anyway it seems like the spread operator is right, although I'm still not used to it.

PR #337, opened 2016-10-01
Normalize timetable config format

yangshun:
- Changed to store `classNo` instead of copying the module lessons in timetable config
- Hydrate the timetable config with module lessons in `TimetableContainer` instead


v3/src/js/utils/timetables.js
```diff
@@ -34,20 +40,41 @@ export const LESSON_TYPE_ABBREV: LessonTypeAbbrev = {
...
+    return _.sample(group)[0].ClassNo;
+  });
+}
+
+// Replaces ClassNo in SemTimetableConfig with Array<Lesson>
+export function hydrateSemTimetableWithLessons(semTimetableConfig: SemTimetableConfig, modules: ModulesMap,
+                                                semester: Semester): SemTimetableConfigWithLessons {
+  return _.mapValues(semTimetableConfig, (moduleLessonConfig: ModuleLessonConfig, moduleCode: ModuleCode) => {
+    const module: Module = modules[moduleCode];
+    return _.mapValues(moduleLessonConfig, (classNo: ClassNo, lessonType: LessonType) => {
```
li-kai:
```
// only hydrate relevant lessons
return getModuleTimetable(module, semester).reduce((accumulated: Array<Lesson>, lesson: RawLesson) => {
  if (lesson.LessonType === lessonType && lesson.ClassNo === classNo) {
    accumulated.push({
      ...lesson,
      ModuleCode: moduleCode,
      ModuleTitle: module.ModuleTitle,
    });
  }
  return accumulated;
}, []);
```
would be clearer and more efficient

v3/src/js/types/timetables.js
```diff
@@ -9,19 +9,25 @@ import type {
   Semester,
 } from 'types/modules';
 
-//  LessonConfig is a mapping of LessonType to an array of Lessons.
+//  ModuleLessonConfig is a mapping of LessonType to ClassNo for a module.
+export type ModuleLessonConfig = {
```
yangshun: I try to avoid calling it map if I can. In this case we can have a convention to call it a config (and the underlying representation just happens to be a map). Remember that the previous structure wasn't really a `LessonTypeToClassNoMap`, it was `LessonTypeToLessonsMap`. This enables us to change the underlying representation without renaming the type.

v3/src/js/utils/timetables.js
yangshun: No doubt it is more efficient, but I would argue that doing a `filter` than a `map` is much clearer as each step/phase is distinct and intention is obvious. Well, the time complexity is the same and N isn't that large after I filter anws 👅

PR #339, opened 2016-10-04
Add dll compilation to development workflow

yangshun:


li-kai:
Looks super good to me 😂 hooray for fast build times

Issue #347, opened 2016-11-14
Fix bookmarks


burnflare:
I just tried bookmarking modules on both Safari and Chrome but the module doesn't actually get bookmarked.

<img width="141" alt="screen shot 2016-11-14 at 4 05 16 pm" src="https://cloud.githubusercontent.com/assets/104150/20257375/29193558-aa84-11e6-9a2b-030ee4f67f57.png">
<img width="189" alt="screen shot 2016-11-14 at 4 05 21 pm" src="https://cloud.githubusercontent.com/assets/104150/20257377/29e9d5fa-aa84-11e6-9b50-760f192a658d.png">

ngzhian:
fixed!

PR #348, opened 2016-11-29
Fix bookmarks #347

ngzhian:


yangshun:
Did this break after we upgraded localforage?

ngzhian:
very possible, someone the callback way doesn't work anymore, so changed it to promise based

yangshun:
Ok merge it!

Issue #317, opened 2016-09-13
Port module info page to v3


yangshun:
_From @yangshun on September 11, 2016 14:57_

https://nusmods.com/modules/CS1010S
- [x] Module Details section
  - [x] CORS Link & IVLE Link
  - [ ] FB Share buttons
  - [x] Semesters offered
  - [x] Semesters exam dates
  - [x] Add to timetable link
  - [ ] Bookmark
- [x] Prerequisites Tree section (KIV, currently broken)
- [x] Reviews & Discussions section
- [x] CORS Bidding History 
- [x] Lessons timetable for each semester

_Copied from original issue: nusmodifications/nusmods-v3#2_

yangshun:
We'll leave out bookmarks and share buttons.

PR #351, opened 2016-12-07
Pimp module

ngzhian:
Partially address #317, under module details section (i don't really want to add the like and share buttons cos i'm not a fan, so that can come later). The rest of the implementation will be in separate PRs.

Added the exam dates, semesters offered, official ivle and cors links, and add to sem/remove to sem buttons.

When module is not in timetable you can add it to the semesters offered:
screen shot 2016-12-08 at 1 12 47 am

If it is already in the timetable you can remove:
screen shot 2016-12-08 at 1 14 56 am


v3/src/js/utils/date.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+export function dateForDisplay(date: String): string {
```
yangshun: We already have a date display function called `formatExamDate` in `modules.js` that we use to display on the timetable page. Should use that for consistency sake.

v3/src/js/views/browse/ModulePageContainer.jsx
```diff
@@ -34,12 +42,68 @@ export class ModulePageContainer extends Component {
...
+        .sort((a, b) => a.Semester - b.Semester)
+        .map(h => ({ semester: h.Semester, date: h.ExamDate })))
+      : [];
+  }
+
+  moduleHasBeenAdded(module: Module, semester: num): boolean {
+    const timetables = this.props.timetables;
+    return (
+      timetables[semester] &&
+      typeof timetables[semester][module.ModuleCode] !== 'undefined'
```
yangshun: Would `!!timetables[semester][module.ModuleCode]` work?

v3/src/js/views/browse/ModulePageContainer.jsx
```diff
@@ -34,12 +42,68 @@ export class ModulePageContainer extends Component {
...
+
+    const semsOffered = this.semestersOffered()
+      .map(sem => `Semester ${sem}`)
+      .join(', ');
+
+    const addOrRemoveToTimetableLinks = this.semestersOffered().map(
+      semester => (
+        this.moduleHasBeenAdded(module, semester) ?
+          (
+            <RemoveModuleButton key={semester} semester={semester} onClick={() =>
```
yangshun: The parens around the component are redundant

v3/src/js/utils/date.js
ngzhian: I was looking for something like that and couldn't find :( Thanks for pointing this out!

v3/src/js/views/browse/ModulePageContainer.jsx
ngzhian: yes it would, thanks!

PR #353, opened 2016-12-08
Cors bidding stats and lessons timetable

ngzhian:
Show cors bidding stats :)

The design is non existent, we can fix in another PR :)

demo

screen shot 2016-12-09 at 11 00 00 am


yangshun:
Wow this looks great!!! Nice work

v3/src/js/views/browse/CorsBiddingStatsTableControl.jsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+const sameFaculty = (stat: BiddingStat, student: Student): boolean => (stat.Faculty === student.faculty);
+const isNew = student => (student.newStudent);
+const PROGRAMME = 'P';
+const GENERAL = 'G';
+const isProgrammeAccount = student => (student.accountType === PROGRAMME);
+const isGeneralAccount = student => (student.accountType === GENERAL);
+
+function isStatRelevantForStudent(stat, student) {
```
yangshun: Would be better to extract these helper/util functions out into `cors-utils.js`

PR #357, opened 2016-12-11
Export as jpeg

ngzhian:
Copied shamelessly from modify.sg, export timetable as jpeg using dom-to-image. Currently lives in a button next to the orientation rotation button.

screen shot 2016-12-11 at 7 42 58 pm

results:
timetable 1

for reference, the timetable-export branch:
nusmods 1

### DIFF
--- a/v3/package.json
+++ b/v3/package.json
@@ -74,6 +74,7 @@
     "bootstrap": "^4.0.0-alpha.4",
     "classnames": "^2.2.5",
     "d3": "^4.4.0",
+    "dom-to-image": "^2.5.2",
     "js-search": "^1.3.2",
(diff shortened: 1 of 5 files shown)


ngzhian:
@xbili it seems like when i cherry pick ur changes, the timetable doesn't have the time, and it's also cut off due to it being fixed width. any clue why?
do you think we can go with the implementation in this PR? I think the image output looks more complete.

xbili:
@ngzhian Hmm I think I exported the wrong `div` last time, hence the missing time. Fixed width was implemented to prevent the JPEG from being cut off when we resize the window.

But this implementation looks much easier. I think can go ahead with this!

PR #361, opened 2016-12-29
Add workshop as lesson type

### LINKED ISSUE #360: Some module types unsupported in URL
I noticed this while loading my own NUS mods timetable.

For example, for the module CFG1010, it appears in the URL like this:

`https://nusmods.com/...&CFG1010[undefined]=7`

This leads to the CFG module appearing in the timetable, but more often than not in the wrong timeslot.

ngzhian:
Fixes #360

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,17 @@ language: node_js
 node_js:
   - 6.6
 before_install:
+  # Repo for Yarn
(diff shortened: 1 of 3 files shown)


yangshun:
Thanks, can help add to v3 in a separate PR too?

PR #367, opened 2017-01-02
[v3] Ical export

ngzhian:
I don't think i support the full subset of what nusmods currently supports, but from my testing it seems sufficient.
supported features: odd/even weeks, comma separated weeks, every week, tutorials excluded from weeks 1,2, and 3, exams.


ngzhian:
I've just tried this for my y4s2 timetable, this works fine, so will merge first :)

Issue #312, opened 2016-09-13
Port team page to v3


yangshun:
_From @yangshun on September 11, 2016 15:13_

https://nusmods.com/team

_Copied from original issue: nusmodifications/nusmods-v3#7_

yangshun:
@zameschua is interested in this.

FireLord1:
@yangshun all this for dear bt

PR #375, opened 2017-01-10
Port team page to v3 (#312)

zameschua:
First time working with eslint so not sure how to resolve these issues
image

### DIFF
--- a/v3/src/js/views/static/TeamContainer.jsx
+++ b/v3/src/js/views/static/TeamContainer.jsx
@@ -1,93 +1,7 @@
 import React from 'react';
 import DocumentTitle from 'react-document-title';
 
-type Props = {
-  name: string,
-  description: string,
-  link: string,
-  gravatar: string,
-  gravatarAlt: string,
-  facebook: string,
-  twitter: string,
-  github: string,
-  linkedin: string,
-}
-
-const teamInfo = [
-  {
-    name: 'Beng (Eu Beng Hee)',
-    description: 'He\'s Beng. Nuff\' said.',
-    link: 'https://www.github.com/ahbeng',
-    gravatar: 'https://www.gravatar.com/avatar/601eddec66c2e48891195af393144c17?s=256',
-    gravatarAlt: 'Beng',
-    facebook: 'https://www.facebook.com/benghee',
-    twitter: 'https://www.twitter.com/ahbengish',
-    github: 'https://www.github.com/ahbeng',
-    linkedin: 'https://www.linkedin.com/in/benghee',
-  },
-  {
-    name: 'Tay Yang Shun',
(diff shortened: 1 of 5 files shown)


yangshun:
Nice first attempt! However, we want to separate the data from the view. The approach should follow V2's, by importing a [JSON file](https://github.com/nusmodifications/nusmods/blob/master/app/scripts/about/team.json) and rendering using a [template](https://github.com/nusmodifications/nusmods/blob/master/app/scripts/about/templates/team.hbs). A for loop should iterate through the team data and render each row according to each member's data.

Reason for this is so that if we want to change the styling, we can easily change the styling for all the rows at once, rather than going to each row and making the same changes N times.

zameschua:
Updated as advised.
I'm not so familiar with react structure but I tried my best to follow the other files in the doc haha

yangshun:
Looks much better! You're on the right track (:

zameschua:
Shifted data to js/data/team.json
removed the extra component
added conditional rendering for social icons

Thanks!!

v3/src/js/views/static/TeamContainer.jsx
```diff
@@ -0,0 +1,61 @@
...
+                </div>
+                <div className="col-sm-9 col-xs-12">
+                  <h3>{teamMember.name}</h3>
+                  <p>{teamMember.description}</p>
+                  <div className="row">
+                    {teamMember.facebook ?
+                      <div className="col-sm-1 col-xs-3">
+                        <a href={teamMember.facebook}><i className="fa fa-facebook-square fa-lg" /></a>
+                      </div>
+                    : false}
```
yangshun: Returning `false` is not very conventional although it works. The two common ways of conditional rendering.

1. `{condition: <Component> : null}`
2. `{condition && <Component>`

The full list can be found in [React docs](https://facebook.github.io/react/docs/conditional-rendering.html). In v3 we tend to lean towards approach 2 above (:

v3/src/js/views/static/TeamContainer.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import DocumentTitle from 'react-document-title';
+
+const teamInfo = require('../../data/team.json');
```
yangshun: Use the absolute paths so we don't have to change the import path if we change the location of the `TeamContainer.jsx` file. Also, use `import` instead of `require` to be more es6-ish.

`import teamInfo from 'data/team.json';`

zameschua:
Updated as advised. Thank you so much for guiding me!

yangshun:
Closes #312

yangshun:
@zameschua Thanks for the effort and being so patient with our code reviews (:

PR #404, opened 2017-06-25
Hot reload 3

li-kai:
Update to hot reload 3. Carousell has been using this for months so it's safe :D


v3/package.json
```diff
@@ -88,6 +87,7 @@
     "react-disqus-thread": "^0.4.0",
     "react-document-title": "^2.0.3",
     "react-dom": "^15.5.4",
+    "react-hot-loader": "next",
```
yangshun: Should not be in dependencies. I noticed it because `v3/flow-typed/package-dep-libdefs.js` was updated :P

yangshun:
Just merge it after fixing it.

li-kai:
@yangshun It's alright :X react-hot-reload gives noop based on NODE_ENV. I'll still move it.

yangshun:
Yeah but if we did stuff like automatically creating a vendor bundle out of `pkg.dependencies`, it may be included in it. Not sure if tree shaking would affect anything but just in case.

v3/package.json
```diff
@@ -62,6 +61,7 @@
     "postcss-loader": "^2.0.5",
     "react-addons-test-utils": "^15.5.1",
     "react-dev-utils": "^0.5.2",
+    "react-hot-loader": "next",
```
yangshun: Don't we wanna lock down to a specific minor version at least?

PR #409, opened 2017-07-22
Port over Developers page from v2

### LINKED ISSUE #309: Port developers page to v3
_From @yangshun on September 11, 2016 16:38_

https://nusmods.com/contribute/developers

Should also integrate the content from https://nusmods.com/contribute into this page, minus the Reviewers part. 

_Copied from original issue: nusmodifications/nusmods-v3#11_

zameschua:
EDIT by @yangshun: Fixes #309


v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,73 @@
...
+    this.state = {
+      developersData: null,
+      isLoading: true,
+    };
+  }
+
+  componentDidMount() {
+    const self = this;
+
+    fetch('https://api.github.com/repos/NUSModifications/NUSMods/contributors').then(response => (
```
li-kai: Please use axios package instead of fetch, as fetch isn't supported in all browsers yet. (This would fail in safari for example)

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,73 @@
...
+      response.json()
+    )).then((json) => {
+      self.setState({
```
li-kai: Since you're using [arrow functions](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions) here, `this.setState` would work.

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,73 @@
...
+        developersData: json,
+        isLoading: false,
+      });
+    });
+  }
+
+  render() {
+    let renderComponent = '';
+
+    if (this.state.isLoading) {
```
li-kai: Could you inline this with ternary operators `this.state.isLoading ? ... : ...`

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,71 @@
...
+        developersData: response.data,
+        isLoading: false,
+      });
+    });
+  }
+
+  render() {
+    let renderComponent = '';
+
+    this.state.isLoading ? (
```
li-kai: Oh woops, I didn't mean it this way. Take a look at how [TimetableDay.jsx](https://github.com/nusmodifications/nusmods/blob/22cf594c7e5c2438f64fb6e64efffb3c21f357bb/v3/src/js/views/timetable/TimetableDay.jsx) does it.

angelsl:
I noticed this hits the GitHub API directly from the user's browser.

It's unlikely it'll ever happen to a normal user, but maybe it'd be good to handle the possibility of a 403 Forbidden if somehow the user has exhausted their API rate limit.

zameschua:
There's a no-nested-ternary on eslint though, so should I switch it back to if, else if, else loops?

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react';
```
yangshun: Do you know how to add flowtypes? Else add a `// TODO: Add Flow here` and someone else will take care of it.

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,64 @@
...
+import DocumentTitle from 'react-document-title';
+import axios from 'axios';
+
+const DEVELOPERS_URL = 'https://api.github.com/repos/NUSModifications/NUSMods/contributors';
+
+class DevelopersContainer extends Component {
+  constructor() {
+    super();
```
yangshun: The standard is:

```
constructor(props) {
  super(props);
  ...
}
```

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,64 @@
...
+            <br /><br />
+            Please reach out to us if you are interested in helping!
+            Join us and make NUS a better place for its students (your friends)!
+            </em></p>
+            <br /><br />
+
+            <div className="row">
+              {this.state.isLoading ?
+                <div className="col-2 offset-md-5">
+                  <i className="fa fa-circle-o-notch fa-spin" style={{ fontSize: '50px' }} />;
```
yangshun: Use `rem` for `font-size`.

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,64 @@
...
+  constructor() {
+    super();
+    this.state = {
+      developersData: null,
+      isLoading: true,
+    };
+  }
+
+  componentDidMount() {
+    axios.get(DEVELOPERS_URL).then((response) => {
```
yangshun: Currently if we exceed the API request limit or Github is down for some reason, the `then` method will never get executed and the spinner will be spinning forever.

Hence, add a `Promise.catch` here and set `isLoading: false` with an error message. You can do something like:

```
axios.get(DEVELOPERS_URL)
.then(response => { 
  this.setState({
    developersData: response.data,
    isLoading: false,
  });
})
.catch(err => { 
  this.setState({
    isError: true, // Show error message on UI. You might want to see what the error in `err` is and render it.
    isLoading: false,
  });
});
```

Alternatively, refactor into:

```
axios.get(DEVELOPERS_URL)
.then(response => { 
  this.setState({
    developersData: response.data,
  });
})
.catch(err => { 
  this.setState({
    isError: true, // Show error message on UI. You might want to see what the error in `err` is and render it.
  });
})
.then(response => { 
  // This block will always get executed regardless of request outcome.
  this.setState({
    isLoading: false,
  });
})
```


https://gist.github.com/jish/e9bcd75e391a2b21206b

v3/src/js/views/static/DevelopersContainer.jsx
zameschua: I know how to use flowtypes but I'm not sure about the convention, ie. which variables require a type checking. Generally I see props / arguments of functions being type-checked but I did not inherit any props from the parent though, and I didn't define any functions in this page.

yangshun: I think it's fine for this page. Can ignore flow

v3/.eslintrc.js
```diff
@@ -58,5 +58,6 @@ module.exports = {
     'jsx-a11y/no-static-element-interactions': 'off',
     // Let git handle the linebreaks instead.
     'linebreak-style': 'off',
+    'no-nested-ternary': 'off',
```
yangshun: Try not to touch the ESLint rules.

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,86 @@
...
+            of its individual contributors and NUS student community. Many student hackers have
+            reported issues, suggested improvements, or even better, write code and contribute patches!
+            <br /><br />
+            Please reach out to us if you are interested in helping!
+            Join us and make NUS a better place for its students (your friends)!
+            </em></p>
+            <br /><br />
+
+            <div className="row">
+              {this.state.isLoading ?
```
yangshun: One way to write the logic here without using nested ternary is:

```
{this.state.isLoading && <Spinner component>}
{this.state.isError && <Error message component>}
{!this.state.isLoading && !this.state.isError && <Success component>}
```

Via using the `&&` operator: https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator

v3/src/js/views/static/DevelopersContainer.jsx
```diff
@@ -0,0 +1,86 @@
...
+    axios.get(DEVELOPERS_URL)
+    .then((response) => {
+      this.setState({
+        developersData: response.data,
+      });
+    })
+    .catch((err) => {
+      this.setState({
+        isError: true,
+        errorMessage: err.message,
```
yangshun: Did you manage to try getting an error and seeing the error payload contains a `message` key?

zameschua: Nope, but I searched online for Github's rror object schema and it seems to contain the field. Should I keep pinging the api and try to induce the error?

yangshun: I think that's good enough (:

v3/src/js/views/static/DevelopersContainer.jsx
zameschua: Damnn I never knew about this

PR #410, opened 2017-07-22
Enable escape button shortcut to close colorpicker

### LINKED ISSUE #350: Escape button shortcut to close colorpicker
Add escape button shortcut to close colorpicker for modules.

zameschua:
Attempts to fix #350 
I had to add an event listener to window, because I can't add a 'onKeyDown' handler to components that can't be focused (div and span in colorpicker).

### DIFF
--- a/v3/src/js/views/components/color-picker/ColorPicker.jsx
+++ b/v3/src/js/views/components/color-picker/ColorPicker.jsx
@@ -10,18 +10,24 @@ import './color-picker.scss';
 
 type Props = {
   onChooseColor: Function,
-  handleEscKeyDown: Function,
+  onDismiss: Function,
 };
 
 class ColorPicker extends Component {
   props: Props;
 
   componentWillMount() {
-    window.addEventListener('keydown', this.props.handleEscKeyDown);
+    window.addEventListener('keydown', this.handleKeyDown);
   }
 
   componentWillUnmount() {
-    window.removeEventListener('keydown', this.props.handleEscKeyDown);
+    window.removeEventListener('keydown', this.handleKeyDown);
+  }
+
+  handleKeyDown = (event) => {
+    if (event.key === 'Escape') {
+      this.props.onDismiss();
+    }
   }
 
   render() {
--- a/v3/src/js/views/components/color-picker/ColorPicker.jsx
+++ b/v3/src/js/views/components/color-picker/ColorPicker.jsx
@@ -8,6 +8,8 @@ import { NUM_DIFFERENT_COLORS } from 'reducers/theme';
 
 import './color-picker.scss';
 
+const ESCAPE_KEYCODE = 27;
+
 type Props = {
   onChooseColor: Function,
   onDismiss: Function,
@@ -25,7 +27,7 @@ class ColorPicker extends Component {
   }
 
   handleKeyDown = (event) => {
-    if (event.key === 'Escape') {
+    if (event.key === 'Escape' || event.keyCode === ESCAPE_KEYCODE) {
       this.props.onDismiss();
     }
   }
--- a/v3/src/js/views/components/color-picker/ColorPicker.jsx
+++ b/v3/src/js/views/components/color-picker/ColorPicker.jsx
@@ -1,6 +1,6 @@
 // @flow
 /* eslint-disable react/prefer-stateless-function */
-import React from 'react';
+import React, { Component } from 'react';
 import type { ColorIndex } from 'types/reducers';
 
 import _ from 'lodash';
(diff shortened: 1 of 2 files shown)


yangshun:
Please provide a short description in your PRs. Minimally describing which issue this PR attempts to fix.

v3/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -28,6 +28,11 @@ type Props = {
 class TimetableModulesTable extends Component {
   props: Props;
 
+  constructor() {
```
yangshun: Remember the `constructor` takes in `props`

v3/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -28,6 +28,11 @@ type Props = {
...
+    super();
+    this.handleEscKeyDown = this.handleEscKeyDown.bind(this);
```
yangshun: I believe we have `autobind` installed, so can just `import` it and do `autobind(this)`.

v3/src/js/views/components/color-picker/ColorPicker.jsx
```diff
@@ -10,24 +10,37 @@ import './color-picker.scss';
...
-              onClick={() => {
-                props.onChooseColor(index);
-              }} />
-          );
-        })}
+class ColorPicker extends Component {
+  props: Props;
+
+  componentWillMount() {
+    window.addEventListener('keydown', this.props.handleEscKeyDown);
```
yangshun: If the `prop` is called `handleEscKeyDown`, then the check for the escape key should be placed inside here. Think about the responsibilities of each component. Should the parent component be in charge of deciding which key is pressed? I doubt so. The parent component should only be in charge of closing the `ColorPicker`. The `ColorPicker` owns the logic for deciding which key is to be pressed.

I would recommend to change the prop on `ColorPicker` to `onDismiss`, and the parent component simply has to pass in `this.cancelModifyModuleColor` to `onDismiss`. (using this is better instead of the prop because it checks for active module before firing the prop.

The `if (e.key === 'Escape') {` line should be shifted into `ColorPicker`.

v3/src/js/views/timetable/TimetableModulesTable.jsx
yangshun: I think you don't need the constructor anymore if you pass in `this.cancelModifyModuleColor`. `this.cancelModifyModuleColor` is defined using class property syntax which is "automatically" binded to the instance itself.

yangshun:
> I can't add a 'onKeyDown' handler to components that can't be focused (div and span in colorpicker).

You can always add `tabindex` attribute/prop to a component to allow it to be focused. But in this case, adding to the `window` is the right thing to do.

v3/src/js/views/components/color-picker/ColorPicker.jsx
zameschua: Ohh when I was learning, I  learnt that all the logic should always be placed in the parent containers, then the display components should try not to contain any logic, which is why I did it like this.

But your logic of assigning responsibilities to different components makes more sense.

v3/src/js/views/timetable/TimetableModulesTable.jsx
li-kai: Nope autobind is removed, but class properties was added instead.

```
// no need for constructor binding, just change your class method to this
handleEscKeyDown = () => {
}
```

yangshun: Ok got it. Good to have multiple pairs of eyes on a PR (:

v3/src/js/views/components/color-picker/ColorPicker.jsx
yangshun: So there are two approaches here, and I think both are fine (or at least I can't really decide which is a clearly better approach):

1. Attach event listener in parent container, check for escape key press in parent container.
2. Attach event listener in children component, check for escape key press within children component.

What you were doing was to attach event listener in children component and check for escape key press within parent container. This is not ideal.

One litmus test you can do is to consider if you had to remove this keyboard functionality, how many places (files/components) do you have to change to completely remove it? Ideally, just in one place.

v3/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -38,6 +43,12 @@ class TimetableModulesTable extends Component {
     }
   }
 
+  handleEscKeyDown(e) {
+    if (e.key === 'Escape') {
```
yangshun: Oh please avoid single variable names unless it's for iterators. I don't know the difference between `event.key` and `event.keyCode`. Do look it up and see if there are cross-browser issues with either.

v3/src/js/views/timetable/TimetableModulesTable.jsx
zameschua: Cool ok I just learnt about the class property syntax. So generally we should avoid `.bind(this)?` 
And use class property syntax (arrow function methods) only when we want to avoid `.bind(this)` and use the normal definition otherwise?

v3/src/js/views/timetable/TimetableModulesTable.jsx
zameschua: W3Schools seems to recommend using the `event.key` property, and there doesn't seem to be any compatibility issues. Should I do `if (event.keyCode === 27 || event.key = 'Escape')` instead?
image

yangshun: Ok that works. Do `const ESCAPE_KEYCODE = 27;` to avoid magic numbers.

v3/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -82,7 +82,8 @@ class TimetableModulesTable extends Component {
                     {this.props.activeModule === module.ModuleCode &&
                       <ColorPicker onChooseColor={(colorIndex: ColorIndex) => {
                         this.props.selectModuleColor(module.ModuleCode, colorIndex);
-                      }} />
+                      }}
+                        onDismiss={this.props.cancelModifyModuleColor} />
```
yangshun: Should be `onDismiss={this.cancelModifyModuleColor}`

v3/src/js/views/components/color-picker/ColorPicker.jsx
```diff
@@ -1,33 +1,54 @@
 // @flow
 /* eslint-disable react/prefer-stateless-function */
```
yangshun: I think you can remove this line

PR #414, opened 2017-07-27
Fix venue casing and identifier

li-kai:
1) Casing was changed when upgrading to new gulp api
2) Empty string is now a venue (I'll investigate)


li-kai:
@yangshun Yeah but I think it's to better to be consistent with the rest of all other APIs which are upper cased.

PR #417, opened 2017-08-02
Refactor/colocate tests

li-kai:
As per title. No functionality changes.

Changed a few rules, because some of the indent rules were conflicting with each other and the end result looked nicer too imo :X


v3/.eslintrc.js
```diff
@@ -22,6 +21,22 @@ module.exports = {
       },
     },
   },
+  overrides: [
+    {
+      files: '**/*.test.{js,jsx}',
+      excludedFiles: '**/!(*.test).{js,jsx}',
```
yangshun: These `excludedFiles` seem to be the complement of the included `files`. I wonder if both are still needed?

li-kai: Indeed they do not need to coexist

PR #419, opened 2017-08-03
Implement favicon support

li-kai:
Generated from https://realfavicongenerator.net/

Opted to not use .ico for now. Browser support is there, and we don't even plan to support IE. Saves us about 10kb.


v3/webpack/webpack.config.dev.js
```diff
@@ -39,17 +40,18 @@ const developmentConfig = merge([
...
+      new webpack.WatchIgnorePlugin([parts.PATHS.node, parts.PATHS.build]),
       // Enable multi-pass compilation for enhanced performance
       // in larger projects. Good default.
       // Waiting on: https://github.com/jantimon/html-webpack-plugin/issues/533
-      new webpack.HotModuleReplacementPlugin({
-        // multiStep: true,
-      }),
+      new webpack.HotModuleReplacementPlugin(
+        {
+          // multiStep: true,
```
yangshun: Is this still needed? If not can remove?

v3/webpack/webpack.config.prod.js
```diff
@@ -28,6 +29,13 @@ const productionConfig = merge([
         title: 'NUSMods',
         template: path.join(parts.PATHS.app, 'index.html'),
       }),
+      // Copy files from static folder over to dist
+      new CopyWebpackPlugin(
+        [
+          { from: 'static' },
+        ],
+        { copyUnmodified: true },
```
yangshun: Trailing commas in a function call is a syntax error in node < 8. Let's do without it if we can.

v3/webpack/webpack.config.dev.js
li-kai: Yep still needed, issue not fixed yet.

yangshun:
:shipit:

PR #421, opened 2017-08-08
Add aria-label to buttons with Font Awesome icons for accessibility

ZhangYiJiang:
## Context 

Font Awesome is used on the page to display icons in some buttons. However some of these icons are included without any text label, which means screen readers will not be able to discern what the icons mean 

## Approach 

In line with Font Awesome best practices, if the icons are semantic (ie. they have meaning), and if there are no text labels, we add ARIA text labels using `aria-label` attributes. We also apply `title` attributes to create tooltips. 

## Reference 

http://fontawesome.io/accessibility/

### DIFF
--- a/v3/src/js/views/static/DevelopersContainer.jsx
+++ b/v3/src/js/views/static/DevelopersContainer.jsx
@@ -56,7 +56,8 @@ class DevelopersContainer extends Component {
             <div className="row">
               {this.state.isLoading &&
                 <div className="col-12 text-center">
-                  <i className="fa fa-circle-o-notch fa-spin" style={{ fontSize: '4rem' }} />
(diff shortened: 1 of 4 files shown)


yangshun:
Thanks @ZhangYiJiang ! a11y is definitely something we want to pursue after we attain feature parity. Created #422 to track this.

ZhangYiJiang:
There's a bit of repetition in the code - it might be a good idea to create a separate `IconButton`  component here. I *believe* props will also enforce devs to always include a label. I'm not sufficiently familiar with React to do this though 😅

yangshun:
I get that there's some repetition, but I'm not a fan of tiny components, and imo wrapping them up into components don't add much value. These buttons are mainly just markup and do not contain much logic and in the event we want to customize it, it becomes tougher. I'd use components to encapsulate larger chunks of markup that require testing.

Issue #398, opened 2017-05-15
Attain > 90 for webpage performance metrics


yangshun:
- Pagespeed - https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fv3.nusmods.com%2F&tab=mobile
- Lighthouse - https://developers.google.com/web/tools/lighthouse/
- Yslow - http://yslow.org/

zameschua:
Interested, but will probably require a lot of guidance for this! 

# Report from PageSpeed Insights
![image](https://user-images.githubusercontent.com/18271085/28496094-e6721540-6f94-11e7-9f78-f3e0fe91a98d.png)
![image](https://user-images.githubusercontent.com/18271085/28496095-f1dbe0dc-6f94-11e7-809d-756d0658ae4c.png)

#Report from lighthouse

## Progressive Web-appness, related #319 
![image](https://user-images.githubusercontent.com/18271085/28496070-423790d6-6f94-11e7-90e8-9c881eaa679c.png)

## Performance
![image](https://user-images.githubusercontent.com/18271085/28496081-af1678de-6f94-11e7-9cf4-31846a55905d.png)
![image](https://user-images.githubusercontent.com/18271085/28496086-c80ffde2-6f94-11e7-9c0c-479b402d4c38.png)

## Accessibility
![image](https://user-images.githubusercontent.com/18271085/28496094-e6721540-6f94-11e7-9f78-f3e0fe91a98d.png)

zameschua:
Will look into integrating Webpack code splitting first

Issue #319, opened 2016-09-14
Make NUSMods into a Progressive Web App


yangshun:
Thanks @ngzhian for sharing this: https://infrequently.org/2016/09/what-exactly-makes-something-a-progressive-web-app/

Some other articles: https://medium.com/@owencm/designing-great-uis-for-progressive-web-apps-dd38c1d20f7#.h9vyvzq9n

ngzhian:
https://github.com/GoogleChrome/lighthouse will be helpful too :)

li-kai:
We're scroing 100% on pwa metric now. We pwa now.

PR #424, opened 2017-08-10
Add router-level code splitting using react-loadable

zameschua:
Working on #398 #319

Slightly improves lighthouse report
Lighthouse report can be further improved by following the PRPL pattern

Before:
image
image

After:
image
image


v3/package.json
```diff
@@ -76,6 +76,7 @@
   },
   "dependencies": {
     "axios": "^0.16.2",
+    "babel-plugin-syntax-dynamic-import": "^6.18.0",
```
yangshun: Should be in `devDependencies`.

v3/src/js/views/Loader.jsx
```diff
@@ -0,0 +1,11 @@
+// @flow
+import React from 'react';
+
+export default function Loader() {
+  return (
+    <div className="col-12 text-center">
```
yangshun: `col` classes must always be wrapped in a `row`, hence it's not a good idea to leave it the parent to ensure that they have a `.row` container. Should add it here. Then again, you don't need a `.row` and `.col` if you're just doing `col-12`. Simple `div` would suffice.

v3/flow-typed/package-dep-libdefs.js
```diff
@@ -1,9 +1,17 @@
 // This file is autogenerated by flow-scripts.
 
+declare module 'babel-plugin-syntax-dynamic-import' {
```
yangshun: After you remove `babel-plugin-syntax-dynamic-import` from `dependencies`, and regenerate this file, this line shouldn't be around.

ZhangYiJiang:
I'm not sure if this is a good idea. This change doesn't seem like it would significantly improve loading time. 

Before: 

```
  203.23 KB     build/vendor.773dbe967b0dc197a3b5.js
  36.52 KB      build/app.2664b4abcacf976da80e.js
```

After: 

```
  203.31 KB  build/vendor.353a389311bffb84d45f.js
  19.79 KB   build/app.2413c47a1a60fcbb164b.js
  10.74 KB   build/3ff70d6b3349cc40471c.js
  6.6 KB     build/3fa02c33b490d2c0c301.js
  1.97 KB    build/0ece4f1a18e432c95372.js
  1.95 KB    build/7c8b75a172bd34d84758.js
  1.79 KB    build/1b4b954b236da1735b8d.js
  1.58 KB    build/068d14c40c31a34be353.js
  1.18 KB    build/8481ed07e8913d1f2c9e.js
  817 B      build/97851965c97bae9fc66e.js
  482 B      build/17a6b53c45fad545ae63.js
```

(CSS and manifest omitted for clarity)

Notice all we've done is split the 36.5 kb app.js down to a bunch of smaller files. The problem is as a percentage of total assets, `app.js` is relatively small. Trimming it further would likely not improve load time, but instead annoy the user as there's now a spinner whenever a new page is loaded for he first time. 

A better option would be to look at `vendor.js`. If you use source-map-explorer, you can see that d3.js takes up a huge amount of space. 

screenshot from 2017-08-11 02-11-59

Since d3 is only used for module details page (dependency graph), selectively loading it *would* make a big impact. (It's probably also a good idea to try to import d3 modules more selectively to allow unused parts of the library to be shaken away, but that's for another time).

yangshun:
I think with code splitting we can exclude some stuff out of vendor bundle. @ZhangYiJiang if you have time would you mind trying that out and see if we can reduce the size of vendor bundle and lazy load d3?

ZhangYiJiang:
Yes, definitely. Sorry my comment wasn't suppose to imply that code splitting as a whole is useless, just that the current approach of splitting the pages is likely to be ineffective. I've actually done vendor library splitting for 3216 before (for Pinpoint.me, the AR portion didn't work on iOS, so we only loaded it on demand for Android users, shaving about 1MB off the original bundle size) - this should be similar.

yangshun:
> Sorry my comment wasn't suppose to imply that code splitting as a whole is useless, just that the current approach of splitting the pages is likely to be ineffective.

No problems, I understand. The groundwork @zameschua has done will nevertheless be useful for enabling the better code splitting approach later on. I personally haven't delved into this much and the project would benefit from your experience (;

v3/.eslintrc.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
   parser: 'babel-eslint',
+  parserOptions: {
```
li-kai: I tried in out locally and removing these additions don't affect anything. Looking through stackoverflow it seems this was used to fix eslint at one point of time.

li-kai:
Nice work. All's left is `git rebase -i upstream/master` (replace upstream with what you named our repo). Replace all but the oldest one with `fixup`, so you are only left with one commit. Remember to remove `/dll` too.

zameschua:
@li-kai Sorry I thought rebasing would be an automatic thing HAHAHA ok on it

zameschua:
**Before
  203.21 KB  build\vendor.82eb0beb7d45e3c82e61.js <-- Contains all vendors including d3.js
  32.98 KB   build\app.6528e664b0f75fcd8393.js
  6.09 KB    build\d314bc55a85689d4a07a.js <-- Contains ModulePageContainer

**After
  127.15 KB  build\vendor.99ffcd2921460b9cbd0c.js <-- Contains all vendor bundles excluding d3.js
  81.96 KB   build\efda2075a88e39938049.js <-- Contains d3.js and ModulePageContainer
  32.97 KB   build\app.3fa6788e70de21d0d011.js

(CSS and Manifest omitted for clarity)

yangshun:
This is great Zames! Perfect! (:

PR #425, opened 2017-08-11
Browse modules page module filters

ZhangYiJiang:
### Context 

v3's http://v3.nusmods.com/modules page is currently a little bare. This PR attempts to bring the page to feature parity with https://nusmods.com/modules. 

### Approach 

This is a clean implementation without referencing how the original code looks. Comments welcome! 

Filters are represented as `Filter` objects which are simple wrappers around a predicate (a function that returns true/false).

```js
class Filter {
  enabled = false;
  label: string;
  fn: (Module) => boolean;
}
```

The predicate `fn` returns if the module is selected. More specialized filters subclass `Filter`. Here's an example for filtering modules according to their level (1000, 2000, etc.)

```js
export type ModuleLevel = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;

class LevelFilter extends Filter {
  level: ModuleLevel;

  constructor(level: ModuleLevel) {
    super(`${level}000`, (module) => {
      const match = module.ModuleCode.match(/(\d+)/);
      if (!match) return false;
      return match[1][0] === String(level);
    });

    this.level = level;
  }
}
```

We want specialized subclasses so that additional information can be made available to views. The original UI uses a plain list of checkboxes, but this is suboptimal. Timeslots for instance can be displayed in a grid instead 

```
           Mon Tue
          ---------
Morning   |   |   |
          ---------
Afternoon |   |   |
          ---------
Evening   |   |   | ...
          ---------
```

For this UI to work, the Filter object needs to have some information about which timeslot it represents. This is provided by the subclass. 

```js
class TimeslotFilter extends Filter {
  semester: Semester;
  type: TimeslotType;
  timeslot: string; // May be better to split into day + time instead 
}
```

After the filters have been selected, we can simply run the modules through all filters and display the ones that pass any of them. Inheritance works here because all filter subclasses can be substituted for each other when used to, well, filter stuff, thus LSP is obeyed.


yangshun:
This is amazing @ZhangYiJiang . I'll review this over the weekend. Thanks for your contributions!

v3/src/js/utils/filters/Filter.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import type { Module } from 'types/modules';
+
+export default class Filter {
+  enabled = false;
+
+  label: string;
+  predicate: (Module) => boolean;
```
yangshun: Just checking, are the parens required?

v3/src/js/utils/filters/Filter.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import type { Module } from 'types/modules';
+
+export default class Filter {
```
yangshun: I think it would be clearer to have a more specific name for the filter. How about `ModuleFilter`?

v3/src/js/utils/filters/LevelFilter.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import Filter from './Filter';
+
+export type ModuleLevel = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
```
yangshun: I think this would be better inside `types/modules.js`

v3/src/js/utils/filters/TimeslotFilter.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import config from 'config/app-config.json';
+import { getModuleSemesterData } from '../modules';
```
yangshun: We generally don't want to import from ancestor paths. You can just import from `utils/modules` as we have configured `src/js` as one of the import paths webpack resolves.

v3/src/js/utils/filters/TimeslotFilter.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+import TimeslotFilter, { TimeslotTypes } from './TimeslotFilter';
+
+test('predicate should filter modules according to their lecture timeslot', () => {
+  // Generate all possible combinations of parameters to test against
+  const days = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday'];
+  const times = ['Morning', 'Afternoon', 'Evening'];
+  const semesters: Array<Semester> = [1, 2, 3, 4];
+  const types = Object.keys(TimeslotTypes);
+  const possibleArgs = Combinatorics.cartesianProduct(days, times, semesters, types);
```
yangshun: Cool beans!

v3/src/js/utils/filters/Filter.js
yangshun: Can we rename this to test instead? A predicate sounds more like a function instead of a method name.

ZhangYiJiang:
Thanks for the part about AND/OR - I didn't play around with the page closely enough to realize that's how it worked. 

I think I'll add a `FilterCollection` class that represents each block of filters. Filters within the block is applied as OR (`Array#some`), and each FilterCollection is applied using AND (`Array#every`). For GraphQL, I talked to Li Kai, and for the first iteration we'll copy the original's logic of eager loading all module data first. 

The biggest concern is whether eager loading ~1.3MB (gzipped) of module data before showing everything vs. doing GraphQL query as the user selects filters is better or worse in terms of user experience. On the one hand GraphQL promises much better initial loading time, but on the other hand being able to play around with filters and getting immediate feedback is also very satisfying.

yangshun:
@ZhangYiJiang We should merge it back into master more often. We prefer to have smaller PRs and short-lived PR branches.

v3/src/js/utils/filters/FilterGroup.js
```diff
@@ -0,0 +1,26 @@
...
+  label: string;
+  filters: { [string]: Filter };
+
+  constructor(label: string, filters: Array<Filter>) {
+    this.filters = _.keyBy(filters, filter => filter.label);
+    this.label = label;
+  }
+
+  test(module: Module): boolean {
+    return _.values(this.filters).some(filter => filter.test(module));
```
li-kai: Curse you flow.

v3/src/js/utils/filters/LevelFilter.js
```diff
@@ -0,0 +1,18 @@
...
+
+import type { ModuleLevel } from 'types/modules';
+import ModuleFilter from './ModuleFilter';
+
+export default class LevelFilter extends ModuleFilter {
+  level: ModuleLevel;
+
+  constructor(level: ModuleLevel) {
+    super(`${level}000`, (module) => {
+      const match = module.ModuleCode.match(/(\d+)/);
```
li-kai: I don't get this, why not just match the first digit?

v3/webpack/webpack.config.dev.js
```diff
@@ -55,6 +55,9 @@ const developmentConfig = merge([
       // do not emit compiled assets that include errors
       new webpack.NoEmitOnErrorsPlugin(),
     ],
+    watchOptions: {
+      aggregateTimeout: 2000,
```
li-kai: 2000 is a bit too much, isn't it? Default is ~350

v3/.eslintrc.js
```diff
@@ -82,6 +82,11 @@ module.exports = {
     'react/require-default-props': 'off',
     // TODO: Replace divs with buttons, but remove all button styling.
     'jsx-a11y/no-static-element-interactions': 'off',
+    'jsx-a11y/label-has-for': ['error', {
```
li-kai: Could you add comments for this? Help let people understand why we change these rules.

v3/src/js/utils/filters/test-filter.js
```diff
@@ -0,0 +1,16 @@
+/* global expect */
+
+// eslint-disable-next-line import/prefer-default-export
+export function testFilter(Filter, module, combinations, valid) {
```
ZhangYiJiang: Adding `.test` causes it to be picked up by Jest, which will try to run the file as a test and fail because it doesn't have any tests, so we can't use that. Will rename it to `filter-test-helpers.js`

v3/webpack/webpack.config.dev.js
ZhangYiJiang: This shouldn't have been committed in the first place, oops

PR #427, opened 2017-08-15
Reduce some stylistic issues to warn outside production

ZhangYiJiang:
By using `process.env.NODE_ENV`, we can drop some common style issues down to **warning** during development. This reduces development friction, since Webpack waits on ESLint **errors** to be resolved before compiling, and most style issues should not force the developer to stop and fix them before proceeding. 

Since CI runs in production, anything that slips through will be caught by Travis anyway, so this is safe.

### DIFF
--- a/v3/.eslintrc.js
+++ b/v3/.eslintrc.js
@@ -1,3 +1,6 @@
+const warnInDevelopment = process.env.NODE_ENV === 'production' ?
+  'error' : 'warn';
+
 module.exports = {
   parser: 'babel-eslint',
   root: true,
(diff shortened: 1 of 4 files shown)


ZhangYiJiang:
Hmmm, I'm not sure what's happening here - seems that Flow remapping isn't working? https://travis-ci.org/nusmodifications/nusmods/jobs/264626242#L526

li-kai:
AH. Yeah if you see in `webpack.dev.config.js`, we map the root only when it is test 😅

PR #428, opened 2017-08-15
Fix dark mode export table header background colour

angelsl:
t2


yangshun:
Thank you! Did you manage to try with normal mode and see if there was no regression?

angelsl:
@yangshun As far as I could.

t1

yangshun:
Perfect, that's all that's needed. Thanks! (:

PR #429, opened 2017-08-15
Fix v3 jpg/ical export

angelsl:


ZhangYiJiang:
Might be a good idea to extract the download data URL logic into its own function.

v3/src/js/actions/export.js
```diff
@@ -1,11 +1,25 @@
...
 import ical from 'ical-generator';
 
 import type { ModuleCode, Module, Semester } from 'types/modules';
 import type { SemTimetableConfigWithLessons } from 'types/timetables';
 import { iCalForTimetable } from 'utils/ical';
 
+function downloadUrl(url: string, filename: string) {
+  const link = document.createElement('a');
+  const body = document.body;
+  invariant(body);
```
ZhangYiJiang: I don't think this is necessary - a guard clause like 

    if (!body) return;

will satisfy Flow's type checker. Adding `assert` as an import also adds ~3.7kb to the file size 😅 

```
  36.69 KB (+3.69 KB)  build/app.62aad2275949e671dcec.js
```

ZhangYiJiang: So this is interesting, because I thought that line won't work at all since [`assert` is a Node module](https://nodejs.org/api/assert.html), but it seems like [Webpack polyfills some of these modules](https://github.com/nusmodifications/nusmods/blob/560accee8620d58b178601f8c61ed8f3dc24e5e9/v3/webpack/webpack.parts.js#L341). However since we're just using it for a not-null check there's no need to import the entire module for this.

PR #431, opened 2017-08-16
Upgrade to Bootstrap 4 beta

ZhangYiJiang:
Bootstrap 4 beta was just released, so of course we need to upgrade to the new hotness! Seriously though, given the scope of the changes and the fact that they've switched bootstrap.com over to v4-beta documentations, I *think* the code will be more stable now, so it's as good a time as any to upgrade. 

The major breaking changes that affects NUSMods v3 I've noticed are -  

- Theme colors are defined as a map and accessed through the `theme()` function now instead of `$brand-*` variables 
- `offset-*` classes have been removed. I've added back `offset-md-1` and `offset-sm-1` manually. The same effect cannot be recreated with the new alignment classes, although they do open up new possibilities - https://getbootstrap.com/docs/4.0/layout/grid/#alignment 
- `btn-secondary` has been changed to a dark gray color. This affects the new student 'toggle' button. I've changed it to `btn-outline-primary` instead 
- `hidden-*` responsive utilities have been removed. `d-*` utilities can replace them in some limit cases, but not for `hidden-xs-down`, so I fixed it by writing the media query into `nav.scss` itself. 

Also fixed some off-by-one bugs with responsive layout - don't use media query `max/min-width` directly with `$grid-breakpoint`, but rather use the `media-breakpoint-*` utilities provided by Bootstrap. There's no size regression - BS4 beta's final CSS is about 700 bytes smaller than alpha-6.

### DIFF
--- a/v3/package.json
+++ b/v3/package.json
@@ -78,7 +78,7 @@
   "dependencies": {
     "axios": "^0.16.2",
     "babel-polyfill": "^6.23.0",
-    "bootstrap": "4.0.0-alpha.6",
+    "bootstrap": "4.0.0-beta",
     "classnames": "^2.2.5",
     "d3": "^4.10.0",
     "dom-to-image": "^2.5.2",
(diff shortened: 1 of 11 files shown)


ZhangYiJiang:
Yeah, I don't like it too. I'm strongly considering moving off BS4 grid, but it's not trivial (and having columns makes it easy to prototype layout)

PR #436, opened 2017-08-22
Minor style fixes

yangshun:
Just some minor style fixes as I was reviewing the latest code. It's not documented anywhere and it's not exactly important, but let's try to follow this format for imports:

1. NPM imports
2. Our own `type` imports
3. Our own non-`type` imports
  3.1 Absolute imports
  3.2 Relative imports

For CSS, sort property names alphabetically. There are a few schools of thoughts regarding sorting order but sorting it alphabetically makes it easier for binary scanning a list of properties. In the event of overrides occurring because of alphabetical order, this rule can be ignored (or perhaps don't use the shorthand).

### DIFF
--- a/v3/src/js/views/browse/ModuleFinderList.jsx
+++ b/v3/src/js/views/browse/ModuleFinderList.jsx
@@ -1,8 +1,9 @@
 // @flow
-
 import React from 'react';
-import FilterGroup from 'utils/filters/FilterGroup';
+
 import type { Module } from 'types/modules';
+
+import FilterGroup from 'utils/filters/FilterGroup';
 import ModuleFinderItem from 'views/components/ModuleFinderItem';
 
 type Props = {
@@ -16,9 +17,9 @@ export default function ModuleFinderList(props: Props) {
 
   return (
     <ul className="modules-list">
-      {filteredModules.slice(0, 30).map((module) => {
-        return <ModuleFinderItem key={module.ModuleCode} module={module} />;
-      })}
+      {filteredModules.slice(0, 30).map(module =>
+        <ModuleFinderItem key={module.ModuleCode} module={module} />,
+      )}
     </ul>
   );
 }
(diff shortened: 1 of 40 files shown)


v3/src/js/views/browse/ModuleFinderList.jsx
```diff
@@ -16,9 +17,9 @@ export default function ModuleFinderList(props: Props) {
 
   return (
     <ul className="modules-list">
-      {filteredModules.slice(0, 30).map((module) => {
-        return <ModuleFinderItem key={module.ModuleCode} module={module} />;
-      })}
+      {filteredModules.slice(0, 30).map(module =>
+        <ModuleFinderItem key={module.ModuleCode} module={module} />,
```
ZhangYiJiang: I don't think there should be a comma here

v3/src/js/views/browse/ModulePageContainer.jsx
```diff
@@ -127,86 +126,69 @@ export class ModulePageContainer extends Component {
...
                 {renderExaminations}
-
                 <dt className="col-sm-3">Semesters Offered</dt>
                 <dd className="col-sm-9">{semsOffered}</dd>
-
-                <dt className="col-sm-3">Offical Links</dt>
+                <dt className="col-sm-3">Official Links</dt>
                 <dd className="col-sm-9">
                   <ul className="nm-footer-links">
                     {ivleLink ? <li><a href={ivleLink}>IVLE</a></li> : null}
```
ZhangYiJiang: Missed converting these from ternary to `&&`

v3/src/js/views/components/ModuleFinderItem.jsx
```diff
@@ -25,42 +26,35 @@ export default function ModuleFinderItem(props: Props) {
               </Link>
             </h2>
           </header>
-
-          <p>{ module.ModuleDescription }</p>
-
+          <p>
```
ZhangYiJiang: If we're moving this information up then it should logically be in the `<header>`

v3/src/js/views/browse/ModuleFinderList.jsx
ZhangYiJiang: Hmmm, so apparently ESLint wants a comma here because it views this as a multiline function. I feel that this is a bit misleading since it can get confused with arrays, but it's a minor problem either way.

Issue #426, opened 2017-08-15
Bring module finder to feature parity 


ZhangYiJiang:
This is a meta issue tracking the progress of the module finder page at `/modules/`. 

### Pull Requests 

- #425 

### Progress 

- [X] Implement `Filter` base class 
- Update data layer
  - [X] `moduleInformation.json` for loading all module data 
  - [ ] `facultyDepartment.json` for mapping of departments to faculties 
- Implement specialized Filter subclasses 
  - [X] `LevelFilter`
  - [x] `TimeslotFilter`
  - [ ] `FacultyFilter`
- Implement views 
  - [X] Filter sidebar and subcomponents 
  - [ ] Update `ModuleFinderContainer` with filtering logic 
  - [ ] Pagination

PR #437, opened 2017-08-24
Modules page filters part three: Serializable filters

ZhangYiJiang:
This is part 3 of ??? for the module finder page at `/modules`. See #426 for history and context 

### New features 

- Filters are (de)serializable from URL query strings now. This allows several use case including 
  - Sharing the module finder page with someone else by simply sending them the URL 
  - Getting back the same module finder page state after browsing to a module page 
- Timeslot filters are represented using a TimeslotTable component now. I'm still unsatisfied with the UI, but this is a good starting point 
- Preclusions, prereqs and coreqs on the module finder page now use LinkModuleCodes 

### Internal changes 

- ModuleFilter and FilterGroup now takes in an additional `id` parameter that is used for serialization. This value should be short, unique and URL safe. This value is also used to key these objects instead of `label` previously. 

### Bug fixes 

- ModuleFilter's `count()` function did not memoize the count value properly. This has now been fixed. 

### Known Issues 

- LinkModuleCodes will raise a unique key warning if the same module is referenced twice 
- Module finder will crash if the Workload string is unparsable 

Both of these will be fixed in the next PR.


li-kai:
@yangshun Need your opinion on this. We're now serialising data from the url. This means we have two sources of truth now - store and url.

We're not able to make the store follow url and url follow store unless we [pass history object to actions](https://reacttraining.com/react-router/core/guides/redux-integration/deep-integration) or use react-router-redux. But [react-router-redux](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux) is still in beta after 7 months.

Alternatively, we could choose another [router library](https://github.com/FormidableLabs/redux-little-router) or simply [write our own](https://medium.freecodecamp.org/you-might-not-need-react-router-38673620f3d).

Of course, the last thing is we simply depend on URL for state, but I prefer not to do that, as we need to bring url support for timetables as well.

ZhangYiJiang:
Probably best to outline what the code in this PR does for the sake of discussion: 

- `ModuleFinderContainer` currently holds filter state locally (filter state refers to which filters are enabled)
- Filter state is initialized in `ModuleFinderContainer`'s constructor, which reads in the search params and passes it down to the individual `FilterGroup` objects, which toggles the relevant filters. This forms the initial filter state. 
- When filter state is updated, `ModuleFinderContainer` updates the query string from each of the `FilterGroup` object. 

On the whole, this works mostly fine. The scenarios outlined in the PR works, since in all of those cases a new `ModuleFinderContainer` is being constructed. 

However, URL and state will desync if the user presses the back button after some filters have been selected, since this updates the URL (removes the query string) but does not cause `ModuleFinderContainer`'s constructor to be called. 

It should be possible to manually bind a listener in `ModuleFinderContainer` to listen to history changes and update state from there. This should take care of the above desync issue, but care must be taken to avoid infinite loops.

yangshun:
Two-way sync is a pain but it's still not as bad; the current NUSMods is actually doing that for the timetable (URL and localStorage). For a while I was trying to do a three-way sync (+ a server) and it was very crazy. I gave up after struggling with it for a few hours. What Yijiang suggested (watching the history) sounds ok. We can try that.

> Of course, the last thing is we simply depend on URL for state, but I prefer not to do that, as we need to bring url support for timetables as well.

What's wrong with the above? Sounds ok to me. I was also about to suggest using URL as the main source of truth. In current NUSMods if the URL differs from localStorage, an alert is shown to prompt the user to decide on which data to be used as the source of truth. For module filtering, I'd say just go with the URL as the source of truth and overwrite the one in the store.

ZhangYiJiang:
Implemented the listener so that changes to the URL also triggers filter state updates. This PR should be ready for review.

v3/src/js/utils/filters/FilterGroup.js
```diff
@@ -1,25 +1,32 @@
...
-import _ from 'lodash';
-import type { Module } from 'types/modules';
+import { keyBy, values } from 'lodash';
 import update from 'immutability-helper';
+
+import type { Module } from 'types/modules';
+
 import ModuleFilter from './ModuleFilter';
 
+export const ID_DELIMINATOR = ',';
```
li-kai: Nit: `DELIMITER`.

v3/src/js/utils/filters/FilterGroup.js
```diff
@@ -43,6 +52,23 @@ export default class FilterGroup<Filter: ModuleFilter> {
     return !!this.activeFilters.length;
   }
 
+  // Query string (de)serialization - this allows the currently enabled filters to be
+  // embedded directly
+  toQueryString(): string {
+    if (!this.isActive()) return '';
```
li-kai: Redundant now :D

v3/src/js/views/browse/module-filters.js
```diff
@@ -1,39 +1,45 @@
...
+    Timeslots.map(([day, time]) => new TimeslotFilter(day, time, 'Tutorial')),
+  ),
 
-export const tutorialTimeslots = new FilterGroup(
-  'Tutorial Time',
-  timeslots.map(([day, time]) => new TimeslotFilter(day, time, 'Tutorial')),
-);
+  [MODULE_CREDITS]: new FilterGroup(MODULE_CREDITS, 'Module Credit', [
+    new Filter('0', '0-3 MC', module => parseInt(module.ModuleCredit, 10) <= 3),
+    new Filter('4', '4 MC', module => module.ModuleCredit === '4'),
```
li-kai: Might miss modules with `3.5 MC` or `4.5 MC` if they existed.

PR #438, opened 2017-08-27
Page layout realignment 

ZhangYiJiang:
### Context 

Currently v3 uses Bootstrap grid for all layout. Unfortunately Bootstrap is not sufficiently flexible for all layout needs - for instance, it doesn't support mixed fixed and flexible width columns, resulting in awkward sizes for fixed width content like the navigation sidebar:

At the lower end of breakpoint-md, the column is too narrow, causing ugly line breaks to appear 

screenshot from 2017-08-27 11-42-49

At large screen widths, the column becomes uncomfortably large (screenshot with 1920px wide screen)

screenshot from 2017-08-27 11-45-54

### Approach 

This PR swaps BS4 grids for the side nav with a fixed width solution using a mix of flexbox and `calc()`.

**breakpoint-lg** 

screenshot from 2017-08-27 11-49-40

screenshot from 2017-08-27 11-50-01

**breakpoint-md**

screenshot from 2017-08-27 11-50-14

screenshot from 2017-08-27 11-50-35

The main benefit of this in addition to the aforementioned improvement in layout is that it reduces the level of nesting and amount of boilerplate markup. The downside is that the CSS is slightly more complicated.

We also use `position: fixed` for the sidebar so that it scrolls with content.

### DIFF
--- a/v3/src/js/views/AppShell.jsx
+++ b/v3/src/js/views/AppShell.jsx
@@ -27,28 +27,24 @@ type Props = {
 // Put outside render because this only needs to computed on page load.
 const weekText = (() => {
   const acadWeekInfo = NUSModerator.academicCalendar.getAcadWeekInfo(new Date());
-  const year = `AY20${acadWeekInfo.year}`;
-  let semester = '';
-  let semesterType = '';
-  let week = '';
+  const parts = [`AY20${acadWeekInfo.year}`];
 
   // Check for null value (ie. during vacation)
   if (acadWeekInfo.sem) {
-    semester = `, ${acadWeekInfo.sem}`;
+    parts.push(acadWeekInfo.sem);
   }
 
   // Hide semester if semester type is 'Instructional'
   if (acadWeekInfo.type !== 'Instructional') {
-    semesterType = `, ${acadWeekInfo.type} Week`;
+    parts.push(`${acadWeekInfo.type} Week`);
   }
 
   // Do not show the week number if there is only one week, eg. recess
   if (acadWeekInfo.num > 0) {
-    week = `, Week ${acadWeekInfo.num}`;
+    parts.push(`Week ${acadWeekInfo.num}`);
   }
 
-  const acadWeekString = `${year}${semester}${semesterType}${week}`;
-  return acadWeekString;
+  return parts.join(', ');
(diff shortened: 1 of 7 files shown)


v3/src/js/views/AppShell.jsx
```diff
@@ -82,41 +78,34 @@ export class AppShell extends Component {
...
+            </NavLink>
+            <NavLink className="nav-link" activeClassName="active" to="/settings">
+              <i className="fa fa-fw fa-lg fa-gear" />
+              <span className="nm-link-title">Settings</span>
+            </NavLink>
+          </nav>
+
+          <main className="main-content">
+            {this.props.fetchModuleListRequest.isPending && !this.props.moduleList.length ?
+              <p>Loading...</p> : null
```
yangshun: Could you convert this to the `&&` version?

v3/src/js/views/AppShell.jsx
```diff
@@ -82,41 +78,34 @@ export class AppShell extends Component {
...
+            }
+            {this.props.fetchModuleListRequest.isSuccessful || this.props.moduleList.length ?
+              this.props.children : null
```
yangshun: Could you convert this to the `&&` version?

v3/src/styles/pages/timetable.scss
```diff
@@ -27,7 +27,6 @@ $timetable-day-rows-height: calc(100% - #{$timetable-day-cell-height});
 
   .timetable-wrapper {
     -webkit-overflow-scrolling: touch;
-    overflow-x: scroll;
```
yangshun: Could you help test this on mobile and make sure that removing this line doesn't break anything? It's supposed to allow the timetable to be scrollable.

ZhangYiJiang: Hmmm, which mobile browser? I have an Android device so I'll need someone else (\*cough* @li-kai \*cough*) to check Safari Mobile. It's removed because removing the `::webkit-scrollbar` hack causes the scrollbar to appear here on Chrome desktop.

li-kai: Just checked, am able to scroll on mobile.

PR #439, opened 2017-08-28
Css sorting

li-kai:
Sort css rules in a semantic fashion, and automatically too with `yarn lint:styles -- --fix`.

Heavily recommend viewing the changes in split mode. Hopefully you'll find the new ones easier to understand.

An upcoming PR will introduce commit hooks to lint and recommend ways of fixing them.


v3/.stylelintrc.js
```diff
@@ -0,0 +1,139 @@
...
+      'dollar-variables',
+      'declarations',
+      'rules',
+      'at-rules',
+    ],
+    // https://github.com/hudochenkov/stylelint-order/blob/master/rules/properties-order/README.md
+    // tldr; Rules follow a logical, semantic ordering to prevent
+    // readers from resetting their context of styles being applied
+    'order/properties-order': [
+      "composes",
```
ZhangYiJiang: /nit wrong quote mark (no idea why ESLint wasn't triggered)

yangshun:
Go ahead and merge,  as long as the ordering can be kept consistent via a lint rule/script

v3/.stylelintrc.js
li-kai: Right...cause config files aren't linted.

ZhangYiJiang:
Looks fine. Do we want to set the `unspecified` option? Also, is there a standard sorting order we can plug into this so we won't have to update the list when new CSS standards appear (eg. grid)?

li-kai:
There's no standard preset for sort order. In fact, there's no preset. 🙊

So setting `unspecified` might be a good move.

PR #440, opened 2017-08-28
Upgrade Flow to v0.54

ZhangYiJiang:
Since Flow got a relatively big update with v0.53, it seems like a good idea to remove the accidentally checked in libdef stub for React and clean up the errors that have accumulated over the last few PRs. Here are issues blocking this:  

- [x] https://github.com/yangshun/flow-scripts/pull/8 - need a version bump from flow-scripts so that we won't accidentally generate the stub again 
- [X] ~~~https://github.com/yannickcr/eslint-plugin-react/pull/1377 - ESLint React plugin needs to be upgraded to support the new Component typedef in v0.53~~~
- [x] https://github.com/nusmodifications/nusmods/compare/upgrade-flow?expand=1#diff-65c8a51f0e20cd366636d726bae01971R76 - I don't know wtf is happening here.


ZhangYiJiang:
- flow-scripts v0.4.0 has been published 
- We ignore eslint-plugin-react for now - assume the changeset will be a major release. Once that is out, the extra `props: Props` and `state: State` in the components can be stripped out 
- According to Flow developers this is a feature, not a bug, so we add a (rather silly) check to convince the typechecker what we're doing is safe

ZhangYiJiang:
Upgrading to Flow v0.54 caused the following piece of code to be marked as unsafe (Object literal does not have ModuleCode property). Unfortunately after poking at it for several hours, including reading the Flow changelog, issue tracker, trying to reproduce it in the Flow online REPL and switching to VS Code, I've not made any progress, so I've just slapped `$FlowFIxMe` on it. 

```
      timetableLessons = timetableLessons.map((lesson) => {
        // Identify the current lesson being modified.
        return areLessonsSameClass(lesson, activeLesson) ?
          // $FlowFixMe Explicitly annotating lesson also does not work
          { ...lesson, isActive: true } : lesson; 
```

PR #441, opened 2017-09-06
Modules page filters part four: Infinite scroll, sticky sidebar 

ZhangYiJiang:
This is part of the project to bring the module search page at `/modules` to feature parity for v3. See https://github.com/nusmodifications/nusmods/issues/426 for the other PRs in this series. 

### Infinite scroll

This is the big feature for this PR. Infinite scroll vs. a more traditional page based view is controversial, and hard to get right. The current implementation is experimental - if it is not up to scratch, it should not be too difficult to pull out the infinite scroll and replace it with a more traditional pagination navigation. That said, I believe the current implementation is **as good as pagination** (this is my litmus test for whether infinite scroll should be kept) in almost all aspects, and works better in a lot others, especially on mobile. 

#### Approach 

The `ModuleFinderList` component handles the splitting of the modules array into pages (chunks of 10 modules). [React Waypoint](https://github.com/brigade/react-waypoint) wraps every page to track which page the user is looking at. 

When the user reaches the last visible page, the number of visible pages is incremented, causing the next set of modules to be rendered. Page state is kept in `ModuleFinderContainer`, which keep tracks of the user's current page, the first rendered page and the number of visible pages. If the user changes filters, the current page resets to zero and we scroll back to top.

React Virtualized (and other off the shelf lazily rendered infinite scroll solutions) cannot be used because we have variable list item heights. 

#### Navigation 

screenshot from 2017-09-06 20-45-42

Dividers are inserted between pages to orientate the user by showing them how many items they've looked at. To preserve the user's current page, as the user scrolls, the URL hash is updated. When the user returns, we only show modules from the current page onward. 

localhost-8080-modules-level 2

The user has to manually trigger the loading of previous pages. This removes the need for us to have to worry about having to manually manage scroll position, which would be difficult and fragile. 

#### Performance 

Initial performance testing using a local production build on mobile and desktop Chrome shows no stuttering and lag during scrolling. 

#### Remaining issue 

The one unavoidable issue is the scrollbar position, which jumps as the new page is rendered. I don't personally think it is a dealbreaker, but we need to do more user testing to figure this out, as well as if this implementation is usable in general. 

### Style updates 

The other part of this PR is to get the styling of various UI elements closer to a release ready stage. 

#### Sticky filter sidebar

On desktop the design uses Amazon's product search sidebar as reference, so the changes are mostly minor tweaks. On mobile the [Google IO conference session page](https://events.google.com/io/schedule/?section=may-17) (remember to resize down to mobile size - the design changes significantly between mobile and desktop) filter controls is used as reference. 

localhost-8080-modules-lecture tuesday-morning mc 4 iphone 6 plus

Note that this is styling only - the popup behavior currently uses `:hover`, which is a placeholder until a proper component can be written for this. 

#### Semester info selector 

The semester info selector need to fulfill two use cases 

- Tell the user which semester the module is available in 
- Allow the user to check class info in another semester 

This means it needs to function as a set of three-state tabs, which is somewhat complex as far as UI elements go. This is what it looks like now 

screenshot from 2017-09-06 20-37-19

the tooltip should at some point be replaced with a proper tooltip component, like say Popper. 

### Other changes 

- Stylelint has been reduced to warning if the env is not production. This reduces development friction, since most of its complaints are stylistic and can be fixed before commit
- Fix `LinkModuleCodes` key warning caused by the same module code appearing more than once


v3/src/js/views/browse/ModuleFinderContainer.jsx
```diff
@@ -31,12 +32,28 @@ type Props = ContextRouter;
...
+  page: PageRange,
   modules: Module[],
   filterGroups: { [FilterGroupId]: FilterGroup<any> },
 };
 
+export function mergePageRange(prev: PageRange, diff: PageRangeDiff): PageRange {
+  const next = _.clone(prev);
+
+  // Current page is SET from the diff object
+  if (diff.current != null) next.current = diff.current;
```
li-kai: nit, just use braces here

v3/src/js/views/browse/ModuleFinderContainer.jsx
```diff
@@ -31,12 +32,28 @@ type Props = ContextRouter;
...
+
+  // Start and pages are ADDED from the diff object
+  ['start', 'loaded'].forEach((key) => {
+    if (diff[key] != null) next[key] += diff[key];
```
li-kai: nit, same here

v3/src/js/views/components/module-info/TimeslotTable.jsx
```diff
@@ -19,46 +34,73 @@ const timeLabels: { [Time]: string } = {
...
+          <tr className="module-timeslot-row">
+            <th />
             {days.map(day => (
-              <td key={`cell-${day}-${time}`}>{ children.get(getTimeslot(day, time)) }</td>
+              <th
+                key={`heading-${day}`}
+                className={classnames('module-timeslot-day', {
+                  'module-timeslot-hover': day === hover.day,
+                })}
+              >{day.slice(0, 3)}</th>
```
li-kai: nit ,  put</th> on the next line, same for the others below.

v3/src/js/views/browse/ModuleFinderContainer.jsx
```diff
@@ -104,16 +124,49 @@ export class ModuleFinderContainerComponent extends Component<Props, State> {
...
+
+      // Scroll back to the top
+      window.scrollTo(0, 0);
     });
   };
 
+  onPageChange = (diff: PageRangeDiff) => {
+    this.setState(prevState => ({
+      page: mergePageRange(prevState.page, diff),
+    }), () => {
```
li-kai: Could you extract this function? Kinda hard to read with so many arrows.

PR #445, opened 2017-09-12
Module page improvement and request reducer refactor 

ZhangYiJiang:
This was originally suppose to be a update mostly about UI, but it turned into a bug hunting exercise. Oh well. 

### UI Improvement 

Most of the UI improvements are on the module page: 

screenshot from 2017-09-12 19-25-14

screenshot from 2017-09-12 19-25-32

I think we should keep most of the old design, which works quite well. The big difference is that we lift important info such as semesters in which the module is available in, and the number of MCs up to the header. These will be links when I figure out how to generate filter URLs outside of module finder pages, so imagine they are orange (and maybe bold). 

The "Add to timetable" button is a placeholder - I would like the final UI to use a dropdown, but we don't have one available. I was also going to port over the navigation sidebar, but this changeset is already big enough, so that got pushed to the next PR. 

### Behind the scene 

`ModulePageContainer` has been split into two components - `ModulePageContainer` now just handles fetching module info JSON and displaying the error page, while `ModulePageContent` contains the actual contents of the page. This allows both network requests to happen at the same time, which should improve the loading speed a little, but more importantly improves separation of concern, as `ModulePageContent` now no longer needs to care about what happens when the module doesn't exist, or fails to load.

### Bug fixes 

- Fixed a pagination/infinite scroll bug where new pages would load in even when scrolling up 
- Fixed 'Page not found' appearing on module page titles by swapping out `react-document-title` with `react-helmet`. This is one of the reason for the large changeset - you can turn whitespace diffs off to see that most of the changes are due to this 
- Fixed the Request reducer - and all components that depend on it 
- Fixed main page content being rendered before module list has been loaded in `AppShell.jsx`. 
- Fixed a module bank bug triggered by race condition due to incorrect rehydration logic. The fix is a bit of a hack - in the long run we should probably swap out the current solution with something like https://github.com/rt2zz/redux-persist/tree/v5/src (v5 is in beta right now)


v3/src/js/actions/helpers.js
```diff
@@ -1,25 +1,25 @@
...
+import type { FSA } from 'types/redux';
 
-export const RESET_REQUEST_STATE: string = 'RESET_REQUEST_STATE';
-export function resetRequestState(domain: string | Array<string>): FSA {
+export const RESET_REQUEST_STATE = 'RESET_REQUEST_STATE';
+export function resetRequestState(domain: string | string[]): FSA {
   return {
     type: RESET_REQUEST_STATE,
-    payload: _.isArray(domain) ? domain : [domain],
+    payload: castArray(domain),
```
li-kai: Awesome, I didn't even know there was such a utility.

v3/src/js/utils/react.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Returns the given array of React nodes joined by a glue string. This
+ * allows eg. inline lists to be created where links are separated by
+ * dots.
+ *
+ * @param {Node[]} nodes
+ * @param {string} glue
+ * @returns {Node}
+ */
+export function join(nodes: Node[], glue: string = ' • '): Node {
```
li-kai: `intersperse` of signature `(any[], delimiter: any)` as a more generic function would be better

v3/src/js/utils/react.js
```diff
@@ -0,0 +1,63 @@
...
+ * except it is safer as the returned result is a React Node, so there's
+ * no need to set innerHTML. This is useful for eg. highlighting search results
+ * with <mark>
+ *
+ * @param str
+ * @param regex
+ * @param replacement
+ * @returns {Node}
+ */
+export function replace(
```
li-kai: `replaceWithNode`

PR #448, opened 2017-09-17
Use <Timetable> for module page timetable

ZhangYiJiang:
This replaces the tabular lesson representation on the module page, which is a straight port from v2 (which in itself is a straight rip from CORS), with the `<Timetable>` component used on the timetable page. IMO this works better, especially if there's a lot of lessons - 

**Before** 

screenshot from 2017-09-17 11-35-20

**After** 

screenshot from 2017-09-17 11-31-38

GES1033

screenshot from 2017-09-17 11-39-13

CS1101S 

--- 

The only feature missing from the old version is the link to the Venue page, but I don't know if anyone uses that. Also the semester picker component is kind of ugly - I'll fix that in another PR.


v3/src/js/views/components/module-info/SemesterPicker.jsx
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      this.props.onSelectSemester();
+    }
+  };
+
+  /**
+   * Map button labels (semester names) to semester data
+   * @returns {{}}
+   */
+  semesterMap(): { [string]: ?{ semester: Semester, semesterData: SemesterData } } {
```
li-kai: imo this data is better served passed down from the parent... avoid recalculation for every step, and makes it so that `onSelectSemester` doesn't have to accept optional props. What do you think?

v3/src/js/views/components/module-info/SemesterPicker.jsx
```diff
@@ -0,0 +1,97 @@
...
+    useShortNames: false,
+  };
+
+  onSelectSemester = (choice: string) => {
+    const chosen = this.semesterMap()[choice];
+
+    if (chosen) {
+      // This makes Flow happy
+      const semester = Number(chosen.semester);
+      if (semester === 1 || semester === 2 || semester === 3 || semester === 4) {
```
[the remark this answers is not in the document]
ZhangYiJiang: Doesn't work - there's no integer data type in JS, so the range `0 < semester <= 4` also contains every float in between the two. Unfortunately Flow's type system doesn't give any way of casting object keys that are coerced into strings back, even when it's obviously safe (as is the case here)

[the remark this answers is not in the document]
ZhangYiJiang: I guess we can drop the enum and make `Semester = number` . It's not very useful anyway since it's unlikely we'll be hard coding any values of `Semester`

v3/src/js/views/components/module-info/SemesterPicker.jsx
ZhangYiJiang: Just to clarify, do you mean we only pass `Semester[]` down as prop instead of `SemesterData[]`?

li-kai: I meant `semesterMap`

PR #449, opened 2017-09-26
Fix various bugs on timetable

li-kai:
Actually just one lol. Since we sort and arrange by start time, we would never had 2 same lessons with same start time on the same row.

Now we only pass `onModifyCell` if it's actually modifiable.

### DIFF
--- a/v3/src/js/views/timetable/TimetableCell.jsx
+++ b/v3/src/js/views/timetable/TimetableCell.jsx
@@ -11,7 +11,7 @@ import styles from './TimetableCell.scss';
 type Props = {
   lesson: Lesson,
   style: Object,
-  onModifyCell: Function,
+  onModifyCell?: Function,
 };
 
 function TimetableCell(props: Props) {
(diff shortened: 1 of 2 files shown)


v3/src/js/views/timetable/TimetableCell.jsx
```diff
@@ -11,7 +11,7 @@ import styles from './TimetableCell.scss';
 type Props = {
   lesson: Lesson,
   style: Object,
-  onModifyCell: Function,
+  onModifyCell: ?Function,
```
ZhangYiJiang: I think this should be `onModifyCell?: Function` instead

v3/src/js/views/timetable/TimetableRow.jsx
```diff
@@ -50,10 +50,11 @@ function TimetableRow(props: Props) {
         };
         return (
           <TimetableCell
-            key={lesson.ModuleCode + lesson.ClassNo}
+            key={lesson.StartTime}
             style={style}
             lesson={lesson}
-            onModifyCell={onModifyCell}
+            // $FlowFixMe When object spread type actually works
+            onModifyCell={lesson.isModifiable && onModifyCell}
```
yangshun: This expressions evaluates to `false` when `lesson.isModifiable` is `false`. Are you sure that's what you want?

Issue #332, opened 2016-09-28
Use new icon loading approach


yangshun:
Say no to icon fonts and use SVG instead
- http://mir.aculo.us/2014/10/31/icon-fonts-vs-inline-svg/

@li-kai suggested inlining SVGs and using these http://google.github.io/material-design-icons/

Can consider these also: https://icons8.com/line-awesome

ZhangYiJiang:
Fixed by #450

PR #450, opened 2017-09-26
Feature/svg icons

li-kai:
Adds icons from feathericons.com

screenshot 2017-09-23 02 37 07


ZhangYiJiang:
Can we use https://github.com/carmelopullara/react-feather instead? I'm really not a fan of copy-pasting code from external libraries directly into our codebase

ZhangYiJiang:
A couple of notes on SVG icons: 

- Using `fill="currentColor"` or `strokeColor="currentColor"` allows the SVG to inherit the outer  element's `color` property, but **only** if the SVG is embedded using `<svg>` - using `` will not work. 
- Unlike an icon font, `font-size` will not resize the icons - obvious in hindsight, but worth repeating. 
- Because of the way the module is packaged, currently we import each required icon into `views/components/icons.jsx`. This has two benefits - 
  1. It allows us to mix our own icons with Feather's icons. This does mean we have to match the props on `react-feather` if we want to keep a consistent interface, but since the only props are `color` and `size`, this is pretty easy 
  2. It ensures only the icons we use are bundled, otherwise Webpack will bundle every single icon. 

Reference: 

- https://css-tricks.com/scale-svg/
- https://icons8.com/ (where I found the LinkedIn icon)

ZhangYiJiang:
Also this closes #332

PR #451, opened 2017-09-26
Add prereq tree visualization

li-kai:
Implements prereq tree by presenting it horizontally. Doing this greatly preserves space, where even in worst case results in:
screenshot 2017-12-22 02 34 14

### DIFF
--- a/v3/src/js/views/modules/ModuleTree.jsx
+++ b/v3/src/js/views/modules/ModuleTree.jsx
@@ -68,7 +68,7 @@ function Tree({ layer, name, branches, isPrereq = false }: TreeDisplay) {
       >
         {isConditional ? formatConditional(name) : <LinkModuleCodes className={styles.link}>{name}</LinkModuleCodes>}
       </div>
-      {branches && <Branch layer={incrementLayer(layer, name)} branches={branches} />}
+      {branches && branches.length > 0 && <Branch layer={incrementLayer(layer, name)} branches={branches} />}
     </li>
   );
 }
--- a/v3/src/js/views/modules/ModuleTree.jsx
+++ b/v3/src/js/views/modules/ModuleTree.jsx
@@ -1,193 +1,97 @@
 // @flow
-
-import React, { Component } from 'react';
+import React from 'react';
+import classnames from 'classnames';
 import _ from 'lodash';
-import d3 from 'd3/build/d3';
+
+import { NUM_DIFFERENT_COLORS } from 'utils/colors';
+import LinkModuleCodes from 'views/components/LinkModuleCodes';
 
 import type { Module } from 'types/modules';
+
+import styles from './ModuleTree.scss';
 
 type Props = {
   module: Module,
 };
 
-export default class ModuleTree extends Component<Props> {
-  prereqRoot: ?HTMLDivElement;
+type TreeDisplay = {
+  layer: number,
+  name: string,
+  isPrereq?: boolean,
+  // TreeDisplay[] will result in infinite loop
+  branches: ?Array<TreeDisplay>,
+};
 
-  componentDidMount() {
-    const isOrAnd = d => d.name === 'or' || d.name === 'and';
-    const modsFilter = d => !isOrAnd(d);
-    const andOrFilter = d => isOrAnd(d);
-    const getHeight = d => (isOrAnd(d) ? 35 : 60);
-    const getWidth = d => (isOrAnd(d) ? 50 : 120);
-    const getX = d => (isOrAnd(d) ? -25 : -60);
-    const getY = d => (isOrAnd(d) ? -17.5 : -35);
-
-    const prereqRoot = this.prereqRoot;
-    if (!prereqRoot) return;
-
-    const SVGHeight = 400;
-    let SVGWidth = prereqRoot.clientWidth;
-    let rectangles;
-    let interact;
-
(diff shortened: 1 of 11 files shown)


ZhangYiJiang:
I'm surprised this works using purely HTML + CSS - good job! Replacing d3 would slim down the page quite a bit, so much so that we might not need to break it out into a separate component after all. 

RE: Styling - I think having three colors to identify current mod, prereq and required by would be useful. The best would be if we have all of the students past timetables and highlight fulfilled and unfulfilled requirements, but I'm not sure how realistic that is. 

RE: Is it useful - I think it would be if we could make it dynamic - ie. you can expand out each node by clicking on it (that would conflict with the action of opening the module page though, hmm), or recursively show requisites, although for high level modules this may be very deep. Also the original can also show the modules that the current module is required by, which makes it more useful, and if we can highlight fulfilled and unfulfilled requirements, that would be the best. 

Misc. comments: 

- Flipping the whole thing onto it's side may make large trees more readable. 
- Instead of AND and OR on the branches, ALL OF and ONE/ANY OF may be clearer, especially for non CS students :P

v3/src/js/views/modules/ModuleTree.jsx
```diff
@@ -1,193 +1,99 @@
...
+      })}
+    >
+      <div
+        className={classnames(styles.node, {
+          [`color-${layer}`]: !isConditional,
+          [styles.conditional]: isConditional,
+          [styles.prereqNode]: isPrereq,
+        })}
+      >
+        {isConditional ? formatConditional(name) : <LinkModuleCodes className={styles.link}>{name}</LinkModuleCodes>}
```
ZhangYiJiang: I don't think this needs `<LinkModuleCodes>` - that component is used when you have text that *may* have module code embedded. If you already know that this will always be a link, then you can just use `<Link>` (with `routes/path.js` to generate the path). 

Also, you can wrap the entire `div` in the `<Link>` this way.

v3/src/js/views/modules/ModuleTree.jsx
```diff
@@ -1,193 +1,99 @@
...
+      {branches && <Branch layer={incrementLayer(layer, name)} branches={branches} />}
+    </li>
+  );
+}
 
-  render() {
-    return <div ref={(div) => { this.prereqRoot = div; }} className="nm-prerequisites-tree" />;
-  }
+function ModuleTree(props: Props) {
+  // $FlowFixMe when we can add properties
```
ZhangYiJiang: What does this mean?

v3/src/js/views/modules/ModuleTree.jsx
```diff
@@ -1,193 +1,99 @@
...
+      <ul className={classnames(styles.tree, styles.root)}>
+        <Tree layer={1} name={modTree.name} branches={_.castArray(modTree.children)} />
+      </ul>
+    </div>
+  );
 }
+
+export default ModuleTree;
+// For testing
+export { incrementLayer };
```
ZhangYiJiang: Can just export the function directly - no need to export an object containing the function

v3/src/js/views/modules/ModuleTree.jsx
li-kai: Not really, some modules are deadends

ZhangYiJiang: Hmmm, good point. In that case we could use absolute positioning to just make the anchor the size of the entire `<div>`. Currently there's a small but noticeable area around the div where it's not clickable, and that bothers me somewhat.

v3/src/js/views/modules/ModuleTree.test.jsx
```diff
@@ -0,0 +1,104 @@
...
+                  ],
+                },
+              ],
+            ],
+          },
+        ],
+      },
+      LockedModules: ['CS5242', 'CS5339', 'CS6281'],
+    };
+    // $FlowFixMe just the two is enough
```
ZhangYiJiang: If you only need these two properties out of `<Module>`, would it be better in terms of design to have just these two as props instead of the entire `Module` object?

v3/src/js/views/modules/ModuleTree.jsx
```diff
@@ -1,193 +1,98 @@
...
+  );
+}
 
-    nodeAll.selectAll('text')
-      .classed({ 'opacity-transition': true, opaque: true, transparent: false });
-    nodeAll.on('mouseover', mouseOver);
-    nodeAll.on('mouseout', mouseOut);
-    nodeAll.on('click', clicked);
-  }
+function Tree({ layer, name, branches, isPrereq = false }: TreeDisplay) {
```
ZhangYiJiang: Would it be better to make this and `<Branch>` plain functions instead of function components?

v3/src/js/views/modules/ModuleTree.jsx
```diff
@@ -1,193 +1,98 @@
...
-      .transition()
-      .duration(1000)
-      .attr('width', getWidth)
-      .attr('height', getHeight)
-      .ease('elastic');
+function Branch({ layer, branches }: { layer: number, branches: TreeDisplay[] }) {
+  return (
+    <ul className={styles.tree}>
+      {branches.map((child) => {
+        return _.castArray(child).map((subchild) => {
```
ZhangYiJiang: /nit can remove `return`

PR #452, opened 2017-09-27
Module page sticky side navigation 

ZhangYiJiang:
This PR brings the side menu from v2 to v3. 

peek 2017-09-27 14-42

It uses [`react-scrollspy`](https://github.com/makotot/react-scrollspy), which works pretty well, and is only about 200 LoC. The API is slightly annoying because you have to write the same ID three times, but it's not too bad. 

### Other changes 

- The prerequisite tree and Disqus thread components have been stripped out entirely, because they throw errors that crashes JS execution on the page. 
- Public path is now configured correctly in Webpack, so the `<base href="/">` in the HTML template is no longer required.


v3/src/styles/pages/modules.scss
```diff
@@ -40,6 +46,45 @@
     font-size: $font-size-sm;
   }
 
+  section {
+    // Account for the fixed header's height. If this is not done, when side menu links are
+    // clicked a portion of the content will be clipped by the fixed header.
+    padding-top: $navbar-height + 1rem;
```
li-kai: Could you give this a variable since it's reference 3 times

v3/src/js/views/browse/ModulePageContent.jsx
```diff
@@ -148,36 +145,45 @@ class ModulePageContentComponent extends Component<Props> {
...
+            <section id="timetable">
               <h2>Timetable</h2>
               <LessonTimetable
                 semestersOffered={this.semestersOffered()}
                 history={module.History}
               />
-            </div>
-
-            <h2>Review and Discussion</h2>
-            <ReactDisqusThread
```
li-kai: Is this intentional?

ZhangYiJiang: Removing `<ReactDisqusThread>`? Yes, because the component was throwing an exception which meant that nothing on the page worked after initial mounting. The prereq tree code has the same problem. I'll drop in a TODO to remind us to replace it with something.

PR #453, opened 2017-09-28
Fix settings page theme bug

ZhangYiJiang:
A CSS specificity bug was caused by #448 changing where the container with the `.theme-*` class sits in the DOM hierarchy. This caused the currently selected theme's colors to override the theme colors in the theme selection UI, causing some... interesting effects. 

This PR adds a more specific selector for each of the theme color button on the settings page, fixing the bug. To do so, I have also extracted the theme colors into a SASS map. 

This is a slightly less clean solution to using CSS variables, but using variables would mean themes are no longer supported for Edge 14 and iOS Safari 9 users - although both are non-current version.


li-kai:
I'm up for the CSS4 solution though. The [article you posted](https://codepen.io/jakealbaugh/post/css4-variables-with-fallbacks-using-sass) is quite convincing 😄

ZhangYiJiang:
I liked that solution too. The only problem is that you'll then need to use `@supports` to remove the setting from the page for those browser's users, which is doable, but personally given that the difference between the two approach is so minor little in practice, I fall on the side of maximizing support.

yangshun:
I would lean towards Yijiang's current approach. But the decision is up to you guys.

PR #455, opened 2017-09-29
Update jest dependency

li-kai:
Just updating dependencies here. No change in functionality.


ZhangYiJiang:
Weird, why didn't Coveralls get triggered?

PR #456, opened 2017-10-04
Module search

### LINKED ISSUE #142: Search by Keyword from the description
I think in the past the search index also get the keyword from the module description. 
Is it removed?

For example, I'm looking for a module that teaches "Delaunay triangulation". (it is listed in CS5237 description)

ZhangYiJiang:
This closes #142 

peek 2017-10-03 20-50

### Performance 

Searching is done entirely on the client side. Performance is acceptable even without any sort of fancy indexing - just plain `String#includes` - the slowest part of the operation is actually rendering. On mobile we may need to disable 'instant' search, but again the performance is actually really good. 

Note that there is more stuttering in development due to Redux logger. The gif above was recorded against a production build. 

### Implementation 

Search is implemented as a filter. This means it plays nice with the rest of the filters, although the code is slightly awkward. The search box keeps the search string in local state and only pushes it into the Redux store on a throttle. This improves perceived responsiveness because rendering is not triggered directly by user input events. 

The Redux store is also useful because the search term highlighting requires the search term, but only several components deeper in the component tree, so pushing it down from the topmost component would be rather awkward. 

### To do

- [x] Sorting by relevance: The current implementation makes this a little awkward, but I think it should be possible by assigning each module a score in `moduleSearch()` based on where the token appears (module code and title are 'worth' more), and `<ModuleFinderContainer>` can keep in its state to sort the modules by. 

- [ ] Better tokenization: Currently tokens are generated by a simple `String#split`, but that means things like `"science fiction"` (notice the quotation marks) will be treated as two words. It should be easy enough to treat things in quotes as verbatim tokens


yangshun:
Is this PR still WIP?

ZhangYiJiang:
It's in an odd place because the code is ready, but I need more performance testing to make sure it performs well on both mobile (esp. low end devices) and desktop. I think it can be merged in this state, although some of the parameters may need tweaking. 

I'll document the performance "hacks" used here 

- The SearchBox component syncs input value with component state, and only pushes the state to Redux separately. This allows the UI to render first before everything else, improving perceived performance 
- The ModuleFinder component uses `performance.now()` to benchmark the amount of time it took to initialize filters for the first time, and only enables 'instant search' when the user is on desktop and the performance is good enough 
- Instant search is where the content gets rendered while the user is typing (ie. `throttle`), with it off `debounce` is used, so content is rendered only after the user has finished typing 

The performance on my computer is pretty good - on production the lag is almost imperceptible (Chrome performance console will still highlight them, but perception is better because of the async rendering I mentioned above). On mobile instant search is impossible to use, because the rendering takes upwards of 1s. Rendering takes about 50% of the time, while filter + sort takes up the rest of the time.

v3/src/js/utils/react.jsx
```diff
@@ -38,6 +67,13 @@ export function replaceWithNode(
   });
 }
 
+export function highlight(str: string, search: string | string[], Tag: string = 'mark'): Node {
+  const terms = castArray(search).filter(Boolean);
+  if (!terms.length) return str;
+  const regex = new RegExp(`(${terms.map(escapeRegExp).join('|')})`, 'ig');
+  return replaceWithNode(str, regex, (match, i) => <Tag key={i}>{match}</Tag>);
```
yangshun: Have you considered https://github.com/bvaughn/react-highlight-words instead? I would prefer to have less of our own code as possible in the repo

v3/src/js/views/browse/ModuleFinderContainer.jsx
```diff
@@ -90,10 +102,23 @@ export class ModuleFinderContainerComponent extends Component<Props, State> {
...
+
   componentDidMount() {
     axios.get(nusmods.modulesUrl())
       .then(({ data }) => {
+        const start = window.performance.now();
         this.filterGroups().forEach(group => group.initFilters(data));
+        const time = window.performance.now() - start;
+
+        this.useInstantSearch = breakpointUp('md').matches && (time < INSTANT_SEARCH_THRESHOLD);
+        console.log(`${time}ms taken to init filters`); // eslint-disable-line
```
yangshun: Maybe `console.info` would be more appropriate.

v3/src/js/views/browse/ModuleFinderContainer.test.jsx
```diff
@@ -61,6 +61,8 @@ function createContainer(initialEntries?: Array<LocationShape | string>): Contai
         history={history}
         location={history.location}
         match={mockMatch}
+        resetModuleFinder={_.noop}
+        searchTerm={''}
```
yangshun: Let's stick to `searchTerm=""` or specify empty string as a default prop of the component.

v3/src/js/views/browse/ModuleFinderList.scss
```diff
@@ -0,0 +1,7 @@
+@import "~styles/utils/modules-entry";
+
+.noModulesIcon {
+  width: 3rem;
```
yangshun: If both `width` and `height` are meant to be the same:

```css
$size: 3rem;
width: $size;
height: $size;
```

v3/webpack/webpack.parts.js
```diff
@@ -204,7 +204,7 @@ exports.minifyJavascript = () => {
         compress: {
           warnings: false,
           // Drop `console` statements.
-          drop_console: true,
+          // drop_console: true,
```
yangshun: We should either remove this or leave it there. Commenting it out is not very clean.

v3/src/js/main.js
```diff
@@ -11,7 +11,7 @@ import '../styles/main.scss';
 const persistedState = storage.loadState();
 const store = configureStore(persistedState);
 store.subscribe(
-  _.throttle(() => {
+  _.debounce(() => {
```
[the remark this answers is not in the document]
ZhangYiJiang: The problem I had was that with `throttle`, this occasionally collided with my timers for search, causing lag because serializing a large object to JSON takes a pretty long time. I'm reducing this to 1s, but a more permanent solution may be to keep a reference to only when the serialized objects have changed.

v3/src/js/utils/react.jsx
ZhangYiJiang: I have not, but that seems to be a pretty big component (5.1kb gziped - https://bundlephobia.com/result?p=react-highlight-words@0.10.0) compared to five lines here

v3/src/js/utils/react.jsx
```diff
@@ -50,3 +86,9 @@ export function replaceWithNode(
 export function noBreak(text: string): string {
   return text.replace(/ /g, NBSP);
 }
+
+export function defer(task: () => any) {
+  window.requestAnimationFrame(() => {
+    window.requestAnimationFrame(task);
```
li-kai: lol the double requestAnimationFrame trick. Technically we don't need window here.

v3/src/js/utils/react.jsx
```diff
@@ -38,6 +67,13 @@ export function replaceWithNode(
   });
 }
 
+export function highlight(str: string, search: string | string[], Tag: string = 'mark'): Node {
+  const terms = castArray(search).filter(Boolean);
+  if (!terms.length) return str;
+  const regex = new RegExp(`(${terms.map(escapeRegExp).join('|')})`, 'ig');
```
li-kai: Could we match start of string instead? Meaning "count" should not match "accounting".

v3/src/js/views/browse/ModuleFinderContainer.jsx
```diff
@@ -90,10 +105,23 @@ export class ModuleFinderContainerComponent extends Component<Props, State> {
...
+
   componentDidMount() {
     axios.get(nusmods.modulesUrl())
       .then(({ data }) => {
+        const start = window.performance.now();
         this.filterGroups().forEach(group => group.initFilters(data));
+        const time = window.performance.now() - start;
+
+        this.useInstantSearch = breakpointUp('md').matches && (time < INSTANT_SEARCH_THRESHOLD);
+        console.info(`${time}ms taken to init filters`); // eslint-disable-line
```
li-kai: 😞  my mac flutters in between 126 - 183ms

v3/src/js/utils/react.jsx
ZhangYiJiang: Hmmm, but search uses `includes`, which matches anywhere. I was afraid things would get even slower if a regex is used.

PR #457, opened 2017-10-04
Tweak timetable styles

ZhangYiJiang:
This fixes several bugs and tweaks the timetable design slightly 

- Bug: Firefox assumes `background-size: 50%` to be shorthand for `background-size: 50% auto` (unlike Chrome, which assumes `background-size: 50% 50%`). This causes the background stripes to break on Firefox 
- Bug: `min-width: auto` was used - I'm not sure if browsers actually accepted this, but it is not valid. The initial value of `min-width` is `0` (or `initial`) 
- Bug: `overflow-wrap` is causing Firefox to break up text in bad places, such as within room names. I'm not sure if this is a browser issue or just how the property is suppose to work, but I've removed it for now (@li-kai - do you know what's the intended use for this?)
- Bug: Removed the `@media (min-width: $breakpoint) and (orientation: landscape)` min width rule. I'm not entirely sure what's the intended purpose of this rule, but it doesn't seem to work properly when triggered. The tweak below should compensate for this. (@li-kai) 
- Tweak: Vertical timetable height is now determined dynamically by the number of hours spanned. Previously this was a fixed number, but that resulted in too little space for timetables that span the full space of 0800-2100 
  
  screenshot from 2017-10-04 13-41-42  
  (What the bug looks like) 

- Tweak: In horizontal mode, cells on the same row now have the same height. This makes the timetable appear neater and match the behavior on vertical mode 

  screenshot from 2017-10-04 13-48-47

- Tweak: All cells now have 1px space from neighbouring cells. This matches v2 design and makes module page timetables more readable:

  screenshot from 2017-10-04 13-50-42

  I've also removed the padding from `TimetableDay` - I'm not entirely sure of their purpose (**Question**: good idea? Maybe mobile only?) 

  screenshot from 2017-10-04 13-52-15

- Feature: Day heading is now sticky on horizontal mode. The size is also collapsed on mobile (sm or below). I'm not sure why this doesn't seem to work on vertical mode - help would be appreciated. **Question**: Do you think this is useful? 

  screenshot from 2017-10-04 13-44-33

  screenshot from 2017-10-04 13-45-11


li-kai:
Regarding `overflow-wrap`, I rather text overflowed than get hidden in timetable. I would make it such that cells filled up height, but that's not possible on vertical mode.

`@media (min-width: $breakpoint) and (orientation: landscape)` is to scale the timetable to height of the window on desktops, that way you can see the entire timetable in vertical mode without scrolling up and down. I think this is quite preferable than scaling each hour (since there's more than sufficient vertical height on a desktop)

padding from `TimetableDay` is something that's done on v2. And also raised by user feedback from https://modsplanner.tk website, because people couldn't differentiate between days easily (even with border).

Sticky is definitely useful. "The size is also collapsed on mobile". Not sure...this looks weird.
On the issue on why it doesn't work on vertical, it's cause sticky is relative to the window. Since the scrollbar is within the timetable, sticky works on horizontal. It fails on vertical because you're scrolling on the viewport.

ZhangYiJiang:
For `overflow-wrap`, this is the bug I see on Firefox (left). Removing the property (right) doesn't seem to cause overflowing text to be hidden, and it still breaks at natural line breaks (hyphens, spaces) 

screenshot from 2017-10-04 14-55-19 screenshot from 2017-10-04 14-57-25

For the media query, I do agree it looks better, but the problem is that right now it is possible to get overflowing cells if you have a 1h lesson with four lines (odd/even week) on small screens (iPad in landscape). Dynamic height ensures there's enough space, although if you don't have any 1h class with odd/even weeks it'll look like it's wasting a bit of space. I wonder if we can render, check for overflow, then readjust the height, although that may be a bit too much work :P

yangshun:
> padding from TimetableDay is something that's done on v2. And also raised by user feedback from https://modsplanner.tk website, because people couldn't differentiate between days easily (even with border).

In v2 it's more of a min-height rather than a padding. If one day has overlapping classes, the padding goes away:

screen shot 2017-10-04 at 5 34 10 pm

ZhangYiJiang:
> "The size is also collapsed on mobile". Not sure...this looks weird.

Would following v2's style be better? 

screenshot from 2017-10-04 18-15-09

li-kai:
@yangshun Seems to be the case, but I fell padding works better with monstrous modules like GER1000.

@ZhangYiJiang yeah that way looks better

v3/src/js/views/timetable/TimetableContainer.scss
```diff
@@ -10,11 +10,10 @@
...
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
-}
 
-// Higher specificity to override bootstrap
-:global(.verticalMode).timetableWrapper {
   @include media-breakpoint-down(sm) {
-    padding: 0 2% 0 0;
+    // Maximize the space taken by the timetable by overflowing into the grid gutters
+    margin-right: -($grid-gutter-width / 2);
```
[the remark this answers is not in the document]
ZhangYiJiang: Added to `.days` as `padding: 0 2vw 0 0`

ZhangYiJiang:
Stickied timing to the left on vertical mode - useful when the timetable is too wide, such as when picking out a large number of choices. This does require `min-width: fit-content`, which is not supported in Edge, but I think we can live with this https://caniuse.com/#feat=intrinsic-width 

screenshot from 2017-10-10 23-00-10

ZhangYiJiang:
Should be good for re-review.

v3/src/js/views/timetable/TimetableCell.scss
```diff
@@ -5,16 +5,18 @@ $cell-opacity: 0.5;
...
+  display: flex;
+  flex-direction: column;
   padding: $cell-padding;
   font-size: $font-size-xxs;
   line-height: 0.85rem;
   text-align: left;
-  vertical-align: top;
-  border: 0;
-  border-bottom: 2px solid;
+  border-width: 0 0 0.2rem;
```
li-kai: Regression? Borders are now a bit too thick for our current visual style 
screenshot-2017-10-12 timetable - nusmods

v3/src/js/views/timetable/TimetableRow.scss
```diff
@@ -14,3 +16,12 @@
     }
   }
 }
+
+:global(.verticalMode) .timetableRow {
+  height: auto;
+  margin: 0 1px 0 0;
+
+  &:empty {
```
li-kai: I think we agreed that width should be consistent between rows. This seems to do the opposite?

Old vs

new

v3/src/js/views/timetable/TimetableDay.scss
```diff
@@ -23,32 +24,62 @@
...
+  @include media-breakpoint-down(sm) {
+    flex: 0 0 $timetable-day-h-size-sm;
+    overflow: hidden;
+
+    .dayNameText {
+      width: 80%;
+      font-size: $font-size-xs;
+      // On small screens, display the day names vertically
+      text-orientation: upright;
+      writing-mode: vertical-lr;
```
li-kai: This is how it looks on iOS 11. 😅 Looking at online examples, all the text are left-facing, how'd you make yours right facing?

v3/src/js/views/timetable/TimetableTimings.scss
```diff
@@ -26,5 +33,9 @@
 }
 
 :global(.verticalMode) .time {
+  display: inline-block;
+  padding: 0 0.25rem 0 0.1rem;
+  border-radius: 0 0.2rem 0.2rem 0;
+  background: rgba($body-bg, 0.75);
```
li-kai: Thoughts on doing this on `.timings` instead?

v3/src/js/views/timetable/TimetableDay.scss
ZhangYiJiang: Ugh, so Safari doesn't support `text-orientation: upright`, so I added in a `@supports not` block below as fallback, but it doesn't seem to be working. Let me try something else instead.

v3/src/js/views/timetable/TimetableRow.scss
ZhangYiJiang: Huh, I thought you intended it to collapse on small screens - the `:empty` block is to add a guard against this, observed on Chrome in responsive design mode 

screenshot from 2017-10-12 23-07-28

PR #460, opened 2017-10-08
Add slug to module page route

ZhangYiJiang:
This changes the route for modules from `/modules/:moduleCode` to `/modules/:moduleCode/:slug?`. The slug is generated from the kebab case version of the module title, and is optional. If a non-canonical URL is visited, we automatically redirect the visitor to the canonical version. The primary purpose of the slug is to improve human readability of the URL and improve SEO.


ZhangYiJiang:
Also, for the record, the module with the longest name (by characters) mounted this year is ES1501Q, *Critical Thinking and Expository Writing: "Why you so like dat?" Singlish and cultural literacy* (no really, that is its actual name), generating the URL `localhost:8080/modules/ES1501Q/critical-thinking-and-expository-writing-why-you-so-like-dat-singlish-and-cultural-literacy`

v3/src/js/views/components/Table.jsx
```diff
@@ -13,14 +13,9 @@ export default function Table(props: Props) {
     <tbody>
       {props.data.map((s, i) => (
         <tr key={i}>
-          {s.map(d =>
-            (<td key={d}>
-              {d}
-            </td>),
-          )}
+          {s.map((d, j) => <td key={j}>{d}</td>)}
```
yangshun: Can we use more meaningful variable names instead of `s`, `j` and `d`?

v3/src/js/reducers/entities/moduleBank.js
```diff
@@ -23,20 +23,20 @@ export type ModuleBank = {
...
-  moduleCodes: Set<ModuleCode>,
+  moduleCodes: Map<ModuleCode, ModuleCondensed>,
 };
 
 const defaultModuleBankState: ModuleBank = {
   moduleList: [], // List of modules
   modules: {}, // Object of ModuleCode -> ModuleDetails
   moduleSelectList: [],
-  moduleCodes: new Set(),
+  moduleCodes: new Map(),
```
li-kai: Is there any way we could simply use `modules` for this?

ZhangYiJiang: No, modules only contains fetched modules. The naming of that variable is bad.

yangshun: Sorry, my bad 😭

v3/src/js/utils/modules.js
```diff
@@ -15,8 +16,8 @@ import type {
 
 import config from 'config';
 
-export function modulePagePath(moduleCode: ModuleCode): string {
-  return `/modules/${moduleCode}`;
+export function modulePagePath(moduleCode: ModuleCode, moduleTitle: ModuleTitle): string {
```
yangshun: Seems redundant to have to pass in the title along with the module code. Makes more sense to just pass in the entire module directly.

v3/src/js/reducers/entities/moduleBank.js
yangshun: Hmm, redux store state should only have serializable data. Have you faced any problems with serializing them?

v3/src/js/utils/modules.js
ZhangYiJiang: Links to modules need to be generated in places where the entire `Module` object is not available - that data needs to be loaded from the backend.

v3/src/js/reducers/entities/moduleBank.js
ZhangYiJiang: The module code cache is used to linkify module codes in module description and *requisite fields. It does that right now by using a regex to pull out valid module codes, then do a lookup against a list of valid module codes (otherwise it'll generate dead links). I think it may be possible that using an array is fast enough (not benchmarked), but this is one of the places where it seems a faster data structure is worth using since this lookup need to be performed many times when displaying list of modules on the module finder page.

v3/src/js/reducers/entities/moduleBank.js
```diff
@@ -23,20 +28,22 @@ export type ModuleBank = {
...
 const defaultModuleBankState: ModuleBank = {
   moduleList: [], // List of modules
   modules: {}, // Object of ModuleCode -> ModuleDetails
   moduleSelectList: [],
-  moduleCodes: new Set(),
+  moduleCodes: {},
 };
 
 function precomputeFromModuleList(moduleList: ModuleList) {
   // Cache a Set of all module codes for fast module existence checking
```
yangshun: This comment is outdated I think

PR #461, opened 2017-10-10
Hello team!

ZhangYiJiang:
Adds myself to the team! 🎆 

Also refactors slightly the static pages by extracting out the common wrapper into a component (thus most of the changes are in indent levels) and adjusting the styles. The responsive square utility is used to pad out the space around images so that their loading does not push content downwards, which is a personal pet peeve of mine.


v3/src/styles/utils/responsive-square.scss
```diff
@@ -0,0 +1,15 @@
+// Responsive square - useful for creating rectangles that scale proportionately with width
+// From https://spin.atomicobject.com/2015/07/14/css-responsive-square/
+.responsive-square {
+  position: relative;
+  display: block;
+  padding-bottom: 100%;
+}
+
+.responsive-square-content {
```
yangshun: Is `responsive-square-content` always meant to be nested inside a `responsive-square`? If it is I think it's ok to nest the styles as well.

PR #463, opened 2017-10-13
Error page and Sentry feedback integration 

ZhangYiJiang:
- Integrates Sentry into the app by capturing errors manually and using the raven-for-redux middleware 
- Add a `<ErrorPage>` component that shows up when an error is captured 

  screenshot from 2017-10-13 12-12-03

- Add a button that activates the [Raven feedback form](https://docs.sentry.io/learn/user-feedback/) to `<NotFoundPage>` and `<ErrorPage>` 
- Push request error into Redux and integrate [Redux Developer Tool](https://github.com/gaearon/redux-devtools)


ZhangYiJiang:
To test `<ErrorPage>`, go to `/modules/`, go offline (use the browser developer tools) and click on any of the modules. To test Raven, edit line 6 in `util/sentry.js` to force Sentry to run in dev (it does not by default). 

``` js
const loadRaven = true; // process.env.NODE_ENV === 'production';
```

With React 16, I think we might be able to do better by using React's error boundary, in which case we can simply throw the error in the component and let the parent catch it instead of handling it inside the component itself like we're doing now, although I'm not entirely sure how that will work with Promises.

v3/src/js/stores/configure-store.js
```diff
@@ -1,11 +1,21 @@
...
+
+/* eslint-disable no-underscore-dangle */
+// For redux-devtools-extensions - see
+// https://github.com/zalmoxisus/redux-devtools-extension
+const composeEnhancers =
+  typeof window === 'object' &&
+  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ?
+    window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
+      // Specify extension’s options like name, actionsBlacklist, actionsCreators, serialize...
+    }) : compose;
```
li-kai: `/* eslint-enable no-underscore-dangle */`

v3/src/js/views/errors/ErrorPage.jsx
```diff
@@ -0,0 +1,62 @@
...
+  error: string,
+  eventId?: string,
+  showRefresh: boolean,
+};
+
+export default class NotFoundPage extends PureComponent<Props> {
+  props: Props;
+
+  static defaultProps = {
+    error: 'something went wrong',
```
li-kai: I feel like "Uh oh... something went wrong. `${error}`" is better?

v3/src/js/views/errors/ErrorPage.jsx
```diff
@@ -0,0 +1,62 @@
...
+            </p>}
+
+            {eventId &&
+            <p>
+              An error report has been made and we will look into this.
+              <button
+                className={styles.link}
+                onClick={() => Raven.showReportDialog({ eventId })}
+              >
+                Tell us more about what happened so we can better fix this.
```
li-kai: We'd really appreciate it if you could tell us ...

v3/src/js/storage/index.js
```diff
@@ -1,14 +1,13 @@
 import _ from 'lodash';
+import Raven from 'raven-js';
 
 // Simple wrapper around localStorage to automagically parse and stringify payloads.
 // TODO: Use an in-memory storage for environments where localStorage is not present,
 //       like private mode on Safari.
 function setItem(key, value) {
-  try {
+  Raven.context(() => {
```
li-kai: Why not use https://docs.sentry.io/clients/javascript/#manually-reporting-errors instead?

ZhangYiJiang:
Fixed the dialog problem with a z-index override in `components/sentry.scss`. It's not clean (Sentry can change their dialog wrapper class name and that'll break this fix), but it should do for now.

PR #467, opened 2017-10-28
CORS bidding notification for v3 

ZhangYiJiang:
This ports over the CORS scrapper script and implements the notification bubble. 

On desktop: 

screenshot from 2017-10-28 13-48-18

On mobile:

screenshot from 2017-10-28 13-49-01

The scraper script has been improved - the data structure it produces now looks like this: 

```
type CorsPeriod = {
  type: 'open' | 'closed',
  start: string,
  startDate: Date, // The scraper produces startTs and endTs, which the client converts to Date objects
  end: string,
  endDate: Date,
}

type CorsRound = {
  round: string,
  periods: CorsPeriod[],
};

type Config = {
  // ... other config values
  corsSchedule: CorsRound[],
}
```

Date string formatting is done entirely during scraping. This means that we need to send both the timestamp and the formatted string to the client, but this additional weight is offset by the removal of string parse/format code on the client side. It also allows us to use Moment's `format()` function, which is much nicer than anything we can write ourselves. 

The scraper now also has a bunch of assertions as well as some sanity checks, so that it can fail fast if the CORS site changes HTML structure or contains invalid data (one of the 2017/18 Sem 1 bidding periods is in 201**6** :man_facepalming: ) so that they can be manually corrected.


ZhangYiJiang:
To manually test the notification bubble, add `?round={CORS bidding round name}` to the URL of any page, eg. `?round=1A`

v3/src/js/views/AppShell.jsx
```diff
@@ -91,6 +109,9 @@ export class AppShell extends Component<Props> {
 
         <div className="main-container">
           <Navtabs />
+
+          <CorsNotification time={this.currentTime()} />
```
li-kai: So the notification will appear on every page? Right now it only shows on timetable iirc.

v3/src/js/views/components/cors-info/CorsNotification.jsx
```diff
@@ -0,0 +1,44 @@
...
+import React, { PureComponent } from 'react';
+import { capitalize } from 'lodash';
+
+import config, { type CorsRound } from 'config';
+import { roundEnd } from 'utils/cors';
+
+import styles from './CorsNotification.scss';
+
+type Props = {
+  time: Date,
```
li-kai: Would be nice to have a default prop of current time.

v3/src/js/views/AppShell.jsx
```diff
@@ -67,6 +74,17 @@ export class AppShell extends Component<Props> {
     }
   }
 
+  currentTime() {
+    // For testing
```
li-kai: Is this gonna be removed when this gets approved ❓

v3/package.json
```diff
@@ -14,6 +14,7 @@
     "lint:styles": "stylelint \"src/**/*.scss\" --syntax scss",
     "test": "jest --coverage",
     "test:watch": "jest --watch",
+    "cors": "node scripts/cors-crawler",
```
li-kai: Hmm...I'm leaning more and more towards moving this to api repo / mono-repo format now. Now we have different versions of cheerio.

v3/package.json
```diff
@@ -34,6 +35,8 @@
     "babel-preset-react-optimize": "^1.0.1",
     "babel-register": "^6.24.1",
     "cache-loader": "^1.0.3",
+    "chalk": "^2.2.2",
```
li-kai: This isn't used any where.

v3/src/js/views/components/cors-info/CorsNotification.scss
```diff
@@ -0,0 +1,33 @@
...
+      display: none;
+    }
+  }
+
+  @include media-breakpoint-up(sm) {
+    position: fixed;
+    bottom: 2rem;
+    left: 2rem;
+    z-index: $cors-notification-z-index;
+    border-width: 0.1rem;
```
li-kai: I think I know why the borders keep bothering me now. It's 3 times as thick as borders used everywhere else (1px). 
capture

v3/src/js/views/components/cors-info/CorsNotification.scss
```diff
@@ -0,0 +1,33 @@
...
+    box-shadow: 0 3px 12px rgba(#000, 0.3);
+
+    :global {
+      // Delay the animation a bit so that the user will not be distracted
+      // by the page loading
+      animation: bounceInUp 1.5s 0.7s backwards;
```
li-kai: Mmm. It's alright I suppose. A tad distracting, but it's important.

v3/src/js/views/components/cors-info/CorsNotification.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const round = this.currentCorsRound();
+
+    // CORS bidding is over - don't show anything
+    if (!round) return null;
+    const period = this.currentPeriod(round);
+    if (!period) return null; // Shouldn't happen
+
+    const isRoundOpen = time >= period.startDate;
+    return (
+      <a href="https://myaces.nus.edu.sg/cors/StudentLogin" target="_blank" rel="noopener noreferrer">
```
li-kai: Nice touch. 👍 I really like it except it's hard to tell it's clickable (and goes to cors). It's fine, we'll refine as we have more notifications.

v3/package.json
ZhangYiJiang: It's used by the build script https://github.com/nusmodifications/nusmods/blob/master/v3/scripts/build.js (eslint wasn't picking this up because of config)

v3/src/js/views/AppShell.jsx
ZhangYiJiang: I think it should be left in, because we can use it for manual testing before every semester, and it's not a lot of code

v3/src/js/views/components/cors-info/CorsNotification.jsx
ZhangYiJiang: Done

v3/src/js/views/components/cors-info/CorsNotification.scss
yangshun: Ya we should stick with 1px. Using `px` is ok for borders

PR #469, opened 2017-10-30
Exclude public holidays from iCal timetable

### LINKED ISSUE #376: iCalendar File export does not account for public holidays in Singapore
The calendar file generated includes lessons on public holidays but there should be no classes then.

ZhangYiJiang:
Closes #376 

This pulls in SG public holidays data from https://github.com/rjchow/singapore_public_holidays and adds them to the list of excluded lesson dates, so lessons on public holidays are excluded 

screenshot from 2017-10-30 12-48-07


v3/src/js/data/holidays.json
```diff
@@ -0,0 +1,35 @@
...
+  "2018-02-17",
+  "2018-03-30",
+  "2018-05-01",
+  "2018-05-29",
+  "2018-06-15",
+  "2018-08-09",
+  "2018-08-22",
+  "2018-11-06",
+  "2018-12-25"
+]
```
yangshun: Add newline

li-kai:
The thing I'm not confident of is that we're sourcing public holidays on a person's repo that may not be maintained, or accurate in the future. Since students have been aware of public holidays themselves for the past 7 years...should we keep it that way?

ZhangYiJiang:
Actually the source is the Ministry of Manpower's website. The external repo is a slightly easier source to consume from if it will be maintained, but even if it won't be the `holidays.json` file can still be kept up to date by hand relatively easily. 

And yeah, it is slightly ridiculous that the "best" machine readable source is a stack of CSVs from some random guy on GitHub, but at least the data can be verified easily from the primary source. 

As for keeping the current behavior, no, this is clearly a bug. We know more than a year in advance the dates when there won't be classes. There's no reason to generate calendar events on those days.

v3/src/js/data/holidays.json
```diff
@@ -0,0 +1,35 @@
...
+  "2016-02-08",
+  "2016-02-09",
+  "2016-03-25",
+  "2016-05-01",
+  "2016-05-21",
+  "2016-07-06",
+  "2016-08-09",
+  "2016-09-12",
+  "2016-10-29",
+  "2016-12-25",
```
yangshun: `2016-12-25` is a Sunday. We should be adding `2016-12-26` instead of `2016-12-25`.

v3/scripts/holidays.js
```diff
@@ -0,0 +1,25 @@
...
+const inputs = fs.readdirSync(IN_DIR);
+const holidays = [];
+
+inputs.forEach((file) => {
+  console.log(`Reading holidays from ${file}`);
+  fs.readFileSync(path.join(IN_DIR, file), 'utf-8')
+    .trim()
+    .split('\n')
+    .slice(1)
+    .map(line => line.split(',')[0])
```
yangshun: While working on the scraper I realized an issue with this. Shouldn't we be using the `Observance` column rather than the `Date` column? The observance is the actual date of the holiday.

ZhangYiJiang:
PR should be ready for review

PR #475, opened 2017-11-03
Update v3 README

ZhangYiJiang:
Feel free to add more stuff to the README by pushing to the branch https://github.com/nusmodifications/nusmods/blob/update-v3-readme/v3/README.md


v3/README.md
```diff
@@ -20,27 +36,90 @@ To run the development build, simply run:
 $ yarn start
 ```
 
-#### Running tests
+This will start Webpack dev server, which will automatically rebuild and reload any code and components that you have changed. 
+
+#### Running tests and linter
+
+We use [Jest] with [Enzyme][enzyme] to test our code and React components.
```
yangshun: I think Jest's link is missing

v3/README.md
```diff
@@ -20,27 +36,90 @@ To run the development build, simply run:
...
+├── static                  - Static assets, eg. favicons
+|                             These will be copied directly into /dist 
+└── webpack                 - Webpack config 
+```
+
+#### Colocation 
+
+Components should keep their styles and tests in the same directory with the same name. For instance, if I have a component called `MyComponent`, the files should look like 
+
+```
```
yangshun: I usually put them in a directory of the same name but in kebab case such as in the components. I think it may be neater to do that? But I'm ok with this too.

ZhangYiJiang: Actually if we do that then shouldn't the structure be 

```
- MyComponent
  |-- index.js              - Component definition 
  |-- MyComponent.scss      - Styles 
  └-- MyComponent.test.jsx  - Tests 
```

PR #476, opened 2017-11-03
Public holidays scraper

yangshun:
A supporting PR for #469 . This PR adds a scraper to fetch public holiday ICS files from MOM website and converts them into a CSV format that is similar to https://github.com/rjchow/singapore_public_holidays, which reduces our external reliance.

### DIFF
--- a/v3/scripts/public-holidays-ics-to-csv.js
+++ b/v3/scripts/public-holidays-ics-to-csv.js
@@ -31,10 +31,10 @@ async function fetchPublicHolidaysIcs(year) {
   return res.data;
 }
 
-years.split(',').map(s => s.trim()).forEach(year => {
-  fetchPublicHolidaysIcs(year).then(icsData => {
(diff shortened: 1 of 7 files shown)


v3/scripts/public-holidays-ics-to-csv.js
```diff
@@ -0,0 +1,79 @@
...
+
+const args = process.argv.slice(2);
+const years = args[0];
+if (!years) {
+  console.error('Please specify the year(s).');
+  process.exit(1);
+}
+
+const OUT_DIR = path.join(__dirname, 'holidays');
+if (!fs.existsSync(OUT_DIR)) {
```
ZhangYiJiang: /nit (Not particularly important since the output from the script is also checked in, so the folder's existence is guaranteed) To do this properly I believe you need something like [mkdirp](https://github.com/substack/node-mkdirp) since `mkdir` is not recursive

PR #481, opened 2017-11-10
Use our own component to display Disqus threads

ZhangYiJiang:
Turns out we don't really need no stinkin' library for this. 

This implementation is a simplified version of disqus-react's [`<DiscussionEmbed>`](https://github.com/ryanvalentin/disqus-react/blob/master/src/DiscussionEmbed.jsx), with most of the support code used to support reactive `shortCode` removed since we don't need that (basically most of `cleanInstance()` and `componentWillUpdate`), and a fix for what looks like a potential race condition in the original code (it checks `document.getElementById('dsq-embed-scr')` to check that the script element is created, but then tries to access  `window.DISQUS`, which may not have been loaded yet).


v3/src/js/views/components/DisqusComments.jsx
```diff
@@ -0,0 +1,55 @@
+// @flow
+
+import React, { PureComponent } from 'react';
+import config from 'config';
```
yangshun: I think this component should not rely on `config`.

v3/src/js/views/components/DisqusComments.jsx
```diff
@@ -0,0 +1,55 @@
...
+      // See https://help.disqus.com/customer/portal/articles/472107
+      window.DISQUS.reset({
+        reload: true,
+        config: this.getDisqusConfig(),
+      });
+    } else {
+      // Inject the Disqus script if we're loading it for the first time, ie. when
+      // window.DISQUS is not set
+      window.disqus_config = this.getDisqusConfig();
+      window.disqus_shortname = config.disqusShortname;
```
yangshun: Should receive the short name as a `prop` instead.

ZhangYiJiang: I think that is better with regards to separation of concern, but worse as an API design because `shortname` will not be reactive - unlike the one from Disqus, which does handle changing shortname, in our case shortname **must** be the same **everywhere** this component is used (since the Disqus script is loaded from `{shortname}.disqus.com`, and only loaded once). In that case it seems better to not have the prop at all.

yangshun:
I still don't think it should be put as part of the component. What if for some reason we want to load Disqus comments from somewhere else in future. But it's a small thing that can be easily refactored out if and when there's a need to. Not a blocker.

ZhangYiJiang:
If that is a requirement we would just use the library directly. Not using it allows us to discard more than 50% of the code in that component, since most of the code in the component is actually there to support that (it's kind of obvious that the Disqus embed code was never properly designed for this if you read the source).

PR #484, opened 2017-11-14
Feature/react 16 list

li-kai:
Implementation has changed in the following:
1. Modules added are still in the list but disabled
1. List full screens in mobile

This helps in searching on mobile.

screenshot 2017-12-18 12 05 04


v3/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -144,7 +115,10 @@ class TimetableModulesTable extends Component<Props> {
               <div className={classnames(styles.moduleActionColumn, styles.moduleDetailsColumn)}>
                 {this.renderModuleActions(module)}
 
-                <Link to={modulePagePath(module.ModuleCode, module.ModuleTitle)}>
+                <Link
+                  href={modulePagePath(module.ModuleCode, module.ModuleTitle)}
```
ZhangYiJiang: Isn't `<Link>` href automatically generated? https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L75

v3/src/js/views/components/Select.jsx
```diff
@@ -0,0 +1,83 @@
...
+          className={classnames(styles.input, 'form-control')}
+          placeholder={this.props.placeholder}
+          title={this.props.label}
+          aria-label={this.props.label}
+          aria-haspopup="false"
+          aria-autocomplete="list"
+          aria-expanded={this.state.isOpen}
+          aria-controls={this.selectId}
+          type="text"
+          dir="ltr"
```
ZhangYiJiang: Some attributes of these seem a little excessive? Did you copy them from `react-select`?

v3/__mocks__/modules/index.js
```diff
@@ -0,0 +1,7 @@
+import ACC2002 from './ACC2002.json';
+import BFS1001 from './BFS1001.json';
+import CS1010S from './CS1010S.json';
+import CS3216 from './CS3216.json';
+import PC1222 from './PC1222.json';
+
+export default [ACC2002, BFS1001, CS1010S, CS3216, PC1222];
```
ZhangYiJiang: You missed GES1021

v3/src/js/reducers/entities/moduleBank.js
```diff
@@ -91,10 +67,17 @@ export function getSemModuleSelectList(
...
   semTimetableConfig: SemTimetableConfig,
 ): ModuleSelectListItem[] {
-  return state.moduleSelectList.filter((item: ModuleSelectListItem): boolean => {
-    // In specified semester and not within the timetable.
-    return _.includes(item.semesters, semester) && !semTimetableConfig[item.value];
-  });
+  return state.moduleList
+    .filter((item) => {
+      // In specified semester and not within the timetable.
+      return item.Semesters.includes(semester);
```
ZhangYiJiang: /nit Can use shorthand notation here to remove the `return`

v3/src/js/reducers/entities/moduleBank.js
```diff
@@ -91,10 +67,17 @@ export function getSemModuleSelectList(
...
+    })
+    .map((mod) => {
+      return {
```
ZhangYiJiang: Ditto above

v3/src/js/views/timetable/TimetableModulesTable.scss
```diff
@@ -26,12 +26,12 @@
...
+}
 
-  @include media-breakpoint-up(md) {
-    .moduleAction {
-      padding: 0.5rem;
-      border: 0;
-      border-radius: 0;
-    }
+@include media-breakpoint-up(md) {
+  .moduleAction {
```
ZhangYiJiang: The selector won't have enough specificity if you move it outside `.moduleActionButtons`

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,39 +1,115 @@
...
+                });
+              return (
+                <li
+                  {...props}
+                  className={classnames(styles.option, {
+                    [styles.optionSelected]: highlightedIndex === index,
+                    [styles.optionDisabled]: module.isAdded,
+                  })}
+                >
+                  {module.ModuleCode} {module.ModuleTitle}
```
ZhangYiJiang: Would be good to highlight the matched part of the string. Be careful of rendering performance though.

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,39 +1,115 @@
...
+                  {module.isAdded && (
+                    <div>
+                      <span className="badge badge-info">Added</span>
+                    </div>
+                  )}
+                </li>
+              );
+            })}
+            <li className={styles.item}>
+              Try CS1010 or Programming. Searching <strong>{this.props.moduleList.length}</strong> modules.
```
ZhangYiJiang: /nit 'CS1010' and 'Programming' should have quotation marks around them.

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,39 +1,115 @@
...
+    if (!inputValue) {
+      return [];
+    }
+    // $FlowFixMe
+    const moduleCodes = createSearchFilter(inputValue).initFilters(this.props.moduleList).filteredModules();
+    const results = this.props.moduleList.filter((mod) => {
+      return moduleCodes.has(mod.ModuleCode);
     });
+    // $FlowFixMe why don't width subtyping help here?
+    return sortModules(inputValue, results.slice(0, 500));
```
ZhangYiJiang: 500 should be extracted out to a constant

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,39 +1,115 @@
...
+  onChange = (selection: any) => {
+    return selection && this.props.onChange(selection);
+  };
+
+  getFilteredModules = (inputValue: string): ModuleSelectList => {
+    if (!inputValue) {
+      return [];
+    }
+    // $FlowFixMe
+    const moduleCodes = createSearchFilter(inputValue).initFilters(this.props.moduleList).filteredModules();
```
ZhangYiJiang: Why use `ModuleFilter` for this? The `ModuleFilter` is nothing more than a wrapper around a predicate, and if you don't need the extra stuff, like the ability to chain multiple filters together, then you should just extract the predicate function out and use that directly. This doesn't work because it is expecting `Module` objects, which these are not.

v3/src/js/views/AppShell.jsx
```diff
@@ -97,15 +95,7 @@ export class AppShell extends Component<Props> {
...
-          <form className="nm-navbar-form">
-            <ModulesSelect
-              moduleList={this.props.moduleSelectList}
-              onChange={(moduleCode) => {
-                this.props.history.push(`/modules/${moduleCode.value}`);
-              }}
-              placeholder="Search modules"
-            />
-          </form>
+          <form className="nm-navbar-form" />
```
ZhangYiJiang: This `<form>` is useless

v3/src/js/views/components/ModulesSelect.scss
```diff
@@ -0,0 +1,79 @@
...
+    left: 0;
+    max-height: 100%;
+  }
+}
+
+.selectList {
+  max-height: calc(100% - #{$input-height});
+  padding: 0;
+  margin: 0;
+  overflow-y: scroll;
```
ZhangYiJiang: Should be `auto`, otherwise the scrollbar will be present even when there's no overflow

v3/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -266,7 +266,7 @@ class TimetableContent extends Component<Props> {
                   <ModulesSelect
                     moduleList={this.props.semModuleList}
                     onChange={(moduleCode) => {
-                      this.props.addModule(semester, moduleCode.value);
+                      this.props.addModule(semester, moduleCode);
```
ZhangYiJiang: /nit can collapse into single line

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,41 +1,156 @@
...
+    this.setState({ isModalOpen: true });
+  };
+  closeModal = () => {
+    this.setState({ isModalOpen: false });
+  };
+  updateMediaQuery = (e: MediaQueryListEvent | MediaQueryList) => {
+    if (e.matches === this.state.isDesktop) {
+      return;
+    }
+    this.setState({ isDesktop: e.matches });
```
ZhangYiJiang: Can move this inside the conditional

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,41 +1,156 @@
...
-        onChange={this.props.onChange}
+      <div className={styles.container}>
+        <label className="sr-only" {...getLabelProps()}>
+          {placeholder}
+        </label>
+        <input autoFocus={this.state.isModalOpen} className={styles.input} {...getInputProps({ placeholder })} />
+        {isOpen && (
+          <ol className={styles.selectList}>
+            {this.getFilteredModules(inputValue).map((module, index) => {
+              const props = module.isAdded
```
ZhangYiJiang: It may be easier to just use an if-else here for the entire `<li>` instead

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,41 +1,156 @@
...
+                  {module.ModuleCode} {module.ModuleTitle}
+                  {module.isAdded && (
+                    <div className={styles.badgeContainer}>
+                      <span className="badge badge-info">Added</span>
+                    </div>
+                  )}
+                </li>
+              );
+            })}
+            <li className={styles.item}>
```
ZhangYiJiang: This needs to be styled differently from the rest of the items somehow

v3/src/js/utils/moduleSearch.js
```diff
@@ -12,32 +11,48 @@ export function tokenize(str: string): string[] {
...
-        return true;
-      }
+export function createSearchPredicate(searchTerm: string): SearchableModule => boolean {
+  const searchRegex = regexify(searchTerm);
 
-      if (module.ModuleDescription) {
-        return module.ModuleDescription.toUpperCase().includes(term);
-      }
+  return function predicate(module: SearchableModule): boolean {
+    if (
```
ZhangYiJiang: I still think this is problematic without tokenization - search `sci fi` or `program metho` produces no results.

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,41 +1,156 @@
...
+type State = {
+  isModalOpen: boolean,
+  isDesktop: boolean,
+};
+
+const RESULTS_LIMIT = 500;
+const DOWNSHIFT_FLAGS = { resetInputOnSelection: true };
+
+class ModulesSelect extends Component<Props, State> {
+  mql: ?MediaQueryList;
```
ZhangYiJiang: I think this can be initialized here directly, then this doesn't need to be an optional

taneliang:
I like the new modal UI on mobile 👍 

Some comments/issues:

5. Bugs 1, 2 and 4 that I listed above have been resolved.
6. Bug 3 (text size change) is still present though. But I don't think this will affect many people so we should just ignore it for now.
7. Mobile: After selecting a module, the search box clears out but the modal still covers the screen.
8. Desktop: The previous search box implementation kept the text box in focus after a module is selected. I think we should keep this behavior - I would think that people want to add multiple mods to their timetable at one go.
9. Mobile + desktop: Deselecting the search box clears any existing search text. May be related to comment 8?
10. iOS: Module list is hidden by on screen keyboard. This, combined with bug 9, means that the modules at the bottom of the list can't be selected on iPhones

v3/src/js/views/components/ModulesSelect.jsx
li-kai: We can't actually because we require access to `window`

v3/src/js/utils/moduleSearch.js
li-kai: Hmm yes that is indeed problematic

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,42 +1,166 @@
...
+      return [];
+    }
+
+    const predicate = createSearchPredicate(inputValue);
+    const results = this.props.moduleList.filter(predicate);
+
+    return sortModules(inputValue, results.slice(0, RESULTS_LIMIT));
+  };
+
+  /* eslint-disable jsx-a11y/label-has-for, jsx-a11y/no-autofocus */
```
ZhangYiJiang: /nit Adding some documentation on why this is necessary would be a good idea

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,42 +1,166 @@
...
 import type { ModuleSelectList } from 'types/reducers';
 
-import { ModulesSearchIndex, ModulesTokenizer } from 'utils/modulesSearch';
+import styles from './ModulesSelect.scss';
 
 type Props = {
   moduleList: ModuleSelectList,
   onChange: Function,
   placeholder: string,
+  isMatchBreakpoint: boolean,
```
ZhangYiJiang: /nit This doesn't sound grammatical. I don't think it's necessary to always use `is` as a prefix for boolean variables.

v3/src/js/views/components/ModulesSelect.jsx
```diff
@@ -1,42 +1,166 @@
...
+                  {...getItemProps({
+                    key: module.ModuleCode,
+                    item: module.ModuleCode,
+                    index,
+                  })}
+                  className={classnames(styles.option, {
+                    [styles.optionSelected]: highlightedIndex === index,
+                  })}
+                >
+                  {`${module.ModuleCode} ${module.ModuleTitle}`}
```
ZhangYiJiang: /nit Would be simpler to just use React's `{}` templates here

ZhangYiJiang:
Oh, almost forgot - the UI also needs to indicate to the user when there's no results found. You can throw that onto the backlog too.

v3/src/js/views/components/ModulesSelect.jsx
yangshun: It's our convention so far, let's stick to it.

ZhangYiJiang: Not really

screenshot from 2017-12-22 00-20-47

It's fine if it fits, but I don't see the point in overthinking the name just to get the prefix.

PR #486, opened 2017-11-17
Implement exam clash alert

### LINKED ISSUE #483: Exam time clash warning
Display an alert below the timetable to indicate that two selected modules have the same exam timing. This can be done purely in the view and no changes to the Redux store is required.

taneliang:
Resolves #483.

screen shot 2017-11-16 at 6 10 13 pm

* Displays alert when there are modules which have clashing exam dates/times
* Displays clashing modules grouped by their exam date/times. This makes it easier for the user to resolve clashes. The clashing modules are displayed in the alert itself so that the user can interact with the modules directly.
* Ignores modules without exams
* Handles multiple clashing dates
* Handles >2 clashing modules on any date
* Preserves existing effective behavior when no clashes are present

Unit tests included for the new function that finds exam clashes. Also enabled import of Bootstrap's alerts SCSS.

### DIFF
--- a/v3/src/js/utils/timetables.js
+++ b/v3/src/js/utils/timetables.js
@@ -19,7 +19,7 @@ import type {
 import type { ModulesMap } from 'reducers/entities/moduleBank';
 
 import _ from 'lodash';
-import { getModuleTimetable } from 'utils/modules';
+import { getModuleTimetable, getModuleSemExamDate } from 'utils/modules';
 
 type LessonTypeAbbrev = { [LessonType]: string };
 export const LESSON_TYPE_ABBREV: LessonTypeAbbrev = {
@@ -196,4 +196,26 @@ export function colorLessonsByType(lessons: Lesson[]) {
 
     return { ...lesson, colorIndex };
   });
+}
+
+// Find all exam clashes between modules in semester
+// Returns object associating exam dates with the modules clashing on those dates
+export function findExamClashes(modules: Array<Module>, semester: Semester): { string: Array<Module> } {
+  const examDates = modules
+    .map(module => getModuleSemExamDate(module, semester))
+    .filter(date => date !== '-');
+  const uniqueExamDates = _.uniq(examDates);
+
+  // Iteratively remove unique exam dates from exam dates.
+  // Not using library/lodash diff functions since they remove
+  // _all_ instances of a unique exam date from examDates.
+  const duplicateDates = examDates; // Array eventually containing clashing exam dates
+  for (let i = 0; i < uniqueExamDates.length; i++) {
+    _.pullAt(duplicateDates, duplicateDates.indexOf(uniqueExamDates[i]));
+  }
+
+  // Fetch modules having exams on the clash dates
+  const clashingModules = duplicateDates.map(date =>
+    modules.filter(module => date === getModuleSemExamDate(module, semester)));
+  return _.zipObject(duplicateDates, clashingModules);
 }
--- a/v3/src/js/utils/timetables.js
+++ b/v3/src/js/utils/timetables.js
@@ -19,7 +19,7 @@ import type {
 import type { ModulesMap } from 'reducers/entities/moduleBank';
 
 import _ from 'lodash';
-import { getModuleTimetable, getModuleSemExamDate } from 'utils/modules';
+import { getModuleTimetable, getModuleSemesterData } from 'utils/modules';
 
 type LessonTypeAbbrev = { [LessonType]: string };
 export const LESSON_TYPE_ABBREV: LessonTypeAbbrev = {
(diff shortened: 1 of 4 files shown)


li-kai:
Of course! This implementation makes so much sense. Can’t believe I spent so long thinking hard about it.

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -102,6 +105,78 @@ class TimetableContainer extends Component<Props> {
...
+    const modules = _.keys(this.props.semTimetableWithLessons)
+      .sort((a, b) => a.localeCompare(b))
+      .map(moduleCode => this.props.modules[moduleCode]);
+    return _.compact(modules);
+  }
+
+  // Separate added modules into sections of clashing modules
+  // Returns array of added modules if there are no exam clashes
+  // Else returns object associating exam dates with the modules clashing on those dates
+  moduleSections(): Array<Module> | { [string]: Array<Module> } {
```
li-kai: This might be simpler as an array of always length 2. You can still have constants with array index. E.g. const CLASH_INDEX = 0;

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -48,6 +49,8 @@ import Timetable from './Timetable';
 import TimetableActions from './TimetableActions';
 import TimetableModulesTable from './TimetableModulesTable';
 
+const NO_CLASH_SECTION_KEY = 'NO_CLASH_SECTION_KEY';
```
li-kai: Using constants. Nice!

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -102,6 +105,78 @@ class TimetableContainer extends Component<Props> {
...
+    if (Array.isArray(moduleSections)) {
+      return renderModuleTable(moduleSections);
+    }
+
+    // Render table of modules that don't clash
+    const nonClashingTable = renderModuleTable(moduleSections[NO_CLASH_SECTION_KEY]);
+    delete moduleSections[NO_CLASH_SECTION_KEY];
+
+    // Render sections of clashing modules, with clashing dates in their headers
+    const clashingSections = Object.keys(moduleSections).sort().map(clashDate => (
```
li-kai: Just inline this within jsx, it’s readable :)

v3/src/js/utils/timetables.js
```diff
@@ -197,3 +197,25 @@ export function colorLessonsByType(lessons: Lesson[]) {
     return { ...lesson, colorIndex };
   });
 }
+
+// Find all exam clashes between modules in semester
+// Returns object associating exam dates with the modules clashing on those dates
+export function findExamClashes(modules: Array<Module>, semester: Semester): { string: Array<Module> } {
+  const examDates = modules
+    .map(module => getModuleSemExamDate(module, semester))
+    .filter(date => date !== '-');
```
yangshun: This part will break if we stop using `-` to indicate no exam date. Just filter using: `_.get(getModuleSemesterData(module, semester), 'ExamDate')`like in `getModuleSemExamDate`.

v3/src/js/utils/timetables.js
```diff
@@ -197,3 +197,25 @@ export function colorLessonsByType(lessons: Lesson[]) {
     return { ...lesson, colorIndex };
   });
 }
+
+// Find all exam clashes between modules in semester
+// Returns object associating exam dates with the modules clashing on those dates
+export function findExamClashes(modules: Array<Module>, semester: Semester): { string: Array<Module> } {
```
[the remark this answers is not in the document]
yangshun: I agree. It took me a while to verify the correctness of the code. It'll be easier to use `_.groupBy` on exam date to get something like:

`{ date1: [module1, module2], date2: [module3] }`

and simply removing the K/Vs where there is only one item in the array.

v3/src/js/utils/timetables.js
```diff
@@ -197,3 +197,25 @@ export function colorLessonsByType(lessons: Lesson[]) {
...
+    .map(module => getModuleSemExamDate(module, semester))
+    .filter(date => date !== '-');
+  const uniqueExamDates = _.uniq(examDates);
+
+  // Iteratively remove unique exam dates from exam dates.
+  // Not using library/lodash diff functions since they remove
+  // _all_ instances of a unique exam date from examDates.
+  const duplicateDates = examDates; // Array eventually containing clashing exam dates
+  for (let i = 0; i < uniqueExamDates.length; i++) {
+    _.pullAt(duplicateDates, duplicateDates.indexOf(uniqueExamDates[i]));
```
yangshun: The number of modules can be small, but this function is highly inefficient - O(n^2). Also, if you have 3 clashing dates, only one of them will be removed and `duplicateDates` will be left with two same dates. Doesn't really matter though because `_.zipObject` will combine them together.

v3/src/js/utils/timetables.js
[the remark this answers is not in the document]
taneliang: Wow I didn't know `_.groupBy` existed. Thanks for the tip!

v3/src/js/views/timetable/TimetableContainer.jsx
taneliang: Do you mean that I should return something like that:
```
[
    [module1, module2], // Modules which don't clash
    {
        clashDate1: [module3, module4],
        clashDate2: [module5, module6],
        ...
    }
]
```

taneliang:
Thanks guys! I also found a potential bug [in the timify tests](https://github.com/nusmodifications/nusmods/blob/master/v3/src/js/utils/timify.test.js#L10) - Line 10 imports something from timetables.test.js, which causes Jest to run tests from both files when testing timify.test.js.

v3/src/js/views/timetable/TimetableContainer.jsx
li-kai: Yep, exactly!

taneliang: You're totally right. That was overcomplicated. Since the function can then be reduced to just 3 lines, I just merged that function with `renderModuleSections`

taneliang:
All done. @yangshun should we polish the UI?

yangshun:
@taneliang No worries, can polish it in a separate PR. Can ship first!

PR #487, opened 2017-12-01
Exit from storage functions unless localStorage exists

### LINKED ISSUE #480: ReferenceError: localStorage is not defined
https://sentry.io/share/issue/5b2abcb6db4047b7a670a7f152c483e0/

```
ReferenceError: localStorage is not defined
  at getItem (./js/storage/index.js:19:5)
  at loadState (./js/storage/index.js:44:12)
  at Object.266 (./js/main.js:14:32)
  at __webpack_require__ (/:1:119)
  at Object.265 (app.0db6babbb66a82a6b4d3.js:1241:18)
...
(12 additional frame(s) were not displayed)
```

luyangkenneth:
Fixes #480

`getItem()` requires a `return undefined;` because of the [consistent-return rule](https://eslint.org/docs/rules/consistent-return).

### DIFF
--- a/v3/src/js/storage/index.js
+++ b/v3/src/js/storage/index.js
@@ -5,6 +5,8 @@ import Raven from 'raven-js';
 // TODO: Use an in-memory storage for environments where localStorage is not present,
 //       like private mode on Safari.
 function setItem(key, value) {
+  if (!window.localStorage) return;
+
   try {
     localStorage.setItem(key, _.isString(value) ? value : JSON.stringify(value));
   } catch (e) {
@@ -13,6 +15,8 @@ function setItem(key, value) {
 }
 
 function getItem(key) {
+  if (!window.localStorage) return undefined;
+
   let value;
   try {
     value = localStorage.getItem(key);
@@ -27,6 +31,8 @@ function getItem(key) {
 }
 
 function removeItem(key) {
+  if (!window.localStorage) return;
+
   try {
     localStorage.removeItem(key);
   } catch (e) {


yangshun:
Hi @luyangkenneth , thanks for your contribution! Unfortunately, this is not the best way to fix the issue. Ideally we want to shim `localStorage` by using an in-memory key/value store in environments where `localStorage` does not exist, such as Safari private browsing mode, so that v3 will still work in those envs. If you're interested in fixing this, modify [storage/index.js](https://github.com/nusmodifications/nusmods/blob/master/v3/src/js/storage/index.js) such that it reads from/writes to a Map/Object if it detects that `window.localStorage` is not present.

ZhangYiJiang:
@yangshun In Safari private browsing mode `window.localStorage` still exists. It throwing exception is a bug - websites shouldn't be able to detect private browsing mode. [This should be fixed in Safari 11](https://bugs.webkit.org/show_bug.cgi?id=157010), and is probably not worth pursuing. 

The bug we're fixing here is when the code is run in a non-browser env, eg. JSDom, where the entire `window.localStorage` object does not exist. I suppose there may be some use cases for replacing it with an in-memory implementation (for integration testing), but I'm a little skeptical about how useful this is compared to simply making the functions do nothing.

luyangkenneth:
Thanks @yangshun and @ZhangYiJiang for your guidance! I'll be glad to implement the in-memory version of `localStorage` if that's the direction we're taking. I'll work on other areas in the meantime 🙂

yangshun:
@ZhangYiJiang I understand the issue. Why not shim `window.localStorage` so that it can let both the tests pass and the app not to crash when user is on private Safari? Private Safari exists not only on MacOS; iOS has a similar bug as well. One day we're gonna get bug reports of someone opening the URL on their iPhone with Safari defaulted to private mode and they wonder why they're not seeing anything (I got a bug report like that at work before).

yangshun:
We'll fix this in another PR. Thanks for your contribution!

Issue #107, opened 2014-07-31
Venn diagram display/selector for semesters filter


ahbeng:
As discussed with @ashrayjain, there might be scenarios where 'Semester 1 only', 'Semester 2 only' and 'Semester 1 AND Semester 2' filters might come in handy. If this set replaced the current filter set, it would take 2 clicks to filter the presumably more common scenario of any modules offered in Semester 1.

A venn diagram representation might be clearer and faster to interact with - tap to select region and drag to select multiple regions. The size/overlap could be adjusted based on the actual module counts. This would probably be done using D3.

[Venn.js](https://github.com/benfred/venn.js) might be handy for this.

yangshun:
Closing in favor of #426

PR #491, opened 2017-12-05
Semester filter and exam clash warning on module finder/info page 

ZhangYiJiang:
This PR implements semester filters and a warning about exam date clashes. 

Regarding #107, I'm not entirely sure if the extra effort is worth it - I can see some use for this, but it seems rare and designing four circle Venn diagram that is actually usable is not my idea of fun. 

The other issue is the UI for the exam clash warning. Currently it looks like this 

screenshot from 2017-12-05 15-10-53
screenshot from 2017-12-05 15-37-19

which works, but I don't think it does the job well enough. I've updated the colors to differentiate it more from theme-primary(), but it still blends in a little. We can revisit this at a later date.


yangshun:
Let's deal with the UI later, it works at the moment.

v3/src/js/views/components/module-info/ModuleExamClash.jsx
```diff
@@ -0,0 +1,64 @@
...
+import type { Module, Semester, ModuleCode } from 'types/modules';
+
+import { AlertTriangle } from 'views/components/icons';
+import { getModuleSemesterData } from 'utils/modules';
+import { getSemesterModules } from 'utils/timetables';
+import { intersperse } from 'utils/array';
+
+import styles from './ModuleExamClash.scss';
+
+const EMPTY_OBJECT = {};
```
yangshun: Not necessary IMO, can just inline this.

v3/src/js/views/components/module-info/ModuleExamClash.jsx
```diff
@@ -0,0 +1,64 @@
+// @flow
```
yangshun: I think this component should not connect to the Redux store at all. Is it feasible to do the clashing computation in some component higher up in the hierarchy that is connected to the store, rather than in this child component?

v3/src/js/views/components/module-info/ModuleExamClash.jsx
```diff
@@ -0,0 +1,64 @@
...
+  semester: Semester,
+  examDate: ?string,
+  modules: Module[],
+};
+
+/**
+ * Shows a warning if the provided examDate clashes with the exam of any modules
+ * that are already taken
+ */
+export class ModuleExamDateComponent extends PureComponent<Props> {
```
yangshun: Should the name be `ModuleExamDateComponent` or `ModuleExamClash`? I don't think it's necessary to have `Component` in the name.

v3/src/js/views/components/module-info/ModuleExamClash.jsx
ZhangYiJiang: It's to stop the component from doing unnecessary rerenders - if it was inlined a new object will be created every time, which will always cause rerender. (And not as I imagine you thinking, a sign of brain damage inflicted by CS2103)

v3/src/js/views/components/module-info/ModuleExamClash.jsx
ZhangYiJiang: Oh, forgot to rename this after renaming the file. It should be `ModuleExamClashComponent`. The additional 'Component' is to stop ESLint from complaining if I try to import just the component minus the Redux HOC into unit tests (If you try to `import { X } from 'X'` when X already has a default export, ESLint will ask if you actually meant to do `import X from X`).

v3/src/js/views/components/module-info/ModuleExamClash.jsx
ZhangYiJiang: I had thought about this, but I don't see any way to make it work - nothing else both of these pages use `state.timetables` at all, and on the module finder in particular this is fairly deeply nested.

v3/src/js/views/components/module-info/ModuleExamClash.jsx
yangshun: Can turn off the ESLint rule for this if you wish. I remember encountering this before but I forgot the way I solved it. Either I turned off the rule or found a workaround. Perhaps you can try seeing where else in the app we're doing something like this.

v3/src/js/views/components/module-info/ModuleExamClash.jsx
yangshun: Ok then

v3/src/js/views/components/module-info/ModuleExamClash.jsx
yangshun: Are you sure about this? It doesn't look like the object is directly used in the React component.

ZhangYiJiang: Oh no, you're right. My bad.

ZhangYiJiang:
This should be ready for review

PR #493, opened 2017-12-07
Dark mode for v3 

ZhangYiJiang:
## Background 

This is a WIP build of dark mode, to bring the feature from v2 to v3.

## Approach 

[Custom CSS properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) is used here to greatly reduce code duplication, and [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) is used to add in the fallback for browsers that doesn't support it.  This approach is much simpler than the SCSS based approach discussed previously. 

screenshot from 2017-12-07 12-54-27
screenshot from 2017-12-07 12-54-41

(The dark checkboxes are from my browser theme, not because of CSS) 

Currently there's no UI for toggling dark mode, so to test run this line of code in the console 

    document.body.classList.add('dark')


v3/src/styles/utils/bootstrap-override.scss
```diff
@@ -0,0 +1,11 @@
+// Override Bootstrap's styles
+body {
+  color: var(--body-color);
+  background: var(--body-bg);
+}
+
+.form-control,
+.form-control:focus {
```
yangshun:
```
.form-control {
  &,
  &:focus {
    color: var(--body-color);
    background: var(--body-bg);
  }
}

ZhangYiJiang: To be honest I think that looks less readable than just repeating `.form-control`

ZhangYiJiang:
> Prefixing all the CSS variables with --nusmods-theme to differentiate if a variable is meant to be used in the theme

I don't understand what you mean by this. Are you talking about the timetable theme colors? I don't think those need to be changed for dark mode. The only colors that are changed so far are shades of gray and the workload component colors

v3/src/styles/utils/bootstrap-override.scss
yangshun: Leave it then.

yangshun:
> Are you talking about the timetable theme colors?

Sorry I mean a mode, not a theme. Got confused by the terminologies. Please `s/theme/mode` the above comment.

PR #496, opened 2017-12-08
New CORS bidding stats UI for v3 

ZhangYiJiang:
### Context 

The current CORS bidding stats UI is basically a cut and paste from the original tabular CORS data, which makes it basically unusable. This PR attempts to simplify the UI to make it somewhat usable. 

### Approach 

When you think about it, what students actually want to know when they access CORS bidding data is actually quite simple - 

- How much bid points should I expect to pay 
- How many slots are available 

The problem is the CORS bidding process is an unholy mess. I started by merging all sectional/lecture groups, and removing the max/min bid points. Then I used first round quotas as an approximation of how many slots there, and sum of all successful bids as an approximation of how many students managed to fill the slots. 

CORS has three rounds and two account types, but if you look at you are allowed to spend your points, there basically are three 'tiers' 

- Program account, new student - used in Round 1C 
- Program account, returning students - used in Round 1A, 1B
- General account - used in Round 2A, 2B

In addition, some modules are required by more than one faculty, so that need to be taken care of too. These two set of numbers make up the summary section 

CS1010S - an oversubscribed module with students from two faculties 

screenshot from 2017-12-08 23-34-25

PS1101E - a half filled module with students from just one faculty 

screenshot from 2017-12-09 00-16-15

This is very much a WIP. The code needs more cleaning up, the quota calculation appears to be incorrect for several modules and there needs to be a way for the user to see all of the numbers. 

I do want to check what you guys think of the direction of the UI and this idea as a whole. And also if I've made any glaring mistakes in understanding how CORS works.


v3/src/js/utils/cors.js
```diff
@@ -1,43 +1,132 @@
...
-        || (!sameFaculty(stat, student) && isGeneralAccount(student));
+      return NEW_STUDENT | RETURNING_STUDENT | GENERAL_ACCOUNT;
     case 'Reserved for [G] in later round':
-      return !sameFaculty(stat, student) && isGeneralAccount(student);
     case 'Not Available for [G]':
-      return sameFaculty(stat, student) && isProgrammeAccount(student);
+      return 0;
     default:
-      throw Error(`unknown StudentAcctType ${stat.StudentAcctType}`);
+      throw Error(`unknown StudentAcctType ${studentAccountType}`);
```
yangshun: Nit: Didn't see any `StudentAcctType` around. Maybe stick to `studentAccountType`?

v3/src/js/types/views.js
```diff
@@ -29,3 +30,39 @@ export type PageRangeDiff = {
 };
 
 export type OnPageChange = PageRangeDiff => void;
+
+/* components/CorsStats */
+
+export type StudentType = number;
+export const NEW_STUDENT = 1;
```
yangshun: Should define a variable for 0 as I see it being used quite a few times in `utils/cors.js`.

Would it be good to add types to `NEW_STUDENT`, e.g. `export const NEW_STUDENT: StudentType = 1`?

v3/src/js/utils/cors.js
```diff
@@ -1,43 +1,132 @@
...
-      return sameFaculty(stat, student) && isProgrammeAccount(student) && !isNew(student);
+      return RETURNING_STUDENT;
     case 'New Students [P]':
-      return sameFaculty(stat, student) && isProgrammeAccount(student) && isNew(student);
+      return NEW_STUDENT;
     case 'NUS Students [P]':
-      return sameFaculty(stat, student) && isProgrammeAccount(student);
     case 'Returning Students and New Students [P]':
-      return sameFaculty(stat, student) && isProgrammeAccount(student);
+      return NEW_STUDENT | RETURNING_STUDENT;
```
yangshun: I think not many beginners are familiar with this form of enums. You might want to add some [reference articles](http://www.alanzucconi.com/2015/07/26/enum-flags-and-bitwise-operators/) as comments either here or in the types file so that future maintainers know what this is doing.

v3/src/js/utils/cors.js
```diff
@@ -1,43 +1,132 @@
...
+      Faculty,
+      Semester,
+      Round,
+
+      StudentType: appliesTo(StudentAcctType),
+      Quota: sumBy(statsGroup, stats => Number(stats.Quota)),
+      Bidders: sumBy(statsGroup, stats => Number(stats.Bidders)),
+      LowestSuccessfulBid: min(statsGroup.map(stats => Number(stats.LowestSuccessfulBid))),
+    };
+  }).filter(stat => stat.StudentType !== 0);
```
yangshun: Magic number. Refer to comment above about 0.

v3/src/js/utils/cors.js
```diff
@@ -1,43 +1,132 @@
...
+        minBid: stat.LowestSuccessfulBid,
+        round: stat.Round,
+      };
+    }
+  }
+
+  stats.forEach((stat) => {
+    if (stat.Bidders === 0) return;
+
+    if (stat.StudentType & NEW_STUDENT) updateSummary(stat, NEW_STUDENT);
```
yangshun: How about this instead:

```js
const studentTypes: Array<StudentType> = [NEW_STUDENT, RETURNING_STUDENT, GENERAL_ACCOUNT];
studentTypes.forEach(studentType => updateSummary(stat, studentType);
```

v3/src/js/views/components/module-info/CorsQuota.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+import styles from './CorsStats.scss';
+
+type Props = {
+  bidders: number,
+  quota: number,
+};
+
+export default function ({ bidders, quota }: Props) {
+  let color = 'success';
```
yangshun: Let's try to avoid building up strings from concatenation/templating as far as possible so as to let it be [greppable](http://jamie-wong.com/2013/07/12/grep-test/). Explicitly repeating `text-` is ok IMO.

v3/src/js/views/components/module-info/CorsRound.jsx
```diff
@@ -0,0 +1,98 @@
...
+  [NEW_STUDENT | RETURNING_STUDENT | GENERAL_ACCOUNT]: 'All students',
+};
+
+function renderBidRow(stats: GroupedBiddingStat) {
+  if (!stats.Bidders) {
+    return <td className={styles.noBids} colSpan="2">No bids</td>;
+  }
+
+  return [
+    <td key="quota">
```
yangshun: Are the keys necessary here?

v3/src/js/views/components/module-info/CorsQuota.jsx
```diff
@@ -0,0 +1,25 @@
+// @flow
```
yangshun: Can we put all these CORS-related components in a separate `cors` folder (either under `components/module-info` or under `components`.

v3/src/js/views/components/module-info/CorsStats.scss
```diff
@@ -0,0 +1,102 @@
...
+  margin-bottom: 0;
+  font-size: 0.8rem;
+}
+
+.facultyRow {
+  margin-bottom: 1rem;
+}
+
+.quota {
+  font-weight: bold;
```
yangshun: There are existing Bootstrap variables for this I believe, use them?

v3/src/styles/bootstrap/variable-overrides.scss
```diff
@@ -7,16 +7,10 @@ $gray-light:                #aeb1b5;
 $gray-lighter:              #d3d6db;
 $gray-lightest:             #f5f7fa;
 
-$theme-colors: (
```
yangshun: Please make sure no one is using this map before removing.

ZhangYiJiang: That was actually an override for Bootstrap's `$theme-colors` map. In beta 2 they made it such that we can override the theme variables individually instead of replacing this map as a whole. Can extract this commit out as a separate PR if need be.

yangshun: No it's fine, I just had to make sure. Thanks!

v3/src/js/types/views.js
ZhangYiJiang: I don't think that actually matters to Flow though, but I guess it makes it clearer to the reader :+1:

v3/src/js/utils/cors.js
ZhangYiJiang: The name in the data structure is `StudentAcctType`. `studentAcctType` is just a local variable name, so I think it's okay

v3/src/js/types/views.js
yangshun: Yep it's for the reader.

v3/src/js/views/components/module-info/CorsRound.jsx
ZhangYiJiang: Yes - not having this triggers a key prop warning. It also happened to the `dt`/`dd` on the module info page. 

```
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `CorsRound`. See https://fb.me/react-warning-keys for more information.
    in td (created by CorsRound)
    in CorsRound (created by CorsSemester)
```

ZhangYiJiang:
It actually doesn't show all of the semesters, but rather only shows the latest three. I've just added a button to show more. The idea behind showing multiple semesters is that generally info from just one semester isn't enough to see if eg. this mod is commonly oversubscribed, or if it was just that one semester.

v3/src/js/views/browse/ModulePageContent.jsx
```diff
@@ -175,6 +168,19 @@ class ModulePageContentComponent extends Component<Props> {
               />
             </section>
 
+            <section id="bidding-stats">
```
li-kai: Gotta update the scrollspy since you moved this down.

v3/src/js/utils/cors.js
```diff
@@ -1,41 +1,134 @@
...
+    if (!summary[stat.Faculty]) {
+      summary[stat.Faculty] = {};
+    }
+
+    // If there are no bidders for a round, we assume anyone could have gotten it for 1 pt
+    // (even though nobody actually bid for it)
+    const minBid = stat.Bidders ? stat.LowestSuccessfulBid : 1;
+
+    // Update the summary with the new minimum bid
+    if (!summary[stat.Faculty][type] || minBid < summary[stat.Faculty][type].minBid) {
```
li-kai: Might want to make it `<=` or approach this a different way. Currently if there are no stats for round 1A, but 1 bid point for round 1C, it'll report it as round 1A - 1 bid pt.

It's mildly confusing haha. (See returning student)
deploy-preview-496--nusmods netlify com_modules_acc1002_financial-accounting iphone 6

v3/src/js/views/components/cors-stats/styles.scss
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  .roundInfo {
+    display: block;
+    font-size: 0.8rem;
+    color: var(--gray-light);
+  }
+
+  .quota {
+    font-size: 1.5rem;
```
li-kai: I'd tune this down just a tad.

v3/src/js/utils/cors.js
ZhangYiJiang: I don't understand what you mean - the results in the screenshots seem correct.

PR #497, opened 2017-12-09
Fix module page infinite loading bug 

ZhangYiJiang:
When a modules page URL uses lowercase module code, eg. https://latest.nusmods.com/modules/cs1101s the page never stops loading because of a small bug in how lowercase module code is handled. 

This PR fixes the bug, and adds some unit tests (which doesn't actually help with the bug in question since the bug happened in `mapStateToProps`, but I thought it might be useful to have them anyway).

### DIFF
--- a/v3/src/js/views/browse/ModulePageContainer.jsx
+++ b/v3/src/js/views/browse/ModulePageContainer.jsx
@@ -16,7 +16,9 @@ import ErrorPage from 'views/errors/ErrorPage';
 import LoadingSpinner from 'views/components/LoadingSpinner';
 import { modulePagePath } from 'utils/modules';
 
-type Props = ContextRouter & {
+type Props = {
+  ...ContextRouter,
+
   moduleCode: ModuleCode,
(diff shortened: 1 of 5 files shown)


ZhangYiJiang:
Can manually confirm that https://deploy-preview-497--nusmods.netlify.com/modules/cs1101s works

v3/src/js/views/browse/ModulePageContainer.jsx
```diff
@@ -16,7 +16,9 @@ import ErrorPage from 'views/errors/ErrorPage';
 import LoadingSpinner from 'views/components/LoadingSpinner';
 import { modulePagePath } from 'utils/modules';
 
-type Props = ContextRouter & {
+type Props = {
+  ...ContextRouter,
```
li-kai: prop spreading works now?

ZhangYiJiang: I have no idea, to be honest. Flow's object union/spread types don't make sense to me.

Issue #458, opened 2017-10-06
Venue names can have slashes in their names 


ZhangYiJiang:
This affects v2 currently, but this needs to be taken into account for v3 too because I think it react-router would suffer from this problem too. 

For venue availability, the routes are defined as `venue/{venue name}`, but if the venue name contains a slash, that gets parsed as part of the URL. Venue's like https://nusmods.com/venues/CQT/SR0622 are not linked to correctly - the correct URL should be https://nusmods.com/venues/CQT%2FSR0622 (notice v2 actually "redirects" you to the incorrect URL when you visit the page).

PR #498, opened 2017-12-09
Venue availability information page

taneliang:
New Venues page. The UI is... rather focused on functionality at the expense of form.

localhost_8080_venues_t

### Features

- Adds route to `/venues/:venue?`
- Displays a list of venues at `/venues`
- Venues searchable by venue name. Search is case-insensitive
- Venues in list expand to show their classes in a timetable. Only one venue can be expanded at any time
- Handles back/forward navigation
- Tests included

### Venues URL

- Expanding a venue also pushes `/venues/<venue>` into the browser history. Some examples:
    - <https://deploy-preview-498--nusmods.netlify.com/venues/LT28>
    - Addresses #458: <https://deploy-preview-498--nusmods.netlify.com/venues/CQT%2FSR0622>
    - Case-insensitivity: <https://deploy-preview-498--nusmods.netlify.com/venues/Cqt%2fsr0622>
- The venue URL works by filtering the list of venues by `<venue>` and expanding that venue if it exists
    - Venue whose name is a substring of other venues: <https://deploy-preview-498--nusmods.netlify.com/venues/LT2>
    - Non-existent venue: <https://deploy-preview-498--nusmods.netlify.com/venues/erc> and <https://deploy-preview-498--nusmods.netlify.com/venues/covfefe>
    - Old malformed URLs generated by v2 will only filter by the component immediately after `/venues/`: <https://deploy-preview-498--nusmods.netlify.com/venues/CQT/SR0622>

### Future Work

The scope of this PR does not include searching venues by their availability - this will be implemented in a future PR.


ZhangYiJiang:
List of bugs - don't worry about fixing all of them, this is just to keep track of them so they can be fixed at some point 

- Scroll position needs to be updated when opening/closing venues, otherwise the difference in height can cause the newly opened timetable to shift off screen 
- Timetable coloring is not done correctly - the colors need to be assigned per module, not per lesson type 
- Timetable arrangement seems incorrect for some venues - eg. https://deploy-preview-498--nusmods.netlify.com/venues/dRm3
- Search results need to be sorted 
- Search query needs tokenization

taneliang:
@ZhangYiJiang To summarize, should the behavior be:
- As user searches, URL will be updated to `/venues?q=<searchTerm>`, with `&day=<searchDay>&startTime=<startTime>&endTime=<endTime>` following in the future.
- When a venue is selected, URL will be updated to `/venues/<venue>?q=<searchTerm>&...`
- The individual query params will only be present if set by the user

taneliang:
I don't think the timetable rendering issue is caused by this PR. Maybe that should be fixed in a separate PR to limit the scope of this PR?

For the scroll position updating, does anyone know of a non-hacky way to scroll the element below the search box? I have [a rather hacky solution](https://github.com/nusmodifications/nusmods/pull/498/commits/53df45c17d24f53e8b13b9638d0bbf8604b173e9#diff-26773a8ec69c3d2246fb59a4b19dbce6R84) for that by just moving the element down by 40px, but this is inelegant and doesn't work super well as the search box is 30px tall on mobile and 40px on tablets and up.

ZhangYiJiang:
> To summarize, should the behavior be: [...]

Sounds correct, but I'm not sure what you mean by 

> The individual query params will only be present if set by the user

---

> I don't think the timetable rendering issue is caused by this PR. 

I'll throw it onto the backlog then. 

---

> For the scroll position updating, does anyone know of a non-hacky way to scroll the element below the search box?

The best way would actually be to read the actual height of the nav header DOM element. Barring that, you can use the [media query util functions](https://github.com/nusmodifications/nusmods/blob/master/v3/src/js/utils/react.jsx) to select the correct height. Or you could just use a large enough number that it would always appear on screen (on larger screens the top will not be aligned properly, but it's a very minor detail). There are also CSS-based solutions, but they're not very clean (add a padding equal to the nav height, then use negative margin to offset that).

taneliang:
> Sounds correct, but I'm not sure what you mean by
>> The individual query params will only be present if set by the user

Sorry the phrasing there was unclear. What I meant was like `q=<searchTerm>` will only be present if the user typed something into the search box. So if the user only selected a time range, the URL will be `/venues?startTime=<startTime>&endTime=<endTime>`.

> The best way would actually be to read the actual height of the nav header DOM element.

Okay I'll do it this way. Not sure why I didn't consider this method. Thanks!

Also one more thing - the color coding of the modules may be confusing as many venues are occupied by more than 8 modules, but there are only 8 colors. v2 avoided that by simply not coloring anything. I think it's clearer with colors than without, but the colors may also be misleading. Thoughts?

image

ZhangYiJiang:
> What I meant was like q=<searchTerm> will only be present if the user typed something into the search box. So if the user only selected a time range, the URL will be /venues?startTime=<startTime>&endTime=<endTime>.

Yup, that's exactly right. 

> Also one more thing - the color coding of the modules may be confusing as many venues are occupied by more than 8 modules, but there are only 8 colors. v2 avoided that by simply not coloring anything. I think it's clearer with colors than without, but the colors may also be misleading. Thoughts?

Well, in the example you have above the problem is caused by double coded GE modules (although I think the problem will remain even after deduplicating them). We'll need some empirical evidence (how many venues suffer from this problem, and confused do users get), but my intuition is that coloring is better, so let's just go with that for now.

taneliang:
> coloring is better, so let's just go with that for now

Agreed. Anyway I've counted, and 259 out of 621 venues have more than 8 unique modules in AY17/18 sem 1, and 253/561 venues also suffer from this in sem 2.

v3/src/js/utils/colors.test.js
```diff
@@ -0,0 +1,71 @@
+// @flow
```
yangshun: Thanks for adding tests 😄

v3/src/js/views/venues/VenueDetailRow.jsx
```diff
@@ -0,0 +1,71 @@
...
+    onClick: noop,
+  };
+
+  arrangedLessons() {
+    if (!this.props.expanded) {
+      return null;
+    }
+
+    const availability: DayAvailability[] = this.props.availability;
+    // const lessons = flatMap(availability, a => a.Classes) // Not using flatMap as it results in a Flow error
```
yangshun: Remove this if not using.

v3/src/js/views/venues/VenueDetailRow.jsx
```diff
@@ -0,0 +1,71 @@
...
+        <h4>
+          <a
+            href={venueHref}
+            onClick={(e) => {
+              e.preventDefault();
+              onClick(name, venueHref);
+            }}
+          >{name}</a>
+        </h4>
+        {lessons ? (
```
yangshun: `lessons && (<Component />)` is shorter.

v3/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -0,0 +1,176 @@
...
+  state: State = {
+    loading: true,
+    venues: {},
+    searchTerm: this.props.urlVenue || '',
+    selectedVenue: this.props.urlVenue || '',
+    selectedVenueElement: undefined,
+  }
+
+  componentDidMount() {
+    axios.get(nusmods.venuesUrl(this.props.activeSemester))
```
yangshun: This is fine for now but I think we should consider putting the venues data into the store so that it can be potentially reused in other places.

v3/src/js/views/venues/VenueDetailRow.scss
```diff
@@ -0,0 +1,20 @@
...
+  padding-top: 1rem;
+  border-top: 1px $gray-lighter solid;
+
+  &:first-child {
+    padding-top: 0.25rem;
+    border-top: 0;
+  }
+
+  a {
+    width: 100%;
```
yangshun: Might want to use `display: block` instead.

v3/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -0,0 +1,176 @@
...
+          {pageHead}
+          <LoadingSpinner />
+        </div>
+      );
+    }
+
+    const venues = this.filteredVenues();
+
+    return (
+      <div className="modules-page-container page-container">
```
yangshun: Is it meant to be `.modules-page-container`?

v3/src/js/views/venues/VenuesContainer.test.jsx
```diff
@@ -0,0 +1,116 @@
...
+    // Unique venue with wrong case
+    component.setState({ searchTerm: 's11-0302' });
+    expect(Object.keys(component.filteredVenues())).toEqual(['S11-0302']);
+
+    // Substring
+    component.setState({ searchTerm: 'T1' });
+    expect(Object.keys(component.filteredVenues())).toEqual(['LT17', 'LT1']);
+
+    // Venue which does not exist
+    component.setState({ searchTerm: 'covfefe' });
```
yangshun: 😂

v3/src/js/views/venues/VenuesContainer.test.jsx
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  test('#onVenueSelect() should change the URL when a venue is clicked', () => {
+    const component = createComponent().instance();
+    component.onVenueSelect('LT17', '/venues/LT17');
+    expect(component.state.selectedVenue).toEqual('LT17');
+    expect(component.props.history.location.pathname).toEqual('/venues/LT17');
+  });
+
+  describe('URL handling', () => {
```
yangshun: Do we wanna test a case where a venue contains a slash?

v3/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -0,0 +1,176 @@
...
+        {pageHead}
+
+        <div className="row">
+          <div className="col-sm-12">
+            <SearchBox
+              throttle={0}
+              useInstantSearch
+              initialSearchTerm={this.state.searchTerm}
+              placeholder="Venues"
```
yangshun: I think a more descriptive placeholder will be better. How about "Search for venues. E.g. LT27"?

v3/src/js/views/venues/VenueList.jsx
```diff
@@ -0,0 +1,44 @@
...
+  const rowRefs: { [Venue]: HTMLElement } = {};
+  const { venues, expandedVenue, onSelect } = props;
+  const lowercaseExpandedVenue = expandedVenue.toLowerCase();
+
+  // Case-insensitive, natural sort of venue names
+  const sortedVenueNames = Object.keys(venues).sort((a, b) =>
+    a.toLowerCase().localeCompare(b.toLowerCase(), 'en', { numeric: true }));
+
+  return (
+    <ul className={styles.venueList}>
```
yangshun: For clarity, it would be good to show a placeholder "No venues found matching <venue name>" when there are 0 results.

v3/src/js/views/venues/VenueDetailRow.jsx
```diff
@@ -0,0 +1,71 @@
...
+  render() {
+    const { name, onClick } = this.props;
+    const lessons = this.arrangedLessons();
+    const venueHref = `/venues/${encodeURIComponent(name)}`;
+    const rootElementRef: Function = this.props.rootElementRef || (() => {});
+
+    return (
+      <li className={styles.venueDetailRow} ref={rootElementRef}>
+        <h4>
+          <a
```
yangshun: Consider wrapping `<h4>` within `<a>` so that it automatically stretches across the whole row without the need for modifying `<a>`'s styles.

v3/src/js/views/venues/VenueDetailRow.jsx
ZhangYiJiang: Actually, if this is a libdef issue, it may be better to file issue/PR against https://github.com/flowtype/flow-typed, and just use a `$FlowFIxMe` to ignore the error while the issue/PR is getting processed.

taneliang: I'm not familiar enough with Flow to know if it's an issue with libdef though. I'll just change it to `$FlowFixMe` first and figure this out later

v3/src/js/views/venues/VenuesContainer.jsx
taneliang: Derp

v3/src/js/views/venues/VenueDetailRow.jsx
taneliang: That's exactly what I was trying to do! Thanks

taneliang:
I've implemented the live query string update. This is the behavior:

- Loading from URL:
  - `/venues/<string>` -> `/venues/<string>?q=<string>`. i.e. navigating to a venue selects venue and filters list by venue name
  - `/venues/<string>?q=<query>` will select venue and filter list by query
- Clicking on venue:
  - `/venues?q=<query>` OR `/venues/<string>?q=<query>` ---*click*--> `/venues/<venue>?q=<query>`

However, if the user searches for `lt`, then clicks on `LT17`, and shares the resulting URL `/venues/LT17?q=lt`, the recipient will see the same venue in a list. This seems a bit strange to me. I'm not sure when anyone will want to share the filtered list when they have just clicked on a venue.

Perhaps we should remove the query part from the URL when a venue is clicked? i.e. `/venues?q=<query>` ---*click*--> `/venues/LT17` ---*query changed*--> `/venues/LT17?q=<newquery>`.

v3/src/js/views/errors/Warning.test.jsx
```diff
@@ -0,0 +1,11 @@
+// @flow
+import React from 'react';
+import { shallow } from 'enzyme';
+import Warning from './Warning';
+
+test('it displays warning triangle and message', () => {
+  const wrapper = shallow(<Warning message="message" />);
+  expect(wrapper.find('AlertTriangle')).toHaveLength(1);
```
ZhangYiJiang: This test seems overspecified. I don't think it's useful to test presentational elements like the warning triangle icon.

v3/src/js/views/venues/VenueDetailRow.scss
```diff
@@ -0,0 +1,16 @@
+@import "~styles/utils/modules-entry.scss";
+
+.venueDetailRow {
+  padding-top: 1rem;
+  border-top: 1px $gray-lighter solid;
```
ZhangYiJiang: Should be `var(--gray-lighter)` now that the dark mode PR has been merged in

ZhangYiJiang:
> Perhaps we should remove the query part from the URL when a venue is clicked? i.e. /venues?q=<query> ---click--> /venues/LT17 ---query changed--> /venues/LT17?q=<newquery>.

My intuition says this will cause problems, because you're desynchronizing URL state from UI state, which will cause problems like the last time. For the time being just leave it - we can revise if it causes problems later.

v3/src/js/views/venues/VenueList.jsx
```diff
@@ -0,0 +1,48 @@
...
+  onSelect: (Venue, string, HTMLElement) => void; // Called with venue name and venue URL (/venues/<venue>)
+};
+
+export default function VenueList(props: Props) {
+  const rowRefs: { [Venue]: HTMLElement } = {};
+  const { venues, expandedVenue, onSelect } = props;
+  const lowercaseExpandedVenue = expandedVenue.toLowerCase();
+
+  // Case-insensitive, natural sort of venue names
+  const sortedVenueNames = Object.keys(venues).sort((a, b) =>
```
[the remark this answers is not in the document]
ZhangYiJiang: Ah, `sortBy` only provides value to the iterating function. I guess you could use `_.entries`, but this is fine too I guess.

v3/src/js/views/venues/VenuesContainer.scss
```diff
@@ -0,0 +1,6 @@
+@import "~styles/utils/modules-entry.scss";
+
+.venuesPageContainer {
+  margin-left: 0;
```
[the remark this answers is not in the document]
taneliang: Oops I copied this chunk over from `.module-page-container` but forgot to remove that

v3/src/js/utils/colors.test.js
```diff
@@ -0,0 +1,71 @@
...
+  test('it should get random color', () => {
+    // We're not actually testing randomness, only that the color indices returned are valid
+    // Check that calling getNewColor with currentColors returns an int
+    // in [0, NUM_DIFFERENT_COLORS] AND not in unexpectedColors
+    function expectValidIndex(unexpectedColors: Array<ColorIndex>, currentColors: Array<ColorIndex>) {
+      expect(without(range(NUM_DIFFERENT_COLORS), ...unexpectedColors))
+        .toContain(getNewColor(currentColors, true));
+    }
+
+    for (let i = 0; i < 100; i++) {
```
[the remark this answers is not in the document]
taneliang: Definitely. I'm already doing that on line 58. Probably wrote this line before I imported `range` 😆

v3/src/js/views/venues/VenueList.jsx
```diff
@@ -0,0 +1,48 @@
...
+    a.toLowerCase().localeCompare(b.toLowerCase(), 'en', { numeric: true }));
+
+  let venueRows = sortedVenueNames.map(name => (
+    <VenueDetailRow
+      key={name}
+      name={name}
+      availability={venues[name]}
+      expanded={name.toLowerCase() === lowercaseExpandedVenue}
+      rootElementRef={(row) => {
+        if (row) {
```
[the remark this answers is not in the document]
taneliang: If I remember correctly, React sometimes called `ref` with null/undefined values. Since `rootElementRef` is passed directly to `ref`, that check is necessary. I may be wrong though; you could try removing the check. If I'm right, It could be made clearer by changing the type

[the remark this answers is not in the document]
ZhangYiJiang: Ah, found the docs 

> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts. ref callbacks are invoked before componentDidMount or componentDidUpdate lifecycle hooks.

PR #500, opened 2017-12-11
Add E-Liang to team

yangshun:
@taneliang Please take over this PR and add on/remove where relevant 😄


v3/src/js/views/static/TeamContainer.jsx
```diff
@@ -27,7 +27,7 @@ function TeamMember({ member }: Props) {
     <section className={styles.teamMember} key={member.name}>
       <div className="row">
         <div className="col-lg-2 col-sm-3 text-center-md">
-          <a href={member.github} className={styles.portrait}>
+          <a href={member.link} className={styles.portrait}>
```
yangshun: Good catch. Thanks

PR #502, opened 2017-12-11
Timetable share/sharing UI

ZhangYiJiang:
This PR implements timetable sharing and syncing. 

screenshot from 2017-12-12 18-15-30

Unlike v2, the URL is not updated automatically with the timetable state. Instead, the URL must be explicitly generated using the share/sync buttons. This keeps the URL of the timetable clean, and stops the page from messing up the browser history. 

screenshot from 2017-12-11 19-17-39

When received, the timetable is made read-only. More sharing options will be added in a future PR. 

### Todo 

- [X] Implement/fix tests 
- [x] Fix module coloring 
- [x] Make timetable read-only when a sync/shared timetable is viewed 
- [ ] Add more options to share the URL (by email, shorten URL, QR code, messaging apps, etc.) - Probably will go into future PRs


v3/src/js/views/components/Modal.scss
```diff
@@ -0,0 +1,27 @@
...
+  margin: 0 auto;
+  background: var(--body-bg);
+  box-shadow: 0 6px 24px rgba(#000, 0.18);
+
+  @include media-breakpoint-down(sm) {
+    top: 4rem;
+  }
+}
+
+.overlay {
```
yangshun: You have to write your own `.overlay` and `.modal` class? It's not provided by the library? :/

ZhangYiJiang: They are, but the library implements them as inline style objects,  and for some reason if you provide your own `className` then that default style is removed

PR #504, opened 2017-12-12
Add Dark Mode toggle to Settings page

yangshun:
screen shot 2017-12-11 at 11 21 32 pm


ZhangYiJiang:
I think calling it 'Night Mode' or 'Dark theme' with an on/off toggle would be clearer 

eg. YouTube's 

screenshot from 2017-12-12 15-24-47

yangshun:
What app is that?

yangshun:
Not inclined to use theme again because we're already using it for the timetable colors.

ZhangYiJiang:
That's YouTube's website. If you're logged in, the option is under your user's icon on the top right hand corner under 'dark theme'

PR #508, opened 2017-12-12
Turn add module buttons into dropdown menu 

ZhangYiJiang:
I'm a bit iffy about this UI. It does mean one less click than before to add to timetable, but it looks kind of ugly 

screenshot from 2017-12-12 22-11-14

Any suggestions?


v3/src/js/views/components/module-info/AddToTimetable.jsx
```diff
@@ -0,0 +1,127 @@
...
+    action(semester, module.ModuleCode);
+  }
+
+  buttonLabel(semester: Semester) {
+    return this.moduleHasBeenAdded(semester)
+      ? `Remove ${config.semesterNames[semester]}`
+      : `Add to ${config.semesterNames[semester]}`;
+  }
+
+  moduleHasBeenAdded(semester: Semester): boolean {
```
yangshun: This should be extracted out into utilities (`utils/modules`?) because it can potentially be reused outside of this component.

v3/src/js/views/components/module-info/AddToTimetable.jsx
```diff
@@ -0,0 +1,127 @@
...
+    const action = this.moduleHasBeenAdded(semester)
+      ? this.props.removeModule
+      : this.props.addModule;
+
+    action(semester, module.ModuleCode);
+  }
+
+  buttonLabel(semester: Semester) {
+    return this.moduleHasBeenAdded(semester)
+      ? `Remove ${config.semesterNames[semester]}`
```
yangshun: Is this missing a from? Should it be `Remove from ...`?

v3/src/js/views/components/module-info/AddToTimetable.jsx
```diff
@@ -0,0 +1,127 @@
...
+      : `Add to ${config.semesterNames[semester]}`;
+  }
+
+  moduleHasBeenAdded(semester: Semester): boolean {
+    const { timetables } = this.props;
+    return timetables[semester] && !!timetables[semester][this.props.module.ModuleCode];
+  }
+
+  otherSemesters(exclude: Semester): Semester[] {
```
yangshun: This can be within `utils/modules` too.

v3/src/js/views/components/module-info/AddToTimetable.scss
```diff
@@ -0,0 +1,7 @@
+.button {
+  // Long selector for specificity
+  :global(.dropdown-toggle.dropdown-toggle-split) {
```
yangshun: If `.dropdown-toggle-split` was invented just for specificity and is not really necessary, I think you can do `.dropdown-toggle.dropdown-toggle` to increase the specificity.

v3/src/js/views/components/module-info/AddToTimetable.jsx
```diff
@@ -0,0 +1,127 @@
+// @flow
```
yangshun: Will be clearer to name this component `AddToTimetableButton`/`AddToTimetableDropdown`.

v3/src/js/views/components/module-info/AddToTimetable.jsx
```diff
@@ -0,0 +1,127 @@
...
+                  aria-haspopup="true"
+                  aria-expanded={isOpen}
+                >
+                  <span className="sr-only">Toggle Dropdown</span>
+                </button>}
+
+              {isOpen &&
+                <div className="dropdown-menu show">
+                  {otherSemesters.map(semester => (
+                    <button
```
yangshun: The dropdown button is too tall. Can we have it the same height as the primary button?

screen shot 2017-12-13 at 9 49 54 pm

v3/src/js/views/components/module-info/AddToTimetable.jsx
yangshun: Indeed, it is also used here: https://github.com/nusmodifications/nusmods/blob/master/v3/src/js/views/modules/ModulePageContent.jsx#L44-L47

v3/src/js/views/components/module-info/AddToTimetable.jsx
ZhangYiJiang: Because the string is too long. On some screen sizes 'Remove from {semester name}' will overflow. It's the same reason why I had to use `button-group-sm`, which is another decision I don't like, but I felt when balanced against reducing the number of clicks needed to add a module this is a good tradeoff. I'm still looking for a solution that doesn't need to make this compromise though...

yangshun: But "Remove Semester 2" doesn't make sense at all... Will it fit if you use "Remove from Sem 2" instead? Removing `ester` should make space for `from`.

ZhangYiJiang:
Switched to using two lines to fit all of the text 

screenshot from 2017-12-15 14-32-04

Issue #411, opened 2017-07-23
Fix navbar at bottom of the screen on mobile


zameschua:
https://uxplanet.org/perfect-bottom-navigation-for-mobile-app-effabbb98c0f

li-kai:
I don't think we should do it at all as it's nearly impossible, see [mobile safari limitations](https://benfrain.com/the-ios-safari-menu-bar-is-hostile-to-web-apps-discuss/). There's a reason why twitter lite's menu bar is at the top 🙁

yangshun:
Oh wow I didn't know this since I don't use iPhone. Indeed I did wonder why Twitter Lite's menu bar is at the top. @zameschua if you feel like it, you can try to shift it to the bottom and see if it is indeed impossible. Else, we can close this issue.

Good find @li-kai !

zameschua:
Haha thanks @li-kai!

https://www.eventbrite.com/engineering/mobile-safari-why/
Eventbrite seems to have done it but requires a workaround. I think can close for now if it's not priority.

yangshun:
With iPhone X shifting the home button to the button, this might be a bad idea now.

Issue #505, opened 2017-12-12
Undo for destructive actions 


ZhangYiJiang:
Some actions on the timetable are destructive - 

- Removing a module from the timetable 
- Replacing the existing timetable with an imported one 

Currently this is handled using a `confirm` prompt, but I think we can do better. It is much better for the user if we simply let them undo their actions. 

I think a [snackbar like the one from Material](https://material-components-web.appspot.com/snackbar.html) would be very useful for this, especially since we're also looking at using the FAB, although suggestions for other UI implementation are welcome too.

PR #510, opened 2017-12-14
Implement FAB for page specific actions 

ZhangYiJiang:
This adds floating action buttons for page actions, which makes NUSMods easier to use on mobile (similar to proposal #411). The FAB comes from [Material Components Web](https://material.io/components/web/) using [RMWC](https://github.com/jamesmfriedman/rmwc) for React bindings. 

screenshot from 2017-12-14 11-37-20 screenshot from 2017-12-14 11-37-31

Since the library is modular, we only need to import the components and styles we need, which makes it fairly lightweight. 

Page actions to implement: 

- [X] Open menu on module info page (we assume the most common action is browsing, not add/remove module from timetable)
- [x] Open filters list on module search page and venues page 
- [ ] Focus the module select box on timetable page (probably will implement after #484 goes through)


v3/src/js/views/modules/ModulePageContent.scss
```diff
@@ -1,9 +1,73 @@
...
+.fab {
+  display: none;
+}
+
+.sideMenu {
+  position: sticky;
+  top: $sticky-top;
+  margin-top: 0.4rem;
+
+  ul {
```
yangshun: You can consider using Bootstrap's `list-unstyled` if it has been included in the main CSS bundle.

.gitignore
```diff
@@ -3,7 +3,7 @@
 .tmp
 .vagrant
 app/bower_components
-vendor
+# vendor
```
yangshun: What's this change for?

ZhangYiJiang: I was originally going to put the material styles under vendor, but I noticed the vendor folder was not properly checked in because this line was in the `.gitignore`. I think this line should probably be removed because it is not specific enough.

yangshun: Ok please remove it then 😂

ZhangYiJiang:
Since the right hand sidebar on both module finder and module info work in the same manner, I've extracted the common parts into a `<SideMenu>` wrapper component which also contains the FAB acting as toggle. 

Because I couldn't fix the way the position sticky search box and position fixed drawer menu interacted with each other, I moved `<CorsNotification>` from `<AppShell>` to the timetable and module info page instead. This does mean that the warning will reappear on every page change on desktop, but I think it's okay. 

re: using RMWC - I intend to use the Snackbar component for #505

ZhangYiJiang:
@taneliang That is a little odd. It's almost certainly an upstream bug - Gmail on Android does not have the same behaviour. I'm guessing what's happening is that the long press is getting converted into a right click on touch screens. It's not a huge issue anyway, because I don't see people trying to long press FABs.

PR #512, opened 2017-12-15
Additional sharing options and short URL support 

ZhangYiJiang:
This PR adds sharing via QR code, email, Telegram and WhatsApp. Additionally, the modal will attempt to shorten the timetable URL when it is opened. 

screenshot from 2017-12-15 12-10-25

If the shortener service is not available for whatever reason, the original URL will be shown instead. 

### Next steps 

- Load all of the timetable actions asynchronously to improve initial page loading time 
- Add animation/transitions 
- Switch to Font Awesome 5 so we can have access to more icons

### DIFF
--- a/v3/src/js/views/timetable/ShareTimetable.jsx
+++ b/v3/src/js/views/timetable/ShareTimetable.jsx
@@ -2,7 +2,6 @@
 
 import React, { PureComponent } from 'react';
 import classnames from 'classnames';
-import select from 'select';
 
 import type { SemTimetableConfig } from 'types/timetables';
 import type { Semester } from 'types/modules';
@@ -28,16 +27,24 @@ export default class ShareTimetable extends PureComponent<Props, State> {
 
   state: State = {
     isOpen: false,
-    action: 'sync',
   };
 
   openModal = () => this.setState({ isOpen: true });
   closeModal = () => this.setState({ isOpen: false });
 
   copyText = () => {
-    if (this.urlInput) {
-      select(this.urlInput);
-      document.execCommand('copy');
+    const input = this.urlInput;
+
+    if (input) {
+      input.select();
+      input.setSelectionRange(0, input.value.length);
+
+      // TODO: Inform the user copy succeeded through the UI, and prompt the user
+      // to manually copy if execCommand fails. Mobile Safari currently doesn't support
+      // execCommand('copy')
+      if (document.execCommand('copy')) {
+        input.blur();
+      }
     }
   };
 
--- a/v3/src/js/views/timetable/ShareTimetable.jsx
+++ b/v3/src/js/views/timetable/ShareTimetable.jsx
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { PureComponent } from 'react';
+import { QRCode } from 'react-qr-svg';
 import classnames from 'classnames';
 
 import type { SemTimetableConfig } from 'types/timetables';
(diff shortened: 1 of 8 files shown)


v3/src/js/views/timetable/ShareTimetable.jsx
```diff
@@ -1,50 +1,163 @@
...
 import { absolutePath, timetableShare } from 'views/routes/paths';
-import { Repeat, Copy } from 'views/components/icons';
+import { Repeat, Copy, Mail } from 'views/components/icons';
 import Modal from 'views/components/Modal';
+import LoadingSpinner from 'views/components/LoadingSpinner';
 
 import styles from './ShareTimetable.scss';
 import actionStyles from './TimetableActions.scss';
 
+type CopyState = 'not copied' | 'success' | 'fail';
```
yangshun: Define constants for these enums. I would also prefer that the strings are capitalized for consistency with rest of the types code:

```
export type Mode = 'LIGHT' | 'DARK';
export const LIGHT_MODE: Mode = 'LIGHT';
export const DARK_MODE: Mode = 'DARK';
```

v3/src/js/views/timetable/ShareTimetable.jsx
```diff
@@ -1,50 +1,163 @@
...
 
   state: State = {
     isOpen: false,
-    action: 'sync',
+    urlCopied: 'not copied',
+    shortUrl: null,
+  };
+
+  loadShortUrl(url: string) {
+    return axios.get('https://nusmods.com/short_url.php', { data: { url } })
```
yangshun: How about just showing the unshortened URL and change to the short URL when the request succeeds? My concern is that an empty input will be shown for poor network conditions. By showing the unshortened URLs first, you can ignore the failure case 🌝

v3/src/js/views/timetable/ShareTimetable.jsx
```diff
@@ -1,50 +1,163 @@
...
+      .then(({ data }) => this.setState({ shortUrl: data.shortUrl }))
+      // Cannot get short URL - just use long URL instead
+      .catch(() => this.setState({ shortUrl: url }));
+  }
+
+  // Event handlers
```
yangshun: This comment is not really necessary imo.

v3/src/js/views/timetable/ShareTimetable.jsx
```diff
@@ -1,50 +1,163 @@
...
+    }
+
+    this.setState({ isOpen: true });
   };
 
-  openModal = () => this.setState({ isOpen: true });
-  closeModal = () => this.setState({ isOpen: false });
+  closeModal = () => this.setState({
+    isOpen: false,
+    urlCopied: 'not copied',
```
yangshun: Use the const/enums.

v3/src/js/views/timetable/ShareTimetable.jsx
```diff
@@ -1,50 +1,163 @@
...
+              type="button"
+              aria-label="Copy URL"
+              onClick={this.copyText}
+            >
+              <Copy className={styles.icon} />
+            </button>
+          </span>
+
+          {this.state.urlCopied === 'success' && <p className={styles.copyStatus}>Link copied!</p>}
+          {this.state.urlCopied === 'fail' && <p className={styles.copyStatus}>Press Cmd + C to copy</p>}
```
yangshun: By only displaying `Cmd + C`, are you assuming mac users only? 😹

v3/src/js/views/timetable/ShareTimetable.test.jsx
```diff
@@ -0,0 +1,70 @@
...
+const timetable = {
+  CS1010S: {
+    Lecture: '1',
+  },
+};
+
+const openModal = wrapper => wrapper.find('button').simulate('click');
+const closeModal = wrapper => wrapper.find(Modal).first().props().onRequestClose();
+
+test('should load short URL when the modal is opened', () => {
```
yangshun: I know we haven't been doing that but let's try this moving forward - Wrap the tests in a

```
describe('MyComponentName', () => { 
  ...
});
```

The additional namespacing helps in test coverage reporting if we ever want to do it.

v3/src/js/views/timetable/ShareTimetable.jsx
ZhangYiJiang: Considered that, but the problem is that this will cause the QR code to change, which is quite visually jarring. I'm adding a timeout, so it should be okay.

v3/src/js/views/timetable/ShareTimetable.jsx
ZhangYiJiang: Technically Safari is the only browser in the support matrix to not support this, but point taken

PR #513, opened 2017-12-16
Settings page improvements 

ZhangYiJiang:
### Features 

- Improve copy for some setting items 
- Show a preview timetable above the theme selection UI 

  
  screenshot from 2017-12-16 12-52-34

- Hide night mode toggle when the browser does not support CSS variables 
- Use CSS transition to make toggling of night mode appear a bit smoother 

### Maintenance 

- Migrate styles to use CSS modules 
- Add `defaultProps` for `isVerticalOrientation` and `onModify` to <Timetable>`


taneliang:
Timetable preview is great. Is the border around the timetable necessary? It causes this ugly double line on the left edge of the timetable.
uglyline

Also `50rem` `max-width` forces the timetable to always scroll. Maybe `60rem` will be better?

ZhangYiJiang:
Hmmm, I was trying to better separate the timetable from the theme options while making sure the user understands that what they're looking at is a preview. I've removed the borders, because on second thought this is probably good enough.

v3/src/js/views/settings/previewTimetable.js
```diff
@@ -0,0 +1,113 @@
+// @flow
+import type { TimetableArrangement } from 'types/timetables';
+
+const previewArrangement: TimetableArrangement = {
```
yangshun: Why not make this part of mocks so that it can potentially be reused in a test?

PR #517, opened 2017-12-16
Fix module side menu highlighting

### LINKED ISSUE #516: Module side menu does not highlight correct item
1. At times, side menu does not highlight correct item even if the item was just clicked.
2. Bidding Stats is never highlighted.

localhost_8080_modules 1

taneliang:
Fixes #516.

Clicking the Prerequisites button still highlights Timetable, as the Prerequisites section is too short (since it's not yet implemented). As long as it is eventually >40px tall, there will be no issues.

I tried setting an offset proportional to `window.innerHeight` as well, but it felt weird, so a constant offset it is.

### DIFF
--- a/v3/src/js/views/modules/ModulePageContent.jsx
+++ b/v3/src/js/views/modules/ModulePageContent.jsx
@@ -194,8 +194,9 @@ class ModulePageContentComponent extends Component<Props> {
           <aside className="col-md-3">
             <nav className="module-side-menu">
               <ScrollSpy
-                items={['details', 'prerequisites', 'bidding-stats', 'timetable', 'reviews']}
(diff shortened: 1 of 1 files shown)


yangshun:
I think the pre-req tree is broken. We can just remove the pre-req section.

taneliang:
It's just a stub section with a TODO, but #451 is supposed to implement it?

ZhangYiJiang:
Hmmm, I'm a bit iffy on this. Can I take over this issue? I think @yangshun's comment https://github.com/nusmodifications/nusmods/pull/510#discussion_r156860196 makes sense - although I wanted to implement it only after that PR goes through to avoid unnecessary conflicts

taneliang:
@ZhangYiJiang Sure!

ZhangYiJiang:
Closed in favor of #535

PR #518, opened 2017-12-16
Implement data migration from v2

### LINKED ISSUE #507: Migrate timetable data from v2 to v3
Since students have already started planning for next semester, it would be irresponsible to launch v3 without some way to bring the data over. 

v2 uses [localForage](https://github.com/localForage/localForage) with IndexedDB for persistence and serializes timetables into what appears to be the same format used in the URL, so we could use the same library and automatically port the data over.

ZhangYiJiang:
Fixes #507 

This sets up data migration from NUSMods v2 by reading off IndexedDB using [localforage](https://github.com/localForage/localForage), and converting the saved data into the format used in v3. 

It also refactors the actions and reducers for timetable imports which I feel are much better now compared to the ones written for the timetable sharing PR. The colors are still an issue though - `SET_TIMETABLE` does not remove the colors from the previous timetable because it can't.


v3/src/js/storage/migrateTimetable.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return timetable;
+}
+
+export default function migrateTimetable(store: Store<*, *, *>) {
+  return Promise.all(
+    migratedKeys.map(([semester, key]) => {
+      let timetable;
+
+      return localforage.getItem(key)
```
yangshun: How will this line behave when the migration is done? i.e. for users who refresh the page after their data has already been migrated.

Do we need to store some flag such as `migratedDone` to not migrate twice?

ZhangYiJiang: Yeah, noticed that problem - fixed it by removing the item after migration and just skipping migration if there's no data to migrate.

yangshun: Please file another task to remove the migration code sometime in Feb 😄

ZhangYiJiang:
Manual testing scripts: https://gist.github.com/ZhangYiJiang/69a47a58b06c25615793b38eaec6d9d1

ZhangYiJiang:
The IndexedDB calls + parsing should be fast enough that it won't be perceivable to the user even though it is async. Loading the additional module data on the other hand takes time, and because of how data is loaded right now there doesn't seem to be a easy way to show a spinner while they are loading, because of how the reducers and actions are structured.

ZhangYiJiang:
Code is ready - now conducting manual testing (by asking people to run the test script)

yangshun:
Looks good. I didn't test it with v2, but we'll test it together before we launch. Final request, could you add TODOs to those files that need to be changed/removed when we are ready to remove the migration code?

PR #520, opened 2017-12-16
Add script to deploy v3 to beta.nusmods.com

taneliang:
Add script to package.json to deploy to [`beta.nusmods.com`](http://beta.nusmods.com). To be changed to `nusmods.com` when it is time to deploy v3 for real.

Uses the v2 `rsync` approach instead of my proposed `rm -rf ../nusmods.com && cp -r dist ../nusmods.com` or @ahbeng's suggested `cp -ru` as:

1. `rsync` > `cp -ru` because `rsync` is able to remove files that have been deleted.
2. `rsync` > `rm -rf && cp -r` because `rsync` is able to remove only the files that have been deleted.
3. `cp -ru` > `rm -rf && cp -r` because `cp -ru` leaves unchanged files alone instead of unnecessarily deleting and copying them.

∴ `rsync` > everything else. ∎


yangshun:
Did you test the command? `beta.nusmods.com` seems to be two levels above the `package.json`

taneliang:
Man that was extremely careless of me. Sorry. Fixed and tested on the server

v3/README.md
```diff
@@ -70,13 +70,16 @@ $ yarn lint:scripts
...
 ```sh
-$ yarn build
+$ yarn build # Build only
+$ yarn deploy # Build and deploy
 ```
 
-This will package and optimize the app for deployment. The files will be placed in the `/dist` directory.
+`yarn build` will package and optimize the app for deployment. The files will be placed in the `/dist` directory.
+
+`yarn deploy` will run `yarn build` and copy the resulting files to the `../../beta.nusmods.com` directory (a directory on the same level as the `nusmods` source root).
```
taneliang: Any suggestions to improve this rather clunky phrasing?

yangshun: How about this:

- `yarn build` will package and optimize the app for deployment. The files will be placed in the `/dist` directory.

- `yarn rsync` will copy files from `dist` to the configured destination folder.

- `yarn deploy` runs `yarn build` then `yarn rsync`.

It's better to compose `deploy` from two separate steps and so that we can run the `rsync` step independently. That's how it's done in v2. 

From my experience in v2, I think `yarn deploy` is a pretty dangerous thing to do because it encourages building and deploying without first testing on v3 staging (successful build step does not mean that the bundled app can run properly). I messed up once or twice when I did that in v2, due to some missing dependency which somehow was not caught. But I guess it's ok to have the `deploy` command as long as we're careful of what it does. We should relook into the deployment workflow to make deployment stable yet easy.

ZhangYiJiang: :+1: on not having a single deploy step command. I made the same mistake for my blog a few times and learnt my lesson. It's better to build on staging, manually check through it (especially since we don't have any sort of e2e tests right now, unit tests can't catch everything), then deploy to production.

yangshun: Then I think we should remove `deploy` altogether?

taneliang: Why don't we replace `yarn deploy` with a `yarn stage`, which builds and deploys to the staging folder. We can then manually deploy it to production from there, or have another script to rsync the staging folder with the production folder. This will at least make staging a one step process

yangshun: Isn't staging already a one-step process? staging.nusmods.com/v3.nusmods.com currently reads from `nusmods/dist/` and `nusmods/v3/dist/` respectively

taneliang: Ah okay I didn't know that. Then maybe we should have an deploy script which only rsyncs files to production, and also confirms with the user before copying? That'll make it safer and easier to deploy to production.

taneliang:
Added `rsync` and `promote-staging` commands. To quote the updated README.md:

> - `yarn build` packages and optimizes the app for deployment. The files will be placed in the `./dist` directory.
> - `yarn promote-staging` deploys `./dist` to the production folder, currently `../../beta.nusmods.com`. It is designed to be safe, executing a dry run and asking for confirmation before deployment.
> - `yarn rsync <dest-dir>` syncs `./dist` to the specified destination folder `<dest-dir>`. It is mainly used by `yarn promote-staging` but could be used to sync `./dist` to any folder.

Tested on the server. Example `yarn promote-staging` output:

```
ubuntu@nusmods:~/nusmods-v3/v3$ yarn promote-staging
yarn promote-staging v1.0.1
$ bash scripts/promote-staging.sh
Promote ./dist to production at ../../beta.nusmods.com.
Staging directory ./dist exists.

Contents of ./dist:
total 7044
drwxrwxr-x  2 ubuntu ubuntu    4096 Dec 17 14:05 .
drwxrwxr-x 12 ubuntu ubuntu    4096 Dec 17 14:05 ..
-rw-rw-r--  1 ubuntu ubuntu    4139 Dec 17 14:05 android-chrome-192x192.png
...
-rw-r--r--  1 ubuntu ubuntu  282306 Dec 17 14:05 workbox-sw.prod.v2.1.0.js.map

Dry running deployment...

> nusmods@3.0.0-alpha rsync /home/ubuntu/nusmods-v3/v3
> rsync -avu --delete-after dist/ "--dry-run" "../../beta.nusmods.com"

building file list ... done
./
_redirects
android-chrome-192x192.png
...
workbox-sw.prod.v2.1.0.js.map

sent 867 bytes  received 87 bytes  1,908.00 bytes/sec
total size is 7,143,029  speedup is 7,487.45 (DRY RUN)
Ready to promote? [yN] y

Promoting...

> nusmods@3.0.0-alpha rsync /home/ubuntu/nusmods-v3/v3
> rsync -avu --delete-after dist/ "../../beta.nusmods.com"

building file list ... done
./
_redirects
android-chrome-192x192.png
...
workbox-sw.prod.v2.1.0.js.map

sent 7,146,516 bytes  received 455 bytes  14,293,942.00 bytes/sec
total size is 7,143,029  speedup is 1.00
All done.
Done in 16.17s.
```

PR #524, opened 2017-12-17
Add link to built commit

### LINKED ISSUE #522: Version display on v3 page
It is currently difficult to check which version of NUSMods has been deployed. We could inject the commit hash/branch name when building and display them in the UI.

taneliang:
Resolves #522.

link

Adds a tiny link to the commit in the footer text. I decided against displaying the commit hash on the page as users don't need to know it.


v3/src/js/views/layout/Footer.jsx
```diff
@@ -18,8 +20,15 @@ function Footer() {
           <li><Link to="/developers">Developers</Link></li>
           <li><Link to="/faq">FAQ</Link></li>
         </ul>
-        {/* TODO: Add last updated timestamp */}
-        <p>Designed and built with all the love in the world by{' '}
+        <p>Designed and{' '}
+          <a
```
yangshun: It's better to do

```
{commitHash ? <a href... >built</a> : 'built'}
```

so that users don't get a non-working link if there's no `commitHash`.

ZhangYiJiang:
Eh, to be honest this seems a little silly :sweat_smile: 

Why not use the build or commit timestamp as the version number?

v3/src/js/views/layout/Footer.jsx
```diff
@@ -3,6 +3,8 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 
 function Footer() {
+  const commitHash = process.env.COMMITHASH;
```
yangshun: It feels wrong to me for React components to read from environment variables. I thought it would affect the jest tests but they seem to be fine. I propose reading `process.env.COMMITHASH` in `main.js` and passing it into the store to be passed as props into the `Footer`. Need more opinions on this though @ZhangYiJiang

v3/src/js/views/layout/Footer.jsx
```diff
@@ -18,8 +20,15 @@ function Footer() {
           <li><Link to="/developers">Developers</Link></li>
           <li><Link to="/faq">FAQ</Link></li>
         </ul>
-        {/* TODO: Add last updated timestamp */}
```
yangshun: This TODO should be kept here. The last updated timestamp is referring to the API data and not really the app bundle.

yangshun:
Build/commit timestamp sounds good to me! I think having *both* will be useful. Timestamps are useful for consumers while link to the GitHub commit is useful for devs. Let's have both?

taneliang:
Yep having both sounds good. I'll add the build timestamp... somewhere.

I don't think timestamps will mean anything to users either though. But it could help us if we were to build the same commit at different times.

v3/src/js/views/layout/Footer.jsx
ZhangYiJiang: I think pushing it into the store is a bit of an overkill for something simple, since this will not be used anywhere else and the build step will replace it with a constant string anyway.

yangshun:
Well if we really wanted to, we could do semver with a proper release engineering process but it's too much of a hassle for a small 3-man project IMO.

ZhangYiJiang:
When I said it seemed a little silly I mean making the word build a link. You can just append 'version 20171217-1032' to the end of the line or on a separate line.

v3/src/js/views/layout/Footer.jsx
yangshun: BTW guys, putting the commit hash into the bundle also means that the `app.js` bundle filename will have a different hash regardless of whether app code was modified or not. But it's only™ 50kb so I think it's ok?

ZhangYiJiang: Hmmm, true, but I think the app bundle pretty much changes with every build anyway

yangshun: Most of the time yes, but if we change non-v3 stuff like the README, write more tests, or some config files, the commit hash changes too.

yangshun:
I like the SO approach - we can immediately tell when we last deployed it.

How about `<date>-<hash>`? E.g. `20171217-bc89de2`.

taneliang:
It is a little silly but it was the nicest place to hide something that only we will use. But we can't do that if we want to display a timestamp. I'll add a line in the footer with `<date>-<hash>`.

v3/src/js/views/layout/Footer.jsx
taneliang: We could just not deploy if we don't change the v3 stuff? Since it's a manual process anyway

yangshun: Yes that's true since it's manual now. Go ahead with this then 😄

taneliang:
Okay, this is what it looks like now.

localhost_8080_about

Also added a `console.log` statement to print the version string when building.

```
yarn run v1.3.2
$ cross-env NODE_ENV=production node scripts/build
Building version 171217-bc89de2
Creating an optimized production build...
```

v3/src/js/views/layout/Footer.jsx
taneliang: Clarified the TODO

yangshun:
In Singapore, the most common date format is `DDMMYY`. Let's either do that or do `YYYYMMDD` so that we don't get confused between YY and DD (takes a little more time to figure out).

taneliang:
The reason I went with `YYMMDD` is so that it's sortable (`DDMMYY` isn't) and doesn't include the "20" prefix which will be in every `YYYYMMDD`. But point noted - I'll change it to `YYYYMMDD`.

v3/webpack/webpack.config.common.js
```diff
@@ -1,13 +1,20 @@
...
 const parts = require('./webpack.parts');
 
 // Used by Webpack to resolve the path to assets on the client side
 // See: https://webpack.js.org/guides/public-path/
 const publicPath = process.env.PUBLIC_PATH || '/';
 
+// Used to inject versioning info into the bundle
+const commitHash = childProcess.execSync('git rev-parse --short=7 HEAD').toString().trim();
+const versionStr = commitHash && `${moment().format('YYYYMMDD')}-${commitHash}`; // Version format: <date>-<hash>
+console.log('Building version', versionStr); // eslint-disable-line no-console
```
ZhangYiJiang: This `console.log` shouldn't be here - Webpack config is not just imported in the build script. Instead, it's best to create a function that constructs the version string, then use it in `scripts/build.js` and `webpack.config.prod.js` (not sure if it is necessary to do it in dev)

v3/src/js/views/layout/Footer.jsx
```diff
@@ -3,6 +3,21 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 
 function Footer() {
+  const { commitHash, versionStr } = process.env || {};
```
ZhangYiJiang: It may be better to just stick to 

```
const commitHash = process.env.commitHash; 
const versionStr = process.env.versionStr; 
``` 

so that `DefinePlugin` can just inline the variables.

v3/src/js/views/layout/Footer.jsx
```diff
@@ -3,6 +3,21 @@ import React from 'react';
...
+
+  const versionSpan = commitHash && versionStr && (
+    <span>
+      Version{' '}
+      <a
+        href={commitHash && `https://github.com/nusmodifications/nusmods/commit/${commitHash}`}
```
ZhangYiJiang: `commitHash` check here is superfluous

PR #526, opened 2017-12-17
Venue availability search for v3 

ZhangYiJiang:
This PR implements the venue availability search. The UI is implemented as a set of three select boxes like in v2, with a button on the side to toggle this. The changes from v2 are that -

- The search box is still open, which means users can filter by _both_ name and availability 
- Time is listed by the hour, instead of half-hour in v2. I made this change because I assume the half-hour options are underused because not many classes occupy half hour blocks 

This PR also updates the UI to move the entire search to the right side as a side menu 

screenshot from 2017-12-20 16-27-34

### Future PRs 

- Adopt a listing UI that's closer to v2's (group by first character, inline list to avoid wasting horizontal space) 
- Highlight on the timetable the selected free time slots


taneliang:
Oops a little long but just coming up with a list of UI issues I could find :) I haven't looked at the code, so I'm approaching this from a users's perspective. I put some ideas for fixing at the bottom.

### Cosmetic issues

1. A huge percentage of the screen is taken up by UI chrome. iOS Safari keeps a persistent UI bar at the top which further reduces the amount of visible content. 
image
1. Text gets clipped in the search box and the Day and To `select` in portrait.
1. It's horizontally scrollable, and the Find free rooms button touches the right edge of the screen. May be related.
scrollable
1. Unusable on phones (e.g. iPhone 5 above, iPhone 8 Plus and Nexus 5X+keyboard here) in landscape
landscape iphone 7 plus
landscape nexus 5x
1. The Find free rooms button is a bit hard to notice on a desktop widescreen.
1. The availability search is a little confusing and could be clearer. Does it mean:
    1. Show venues which are free for the entire time specified
    1. Show venues which are free at some point during the time specified

### Technical

1. When the Find free rooms button/area is deactivated, it's unclear if the locations are still filtered by the selected times. The URL says it is, but I can't really tell by looking at the UI. Also, clicking on a venue next causes the availability queries to disappear.
1. When the free room filter boxes are shown, they're prefilled with some values. It appears that we are filtering by those values immediately, but it isn't obvious and the URL isn't updated. Again, a bit hard to tell if the list is being filtered.

### Some ideas

1. Perhaps we could have a summary box of some sort, stating what is being searched. Something like "Showing **searchterm** venues free **Tuesdays 1800** through **2000**"? This makes it obvious that the list has been filtered and clarifies what is being searched.
1. This summary can be the only thing stickied on `xs` and `sm` screens, so that the search box and availability search boxes could be scrolled offscreen. This makes the UI usable on small phones.
1. The summary box can have a button to show/scroll back to the search box and availability selector things. This maintains the ability to edit the filters while scrolling. Not entirely sure if this is worth the screen real estate though. iPhones can scroll to the top with a tap of the status bar, and Android phones typically have huge screens. Everyone else can just scroll.
1. I'm not sure how the tab idea will work on phones because it sounds like it'll still take up quite a bit of space? But it may be friendlier than the current UI, and it should still be usable if coupled with a summary.

Also, unrelated but our logo is pixellated on HiDPI screens lol.

ZhangYiJiang:
The global sticky header needs improvements (this is in the blocker list, although it probably should be filed under its own issue). This would improve the space issue on mobile. The options I'm looking at are 

    - No fixed header at all 
    - Sticky only the navtabs 
    - Use something like [headroom](https://github.com/KyleAMathews/react-headroom) to only show on scroll up 

> When the Find free rooms button/area is deactivated, it's unclear if the locations are still filtered by the selected times. 

This is a bug - I'm fixing. 

> Unusable on phones in landscape

Unstickying the global header on landscape will probably fix this 

> Perhaps we could have a summary box of some sort 

I like this, although the UI work is going to be a bit more extensive.

v3/src/js/utils/timify.js
```diff
@@ -20,6 +21,17 @@ export function convertIndexToTime(index: number): LessonTime {
   return (hour < 10 ? `0${hour}` : hour.toString()) + minute;
 }
 
+export function timestamp(time: number): LessonTime {
```
yangshun: Could you add tests? Shouldn't be hard to do so.

v3/src/js/utils/timify.js
```diff
@@ -20,6 +21,17 @@ export function convertIndexToTime(index: number): LessonTime {
...
+  return padStart(String(time), 4, '0');
+}
+
+export function formatTime(hour: number): string {
```
yangshun: We have quite a number of time formatting functions here and there, so I think the naming for this function can be more specific.

v3/src/js/views/venues/AvailabilitySearch.jsx
```diff
@@ -0,0 +1,100 @@
...
+type Props = {
+  className?: string,
+  isEnabled: boolean,
+  searchOptions: VenueSearchOptions,
+  onUpdate: (VenueSearchOptions) => void,
+};
+
+// The starting time of lessons
+const CLASS_START_HOUR = 8;
+const CLASS_END_HOUR = 22;
```
yangshun: Don't we have some classes ending at midnight? 😂

v3/src/js/views/venues/AvailabilitySearch.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+        <div className="form-group">
+          <label htmlFor="venue-time">From</label>
+          <select
+            id="venue-time"
+            className="form-control"
+            value={searchOptions.time}
+            onChange={evt => this.onUpdate(evt, 'time')}
+          >
+            {range(CLASS_START_HOUR, CLASS_END_HOUR - 1).map(hour => (
```
yangshun:
```
range(CLASS_START_HOUR, CLASS_END_HOUR - 1)
```

I think you can cache the result as a variable outside of the component since it is not dynamic.

v3/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -115,25 +135,11 @@ export class VenuesContainerComponent extends Component<Props, State> {
...
-
-    const lowercaseSearchStr = searchTerm.toLowerCase();
-    return pick(venues, Object.keys(venues).filter(name =>
-      name.toLowerCase().includes(lowercaseSearchStr)));
-  }
+  };
 
   render() {
-    const { loading, error } = this.state;
+    const { loading, error, isAvailabilityEnabled, searchOptions } = this.state;
```
yangshun: Why not combine all of these destructuring?

v3/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -148,26 +154,80 @@ export class VenuesContainerComponent extends Component<Props, State> {
...
-              onSelect={this.onVenueSelect}
-            />
+          <div className="col-md-8 col-lg-9">
+            {size(venues) === 0 ?
+              <div>
+                <Warning message="No matching venues found" />
+                <p className="text-center text-muted">
+                  There {unfilteredCount === 1
+                    ? 'is a venue that is'
+                    : `are ${unfilteredCount} venues that are`} not shown
```
yangshun: The extra code we have to add for the sake of English 🌝

v3/src/js/views/venues/AvailabilitySearch.jsx
```diff
@@ -0,0 +1,100 @@
...
+        <div className="form-group">
+          <label htmlFor="venue-duration">To</label>
+          <select
+            id="venue-duration"
+            className="form-control"
+            value={searchOptions.duration}
+            onChange={evt => this.onUpdate(evt, 'duration')}
+          >
+            {range(1, CLASS_END_HOUR - searchOptions.time).map(hour => (
+              <option key={hour} value={hour}>{formatTime(searchOptions.time + hour)} ({hour} hr)</option>
```
yangshun: Wanna pluralize display of the `hr` as well?

v3/src/js/views/venues/AvailabilitySearch.jsx
ZhangYiJiang: That's actually suppose to be the last time any class can start. I've moved the code around to make that clearer. (This is kind of a guess - I've never seen a class start after 1830, but I don't know about faculties).

PR #528, opened 2017-12-18
Scroll to top upon navigation

### LINKED ISSUE #523: Scroll position preserved on page navigation
Pages should be automatically scrolled to the top when they are opened, as it does not make sense to start viewing the middle of the page.

### Steps to reproduce

1. Navigate to <https://latest.nusmods.com>.
1. [Shrink browser viewport height](https://developers.google.com/web/tools/chrome-devtools/device-mode/emulate-mobile-viewports) if necessary.
1. Scroll to bottom.
1. Click Settings. The Settings page will be loaded but you'll be at the bottom of the page.

Can be reproduced with static pages as well - switch between About, Team and FAQ.

yangshun:
Fixes #523 by adding a `withScrollToTop` HOC. 

This HOC takes in two flags:
- `onComponentWillMount` - Whether to scroll to top upon navigation (`componentWillMount`)
- `onComponentDidUpdate` - Whether to scroll to top upon update (`componentDidUpdate`). This flag is necessary in cases where the component is not unmounted but a navigation event occurs, for example when we switch between semesters on `/timetable`.

Having these two flags allows the flexibility that the venue page needs - Scrolling to top upon navigation but not scrolling to top when there's a non-page transition navigation, which occurs when we click on the individual venues. Technically speaking we don't need to do scroll to top upon navigation because the page content clears out, but just leaving it there to show that it's possible.


v3/src/js/views/components/ScrollToTop.jsx
```diff
@@ -0,0 +1,42 @@
+// @flow
+import type { ContextRouter, Location } from 'react-router-dom';
+
+import { Component } from 'react';
+import { withRouter } from 'react-router-dom';
+
+type Props = {
+  ...ContextRouter,
+  location: Location,
+  onComponentWillMount: boolean,
```
yangshun: Flow throws errors for these two props. If I remove them Flow stops throwing error. I can't figure out what the hell is going on here.

v3/src/js/views/components/ScrollToTop.jsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  componentWillMount() {
+    if (this.props.onComponentWillMount) {
+      scrollToTop();
+    }
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (this.props.onComponentDidUpdate &&
```
ZhangYiJiang: /nit I think the prop name is slightly misleading - `onPathChange` or something similar may be clearer

v3/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -180,13 +182,16 @@ export class TimetableContainerComponent extends PureComponent<Props, State> {
...
-      />
+      <div>
+        <TimetableContent
+          semester={semester}
+          timetable={displayedTimetable}
+          colors={colors}
+          header={header}
+          readOnly={!!importedTimetable}
+        />
+        <ScrollToTop onComponentDidUpdate={false} />
```
taneliang: Would this be clearer if `ScrollToTop`'s props were both false by default? It's more obvious what `<ScrollToTop onComponentWillMount />` does compared with `<ScrollToTop onComponentDidUpdate={false} />`. I think the clarity is worth the additional typing for the other two uses of `ScrollToTop`.

taneliang: Also /nit should this line be placed right below the `<div>` to be consistent with the other 2 uses?

yangshun:
It's not that weird actually - https://youtu.be/-VC7o6wZ_a8?t=3m14s

I'll fix for the other comments.

PR #529, opened 2017-12-18
Header improvement on mobile 

### LINKED ISSUE #527: Improve header UI on mobile
The current two row v3 mobile header takes up too much space. I'm opening the discussion to what we can do to improve it 

For the space problem, I see three possible solutions 

- **Don't sticky the header at all** - the navigation buttons are not frequently used enough to warrant  making them sticky. The downside is that on longer pages this may make it harder to navigate to other pages 
- **Sticky only the navtabs** - since the top bar only has the NUSMods logo, which takes you back to the timetable anyway, we can make only the navtabs sticky 
- **Show only on scroll up** - using something like [React Headroom](https://github.com/KyleAMathews/react-headroom), we can implement the previous idea, but adopt a more app-like approach of only showing the navtabs when the user scrolls upwards 

Comments?

ZhangYiJiang:
This fixes #527 

This implements solution 2. Personally I actually prefer solution 3, since on pages like venue and module finder which are more interactive it seems better to give as much space as possible to the page UI instead of the global UI. 

However, with the current page structure solution 3 is much more technically challenging to implement correctly - the current method of sticking bits of the UI to the navbar is basically a hack. Maybe with React 16 portals we could do better, but we will just *stick* to this solution for now.


yangshun:
@ZhangYiJiang You mean on mobile?

ZhangYiJiang:
Yeah. I need to check the size of the soft keyboard, which you can't do on a desktop browser, and just play around with it to see if it feels good. I could use Chrome's USB debugging bridge, but I'm laaaaazy to get out my USB cable

v3/src/js/views/layout/Navtabs.scss
```diff
@@ -9,30 +9,52 @@
...
+  }
+
+  @include media-breakpoint-down(sm) {
+    position: sticky;
+    top: 0;
+
+    // Fake shadow - can't use box shadow because they will appear on top of the
+    // top nav bar
+    &::after {
+      $shadow-height: 1rem;
```
taneliang: Is this a little tall? We should probably use the same height as the search box so that it looks consistent, especially since this shadow is hidden when the search box is shown, implying that both the navbar and the search box are on the same level.

deploy-preview-529--nusmods netlify com_modules iphone 5

ZhangYiJiang:
The shadow thing on navigating in and out has been bothering me for a while now, but I can't figure out what's causing it.

taneliang:
@ZhangYiJiang Yeah it wasn't caused by this PR. It's a small issue but I'll file an issue later

PR #535, opened 2017-12-19
Improve module page menu

### LINKED ISSUE #516: Module side menu does not highlight correct item

ZhangYiJiang:
This fixes #516 

- Migrate module info page styles to CSS modules 
- Use a map to store the labels and IDs for the side menu 
- Add offset to scrollspy to ensure the menu items are updated correctly when scrolling through the page


v3/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -29,7 +32,18 @@ type State = {
...
+export const SIDE_MENU_LABELS = {
+  details: 'Details',
+  // TODO: Remove this when the prerequisite tree is ready
+  // prerequisites: 'Prerequisites',
+  timetable: 'Timetable',
+  cors: 'Bidding Stats',
+  reviews: 'Reviews',
+};
+
+export const SIDE_MENU_ITEMS = mapValues(SIDE_MENU_LABELS, kebabCase);
```
yangshun: You can consider using an array of objects instead of having two separate variables.

```js
const SIDE_MENU = [
  { key: 'details', value: 'Details' },
  { key: 'timetable', value: 'Timetable' },
  ...
];
```

ZhangYiJiang: But we still need to reference the key from  `<section id={...}>`, and doing `SIDE_MENU[0].key` feels wrong.

PR #536, opened 2017-12-19
Stop HistoryDebouncer when the incoming path is same as current path

### LINKED ISSUE #530: History messed up for modules and venues
To reproduce:
1. click "modules" or "venues" and try to go back to the previous screen, you need 2 to do so.
2. click "modules", "venues", go back to "modules" then try to go forward to "venues" via history, but it is wiped.

ZhangYiJiang:
This fixes #530. The bug was due to the fact that `history.push` will push items into the browser history even when they match the current path. This fixes that by simply ignoring any calls to `HistoryDebouncer#push` when the next path matches the current one. 

Importing `createHistory` from the transitive dependency history is necessary because the API doesn't expose the function we need to otherwise check that the path is the same.

### DIFF
--- a/v3/src/js/utils/HistoryDebouncer.js
+++ b/v3/src/js/utils/HistoryDebouncer.js
@@ -1,12 +1,7 @@
 // @flow
 
-import type { LocationShape } from 'react-router-dom';
-
-// A subset of RouterHistory from react-router-dom
-type HistoryManipulator = {
-  push(path: string | LocationShape, state?: any): void,
-  replace(path: string | LocationShape, state?: any): void,
-};
+import type { LocationShape, RouterHistory } from 'react-router-dom';
+import { createPath } from 'history'; // eslint-disable-line import/no-extraneous-dependencies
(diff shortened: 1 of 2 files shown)


taneliang:
Doesn't solve 4 completely though: Click "venues", click a venue, click another venue, go back. You're back at /venues.

Also this PR seems to break the venue auto scroll; clicking a venue auto scrolls up a bit too high.

ZhangYiJiang:
> Doesn't solve 4 completely though: Click "venues", click a venue, click another venue, go back. You're back at /venues.

That's what the debouncer do. 

> Also, if the latest changes in master haven't been merged in, this PR seems to break the venue auto scroll; clicking a venue auto scrolls up a bit too high.

Unlikely to have been caused by this PR. I can fix it in #526.

taneliang:
> That's what the debouncer do.

Good point. Unrelated to this PR then, but do you think venue selections shouldn't be debounced?

Issue #521, opened 2017-12-16
Improve client-side data layer 


ZhangYiJiang:
### The problems 

- **Not enough caching** - this results in too many requests being made. You can see this evident on the module and venue search pages, which reload the data on every visit even in a single session. Te The module list is currently cached on localStorage, but there isn't enough space for module and venue data. 

- **Inefficient persistence** - currently the storage layer always persists the all keys that need to be persisted when anything in the store changes. This is really inefficient - it should only persist things when keys that need to be persisted changes. 

- **Difficult for UI to handle progress** - when I dispatch an API request into Redux, it is difficult to check on the progress of the request. The request middleware/reducer is not very well designed - for instance, when the page is loaded, data for all modules on the timetable is refreshed, which means multiple `FETCH_MODULE` actions are dispatched. However, because of the way the reducer is designed, the UI can't distinguish how many calls have been successful. 

- **Difficult for UI to handle failure** - see #506, it's caused by the same problem as above 

### Proposed solutions 

- **Caching** - two layer cache, although cache invalidation may be a problem
    - Redux for in memory caching, which means anything that has been downloaded in the current session like venue information should be kept in the Redux store for the current session
    - Service Worker for larger stuff like venue information that shouldn't need to be updated more than once per day. This also conveniently allows for offline access. 
- **Persistence** - upgrade to Redux Persist 
- The other two issues probably require a rewrite of the request middleware/reducer

PR #539, opened 2017-12-20
Offline support using service worker 

ZhangYiJiang:
Basic offline support via Workbox. Tested successfully on Chrome desktop.

- Fallback to `index.html` for all navigation events that generate an error 
- Caches all calls to `nusmods.com/api` - this allows the app to work offline, since `moduleList.json` and all modules on the timetable are always loaded on initial page load, this is enough data for us to render the timetable offline. 

  The caching strategy is staleWhileRevalidate, which means the request will always be fulfilled immediately from cache if available, and the cache is then refreshed in the background. This is relatively safe since most of the time the data does not change radically between visits.  

### Wishlist 

- Pre-fetch data needed for venue search in the background (module search is not very useful since the module page requires network anyway, but venue search is pretty useful) 
- Better warning/error for when the user is offline


v3/src/index.html
```diff
@@ -7,6 +7,7 @@
   <link href="${htmlWebpackPlugin.files.publicPath}favicon-16x16.png" rel="icon" type="image/png" sizes="16x16">
   <link href="${htmlWebpackPlugin.files.publicPath}safari-pinned-tab.svg" rel="mask-icon" color="#ff5138">
   <link href="${htmlWebpackPlugin.files.publicPath}manifest.json" rel="manifest">
+  <link href="${htmlWebpackPlugin.options.apiBaseUrl}/moduleList.json" rel="preload">
```
li-kai: `<link ... as="fetch">` https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#The_basics (scroll down a bit)

v3/src/js/views/errors/ErrorPage.jsx
```diff
@@ -42,7 +42,7 @@ export default class NotFoundPage extends PureComponent<Props> {
             {this.errorMessage()}
           </h1>
 
-          {showRefresh &&
+          {showRefresh && navigator.onLine &&
```
li-kai: We shouldn't show the report dialog either right?

v3/src/js/reducers/app.js
```diff
@@ -14,16 +14,18 @@ import {
...
+import { SET_ONLINE_STATUS } from 'actions/online';
 
-const defaultAppState: AppState = {
+const defaultAppState = (): AppState => ({
   // Default to the current semester from config.
   activeSemester: config.semester,
   // The lesson being modified on the timetable.
   activeLesson: null,
-};
+  isOnline: navigator.onLine,
```
li-kai: I think we can assume the user is online when loading the app?

ZhangYiJiang: We can't, actually, because the entire app can now be run off service worker offline

v3/src/index.html
```diff
@@ -1,24 +1,26 @@
...
-  <meta name="og:image" content="${htmlWebpackPlugin.files.publicPath}share.png">
-  <meta name="og:site_name" content="NUSMods">
-  <meta name="og:type" content="website">
-  <!-- Open Search -->
-  <link href="${htmlWebpackPlugin.files.publicPath}opensearch.xml" rel="search" type="application/opensearchdescription+xml">
-</head>
-<body>
-  <div id="app"></div>
-</body>
+  <head>
```
li-kai: Wow can't believe none of us caught this before

ZhangYiJiang:
What works: 

- Viewing/changing lessons on the timetable when offline 
- Viewing added modules when offline 

Can be improved 

- Adding/removing modules - adding any module that has not been previously added/visited will result in an uncaught error - this can be improved on after #484 is merged 
- Venue info should be prefetched somehow 
- Visiting module info page of any module while offline will result in an infinite loading spinner - #521 is probably needed for this

PR #542, opened 2017-12-21
Update packages to react 16

### LINKED ISSUE #454: Update React to v16
This issue is tracking things that are blocking the upgrade to React 16 

- [x] react-disqus-thread needs to be replaced (disqus-react is the replacement, but currently blocked by https://github.com/ryanvalentin/disqus-react/pull/4)
- [ ] react-virtualized-select is throwing PropTypes and createClass warnings. This is weird because from what I can see of the  source code, the package is doing things correctly - this needs further investigation

li-kai:
Fixes #454 

Free free to commit directly over this PR


v3/src/js/views/layout/__snapshots__/Navtabs.test.jsx.snap
```diff
@@ -6,6 +6,7 @@ exports[`renders into nav element 1`] = `
 >
   <NavLink
     activeClassName="linkActive"
+    ariaCurrent="true"
```
ZhangYiJiang: I think `aria-*` attributes uses kebab-case, not camelCase - https://reactjs.org/docs/accessibility.html

ZhangYiJiang: Wait... where did this attribute even come from?

ZhangYiJiang:
Using the `<>` shorthand for `<Fragment>` isn't supported yet because we're on Babel 6, and I don't want to try to deal with upgrading to Babel 7 yet, especially since it's still in beta

ZhangYiJiang:
I believe Travis falling is due to it being on Node 6 - I can run this fine in Node 8 locally. If it would take too long to review #537 I can cherry pick out the Travis config commit from that PR into a new one so this can pass.

PR #544, opened 2017-12-22
Fix iOS scrolling on sidemenu and other places 

ZhangYiJiang:
This fixes #544 

- Uses proprietary `-webkit-overflow-scrolling` property to trigger kinetic scrolling on iOS 
- Uses `no-scroll` package with a wrapper function to disable scrolling when modal and side menu is opened. The wrapper is needed because the scroll disabling hack doesn't play well with our position sticky navigation, so we need to add some additional styles to the sticky elements when it is active.

### DIFF
--- a/v3/src/js/utils/no-scroll.js
+++ b/v3/src/js/utils/no-scroll.js
@@ -2,6 +2,11 @@
 
 import noScroll from 'no-scroll';
 
+/**
+ * Wrapper around the simple no-scroll package that also toggles a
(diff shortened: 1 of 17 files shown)


ZhangYiJiang:
@li-kai It fixes the smooth scrolling issue, but not the issue with the rest of the page being scrollable.

ZhangYiJiang:
@li-kai Can I trouble you to do another check to see if everything works on iOS?

PR #545, opened 2017-12-22
Fix filter dropdown crash

### LINKED ISSUE #546: Faculties and departments <select> crashes on select
Selecting an item in the faculties and departments `<select>` causes the page to crash. This bug does not occur in viewports larger than `xs` as a custom selector element is rendered instead of `<select>`.

@ZhangYiJiang has determined that the crash occurs at https://github.com/nusmodifications/nusmods/blob/master/v3/src/js/views/components/filters/DropdownListFilters.jsx#L175 due to a bug in React, and an issue has been filed with React at https://github.com/facebook/react/issues/11911.

ZhangYiJiang:
Fixes #546, the crash reported in https://sentry.io/share/issue/af793a00e6074c7fba6aa00e25c95480/ and https://github.com/facebook/react/issues/11911 

This PR works around the issue by simply interpolating the string. It also shifts some test code around.to hopefully make testing filters in the future less painful.

### DIFF
--- a/v3/src/js/test-utils/filterHelpers.js
+++ b/v3/src/js/test-utils/filterHelpers.js
@@ -1,0 +1,25 @@
+// @flow
+import type { Module, ModuleCode } from 'types/modules';
+import FilterGroup from 'utils/filters/FilterGroup';
+import Filter from 'utils/filters/ModuleFilter';
+
+// eslint-disable-next-line import/prefer-default-export
+export function createModule(code: ModuleCode): Module {
+  return {
+    ModuleCode: code,
+    ModuleTitle: 'Test Module',
+    History: [],
+    Types: [],
+    Department: 'Test Department',
+    ModuleCredit: '4',
+    AcadYear: '16/17',
+    CorsBiddingStats: [],
+  };
+}
+
+export function createGroup(modules: Module[] = []) {
+  return new FilterGroup('test', 'test', [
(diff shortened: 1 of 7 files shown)


ZhangYiJiang:
https://sentry.io/nusmods/v3/issues/422686860/

ZhangYiJiang:
Hmmm, that wasn't enough - I've managed to trigger the crash again by changing one of the checkboxes on the list outside the `<select>` (which caused the `<option>`s inside the select to rerender). Let me write another test case and try to find another workaround.

ZhangYiJiang:
https://sentry.io/nusmods/v3/issues/422774085/

v3/src/js/test-utils/filterHelpers.js
```diff
@@ -0,0 +1,25 @@
...
+export function createModule(code: ModuleCode): Module {
+  return {
+    ModuleCode: code,
+    ModuleTitle: 'Test Module',
+    History: [],
+    Types: [],
+    Department: 'Test Department',
+    ModuleCredit: '4',
+    AcadYear: '16/17',
+    CorsBiddingStats: [],
```
li-kai: Unfortunately with the prereqTree merged means you need to add `ModmavenTree` in

ZhangYiJiang:
Confirmed fixed using manual testing.

PR #552, opened 2017-12-23
Exam module filter and clear filters button in module finder 

ZhangYiJiang:
This implements a filter for modules with no exams. I plan to implement more filters in the future for modules that don't clash with the student's current timetable, but I need to figure out how that will look like. 

It also implements a 'clear filters' button on the header, like in v2.


v3/src/js/utils/moduleFilters.js
```diff
@@ -22,6 +22,7 @@ export const MODULE_CREDITS = 'mc';
 export const SEMESTER = 'sem';
 export const FACULTY = 'faculty';
 export const DEPARTMENT = 'department';
+export const EXAMS = 'exam';
```
yangshun: nit: `exams`?

v3/src/js/views/modules/ModuleFinderContainer.jsx
```diff
@@ -267,6 +268,11 @@ export class ModuleFinderContainerComponent extends Component<Props, State> {
       filteredModules = sortModules(this.props.searchTerm, filteredModules);
     }
 
+    const filterProps = {
```
yangshun: 👍

v3/src/js/utils/moduleFilters.js
ZhangYiJiang: Well, all of the others are singular, because I want the query string to be kept short

PR #553, opened 2017-12-24
Add global search

li-kai:
This implementation should make it clear that it is not for adding modules but searching and navigation.

Additionally I also flattened entities and added VenueBank


yangshun:
Some points for improvement:

- Shift the search bar to beside the logo and make the week indicator flushed to the right.
- The `Try "GER1000" or "LT".` can be part of the placeholder if you shift the search bar to the left. There will be ample space to show.
- I suggest not showing any result until the user has entered at least two characters. Results for one character search terms are hardly meaningful.
- If there are no results for a category, the category header should not show up.

PR #554, opened 2017-12-24
Include module table in timetable image export

yangshun:
I think people would complain if the timetable image export didn't include the exam timetable, which students were used to having, so I'm including it here. There are some hacks used:

1. I have to use `document.getElementById` instead of using `ref`. Somehow it crashes when I shift the `ref` higher up. I think the current approach is still feasible.
2. The stuff to be excluded will be temporarily (< 1s) hidden from view while `domtoimage` converts the DOM to an image. It's ugly but it's a reasonable trade-off IMO. Try it out in the Netlify preview and see for yourself.

Thoughts?

### DIFF
--- a/v3/src/js/actions/export.js
+++ b/v3/src/js/actions/export.js
@@ -29,10 +29,15 @@ export function downloadAsJpeg(domElement: Element) {
 
     // Temporarily add a stylesheet to hide the elements that we don't want
     // to show in the exported image.
-    const style = document.createElement('style');
+    const style: HTMLStyleElement = document.createElement('style');
     style.appendChild(document.createTextNode('')); // WebKit hack :(
-    document.head.appendChild(style);
-    style.sheet.insertRule(`.${TIMETABLE_EXPORT_HIDE_CLASS} { display: none; }`, 0);
+    const head = document.head;
+    if (!head) return null;
+    head.appendChild(style);
+    if (style.sheet) {
+      // $FlowFixMe - https://github.com/facebook/flow/issues/2696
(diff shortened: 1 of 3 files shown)


v3/src/js/actions/export.js
```diff
@@ -20,14 +20,33 @@ function downloadUrl(url: string, filename: string) {
...
+export const TIMETABLE_EXPORT_HIDE_CLASS = 'timetable-export-hidden';
 export function downloadAsJpeg(domElement: Element) {
   return (dispatch: Function) => {
     dispatch({
       type: `${DOWNLOAD_AS_JPEG}_PENDING`,
     });
 
-    const style = { margin: '0', marginLeft: '-0.25em' };
-    return domtoimage.toJpeg(domElement, { bgcolor: '#fff', style })
+    // Temporarily add a stylesheet to hide the elements that we don't want
```
ZhangYiJiang: Would it be better to do this in the CSS file - define the styles in the CSS file and toggle the class instead?

taneliang:
This breaks timetable export on mobile (at least on Chrome on Android) though; only the visible part of the timetable gets captured, along with a scrollbar.

ZhangYiJiang:
Can reproduce. I think what is happening because you're capturing the parent of the timetable instead of the timetable itself. Maybe try temporarily setting the timetable back to `overflow: visible` to ensure the overflow is captured, or maybe use absolute positioning so that the element is not clipped

yangshun:
Do you mean overlay over the entire page?

yangshun:
Closed in favor of https://github.com/nusmodifications/nusmods/pull/557

PR #557, opened 2017-12-24
Try out html2canvas library

li-kai:
Well just so happens they have a new rewrite that allows you to ignore elements without having to interfere with user \o/

Docs: http://html2canvas.hertzen.com/getting-started


v3/src/js/actions/export.js
```diff
@@ -26,20 +26,19 @@ export function downloadAsJpeg(domElement: Element) {
...
-      })
-      .catch((err) => {
-        console.error(err); // eslint-disable-line
-        dispatch({
-          type: `${DOWNLOAD_AS_JPEG}_FAILURE`,
-        });
+    const config = { foreignObjectRendering: true, logging: process.env.NODE_ENV === 'development' };
+    return html2canvas(domElement, config).then((canvas) => {
+      const dataUrl = canvas.toDataURL('image/png');
+      downloadUrl(dataUrl, 'timetable.jpeg');
```
ZhangYiJiang: Should be `.png`, not JPEG

yangshun:
Not sure why, but the exported image seems cut off

timetable 25

yangshun:
This works on Safari! There's still the problem of timetable clipping and weird visual artifacts on the left but it's still better than the current version. Let's use yours instead over https://github.com/nusmodifications/nusmods/pull/554?

timetable-3

v3/src/js/actions/export.js
```diff
@@ -26,20 +26,19 @@ export function downloadAsJpeg(domElement: Element) {
...
+      dispatch({
+        type: `${DOWNLOAD_AS_JPEG}_SUCCESS`,
```
yangshun: We should rename the action and type to `downloadAsImage`/`DOWNLOAD_AS_IMAGE` once and for all.

v3/src/js/actions/export.js
```diff
@@ -26,20 +26,19 @@ export function downloadAsJpeg(domElement: Element) {
...
       });
+    }).catch((err) => {
+      console.error(err); // eslint-disable-line
```
yangshun: Probably log the error in Sentry here

li-kai:
Closed in favor #596

PR #558, opened 2017-12-24
Improve v2 timetable import

ZhangYiJiang:
Converts the migrate timetable function into a Redux action. 

We use Redux state to determine if v2 timetable needs to be migrated - the default value of `settings.isV2TimetableMigrated` is false, and we set it to true after migration is complete, which will then be persisted. This also means we don't need to remove the imported key, which makes the migration process a bit safer since we can rerun it if we discover any issue later on.


yangshun:
I'm not sure whether adding into Redux is a good idea. This key is meant to be temporary but because we're integrating with the whole Redux flow, we have to add so much boilerplate when doing something like `localStorage. isV2TimetableMigrated` would suffice. It's hackier but it'll make things easier to remove especially when it's temporary?

ZhangYiJiang:
I've moved the state out of the Redux store (it does mean the UI will have to guess when migration is finished by watching the loaded state of the timetable modules, but it's not too big a deal, and you're right that it reduces quite a lot of boilerplate). 

I think keeping the migrate function as a Redux action is better because it keeps the call site in the view layer a bit cleaner, since it can dispatch actions without needing the caller to inject them in.

ZhangYiJiang:
Migration test script have been updated - https://gist.github.com/ZhangYiJiang/69a47a58b06c25615793b38eaec6d9d1 

Please help to manually test this, thanks.

PR #561, opened 2017-12-24
Cleanup static pages

yangshun:


v3/src/js/views/static/ContributorsContainer.jsx
```diff
@@ -7,34 +7,36 @@ import Loader from 'views/components/LoadingSpinner';
...
 import StaticPage from './StaticPage';
 
-const DEVELOPERS_URL = 'https://api.github.com/repos/NUSModifications/NUSMods/contributors';
+const CONTRIBUTORS_URL = 'https://api.github.com/repos/NUSModifications/NUSMods/contributors';
 
 type Props = {};
 
 type State = {
-  developersData: ?[Object],
+  contributors: ?[Object],
```
ZhangYiJiang: Hmmm, I think this type is wrong - it should be `?Array<Object>` (it currently is a tuple with a single object)

yangshun: I'll fix it

PR #564, opened 2017-12-25
Attempt to fix download on all platforms

li-kai:
Got most of this from Filesaver.js. Here's to hoping this works on all platforms.


yangshun:
Is it meant to fix Safari?

PR #566, opened 2017-12-25
Touchscreen UI improvements for module finder 

### LINKED ISSUE #547: Custom faculty/department selector unusable on iPad
1. When keyboard is down, selecting an item brings the keyboard up by focusing the text field. The text field is sometimes hidden behind the keyboard when it's focused. Focusing the text field on select is a good idea except in this situation.

2. When keyboard is up, selecting an item selects something else *sometimes*. On iPad (5th generation), I couldn't reproduce this in portrait mode. May be partly caused by https://hackernoon.com/how-to-fix-the-ios-11-input-element-in-fixed-modals-bug-aaf66c7ba3f8?

ZhangYiJiang:
Fixes #547 

This PR refactors the CSS utilities to return CSS media queries represented in object format instead of strings to allow them to be composed together, via the [json2mq](https://github.com/akiran/json2mq) library. It also makes the `makeResponsive` HOC take in a media query object / string instead of just Bootstrap breakpoints. 

Using `pointer: coarse` media query, which has [good support on Chrome and Safari](https://caniuse.com/#feat=css-media-interaction), we make the module filter UI enable touch styling (larger surfaces, use native `<select>` instead of the custom combo-box for selecting department/faculty filters) when the device is a touchscreen instead of just targeting smaller screens. 

screenshot from 2017-12-25 19-58-44

(Desktop Chrome simulating iPad)

### DIFF
--- a/v3/src/js/utils/css.js
+++ b/v3/src/js/utils/css.js
@@ -4,6 +4,8 @@
 import { entries } from 'lodash';
 
 export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+export type QueryObject = { [string]: string | number | boolean } | QueryObject[];
+
 const breakpoints: { [Breakpoint]: number } = {
   xs: 0,
   sm: 576,
@@ -19,18 +21,18 @@ function nextBreakpoint(size: Breakpoint): ?Breakpoint {
   return breakpointEntries[nextBreakpointIndex][1];
 }
 
-export function breakpointDown(size: Breakpoint): string {
+export function breakpointDown(size: Breakpoint): QueryObject {
   const nextSize = nextBreakpoint(size);
-  if (nextSize == null) return 'all';
-  return `(max-width: ${breakpoints[nextSize] - 1}px )`;
+  if (nextSize == null) return { all: true };
+  return { maxWidth: breakpoints[nextSize] - 1 };
 }
 
-export function breakpointUp(size: Breakpoint): string {
-  return `(min-width: ${breakpoints[size]}px)`;
+export function breakpointUp(size: Breakpoint): QueryObject {
+  return { minWidth: breakpoints[size] };
 }
 
-export function touchScreenOnly(): string {
-  return '(pointer: coarse)';
+export function touchScreenOnly(): QueryObject {
+  return { pointer: 'coarse' };
 }
 
 export function supportsCSSVariables() {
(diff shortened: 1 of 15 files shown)


v3/src/js/utils/css.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+
+// Define media breakpoints
+import { entries } from 'lodash';
+
+export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+export type QueryObject = { [string]: string | number | boolean } | QueryObject[];
+
+const breakpoints: { [Breakpoint]: number } = {
```
yangshun: Could we add a comment here to keep it in sync with Bootstrap's values?

v3/src/js/views/components/filters/DropdownListFilters.jsx
```diff
@@ -91,7 +92,30 @@ export class DropdownListFiltersComponent extends PureComponent<Props, State> {
         </h4>
 
         {/* Use a search-select combo dropdown on desktop */}
```
taneliang: Is this comment supposed to be at the `Downshift` line instead?

v3/src/js/views/modules/ModuleFinderContainer.jsx
```diff
@@ -147,7 +148,7 @@ export class ModuleFinderContainerComponent extends Component<Props, State> {
         } else {
           // By default, only turn on instant search for desktop and if the
           // benchmark earlier is fast enough
-          this.useInstantSearch = breakpointUp('md').matches && (time < INSTANT_SEARCH_THRESHOLD);
+          this.useInstantSearch = window.matchMedia(breakpointUp('md')).matches && (time < INSTANT_SEARCH_THRESHOLD);
```
taneliang: I think this disabled instant search on iPad, which was working great before.

Actually is there any reason for disabling instant search on mobile? I don't think instant search should be disabled on fast mobile devices at all. Instant search is a much better experience IMO.

ZhangYiJiang: Main reason for disabling instant search on mobile is that there's almost no space left on screen when the soft keyboard is up - you can at most see the title of the first result, which is not very useful

taneliang: That makes sense. Maybe disable for `xs` devices then? There's a lot of space on tablets

taneliang:
Huh that did not enable instant search on my iPad. It's still enabled on latest though.

The navtabs size is fixed, but the `<select>` and lecture/tutorial time selector is still squashed.

PR #569, opened 2017-12-26
Install Google Analytics and autotrack

### LINKED ISSUE #489: Integrate Google Analytics into NUSMods R
I'm thinking we can use the [react-ga](https://github.com/react-ga/react-ga) wrapper library for this. Please let me know if you're aware of a better alternative!

#### Relevant links:
- https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications
- [v2 Event Tracking - Google Docs](https://docs.google.com/document/d/126ycJnWR87LSHEm1xhxUeG5xu5jF9-c9UyaI42y90sQ/edit) (I got this from #33)

## Tracked events in v2

After examining the codebase, I found that there are 22 tracked events in v2, and 5 of them have been ported to v3 (to varying extents of similarity).

<details>
<summary>Click to expand</summary>

#### Export timetable
- [x] JPEG
- [ ] PDF
- [ ] HTML
- [x] iCal
- [ ] Excel

#### Preferences
- [x] Change faculty or new student status

#### Add module
- [x] From module page
- [ ] From module finder

#### Modules filter
- [x] Click on checkbox
- [ ] Click on hyperlink (I think this refers to the accordion headings?)

#### Misc
- [ ] Click on `Back to top` button

#### Keyboard shortcuts
- [ ] Timetable Search (`.`)
- [ ] Main Search (`\`)
- [ ] Navigate to page
- [ ] Navigate to anchor section
- [ ] Toggle dark mode (`x`)
- [ ] Change theme (`left`, `right`)

#### Bookmarks
- [ ] Add bookmark from module page
- [ ] Add bookmark from module finder
- [ ] Remove bookmark
- [ ] Access module from bookmarks
- [ ] Display bookmarks
</details>

## Next steps

I feel that v3 would need an updated set of events we want to track, given the new UI. Also, I don't think we should use the exact same messages as the tracking calls in v2, because then we wouldn't know whether a call came from a v2 or v3 session (unless we use a different tracking ID?)

So perhaps this task needs a bit more definition, and as a new contributor I'm not sure how this process works or how I can be involved. But in the meantime, I could go ahead and figure out how to integrate GA into v3, without adding any trackers.

What do you guys think? 🙂

ZhangYiJiang:
This PR integrates [autotrack](https://github.com/googleanalytics/autotrack) to automatically track page view changes and outbound links. Still need to add events to buttons and links so they can be captured like in v2. I'm not entirely familiar with GA other than the basics, so help would be appreciated on this.


v3/src/js/utils/google-analytics.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import config from 'config';
+
+window.addEventListener('load', () => {
```
yangshun: Could we track in production only after verifying that it works?

v3/src/js/config/app-config.json
```diff
@@ -6,6 +6,7 @@
   "corsUrl": "https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2016/2017&sem_c=2&mod_c=",
   "ivleUrl": "http://ivle.nus.edu.sg/lms/public/list_course_public.aspx?code=<ModuleCode>",
   "disqusShortname": "nusmods-prod",
+  "googleAnalyticsId": "UA-33503218-1",
```
yangshun: Can we use the new tracking ID for R, aka `UA-33503218-6`?

PR #573, opened 2017-12-26
Fix 'cannot declare a variable twice' bug in Safari 10 

ZhangYiJiang:
This is a bug in Safari 10. The Uglify JS mangle option should hopefully prevent it.


ZhangYiJiang:
Sentry issue: https://sentry.io/nusmods/v3/issues/424125331/

PR #575, opened 2017-12-26
Disable download on iOS devices and fix timetable action UI 

li-kai:
Fixes #564 

Look at this nice "support" https://caniuse.com/#feat=download for download.

Also attempts to fix responsiveness on all screen sizes.

# iOS devices

## iPad
localhost_8080_timetable_sem-2 1
## iPhone
localhost_8080_timetable_sem-2 3

# Non-iOS devices
## <=320px devices (super rare now)
localhost_8080_timetable_sem-2 4
## Mobile phones
localhost_8080_timetable_sem-2 5
## md up
localhost_8080_timetable_sem-2 7
## md up vertical mode
localhost_8080_timetable_sem-2 8
## lg up vertical mode
localhost_8080_timetable_sem-2 9

### DIFF
--- a/v3/src/js/actions/export.js
+++ b/v3/src/js/actions/export.js
@@ -8,16 +8,13 @@ import { iCalForTimetable } from 'utils/ical';
 
 function downloadUrl(url: string, filename: string) {
   const link = document.createElement('a');
-  const body = document.body;
-  if (!body) return;
 
   link.download = filename;
   link.href = url;
-
-  body.appendChild(link);
-  link.click();
-  body.removeChild(link);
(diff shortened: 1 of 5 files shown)


ZhangYiJiang:
I know it's barely visible, but the incorrect border radius on the buttons does bother me slightly. Not enough to be the one to go fix it though 😅

ZhangYiJiang:
Hmm, I just realized that Bootstrap actually supports vertical button groups. Consider using that instead of hacking it like the current solution? https://getbootstrap.com/docs/4.0/components/button-group/#vertical-variation

PR #579, opened 2017-12-26
Hotfix blank timetable import when imported data has no lesson config 

ZhangYiJiang:
This fixes a bug where a user may get a blank timetable import if the imported data is blank (because it was saved when there's no lesson data available). I think we may have this issue elsewhere too, but this hotfix tries to fix the most immediate case first.

### DIFF
--- a/v3/src/js/actions/timetables.js
+++ b/v3/src/js/actions/timetables.js
@@ -1,5 +1,5 @@
 // @flow
-import { flatMap, isEmpty } from 'lodash';
+import { flatMap, isEmpty, each } from 'lodash';
 import localforage from 'localforage';
 
 import type { ModuleLessonConfig, SemTimetableConfig } from 'types/timetables';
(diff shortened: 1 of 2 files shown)


ZhangYiJiang:
Manually confirmed that the patch works using this script - notice the sem 2 timetable data has no lessons 

``` js

var s = document.createElement('script');
s.src = 'https://cdnjs.cloudflare.com/ajax/libs/localforage/1.5.5/localforage.min.js';
s.onload = function() {
  const data = [
  [
    "timetable/2017-2018/sem1:queryString",
    "CS4243[LAB]=4&CS4243[LEC]=1"
  ],
  [
    "timetable/2017-2018/sem2:queryString",
    "GEH1067=&GEH1067=&CS3235=&CS3235=&CS5331=&CS4257=&CS3247=&CS3247=&CS3247="
  ],
  [
    "timetable/2017-2018/sem3:queryString",
    "CS1010X="
  ],
  [
    "timetable/2017-2018/sem4:queryString",
    "PC3130="
  ]
];
  let count = 0;
  data.forEach(([key, queryString]) => {
    localforage
      .setItem(key, queryString)
      .then(() => {
        count++;
        if (count === data.length) {
          localStorage.clear();
          window.location.reload();
        }
      });
  });
};
document.body.append(s);
```

PR #586, opened 2017-12-26
API last updated timestamp

taneliang:
Adds API last updated timestamp to footer.

- Formatted using native JS Date API
- "Data current as at..." string only shown after the module list is refreshed.

localhost_8080_timetable_sem-2 iphone 7


li-kai:
Not a big deal but the date should be at the top of the rest of the footer words.

taneliang:
@li-kai Yi Jiang wants to emphasise it further by putting it in a box or something. I'll play around with it when I get back to Singapore

PR #587, opened 2017-12-26
Remove old KeyboardShortcuts.js

taneliang:
Deleted an unused file that is still lying around.


ZhangYiJiang:
Could you also delete the test file? I think I forgot to delete that too, ps https://github.com/nusmodifications/nusmods/blob/b8c2d9d0df1c62eca224281c280115c6f2eca694/v3/src/js/utils/KeyboardShortcuts.test.js

PR #599, opened 2017-12-27
Show author on apps page (and remove tags)

yangshun:
Author name was missing in the porting. Removed tags because they seemed kinda useless if we can do a full text search anyway.

### DIFF
--- a/v3/src/js/views/static/AppsContainer.jsx
+++ b/v3/src/js/views/static/AppsContainer.jsx
@@ -2,7 +2,6 @@
 
 import React, { Component } from 'react';
 import axios from 'axios';
-import classnames from 'classnames';
 
 import Loader from 'views/components/LoadingSpinner';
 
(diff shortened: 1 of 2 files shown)


v3/src/js/views/static/AppsContainer.scss
```diff
@@ -15,11 +15,15 @@
...
+  }
+
   @include media-breakpoint-down(xs) {
     margin: 3rem 0;
   }
 }
 
-.tagBadge:not(:last-child) {
-  margin-right: 3px;
+.appName {
```
taneliang: Placing this outside `.appIcon` means that the app name is left aligned on `xs` though. I think the left aligned name looks better, but the icon ends up centered while everything else is left aligned. May be weird

v3/src/js/views/static/AppsContainer.jsx
```diff
@@ -2,23 +2,22 @@
...
-    repository_url?: string,
-    icon_url: string,
-    tags: Array<string>,
+  name: string,
+  description: string,
+  author: string,
+  url: string,
+  repository_url?: string,
+  icon_url: string,
+  tags: Array<string>,
```
taneliang: How was this not caught by ESLint 😱

PR #602, opened 2017-12-27
Add Issue Template for better community standards

yangshun:
Ripped off [React's](https://github.com/facebook/react/blob/master/.github/ISSUE_TEMPLATE.md).


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+<ANSWER HERE>
+
+**What is the expected behavior?**
+
+<ANSWER HERE>
+
+**Further information**
+
+- Which version of NUSMods are you currently using? (Check the footer for the version number, 201XXXXX-XXXXXXX)
+- Did this work in previous versions of NUSMods? Yes/No
```
ZhangYiJiang: Would this refer to major versions, ie. v2 and R, or just 'a few weeks ago', since we do deploys almost daily?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+- Which browser and version are you using? Chrome 62/Firefox Quantum 57/etc
```
ZhangYiJiang: /nit Should add 'eg. ' in front of the examples.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+
+**Do you want to request a *feature* or report a *bug*?**
+
+<ANSWER HERE>
+
+**What is the current behavior?**
+
+<ANSWER HERE>
+
+**If the current behavior is a bug, please provide the steps to reproduce and if possible, a minimal demo of the problem via a screenshot or a video recording (on MacOS you can use QuickTime).**
```
taneliang: /nit s/MacOS/macOS

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+<ANSWER HERE>
+
+**What is the expected behavior?**
+
+<ANSWER HERE>
+
+**Further information**
+
+- Which version of NUSMods are you currently using? (Check the footer for the version number, 201XXXXX-XXXXXXX)
+- Did this work in previous versions of NUSMods R? Yes/No
```
taneliang: Should include a Not Sure option so that people won't get stuck here and quit filing this issue. But we'll probably get a lot of not sures if we do that so...

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+- Which browser and version are you using? E.g. Chrome 62/Firefox Quantum 57/etc
+- Which device and operating system are you on? E.g. MacBook Pro/Surface Pro/etc
```
taneliang: These are examples of devices only though. We could have a separate question for OS?

.github/ISSUE_TEMPLATE.md
taneliang: Also I think we should be asking if it worked in previous versions of NUSMods instead?

PR #605, opened 2017-12-27
Add Prettier support

li-kai:
Adds prettier support, so no more code review problems.

Regarding scss files (and maybe markdown), the plugin support isn't really there yet (and scss is really hard to do wrong). I'll look into it in a future PR when ecosystem is more mature.

Check via commits.


v3/README.md
```diff
@@ -107,6 +107,8 @@ $ yarn lint:styles
 $ yarn lint:src
 $ yarn lint:test
 $ yarn lint:scripts
+# Append `--fix` to fix lint errors automatically
+# e.g. yarn lint:src --fix
```
ZhangYiJiang: Need to warn against `yarn lint:styles --fix` - StyleLint has outstanding bugs that causes it to autofix ignored lines, which messes up the themes file

PR #607, opened 2017-12-28
Adds basic timetable validation to import/migration 

ZhangYiJiang:
This fixes the bug where if the user tries to import a v2 timetable that somehow contains a module that does not exist, the timetable will never finish loading because it is forever trying to load a module that isn't there. This is a bit of a stopgap measure as there is no UI to warn the user, but at least it renders the app usable (instead of just hanging on the user).


v3/src/js/actions/timetables.js
```diff
@@ -97,10 +97,15 @@ export function setTimetable(
...
   timetable?: SemTimetableConfig,
   colors?: ColorMapping,
-): FSA {
-  return {
-    type: SET_TIMETABLE,
-    payload: { semester, timetable, colors },
+) {
+  return (dispatch: Function, getState: Function) => {
+    const validatedTimetable =
+      timetable && validateTimetableModules(timetable, getState().moduleBank.moduleCodes);
```
li-kai: I don't quite get this statement, if timetable is valid it'll return an array, which isn't the valid type for timetable

ZhangYiJiang: Derp 😞 . We really need to find a way to type our action payloads somehow

v3/src/js/utils/timetables.js
```diff
@@ -222,6 +223,17 @@ export function findExamClashes(
   return _.omitBy(groupedModules, (mods) => mods.length === 1); // Remove non-clashing mods
 }
 
+export function validateTimetableModules(
```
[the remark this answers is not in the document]
ZhangYiJiang: It does the opposite of that - it strips invalid modules from timetable. The signature is 

```
(SemTimetableConfig) => [SemTimetableConfig, ModuleCode[]]
``` 

where the second parameter consists of all of the removed module codes

Issue #594, opened 2017-12-27
Global search enhancements 


ZhangYiJiang:
## Features / v2 parity 

- [X] Link to more results in module finder and venues page as the last result - this addresses the issue that the search shows too few results compared to v2 
- [X] Alternatively, consider making it display more results in a scrollable list 
- [x] Consider displaying module semester availability in the results - again, v2 parity 
- [X] Consider making it wider on wider screens so that module titles don't wrap 
- [ ] Enable global search on mobile 

## Bugs 

- [x] Fix search term highlighting - currently it highlights the search term verbatim, which is not how the search algorithm works

li-kai:
@ZhangYiJiang I believe we can close this?

ZhangYiJiang:
@li-kai That last one is important though. Maybe open a new separate issue and flesh out the design?

PR #613, opened 2017-12-28
Add option to go to module finder from global search and show more results 

ZhangYiJiang:
Part of #594 

Would like some opinion on this. 

The goal of this is to address the issue/v2 parity for global search. The issue is that the search is limited to 7 results max, which isn't enough in some cases. Just presenting all of the results is not necessarily the best idea, because there may be venues below, and for the results of a search for eg. 'com' there would be too many modules to scroll through, making it almost impossible to get to the venues. 

### Implementation 

- Make the last option a link to the 'full' module search page - this is implemented here 
- Make the last option load more modules + add max height to limit the size of the results list 
- Show more results when there's only one type of result found

The rationale for this is that when we know for sure the user just wants one type of result, we can display a longer list since we aren't limited by the keyboard navigation issue as mentioned above


li-kai:
How about using the header? 

screenshot 2017-12-28 16 27 56

ZhangYiJiang:
I think this is a good idea. Let me give it a try and see how well it works. Medium's implementation is mouse only, but I think we should allow keyboards

Issue #600, opened 2017-12-27
Pulsing TimetableCell overlapped by the day name cell


kiangkuang:
The selected TimetableCell is overlapped by the day name cell during pulse animations
![image](https://user-images.githubusercontent.com/7030358/34376506-faeb1d24-eb26-11e7-9a53-7c53ff29453d.png)

ZhangYiJiang:
Thanks for reporting this! I agree that this visual blemish is slightly annoying. However, fixing this is non-trivial, as the module also needs to stay *under* the day name when the timetable is scrolled to the right. If you would like to help us with this, we can help you get set up - just drop into [our Telegram group](https://telegram.me/nusmods) and talk to us. Thanks!

kiangkuang:
Ah, I see the dilemma. I will see if I can do a JS solution where the z-index is set higher _only_ while not scrolled right to keep the current scrolling behavior intact.

ZhangYiJiang:
Sure. You might want to start with the `<Timetable>` and `<TimetableDay>` components in in https://github.com/nusmodifications/nusmods/tree/master/v3/src/js/views/timetable

taneliang:
Fixed by #616

PR #616, opened 2017-12-28
Fix pulsing TimetableCell animation overlap

kiangkuang:
Fix for #600 

The idea is that it will _only_ increase z-index of TimetableCell to 100 if the TimetableContent is not scrolled horizontally, else set it to 75.
This allows the pulsing animation to properly overlap the day name while in the default unscrolled view, while also allowing the day name to overlap the TimetableCell when scrolled horizontally for visibility.


www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -75,11 +75,25 @@ type Props = {
 
 class TimetableContent extends Component<Props> {
```
ZhangYiJiang: Since you're introducing State into this component, you need to define `State` and add it to `Component<Props, State>`.

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -75,11 +75,25 @@ type Props = {
...
   timetableDom: ?HTMLElement;
+  timetableWrapperDom: ?HTMLElement;
+
+  state: State = {
+    isScrolledHorizontally: false,
+  };
+
+  componentDidMount() {
+    this.timetableWrapperDom.addEventListener('scroll', this.handleScroll);
```
ZhangYiJiang: `this.timetableWrapperDom` is nullable, so you need to check if it exists before calling anything on it.

ZhangYiJiang:
You may also want to use an editor or IDE that integrates with Flow so you can see the Flow errors.

yangshun:
Run `npm run ci` locally to test if anything breaks - https://travis-ci.org/nusmodifications/nusmods/jobs/322496658

kiangkuang:
I've made the changes you've pointed out and got CI to pass.
When I have time I'll check out if the same can be achieved using React Waypoint to simplify the scroll detection.

yangshun:
Thanks for the fix, it has been deployed!

Issue #604, opened 2017-12-27
Add module box enhancements 


ZhangYiJiang:
- [ ] Add FAB on mobile which will focus the add module box when selected 

- [ ] Remove hint text (Try ...) and only show module count when text is entered 

- [ ] Automatically select first item so that hitting enter will immediately add the first item to timetable 

- [ ] Make it more noticeable on desktop by using v2's input group format

li-kai:
![screenshot 2017-12-27 20 48 08](https://user-images.githubusercontent.com/14850387/34382063-6255d6b4-eb47-11e7-98a7-557ed0c88473.png)

For point 4, what do you think of this approach instead?

ZhangYiJiang:
Do you mean we should try to teach the user what the add module box looks like when they first use the app, like an onboarding process, when the timetable is first empty? I like the idea, but you do need to add some text to the effect of 'get started by adding your first module', because otherwise it would be a bit weird to see a text box in the middle of the timetable. 

The other problems I see is that 

- This won't work for existing users who already have a timetable 
- This doesn't necessarily teach the user where the add module box is, since it's not in the usual position

PR #617, opened 2017-12-28
Quick add module box improvements 

ZhangYiJiang:
Part of #604 

This implements these simple improvements 

- Adds a 'no modules found' message when search finds no results 
- Remove tip from desktop 
- Select first item by default


www/src/js/views/timetable/ModulesSelect.jsx
```diff
@@ -131,6 +133,11 @@ class ModulesSelect extends Component<Props, State> {
             <strong>{this.props.moduleList.length}</strong> modules.
           </div>
         )}
+        {showNoResultMessage && (
+          <div className={styles.tip}>
+            No modules found for <strong>&apos;{inputValue}&apos;</strong>.
```
li-kai: Use "&quot;" instead, it makes more sense. &apos is for "I'm" etc

PR #623, opened 2017-12-29
Server-side timetable exports 

### LINKED ISSUE #596: Server-side screenshot rendering for PDF and image timetable export
We've had issues trying to get client side timetable export working consistently. Here's a proposal to switch to server side rendering like v2 

### Client side 

- When the user requests for an export, we serialize the relevant Redux store (the current timetable config, theme color map, hidden modules, dark mode, timetable orientation) and send it to the server 

### Server side 

- Install [Puppeteer](https://github.com/GoogleChrome/puppeteer), which runs a Headless Chrome instance (API reference - https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md) 
- Create a second Webpack entry point that consists of only `<TimetableContainer>` 
- Create a small Express server that listens to incoming requests 
- Deserialize the state from incoming requests, push it into the Redux store 
- Take screenshot and send it back to the client 

This is surprisingly simple, and should be very robust because we're using the same rendering engine as the client. This will also allow us to support more features like 

- PDF export 
- Multiple sizes 
- Include module table in the export 

The only difference I can foresee is that font may be different because currently we use the OS default font, which obviously the server cannot support, but this is a small issue.

ZhangYiJiang:
Fixes #596. 

## Overview 

- When the user selects an export option, the app serializes select parts of the store, JSON stringifies it, and sends it to the server via a GET endpoint under `/export`
  - GET seems to be the appropriate choice in terms of the verb. The length of the query string isn't too big a concern - [all modern browsers supports sending really long query strings](https://stackoverflow.com/q/812925/313758)
  - JSON is chosen because trying to (de)serialize nested data structures from query string results in a sticky mess, and also because Flow hates me 
- The export server launches a headless Chromium instance when it is started, injects the state into the page, and takes a screenshot/pdf and send it back. This makes it fast as Puppeteer does not need to be relaunched for every render, and the page does not need to be reloaded. 
 

##  Client side

- Created a `SET_EXPORTED_DATA` action that replaces various states with imported state  
- Added a class that hides UI elements that are not relevant in the screenshot 
- Create `ExportMenu` component that creates the dropdown menu to hold export options 
- Add config to proxy calls to export server to the local version 

## Server side

- Create `/export` that contains a Koa app; uses Nodemon to automatically restart it during development, and when its assets are updated 
- During development, the server proxies Webpack dev server from `/www`. During production, it reads data off a built version of the page. 
- The app reads module data from API during development, and local filesystem during production 
- The app has two GET routes - `/pdf` and `/image` - that takes data via query strings. The server 

## Known issues 

- Custom color does not work 
- Need to return a proper error page when an error is caught in Koa


export/config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
yangshun: This is meant to be changed in production? Should it be checked in to git then?

export/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "nusmods-export",
+  "version": "0.0.1",
+  "description": "Server-side timetable rendering for export service ",
```
yangshun: /nit Trailing space

PR #588, opened 2017-12-26
[WIP] Indicate current timeslot on Timetable

zameschua:
### Motivation
This is a port from v2 where the current day would be highlighted on the timetable.
A new prop, `isCurrentDay`, is being passed to TimetableDay. (Uses Javascript `Date` object to calculate)

### Progress
Currently it looks like this
image

Works in vertical mode
image

Works in night mode as well
image

I aim to get it to look like this screenshot from @ZhangYiJiang but not sure what's the best way to insert the line indicator for the current time.
image

Also I'm trying to learn so let me know if there's anywhere to improve! 🙏

PR #624, opened 2017-12-29
Indicate current time on timetable

zameschua:
### Motivation
This is a port from v2
Replaces #588 (because I'ma Git noob)

*** Note that the time indicator will always follow Singapore's time zone

Edit: also realised I should be doing npm run ci, and config my linter to work with flow. Will squash my commits before merging.

### Current progress
Trying to get the current time indicators to not-duplicate when the rows are stacked like that. Not sure how though, any tips? 😢 
image

Edit: Got rid of the duplicate time indicators!
image

#### Horizontal mode with obstruction
image
#### Horizontal mode without obstruction
image

#### Vertical mode with obstruction
image
#### Vertical Mode without obstruction
image

#### Night mode
image


www/src/js/utils/timify.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import type { DayText, LessonTime, Lesson } from 'types/modules';
+import moment from 'moment-timezone';
```
ZhangYiJiang: Hmmm, currently we only use moment in tests. Moment (and Moment Timezone) is a pretty large library, and I would prefer not having to add it to the bundle. For our purpose I think we can assume the user's local time is in Singapore. Alternatively, use the UTC functions and offset by +8 (aren't you happy Singapore doesn't have DST 😉?)

www/src/js/views/timetable/CurrentTimeIndicator.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.circularPart {
+  position: absolute;
+  top: -0.25rem;
+  left: -0.25rem;
+  width: 0.5rem;
+  height: 0.5rem;
+  border-radius: 50%;
+  background: $primary;
```
ZhangYiJiang: Should use `theme-color()` instead

www/src/js/views/timetable/TimetableDay.jsx
```diff
@@ -26,6 +27,7 @@ function TimetableDay(props: Props) {
     backgroundSize: `${size}% ${size}%`,
   };
 
+  if (props.isCurrentDay) rowStyle.boxShadow = 'inset 0 0 0 99999px rgba(128, 128, 128, 0.1)';
```
ZhangYiJiang: Is it possible for this to be a class? We're using CSS in the server-side export PR to hide elements from export, so this needs to be a class for that to work

www/src/js/views/timetable/TimetableTimings.jsx
```diff
@@ -12,13 +12,18 @@ type Props = {
 
 function TimetableTimings(props: Props) {
   const range = _.range(props.startingIndex, props.endingIndex);
+
+  const currentDayStyle: Object = {
+    color: 'text-primary',
```
ZhangYiJiang: Ditto above. Also, `text-primary` is not a valid value for `color` - that's why the color isn't changed. `text-primary` is a class - you can apply it with `classnames` - `className={classnames(styles.time, 'text-primary')}`

www/src/js/views/timetable/CurrentTimeIndicator.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+:global(.verticalMode) .currentTimeIndicator {
+  width: 100%;
+  height: 0%;
+}
+
+.circularPart {
+  position: absolute;
+  top: -0.25rem;
```
ZhangYiJiang: Can extract `0.5rem` out to a SASS variable

www/src/js/views/timetable/CurrentTimeIndicator.scss
```diff
@@ -0,0 +1,38 @@
...
+  width: 0.5rem;
+  height: 0.5rem;
+  border-radius: 50%;
+  background: $primary;
+}
+
+.linePart {
+  position: relative;
+  top: 0%;
+  left: -0.05rem;
```
ZhangYiJiang: Ditto above - extract to variable

www/src/js/views/timetable/TimetableTimings.jsx
ZhangYiJiang: Actually, why is this here at all? I thought we're highlighting the current *day*, not *hour*.

ZhangYiJiang:
Also, there's no need to manually squash commits - GitHub does that automatically when merging.

www/src/js/views/timetable/CurrentTimeIndicator.scss
yangshun: Change to -0.2rem. That makes it middle. Since they're not the same, you don't have to extract as a variable anymore. However, extract the width/height.

zameschua: @yangshun I tried -0.2rem and the circle was off center on my browser. Why -0.2rem though, when the height and width is 0.5, wouldn't the middle of the circle be at -0.25rem?

zameschua:
image
Got rid of the duplicate time indicators!
Also got rid of the timezone check.

Ready for merging (I hope)

yangshun:
The line is obscured behind cells. It shouldn't be the case.

zameschua:
@yangshun I put it in front of the cells before but it blocks the information sometimes (location, tutorial group), which is why I changed the Z index to place it behind the cells. Also Google Calendar puts the line behind the cells which is why I'm following their design too.

I can change the Z-index to put it in front though!

yangshun:
My Google Calendar put the line in front of the cells. Which version of Calendar are you using?

I think most calendars put the line in front of the cells:




screen shot 2017-12-30 at 1 36 14 pm

In cases where they don't, the line will be across the whole week:

yangshun:
Another good thing to have will be to start a timer that runs every minute to update the position of the line, or minimally update the line when the view comes into focus. If the user does not refresh the page after initial page load, the line will not move and the timing indicated will be outdated.

zameschua:
@ZhangYiJiang Gimmie a while I'm working on the changes proposed by @yangshun !

zameschua:
### Changes
1. I shifted the logic for calculating the margin offset for the `currentTimeIndicator` to `Timetable` so that it doesn't get re-calculated by every row
2. `forceUpdate()` is called on `Timetable` every minute. I know it's not the most elegant solution but the other solutions I tried seem worse

Also currently there's a bug with the `currentTimeIndicator` being out of position, I'll check it out

li-kai:
Closing because circle ci isn't triggering

www/src/js/views/timetable/Timetable.jsx
```diff
@@ -20,20 +26,52 @@ type Props = {
...
+  }
+
+  componentDidMount() {
+    this.interval = setInterval(() => {
+      this.forceUpdate();
+    }, 60000);
+  }
+
+  componentWillUnmount() {
+    if (this.interval !== null) {
```
yangshun: You don't have to check here. `clearInterval` is a no-op if `this.interval` is `null`.

www/src/js/views/timetable/Timetable.jsx
```diff
@@ -44,16 +82,19 @@ class Timetable extends PureComponent<Props> {
         <div className={styles.container}>
           <TimetableTimings startingIndex={startingIndex} endingIndex={endingIndex} />
           <ol className={styles.days}>
-            {schoolDays.map((day) => (
+            {schoolDays.map((day, index) => (
               <TimetableDay
                 key={day}
                 day={day}
+                isCurrentDay={index === currentDayIndex}
```
yangshun: This is too many props just to display the time indicator. Ideally, we use one prop just to decide whether to show the line. Having too many props makes it easy to miss out one of them in case it is used elsewhere. `currentTimeIndicatorVisible` and `currentTimeIndicatorStyle ` should not be passed into `TimetableDay` if it's not current day, so that makes `isCurrentDay` a redundant.

The only prop needed is `currentTimeIndicatorStyle`. Make `Timetable` control whether a `currentTimeIndicatorStyle` object is passed in. Do the rendering logic check inside `Timetable`  rather than inside `TimetableDay` so that the logic check is only done *once* per render rather than in each `TimetableDay` which is not necessary.

```js
const isCurrentDay = index === currentDayIndex;
return (
  <TimetableDay
    key={day}
    day={day}
    ...
    currentTimeIndicatorStyle ={isCurrentDay && currentTimeIndicatorVisible ? currentTimeIndicatorStyle : null}
    ...
  />
);
```

www/src/js/views/timetable/TimetableDay.jsx
```diff
@@ -35,6 +41,10 @@ function TimetableDay(props: Props) {
         <span className={styles.dayNameText}>{props.day.substring(0, 3)}</span>
       </div>
       <div className={styles.dayRows} style={rowStyle}>
+        {props.isCurrentDay &&
```
yangshun:
```js
{props.currentTimeIndicatorStyle && <CurrentTimeIndicator style={props.currentTimeIndicatorStyle} />}
```

www/src/js/views/timetable/Timetable.jsx
zameschua: `isCurrentDay` prop is currently being used to decide if the current `timetableDay` should be shaded so I think needa keep that

yangshun: My bad, I missed that out.

www/src/js/views/timetable/Timetable.jsx
zameschua: Flow doesn't allow me to pass null value into clearInterval

yangshun: Ok then leave it as it is lol

PR #625, opened 2017-12-29
Redirect users with v2 timetable URLs to homepage

ZhangYiJiang:
I noticed some bug reports with URLs that appear to be v2 timetable URLs (previously bookmarked, perhaps?). These are not handled properly in R and result in a 404 page. This PR detects those URLs and redirects them back to the homepage.


yangshun:
Possible to use RR to route instead? v2 url has two params while R only has one. Is that sufficient for RR to differentiate?

ZhangYiJiang:
@yangshun Not possible because we do have a second param - the `share` action

yangshun:
Ok got it

Issue #585, opened 2017-12-26
RFC: New view for exam schedule 


ZhangYiJiang:
One of the common complaint about R is that the new card based interface makes it hard to compare exam schedule compared to the old tabular UI. This is true, but the old UI was never very good for this either. For instance, it was not clear at a glance how far apart my exams are, or which days of the week they fell on. 

I propose a new view which students can toggle to display the modules in a calendar format (ie. a grid), like the Google Calendar month view

![screenshot from 2017-12-27 01-18-31](https://user-images.githubusercontent.com/445650/34361312-e4158c5c-eaa3-11e7-827b-e46639bc87c0.png)

This allows students to see immediately how heavy their exam schedule is, how much time they have in between, which day of the week they fall on, etc.

PR #636, opened 2018-01-01
Modules table

li-kai:
Make modules table table on big screens and up.

Still have no idea how to make the button group spacing small.


li-kai:
After much discussion, we are going to kill this over #585.

Reasons for initiating this PR were to let users sum MCs with ease and to sort by exam dates.
While a table solves the two pretty easily, making the solution responsive, and exam clashes intuitive become harder.

TLDR: We'll review this again as needed.

PR #637, opened 2018-01-01
Semester in results and "no result" message for global search 

ZhangYiJiang:
This adds two features - 

## Semester in global search results 

Pretty self explanatory. The result is a bit ugly, to be honest. Maybe adjusting the colors would help. Performance seems fine, at least on desktop. We may need to virtualize the list on mobile. 

screenshot from 2018-01-01 23-48-14

## "No result" message 

Inform the user that their search returned no result, and show links to the full modules and venue search. 

screenshot from 2018-01-01 23-48-29

The code is a bit longer than I'd like, but I'm not sure how I can shorten it.

### DIFF
--- a/www/src/js/views/components/filters/styles.scss
+++ b/www/src/js/views/components/filters/styles.scss
@@ -113,11 +113,6 @@ $h-padding: $grid-gutter-width / 2;
     }
   }
 
-  mark {
-    padding: 0;
-    border-bottom: 1px solid var(--yellow);
-  }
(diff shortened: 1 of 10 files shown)


ZhangYiJiang:
I don't think max-length is necessary. The thing only breaks if you try to type in too many letters without a space, and that's enough of an edge case that I don't really care for it. We could also use `text-overflow: ellipsis` if we really want.

PR #638, opened 2018-01-02
Improve side menu styling 

### LINKED ISSUE #548: Module filter pane covers footer
E.g. https://latest.nusmods.com/modules?faculty=duke-nus-medical-school&level=6&mc=8

latest nusmods com_modules_faculty duke-nus-medical-school level 6 mc 8

ZhangYiJiang:
Fixes #548 and some other bugs 

Switches to `position: sticky` on md and above to avoid the clipping issue, and also notification pushing the side menu on module info page down. 

Also create a `touchscreen-or-breakpoint-down` mixin - SCSS mixins involving `@media` rules do not compose well, so unfortunately we have to define a new mixin. This fixes the styling of the module search menu on sm and below on desktop browsers (which should be quite rare, but the bug still bothers me).


ZhangYiJiang:
I think those are fine. They're minor, and like I said I don't actually expect many users to have this viewport configuration

PR #645, opened 2018-01-02
Implement undo for destructive actions

### LINKED ISSUE #505: Undo for destructive actions

taneliang:
Resolves #505.

Implements a "reducer", `unredo`, to undo/redo. `unredo` is a pure function which takes in the previous and present states as well as an action, and returns an updated state after performing these 2 main steps:

1. Update the undo/redo history. This is done by the `computeUndoStacks` function, another "reducer".
2. If the action === undo/redo, apply the present state computed in step 1 to the configured key path.

We only store the undo/redo history for certain actions, and only some key paths of the app state are stored for undo. These can be configured. `unredo` is created by calling the `undoHistory` function with the config object.

### Notes

- We don't watch the app state itself for changes. Although that is ideal, I think it'll be slow as we'll need to diff the state on every action. As such, we only watch for relevant action types.
- This will only work reliably as long as the persisted key paths are ONLY modified by those actions. This is because we only record changes to that data when those actions are performed.
- This approach allows for anything in the Redux store to be undone - just set the actions and key paths to persist. Although it'll be kind of a global undo/redo than a localized undo/redo system.
- Redo and history limiting mechanisms were implemented but not used.
- Changes to Notification:
    - A priority notification feature has been added, allowing a notification to... [trump all notifications](https://youtu.be/Iimj0j4NYME), if you will. This allows a user to spam delete a bunch of modules, and for the order of the notifications to match the undo order.
    - Notifications will now also self-pop when the action button is clicked. This can be disabled by returning `false` in the handler.
    - An optional `willClose(discarded: boolean, actionClicked: boolean)` callback has been added to `NotificationOptions`. This is used to pop undo states off the undo history past stack so that the undo messages will correspond to the modules that the undo button will bring back.

### Todo

- [x] Undo on timetable
- [x] Use snackbar UI when implemented.
- [x] Make import undoable
- [x] Tests

### Possible future additions

We may want to add the ability to undo groups of actions separately, e.g. undo settings changes and timetable changes separately, etc. This should be easily achieved by storing a `UndoHistoryState` object for each group, and passing a group ID to the undo/redo functions.


www/src/js/actions/undoHistory.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import type { FSA } from 'types/redux';
+
+export const UNDO = 'UNDO';
+export function undo(): FSA {
+  return { type: UNDO, payload: {} };
```
ZhangYiJiang: `payload` should be `null`

www/src/js/reducers/undoHistory.js
```diff
@@ -0,0 +1,131 @@
...
+      present: pick(presentAppState, config.keyPathsToPersist),
+      future: [],
+    };
+  }
+
+  switch (action.type) {
+    case UNDO: {
+      // Abort if no past, or present is unknown
+      if (past.length === 0 || !present) return state;
+      const previous = past[past.length - 1];
```
ZhangYiJiang: /nit Personally I prefer `_.last` as it is slightly more readable

www/src/js/views/components/Notification.jsx
```diff
@@ -131,7 +144,13 @@ export class NotificationComponent extends Component<Props, State> {
                 <button
                   type="button"
                   className="mdc-snackbar__action-button"
-                  onClick={shownNotification.action.handler}
+                  onClick={() => {
+                    this.setState({ actionClicked: true });
+                    const action = shownNotification.action;
+                    // Don't auto-close if handler returns false
```
ZhangYiJiang: Comment is wrong, although I think returning false to prevent default is more logical. You can do that by using `=== false`. 

Also I think you can get away with not checking for `action` being null if you destructure it at the start of the function, or maybe if you make it covariant in the state

www/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -131,7 +135,17 @@ export class TimetableContainerComponent extends PureComponent<Props, State> {
             <button
               className="btn btn-success"
               type="button"
-              onClick={() => this.importTimetable(semester, timetable)}
+              onClick={() => {
+                this.importTimetable(semester, timetable);
+                this.props.openNotification('Timetable imported', {
```
ZhangYiJiang: Can this be moved into `this.importTimetable`?

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -145,7 +156,22 @@ class TimetableContent extends Component<Props, State> {
         }))}
         horizontalOrientation={horizontalOrientation}
         semester={this.props.semester}
-        onRemoveModule={(moduleCode) => this.props.removeModule(this.props.semester, moduleCode)}
+        onRemoveModule={(moduleCode) => {
+          this.props.removeModule(this.props.semester, moduleCode);
```
ZhangYiJiang: May want to extract this entire block into its own class function

www/src/js/reducers/index.js
```diff
@@ -23,13 +27,20 @@ export type State = {
...
   moduleFinder: ModuleFinderState,
+  undoHistory: UndoHistoryState,
 };
 
 // $FlowFixMe: State default is delegated to its child reducers.
 const defaultState: State = {};
+const unredo = undoHistory({
+  reducerName: 'undoHistory',
+  actionsToWatch: [ADD_MODULE, REMOVE_MODULE, SET_TIMETABLE],
+  keyPathsToPersist: ['timetables'],
```
ZhangYiJiang: `persist` may be a little misleading. Perhaps just `whitelist` would do. Also I believe `theme.colors` should also be watched.

ZhangYiJiang:
Hmmm, the basic undo doesn't seem to be working - 

1. Add a module
2. Make some modification - change color, move one of the lesson around 
3. Delete module 
4. Undo
5. Observe that although the module is restored, the changes made in step 2 are not restored 

I'm trying to figure out what's happening here. For instance, if I refresh between 2 and 3, the changes in 2 are restored.

www/src/js/reducers/undoHistory.js
```diff
@@ -0,0 +1,116 @@
...
+  keyPaths.forEach((path) => {
+    const presentValue = get(present, path);
+    if (presentValue) set(newState, path, presentValue);
+  });
+  return newState;
+}
+
+// Given a config object, returns function which compute new state after
+// undoing/redoing/storing present as required by action.
+export default function undoHistory(config: UndoHistoryConfig) {
```
ZhangYiJiang: Since this is a higher-order function that creates undo reducers, you may want to call it that, eg. `createUndoReducer`

www/src/js/views/timetable/TimetableContainer.jsx
```diff
@@ -103,6 +107,15 @@ export class TimetableContainerComponent extends PureComponent<Props, State> {
     const colors = fillColorMapping(timetable, this.props.colors);
     this.props.setTimetable(semester, timetable, colors);
     this.clearImportedTimetable();
+
+    this.props.openNotification('Timetable imported', {
+      timeout: 12000,
+      overwritable: true,
+      action: {
+        text: 'Undo',
+        handler: () => this.props.undo(),
```
ZhangYiJiang: /nit Can set `handler: this.props.undo`

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -126,6 +136,30 @@ class TimetableContent extends Component<Props, State> {
...
+      return;
+    }
+
+    this.props.removeModule(this.props.semester, moduleCode);
+    this.props.openNotification(`Removed ${moduleCode}`, {
+      timeout: 12000,
+      overwritable: true,
+      action: {
+        text: 'Undo',
+        handler: () => this.props.undo(),
```
ZhangYiJiang: Ditto above

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -313,6 +347,7 @@ class TimetableContent extends Component<Props, State> {
                       <ModulesSelect
                         moduleList={this.props.semModuleList}
                         onChange={(moduleCode) => {
+                          this.props.popNotification();
```
ZhangYiJiang: Is this still necessary?

taneliang: It is. Otherwise this bug will be introduced:

1. Remove a mod.
2. Quickly add another mod.
3. Click undo. The mod added in step 2 disappears.

This line ensures that step 3 is not possible.

ZhangYiJiang: Ah. I know some implementations (but not the reference) dismisses notification on any sort of user interaction with the page. This is probably fine for now though.

taneliang: I tried removing this line and the notification isn't automatically dismissed. Anyway I don't think we want our notifications dismissed on any interaction, otherwise the theme changing notifications may be dismissed too quickly

PR #646, opened 2018-01-02
Add notification component

ZhangYiJiang:
This integrates the [snackbar component from Material Components Web](https://material-components-web.appspot.com/snackbar.html). Currently it is used when

- Dismissing the CORS notification - the user is given a link to the settings page 
- Switching themes using keyboard 
- Toggling night mode using keyboard 

We're also ditching RMWC, because frankly I'm not seeing much of a point. The API for the snackbar makes more sense when it's imperative, so instead to open the notification, you simply dispatch the `openNotification` action. 

We only use the styles from Material, not code because the reference implementation suffers from a relatively serious bug (https://github.com/material-components/material-components-web/issues/1398) which cannot be easily worked around, and the API as a whole is just not very good (we need a `.close()` method - https://github.com/material-components/material-components-web/issues/1466). The nitty gritty implementation details are documented here: https://github.com/nusmodifications/nusmods/pull/646/files#diff-f86c691506aa3598a615314df7df403cR25 

Compared to the reference design we do not have focus management, which is unfortunate (bad for accessibility), but at least the dang snackbar actually closes properly. 

Also, currently the settings button doesn't scroll to the correct section. I'll fix that in another PR.


ZhangYiJiang:
I think it may be better to reduce the amount of time it stays open. I do think it's important to have a notification just to avoid surprising the user if they accidentally hit z/c and suddenly their entire timetable changes color. 

It's also mildly useful to tell them the theme name so that they can come back to it if they want to in the future.

li-kai:
Hmm that is mildly useful indeed. How about debouncing them?

taneliang:
I think it's still confusing to only be told that the theme changed and not be told what it was changed from, why it was changed, and how to undo. Maybe it'll be clearer if the toast said "Theme changed from Tomorrow to Railscast because you pressed `z`. Hit `c` to change back." A little verbose though.

Also found a bug: If you change the theme and change browser tabs while the toast is still up, the toast won't get dismissed.

ZhangYiJiang:
I think we can show the long message when the key is first pressed. It's still too long to fit into the snackbar though. 

I can reproduce the tab switching bug, and it looks like an upstream bug, which is really quite annoying https://github.com/material-components/material-components-web/issues/1398. We may be able to force some sort of workaround, let me investigate further.

taneliang:
Maybe we could add a "Why" button after the message if we have a keyboard shortcuts help modal. That'll also help when we add more shortcuts

www/src/js/types/reducers.js
```diff
@@ -10,11 +10,24 @@ import type {
 import type { Mode } from 'types/settings';
 
 /* app.js */
+
+export type NotificationOptions = {
+  timeoutInMs?: number,
```
yangshun: `timeoutInMs` <- the ms is pretty redundant. Most time settings in JS land are in ms anyway.

www/src/js/views/AppShell.jsx
```diff
@@ -67,8 +70,11 @@ const weekText = (() => {
 })();
 
 function setMode(mode: Mode) {
-  if (!document.body) return;
-  document.body.classList.toggle('mode-dark', mode === DARK_MODE);
+  const { body } = document;
+  if (!body) return;
+
+  body.classList.toggle('mode-dark', mode === DARK_MODE);
```
yangshun: nit: `const isDarkMode = mode === DARK_MODE;`

www/src/js/views/components/Notification.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+  showSnackbar() {
+    if (this.snackbar && this.props.notification) {
+      const { message, timeoutInMs, action, multiline } = this.props.notification;
+      const actionParams = action ? { actionText: action.text, actionHandler: action.handler } : {};
+
+      this.snackbar.show({
+        multiline,
+        message,
+        timeout: timeoutInMs,
```
yangshun: Without the `inMs`, you can just do `timeout,`! 😄

www/src/js/views/components/Notification.jsx
```diff
@@ -0,0 +1,71 @@
...
+        aria-atomic="true"
+        aria-hidden="true"
+        ref={(r) => {
+          this.element = r;
+        }}
+      >
+        <div className="mdc-snackbar__text" />
+        <div className="mdc-snackbar__action-wrapper">
+          <button type="button" className="mdc-snackbar__action-button" />
+        </div>
```
taneliang: Should probably not render this action if there's no action - it's widens the toast with empty space

ZhangYiJiang:
@taneliang @yangshun These are the requirements/limitations of the package. Material components are not super customizable. I'll try to dig into the package a little, but I won't be surprised if these aren't options.

yangshun:
Highly tempted to tell you to write your own and not use mdf but if we do that we'll soon be able to release `nusmods-material-design-components`. Use your own best judgement 👍

www/src/js/views/components/Notification.jsx
```diff
@@ -0,0 +1,129 @@
...
+        })}
+        aria-live="assertive"
+        aria-atomic="true"
+        aria-hidden={!!shownNotification}
+        onTransitionEnd={this.onTransitionEnd}
+        ref={(r) => {
+          this.element = r;
+        }}
+      >
+        {shownNotification ? (
```
taneliang: Use `!!shownNotification && (` here?

ZhangYiJiang: Tried that, doesn't work. For some reason React doesn't like DOM elements having `null` as children (this may be a bug in the React libdef though).

ZhangYiJiang: No wait, I take that back - it works. Hmmm, not sure what error I was seeing earlier on then

www/src/js/views/components/Notification.jsx
```diff
@@ -0,0 +1,129 @@
...
+    this.transitioning = true;
+    clearTimeout(this.timeoutId);
+    this.setState({ isOpen: false });
+  };
+
+  render() {
+    const { shownNotification, isOpen } = this.state;
+
+    return (
+      <div
```
yangshun: Does doing `!!shownNotification && <div>...</div>` here work?

ZhangYiJiang: For `transition` to work, I believe the parent `div` need to be kept in the DOM. Let me double check this.

PR #651, opened 2018-01-02
Split export related code into their own bundle

ZhangYiJiang:
This shaves ~11kb off the two main bundles gziped by bundling `ical-generator` and `react-qr-svg` together into a lazy loaded bundle. Since both of these are async already, the perceived performance impact should be negligible.   

```
File sizes after gzip:

  114.33 KB (+5.8 KB)  build/vendor.91f9be154916caa4e35a.js
  58.72 KB (+5.81 KB)  build/app.6e114a37a62538adf0cb.js
  21.62 KB             build/app.6e114a37a62538adf0cb.css
  15.97 KB             build/a71072d948948dc0bfcd.js
  3.32 KB              build/4f24304e46dc5f28dea3.js
```

### DIFF
--- a/www/src/js/actions/export.js
+++ b/www/src/js/actions/export.js
@@ -1,10 +1,7 @@
 // @flow
-import ical from 'ical-generator';
-
 import type { Module, Semester } from 'types/modules';
 import type { State } from 'reducers';
 
-import iCalForTimetable from 'utils/ical';
 import { hydrateSemTimetableWithLessons } from 'utils/timetables';
 import type { ExportData } from 'types/export';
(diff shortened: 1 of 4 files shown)


ZhangYiJiang:
Since the component is just a class, the component in `this.QRCode` gets destroyed (since it's a property on the instance), but I believe Webpack's `import()` implementation should cache the loaded file so it should be fine (let me do some research and double check).

ZhangYiJiang:
Can confirm that the bundle will only be downloaded once, no matter how many times `import()` is called. 

screenshot from 2018-01-03 10-03-28

Issue #598, opened 2017-12-27
No feedback on successful addition of module to timetable on mobile


yangshun:
Steps to reproduce:

1. View site on mobile
2. Go to /timetable
3. Tap on "Add module to timetable" input
4. Key in some text and add any module
5. Search term and results disappear with no indication of success/failure.

One solution I have is to retain the search term and results screen so that the `Added` label appears and the user will know that the operation was successful. I think it's ok to leave the search term and results there as people usually add modules from the same faculty anyway.

![image 2017-12-27 00 24 02](https://user-images.githubusercontent.com/1315101/34375616-4095805c-ea9c-11e7-8bd5-c248d237523a.jpg)

ZhangYiJiang:
It's actually a problem on desktop too, since the process is async there may be a small lag between hitting enter on the keyboard or clicking and the item actually appearing, so we need some sort of spinner to inform the user. Here's my idea - use the dropdown menu itself to do this - 

- When an item is entered, the hint text turns into a small spinner and the text 'Adding {ModuleCode} {ModuleTitle}...'. On desktop this can be bigger (icon above text instead of inline) since we have more space. 
- Once the module is added, on mobile the hint text returns. On desktop the modal can just clear itself. 

I think this is better than using a toast stuck to the bottom of the screen (which also doesn't work on iOS). Comments?

yangshun:
That sounds really good!

taneliang:
Implemented in #655

ZhangYiJiang:
I actually don't think this should be closed, because there's still the async problem - basically modules are added asynchronously, but currently there's no spinner, so it looks as if your input is lagging when you add a module not in the module bank. We could open another issue for that though.

taneliang:
That's true. That's a different issue though. I'll open one for that

PR #655, opened 2018-01-03
Make ModulesSelect remain open after selection

nexolute:
This PR is made with respect to #598.

I had implemented @yangshun original solution:

> One solution I have is to retain the search term and results screen so that the Added label appears and the user will know that the operation was successful. I think it's ok to leave the search term and results there as people usually add modules from the same faculty anyway.


www/src/js/views/timetable/ModulesSelect.jsx
```diff
@@ -33,12 +33,25 @@ class ModulesSelect extends Component<Props, State> {
...
     // Refocus after choosing a module
     if (this.input) this.input.focus();
-    if (selection) this.props.onChange(selection);
+    if (selection) {
+      this.props.onChange(selection);
+    }
+  };
+
+  onStateChange = (changes: any) => {
+    if (Object.prototype.hasOwnProperty.call(changes, 'inputValue')) {
```
ZhangYiJiang: Just use `_.has` for this. Also, combine the two `if` statements

www/src/js/views/timetable/ModulesSelect.jsx
```diff
@@ -33,12 +33,25 @@ class ModulesSelect extends Component<Props, State> {
...
+      if (!Object.prototype.hasOwnProperty.call(changes, 'selectedItem')) {
+        this.setState({
+          inputValue: changes.inputValue,
```
ZhangYiJiang: I see two bugs (not entirely sure if they are caused by this specific line, will investigate a bit further when I have time) 

- When blurring after selecting an item, this causes the input value to be set to the last select item's value (ie. it's module code), not the last input value 
- Cursor position is reset if you try to edit the middle of the search query

ZhangYiJiang:
Oh, and ignore the CircleCI failure - that's because of a config error on our part I think

nexolute:
Here's what I changed in the latest commit 4e95fac:

- Turns out the cursor issue is a [known issue](https://github.com/paypal/downshift/issues/217). I had fixed it accordingly.
- According to downshift documentation, props of rendered element should be passed as arguments into the [prop getter](https://github.com/paypal/downshift#prop-getters). I had changed it accordingly.
- I refactored the class.

www/src/js/views/timetable/ModulesSelect.jsx
```diff
@@ -22,37 +22,63 @@ type Props = {
 };
 
 type State = {
-  isModalOpen: boolean,
   isOpen: boolean,
+  isModalOpen: boolean,
+  inputValue: string,
+  selectedItem: any,
```
ZhangYiJiang: This should probably match the type of `item` property that was passed into `getItemProp`, which is `ModuleCode`. Since it is nullable, this should be `?ModuleCode`

yangshun:
@nexolute Good stuff, thanks for the PR!

ZhangYiJiang:
@nexolute Thanks for this PR, and sorry it took so long!

PR #657, opened 2018-01-03
Small UI fixes and improvements 

### LINKED ISSUE #570: Improve Bootstrap component styling in night mode
The following UI components can be better styled in night mode: 

- [x] Alerts (background, border, font colors) 
- [ ] Form elements (border color) 
- [x] Modal and elevated elements (darken shadow, border) 
- [ ] Dropdown menu (menu item hover/selected state, elevation) 

A new set of CSS variables will need to be defined for each of these colors, then the original styles should be overriden. Alternatively, ~~~explore using Bootstrap's internal mixins to do this so we don't get duplicate CSS declarations~~~ not possible - the alert mixins uses `mix` internally, which means we can't use CSS variables as parameters.

ZhangYiJiang:
Fixes #570 and a bunch of other stuff - the big one being 

### Upgrade to Bootstrap 4 beta 3 

Main changes here: https://github.com/twbs/bootstrap/issues/24439. The big changes for us are 

- Updating checkbox markup structure 
- Update input-group class names and structure 
- Remove `cursor: pointer` on buttons 

Also need to add a workaround for https://github.com/twbs/bootstrap/issues/25158. This seems to be the last Bootstrap beta, because the next ship list I can see is for 4.0 release 🎉 

### Night mode 

- Switch more Bootstrap variables to CSS variables for better compatibility 
- Use `dropdown-selected` in other dropdown menus - this overrides Bootstrap's default colors and styles 
- Add some shadow colors to night mode so they can be made darker in night mode 

### Tweaks 

- Adjust gray-lightest in night mode to make it closer to body-bg 
- Hide networked export options in ExportMenu when offline 
- Add self-contained Counter utility class for Downshift item highlighting 
- Add `@night-mode` and `@vertical-mode` mixins that allows for nesting, which makes the code more readable 
- Add extra padding and line between items when targetting touchscreens for `.dropdown-item` 

### Bug fixes 

- Fix React key warning on Settings page for bookmark module list 
- Fix content shifting upwards when no-scroll is active on mobile 
- Remove bg color on nav tabs on md and above


taneliang:
Bad commit message there. I actually changed the 404 page.
image

www/src/styles/components/buttons.scss
```diff
@@ -1,7 +1,3 @@
-.btn {
```
yangshun: Do they do this by default now?

www/src/styles/bootstrap/style-override.scss
```diff
@@ -47,6 +58,16 @@ mark {
     color: #fff;
     background: theme-color();
   }
+
+  @include touchscreen-only {
+    padding-top: 0.5rem;
```
yangshun: Extract into variable or use shorthand

www/src/styles/components/buttons.scss
ZhangYiJiang: Yup - see the v4.0.0b3 ship list

www/src/styles/bootstrap/style-override.scss
ZhangYiJiang: Hmmm, not using shorthand because don't want to override/conflict with any downstream users who may want less horizontal padding. Can extract out to variable.

Issue #606, opened 2017-12-27
Implement client-side data validation


ZhangYiJiang:
Timetable configuration data on the client-side may become invalid for a number of reason

- Lessons get dropped (when there's not enough students in a class) 
- Modules disappear 
- Lesson timetable get published or updated 

etc. etc. Whenever the client receives some data from the API, it should validate its existing timetable configuration data and implement some sanity checks: 

- When `moduleList.json` is received, check that all modules in the timetable still exists 
- When `modules/<module code>.json` is received, check that 
  - All lessons are filled 
  - All class no. are valid 

If these invariants are violated, the app should inform the user in some way (which is why the issue has the user interface label).

PR #661, opened 2018-01-04
Fill timetable blanks after refreshing module data from server

ZhangYiJiang:
This is a temporary fix for the problems described in #606 

It only fixes it for the case where the incoming module data describes more lessons than the timetable already has, and implements no UI for it (in fact, because of the way it is implemented, affected students may actually see the new lessons suddenly pop into existence). 

However, it is still better than the alternative, where the missing lessons never gets filled until the module is removed and added back again.


ZhangYiJiang:
Manually confirmed this is working by manually editing the Redux state and observing the page fill in the missing lessons 

screenshot from 2018-01-04 16-36-08

Issue #264, opened 2016-06-26
Distinguish "No exam timetable data" from "No Exam"


xinan:
At this period of time many modules on CORS are updated for the next semester but not the exam timetable. (I think this happens every year.) Instead of showing "No exam" for all of them, maybe we should be able to distinguish the situation of "No exam" and "No exam timetable info can be retrieved".

yangshun:
Workaround: 97c6f15

ahbeng:
So the issue is that the exam isn't necessarily populated in CORS but may be populated in exam timetable later?

I'm not sure it'd be better to display this as an inline message vs. a temporary global message when the exam timetable hasn't been published. Since it'd just be the same for every module that doesn't have an exam time listed in CORS yet.

But maybe a tooltip note for "No Exam" would be less intrusive. Would need to update NUSMods API to distinguish between no exam in CORS and no exam in exam timetable.

xinan:
Hmm, it is not that no exams in CORS but there is in exam timetable. It is that the module informations are updated first, and the exam information are only updated when the CORS schedules are announced. So during this gap, all modules shows "No exam".

yangshun:
I think it'll be fine to keep it undefined rather than giving the user potentially false information. Perhaps render a dash? 😂

taneliang:
@yangshun Uhoh #571 changes this back to "No Exam". Should it be changed back to "Exam: -"?

taneliang:
NUSIT informed us that their system only knows if there is an exam if someone sets the exam date on the system, and thus it won't be possible to know if the mod actually doesn't have an exam or if the exam date is just unknown. They also mentioned that exam dates can be set and changed at any time, even during the semester, and so unfortunately the assumption that all the exam dates are set at once is not true.

PR #662, opened 2018-01-04
Add "No Exam" message to module info page

### LINKED ISSUE #660: It is not obvious from the module info page that a module has no exam

Feature

**What is the current behavior?**

On the new module info page of a single module, if the module has an exam, the exam date and time will be shown on top of the "Add to Semester X" button. However, if there is no exam for that module, no clue is given. I spent quite a while looking for exam info and in the end figured out that that is the behavior of no exam.

Open the module info page of a module that has no exam. E.g. [MUA2201](https://nusmods.com/modules/MUA2201/keyboard-literature-i) and [CS5218](https://nusmods.com/modules/CS5218/principles-of-program-analysis)

It should be obvious to the user that the module has no exam.

**Further information**

- Which version of NUSMods are you currently using? 20180104-cdde0ba
- Did this work in NUSMods BEFORE the R release/redesign? Yes
- Did this work in NUSMods AFTER the R release/redesign? No
- Which browser and version are you using? Mobile Safari on iOS 11.2.1
- Which device are you using? E.g. 1st gen iPad Pro 12.9
- Which operating system and version are you on? iOS 11.2.1

taneliang:
Resolves #660.

Display "No Exam" in the module page when a module has no exams.

localhost_8080_modules_cs5218_principles-of-program-analysis 1

### DIFF
--- a/www/src/js/views/modules/ModulePageContent.jsx
+++ b/www/src/js/views/modules/ModulePageContent.jsx
@@ -70,11 +70,7 @@ export class ModulePageContentComponent extends Component<Props, State> {
     if (exams.length === 0) {
       return (
         <div key="no-exam" className={styles.exam}>
-          <h3 className={styles.descriptionHeading}>No Exam Info</h3>
-          <p className={styles.noExamWarning}>
-            <strong>Caution</strong> Either {ModuleCode} has no exams or {ModuleCode}&#8217;s exam
-            data has not been released.
-          </p>
+          <h3 className={styles.descriptionHeading}>No Exam</h3>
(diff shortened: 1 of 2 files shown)


ZhangYiJiang:
No, I don't think this is correct. If I understand #264 correctly, the problem is that the exam data may become available only later, but we usually know when that 'later' is since the exam data becomes available for all modules all at once.

taneliang:
Closing as a deeper fix was implemented in dd279c93b28db77e3b5ba16c0c12dc135efca3e3

Issue #658, opened 2018-01-03
Searching for module from search bar and navigating to it results in 404


yangshun:
**Do you want to request a *feature* or report a *bug*?**

Bug

**If the current behavior is a bug, please provide the steps to reproduce. If possible, please upload a screenshot or link to a video recording (on macOS you can use QuickTime) demonstrating the problem.**

1. Searching for module from search bar. Say, "Linear Algebra"
2. Click on "MA2101 Linear Algebra II" 
3. The route changes but the page content shows a 404

Unfortunately, I can't reproduce it reliably and it only happened like twice across 4-5 days. Tried with incognito mode and different browsers as well but unable to reproduce.

**What is the expected behavior?**

Should navigate to the module page without error.

**Further information**

- Which version of NUSMods are you currently using? 20180103-5cd07f3
- Did this work in NUSMods BEFORE the R release/redesign? Yes
- Did this work in NUSMods AFTER the R release/redesign? Unsure
- Which browser and version are you using? E.g. Chrome 63
- Which device are you using? E.g. MacBook Pro (15-inch, 2017)
- Which operating system and version are you on? E.g. macOS 10.12.6

ZhangYiJiang:
Hmmm, I haven't experienced it myself. Are you sure it's a 404 page and not an error page (ie. because the chunk / JSON loading failed)?

yangshun:
I can't really remember. It happened too fast and both times I quickly tried to search for another module and the next attempt worked. Doesn't help that the 404 page and the error page looks a little similar 😂 

It's probably the error page. It makes more sense to be an error page rather than a 404.

mhq199657:
I also encountered it several times.

taneliang:
@mhq199657 Do you know a way to reproduce this? Also a screenshot will be helpful so that we know whether we're getting a 404 or error page 😂

mhq199657:
Sorry, just like what Yang Shun has said, it happened quite sporadically. But I recall there is one occasion that it showed an "Uh oh..." page so I think Sentry should have recorded the error.

ZhangYiJiang:
Uh oh... most likely means the chunk loading has failed. I can make it give the `import()` at least one or two retries, just to make sure it's not a network hiccup

luyangkenneth:
This happened to me as well. I used the top search bar for the first time, clicked on `view all`, and then clicked on the first result. I then saw the "uh oh / refreshing this page might help" message.

yangshun:
[...]

PR #668, opened 2018-01-05
Retry ModulePageContent import on online chunk loading failure

taneliang:
Attempted fix for #658 by implementing point 2 in https://github.com/nusmodifications/nusmods/issues/658#issuecomment-355544612: Retry the `import()` promise instead of just running it once.

The promise is now called three times if we're online and encountering the chunk loading error.

### DIFF
--- a/www/src/js/utils/promise.test.js
+++ b/www/src/js/utils/promise.test.js
@@ -5,7 +5,7 @@ describe('#retry()', () => {
   test('should retry failing fn up till limit if shouldRetry returns true', async () => {
     const mockFailingFn = jest.fn().mockReturnValue(Promise.reject('error'));
     await expect(retry(3, mockFailingFn, () => true)).rejects.toThrow('error');
-    expect(mockFailingFn.mock.calls).toHaveLength(4);
+    expect(mockFailingFn).toHaveBeenCalledTimes(4);
   });
 
   test('should stop retrying if shouldRetry returns false', async () => {
@@ -18,7 +18,7 @@ describe('#retry()', () => {
       .mockReturnValue(false);
 
     await expect(retry(3, mockFailingFn, shouldRetry)).rejects.toThrow('error');
-    expect(mockFailingFn.mock.calls).toHaveLength(2);
+    expect(mockFailingFn).toHaveBeenCalledTimes(2);
   });
 
   test('should stop retrying if promise resolves', async () => {
@@ -29,12 +29,12 @@ describe('#retry()', () => {
       .mockReturnValue(Promise.resolve());
 
     await expect(retry(3, mockFailingFn, () => true)).resolves;
-    expect(mockFailingFn.mock.calls).toHaveLength(2);
+    expect(mockFailingFn).toHaveBeenCalledTimes(2);
   });
 
   test('should not retry successful promises', async () => {
     const mockSucceedingFn = jest.fn().mockReturnValue(Promise.resolve());
     await expect(retry(3, mockSucceedingFn, () => true)).resolves;
-    expect(mockSucceedingFn.mock.calls).toHaveLength(1);
+    expect(mockSucceedingFn).toHaveBeenCalledTimes(1);
   });
 });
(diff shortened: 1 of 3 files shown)


www/src/js/views/modules/ModulePageContainer.jsx
```diff
@@ -30,6 +30,21 @@ type State = {
   error?: any,
 };
 
+// Recursively retry fn until success, maxRetries has been reached, or shouldRetry returns false.
+// Based on https://stackoverflow.com/a/30471209/5281021
+function retry(
```
ZhangYiJiang: Can extract this out to utils since we may want to use this elsewhere too

www/src/js/views/modules/ModulePageContainer.jsx
```diff
@@ -30,6 +30,21 @@ type State = {
...
+  maxRetries: number,
```
ZhangYiJiang: /nit should just be `retries` I think? Not sure why `max` is needed

www/src/js/utils/promise.test.js
```diff
@@ -0,0 +1,50 @@
...
+      .mockReturnValueOnce(true)
+      .mockReturnValue(false);
+
+    return retry(3, mockFailingFn, shouldRetry).catch((error) => {
+      expect(error).toBe('error');
+      expect(mockFailingFn.mock.calls).toHaveLength(2);
+    });
+  });
+
+  test('should not retry if promise returns true', () => {
```
ZhangYiJiang: I think this test description is wrong

www/src/js/utils/promise.test.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import { retry } from './promise';
+
+describe('#retry()', () => {
+  test('should retry up till limit if shouldRetry returns true', () => {
+    expect.assertions(2);
```
ZhangYiJiang: Rather than doing this, why not try using `async` / `await` with `expect().toThrowErrors()` (not sure if that works with async functions), or alternatively, `.rejects` (https://facebook.github.io/jest/docs/en/expect.html#rejects)

www/src/js/utils/promise.test.js
taneliang: #sleepycoding

www/src/js/utils/promise.test.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import { retry } from './promise';
+
+describe('#retry()', () => {
+  test('should retry failing fn up till limit if shouldRetry returns true', async () => {
+    const mockFailingFn = jest.fn().mockReturnValue(Promise.reject('error'));
+    await expect(retry(3, mockFailingFn, () => true)).rejects.toThrow('error');
+    expect(mockFailingFn.mock.calls).toHaveLength(4);
```
ZhangYiJiang: /nit Can use `.toBeCalledTimes(2)` - sorry I didn't catch this just now

www/src/js/views/modules/ModulePageContainer.jsx
```diff
@@ -53,7 +54,13 @@ export class ModulePageContainerComponent extends PureComponent<Props, State> {
   componentDidMount() {
     this.fetchModule(this.props.moduleCode);
 
-    import('views/modules/ModulePageContent')
+    // Try importing ModulePageContent thrice if we're online and
+    // getting the "Loading chunk x failed." error.
+    retry(
+      3,
+      () => import('views/modules/ModulePageContent'),
+      (error) => error.message.indexOf('Loading chunk ') === 0 && window.navigator.onLine,
```
ZhangYiJiang: /nit Can use `String#includes`

www/src/js/utils/promise.test.js
taneliang: Ah yes I was sure this function existed but couldn't find the docs for it

www/src/js/views/modules/ModulePageContainer.jsx
taneliang: I wanted to make sure that "Loading chunk " was at the front of the error message, just in case there's some other error that had that exact string somewhere in it. But that's an edge case I guess.

Issue #473, opened 2017-11-02
Refactor hidden modules


yangshun:
Hidden modules is currently stored in `settings` object as inside the `hiddenInTimetable` array. That's not the best representation as the hidden/show state of a module should be put inside the semester's timetable instead and be specific to a semester instead of an app-wide setting. Also, there's a bug where hidden modules that get removed are still within the `hiddenInTimetable` array.

li-kai:
I have a suggestion, we could simply use a class name similar to our color scheme like `color-nil` or something. That way, invisible can be a color in our picker, and make changes accordingly instead of separating the picker component in two parts (the square and picker).

yangshun:
So you're saying we have a 9th color? Do we want to hide the details or simply make the cell transparent (but still show the words)?

li-kai:
Yep, something like the top left icon 
<img width="225" alt="screenshot 2017-11-02 14 20 09" src="https://user-images.githubusercontent.com/14850387/32312477-fb99a48a-bfd8-11e7-85fb-2e0d7c12b29a.png">

We still hide the details, but that "special handling" wouldn't be in the module details component any more, but in TimetableContainer instead.

li-kai:
Oh I forgot one more thing. Right now if they try to change a hidden modules' color, they wouldn't know why it's not changing as they're still able to open the picker. A transparent color helps us save space too, one fewer icon, less code.

PR #669, opened 2018-01-06
Data layer improvements - Redux Persist, nesting colors to under timetables and improved typing 

### LINKED ISSUE #488: Refactor persistence mechanism with redux-persist
https://github.com/rt2zz/redux-persist

As mentioned by @ZhangYiJiang in the telegram chat. I'm not super familiar with redux at this point, so I'll just put this here for posterity.

### LINKED ISSUE #511: Refactor theme store to account for semester-specific module colors
There's a bug where:

1. Add module X to sem 1
2. Add module X to sem 2
3. Remove module X from sem 1
4. Module X in sem 2 loses its color

This is because the `theme` reducer maps a module to a color. There’s no notion of semesters in it. I refactor should take place such that a color is tied to each module within a semester instead. Two ways to do it:

1. Add a color property into `Module` object.
2. Differentiate by semesters within the `theme` reducer.

I'm leaning towards the first approach because eventually we need to differentiate by acad years also and doing the first approach is more future proof when that change happens.

ZhangYiJiang:
This is unfortunately rather bigger than I wanted this PR to be. 

## Redux Persist 

Fixes #488 

Switch to Redux Persist for persistence. The advantage of this library is 

- Migration support - can change shape of data in a versioned manner 
- Multiple backends - in case we want to switch to localforage in the future 
- Split persistence configuration - can separately configure different branches of the state tree

To use persist, export a `persistConfig` object from the reducer, then in `reducers/index.js`, wrap the reducer with the `persistReducer(key: string, reducer: Reducer, options?: Object)` function 

``` js
// Before 
import settings from './settings';

// After 
import settingsReducer, { persistConfig as settingsPersistConfig } from './settings'; 

const settings = persistReducer('settings',  settingsReducer, settingsPersistConfig); 
```

This is a little more verbose than I would have liked. Unfortunately this seems to be the cleanest way, because importing the storage object from Redux Persist has [unwanted side effects](https://github.com/rt2zz/redux-persist/blob/master/src/storage/index.js). We can't just do the wrapping in each of the reducers files and export both the original reducer (for unit test) and the wrapped reducer as default, like we do with HOC wrapped components. 

I'll be writing a script to help test this manually. 

## Migrate colors and hidden modules to under timetables 

Fixes #511, #473 
 
This is necessary because Redux Persist adds keys to the root of each persisted branch, which means the old shape of `timetables` would no longer work. I would have rather made this change in a separate PR, but since we have to do this anyway let's just rip out the plaster and get on with it. 

This changes the state from 

```
{
  timetables: { [Semester]: SemTimetableConfig } 
}
```

to 

```
{
  timetables: { 
    timetableConfig: { [Semester]: SemTimetableConfig },
    colors: { [Semester]: ColorMapping },
    hidden: { [Semester]: ModuleCode[] },
    academicYear: string
  }
}
```

This change also removes the need for `theme-middleware`, so that has been removed. 

## Other maintenance 

- Code that only runs once when the app is first started have been moved to under `src/js/bootstrapping` (not `bootstrap`, to avoid confusion with the CSS framework). This makes more sense than leaving them in places like `utils`. 

- A few untyped files have been typed, and some types have been tightened. This meant that `:any` has to be added in a few places (mostly tests). I'll be making more extensive changes in another PR, since this PR is already too big 

- All `localStorage` keys have been moved to under `storage/keys.js` - this makes it easy to see what keys have already been used, to prevent collisions

Please help to test this thoroughly - if all goes well, the users should never know these changes have happened at all.


ZhangYiJiang:
Test script: https://gist.github.com/ZhangYiJiang/ee0b901fdf6e2190b09a5b06aa13d8eb 

```
var c = `
localStorage.clear();
localStorage.setItem('reduxState', JSON.stringify(${localStorage.getItem('reduxState')}));
window.location.reload();`;

if (typeof copy === 'function') {
  console.info('Go to https://deploy-preview-669--nusmods.netlify.com/ and paste the code that has been copied into the console');
  copy(c);
} else {
  console.info('Go to https://deploy-preview-669--nusmods.netlify.com/ and paste this into the console');
  console.log(c);
}
```

ZhangYiJiang:
@li-kai I'm not sure how much benefit there is in terms of simplifying timetable code. Change detection (a la #606) can be done when the data is fetched on initial page load. The main downside I see is that this can make `lessonConfig` really big for a module like GER1000.

Issue #538, opened 2017-12-20
Venue page enhancements 


ZhangYiJiang:
Tracking some enhancements for the venues page that would be annoying to keep track of if split into separate issues 

- [ ] Highlight period of time searched for when availability search is on 
- [ ] Merge dual coded modules on the timetable 
- [ ] Adopt more of v2's by grouping venues according to their first character/prefix strings 
- [ ] Add some way to close the timetable

PR #670, opened 2018-01-07
Switch venues page to use two panel layout 

ZhangYiJiang:
Part of #538 

On md and above, a split left pane with the venues list and a right pane with the selected venue's details 

screenshot from 2018-01-07 13-12-57

On xl, we have enough space for the timetable to be displayed horizontally 

screenshot from 2018-01-07 13-12-41

On sm and below, the venue list takes up the entire screen, and the venue detail appears in a modal 

screenshot from 2018-01-07 13-13-09

The venue search box is no longer sticky, because that would take up too much space on the venue list. 

The venues list now displays items inline and groups them by their first letter, like in v2. This layout change simplifies the structure of the venue related views, although VenueContainer now has more logic in it because we need to lift some of the logic from subviews up. 

### Non-UI changes 

- The URL handling code should behave the same as before, although it is simplified to use `componentDidUpdate`, and the search term is no longer pre-filled if a venue is selected. In addition, the entire system now uses the URL as the single source of truth, which simplifies matters somewhat. 

- Internally instead of `VenueInfo` (`{ [Venue]: DayAvailability[] }`, we now use `VenueDetailList` (`[Venue, DayAvailability[]][]`) - this is basically `entries(venueInfo)`, which is easier to work with since we can now easily sort and filter without resorting to sorting the keys then picking. 

- `config.semester` is used to load venue data instead of `state.app.activeSemester` - activeSemester is used for the timetable only, and showing a previous semester's venue data is unlikely to be useful 

- Using `<Helmet>` we can now change the page title based on the selected venue. 

### Todo 

- [X] The venue switcher in the top of venue details needs to be made to work 

### Pushed to backlog 

- [ ] The venue list should scroll to the selected venue item when the page is loaded 
- [ ] A 'back to top' button need to be added to the venue list when it is scrolled down


yangshun:
Nice work! The browser autocomplete dropdown keeps flashing (appears and disappears) for me. Maybe we can do `autocomplete="off"`? If necessary I can take a screen capture to demonstrate the problem. Not sure if others are seeing it too.

ZhangYiJiang:
That is really odd! Did that happen previously? I don't think I touched the search input code at all for this PR.

yangshun:
Oh yes that's happening in prod now as well. I realized the autocomplete list that I see on Venues page is the same as on https://nusmods.com/modules. Are you seeing the flashing as well?

ZhangYiJiang:
I can't - in fact I can't trigger autocomplete at all on the form. You should open a new issue with all of the details so we can triage.

taneliang:
You broke your footer hack 😂

screen shot 2018-01-13 at 10 45 06 am

ZhangYiJiang:
Yeah, I know. There's really no winning here 😭

PR #671, opened 2018-01-07
Fix scroll to hash on settings and module info page 

ZhangYiJiang:
Adds a small util function to handle scrolling to the correct section of the page when there's a hash fragment in the URL. 

This is broken on `<ModulePageContent>` because the content is loaded asynchronously, so the element with the ID is not present at initial page load. We fix this by doing an additional check during `componentDidMount`. 

It is also broken on `<SettngsContainer>` - not entirely sure why, but adding the same function to `<ScrollToTop>` works.

### DIFF
--- a/www/src/js/utils/react.jsx
+++ b/www/src/js/utils/react.jsx
@@ -70,6 +70,14 @@ export function wrapComponentName(Component: ComponentType, wrapper: string): st
   return `${wrapper}(${Component.displayName || Component.name || 'Component'})`;
 }
 
+/**
+ * Small utility function to scroll to an element with ID matching the URL hash if
+ * both are present.
+ *
+ * This mimics traditional webpage behavior and should be used in componentDidMount() when
+ * the component is not loaded on initial page load (ie. the element is not in the DOM when
+ * the page is initially loaded), but has content that can be linked to via hashes.
+ */
 export function scrollToHash() {
   const hash = window.location.hash;
   if (hash) {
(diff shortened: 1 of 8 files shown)


www/src/js/utils/react.jsx
```diff
@@ -70,6 +70,16 @@ export function wrapComponentName(Component: ComponentType, wrapper: string): st
   return `${wrapper}(${Component.displayName || Component.name || 'Component'})`;
 }
 
+export function scrollToHash() {
```
yangshun: This util function is not specific to React, so probably more appropriate to put it elsewhere?

www/src/js/views/components/ScrollToTop.jsx
```diff
@@ -28,6 +28,10 @@ export class ScrollToTopComponent extends Component<Props> {
     }
   }
 
+  componentDidMount() {
+    scrollToHash();
```
yangshun: I think this `scrollToHash` behavior should be opt-in, like with `onComponentWillMount` and `onPathChange`.
 
Not sure if `ScrollToTop` is still a semantic name for this component if it has different behaviors now.

ZhangYiJiang: Hmmm, I've checked, and it seems the bug in SettingsContainer is caused by ScrollToTop, ie. if you remove ScrollToTop then link to hash works again. I've updated the component to check for hash in `location` so that it doesn't scroll to top if that is present

www/src/js/utils/react.jsx
ZhangYiJiang: I think it's reasonable. I expect this to be used only in `componentDidMount` lifecycle methods anyway, given how specific it is.

PR #676, opened 2018-01-09
Fix CSS variable fallback

ZhangYiJiang:
So this is an interesting bug. 

The problem is that CSS Modules and PostCSS Custom Properties don't play well with each other -  Custom Properties needs the global `:root` style to determine the fallback style, but because we modularized our CSS, the `:root` declaration isn't in the same post-SASS-processed CSS file, which means the PostCSS plugin won't work. The solution is to include `css-variables.scss` in `modules-entry.scss`, but this leads to a ton of duplicate `:root` declarations. 

That in itself is not too big of a deal, since gzip works really well with duplicate blocks of text, so it can compress them relatively well. However, this makes development quite a bit more annoying, because now you have a ton of duplicate `:root` rules polluting every element's style (`:root` is inherited by everything). So the solution is... more PostCSS! Specifically, a custom PostCSS plugin that strips out the extra rules from the modular CSS. It's not very elegant (read the code for all the gory details), but it works out pretty well. 

So yeah. That's that. Also, we've learnt that 

- Disabling warnings may not be the best idea (if we had warnings turned on for Custom Properties it would have caught this. On the other hand the warnings are too noisy, and can't be disabled on a per type basis. I'll be making a PR to fix this)
- cache-loader means updating PostCSS config requires deleting the cache 
- CSS modules isn't free, sigh, and interop between different systems is hard 
- We really should have used done proper browser testing, and not just declared that something was working just by looking at the code (I discovered this bug when playing around with Browserstack)


www/scripts/postcss-single-root.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * This leads to a huge number of ':root' styles being emitted, which
+ * isn't a huge problem in production since gzip can compress duplicate
+ * text blocks for free, but does make development harder, so we strip
+ * the extra ':root' rules using a PostCSS plugin.
+ */
+
+const postcss = require('postcss'); // eslint-disable-line import/no-extraneous-dependencies
+
+const rootSelectors = [':root', 'body.mode-dark'];
```
yangshun: Not a blocker, but can the selectors a configurable option that is passed in? Then perhaps you can publish this postcss plugin 😄

www/src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
+/* single-root-entry */
+// DO NOT REMOVE THE COMMENT ABOVE - used by scripts/postcss-single-root.js
```
yangshun: There's no need for this comment if we write it like a pragma:

```
// @postcss-single-root-entry
```

Feel free to change the wording. I think this makes the intention obvious.

www/scripts/postcss-single-root.js
```diff
@@ -0,0 +1,40 @@
...
+
+module.exports = postcss.plugin('postcss-single-root', () => (css) => {
+  let isMainCSS = false;
+
+  // eslint-disable-next-line consistent-return
```
yangshun: Might as well remove this `eslint-disable` and add `return true;` in the function.

ZhangYiJiang:
Confirm working in Edge 14 on Browserstack 

screenshot from 2018-01-09 10-51-34

PR #680, opened 2018-01-10
Delay UrlChangeTracker.handleUrlChange invocation

yangshun:
Added a timeout to `UrlChangeTracker.handleUrlChange` because React 16 (and consequently React Helmet's) render update is not synchronous. 

In fact, `UrlChangeTracker.handleUrlChange()` is already made async via `setTimeout(..., 0)` but it's still insufficient. A timeout of 100ms delays the GA invocation sufficiently long enough for React to do its stuff. 

Why 100ms? It's an arbitary value actually. It's short enough for us not to miss any page change event, but long enough for React to complete its rendering. This works well enough for most cases.

**Before**

screen shot 2018-01-09 at 9 15 01 pm

**After**

screen shot 2018-01-09 at 9 11 36 pm

### DIFF
--- a/www/src/js/bootstrapping/google-analytics.js
+++ b/www/src/js/bootstrapping/google-analytics.js
@@ -15,6 +15,7 @@ export default function initializeGA() {
       // delays the GA invocation sufficiently long enough for React to do its stuff.
       // 100ms is an arbitary value which works well enough for most cases.
       const HANDLE_URL_CHANGE_TIMEOUT = 100;
+      // eslint-disable-next-line
       window.gaplugins.UrlChangeTracker.prototype.handleUrlChange = function(historyDidUpdate) {
         setTimeout(() => {
           origHandleUrlChange.call(this, historyDidUpdate);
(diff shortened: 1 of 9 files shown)


www/src/js/utils/google-analytics.js
```diff
@@ -8,6 +8,20 @@ export default function initializeGA() {
       import(/* webpackChunkName: "autotrack" */ 'autotrack/lib/plugins/outbound-link-tracker'),
       import(/* webpackChunkName: "autotrack" */ 'autotrack/lib/plugins/url-change-tracker'),
     ]).then(() => {
+      const origHandleUrlChange = window.gaplugins.UrlChangeTracker.prototype.handleUrlChange;
+      // This timeout is needed because React (and consequently React Helmet's) render
+      // update is not synchronous. In fact, UrlChangeTracker.handleUrlChange is already
```
ZhangYiJiang: Is this still necessary with `defer={false}`?

yangshun:
Yes unfortunately. Delay false simply make it not use
`requestIdleCallback`. If we remove delay false, we will have to increase
the timeout even more.

PR #682, opened 2018-01-10
Improve data layer engineering

ZhangYiJiang:
A lot of very minor changes that were holdover from the previous PR 

- Fix the Raven middleware `stateTransformer` option 
- Fix hidden modules in timetable export 
- Tightened Redux related typedefs, including making the state read-only, typing thunk's `getState` parameter, and the state parameter in `mapStateToProps` 
- Remove auth tokens from request middleware, since it is not used by, well, anything 
- Add Flow typing to a bunch of places 
- Update README

### DIFF
--- a/www/README.md
+++ b/www/README.md
@@ -163,7 +163,7 @@ $ yarn promote-staging # Promote ./dist to production
 │   │       ├── timetable    - Timetable builder related components
 │   │       └── venues       - Venues page related components
 │   └── styles
-│       ├── bootstrap        - Bootstraping, uh, Bootstrap
+│       ├── bootstrap        - Bootstrapping, uh, Bootstrap
 │       ├── material         - Material components 
 │       ├── components       - Legacy component styles
 │       │                      (new components should colocate their styles)
--- a/www/README.md
+++ b/www/README.md
@@ -133,43 +133,46 @@ $ yarn promote-staging # Promote ./dist to production
 ## Project Structure
 
 ```
-├── scripts                 - Command line scripts to help with development
+├── scripts                  - Command line scripts to help with development
 ├── src
 │   ├── img
 │   ├── js
-│   │   ├── actions         - Redux actions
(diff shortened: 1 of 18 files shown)


www/src/js/views/layout/Footer.test.jsx
```diff
@@ -1,9 +1,11 @@
+// @flow
+
 import React from 'react';
 import { shallow } from 'enzyme';
 
 import { FooterComponent } from 'views/layout/Footer';
 
 test('is a footer element', () => {
-  const actual = shallow(<FooterComponent />);
+  const actual = shallow(<FooterComponent toggleFeedback={jest.fn} lastUpdatedDate={null} />);
```
yangshun: Why are these props needed? Can they be defaultProps within the component instead?

www/README.md
```diff
@@ -133,43 +133,46 @@ $ yarn promote-staging # Promote ./dist to production
...
-│       ├── bootstrap       - Bootstraping, uh, Bootstrap
-│       ├── components      - Legacy component styles
-│       │                     (new components should colocate their styles)
-│       ├── layout          - Site-wide layout styles
-│       ├── pages           - Page specific styles
-│       └── utils           - Utility classes, mixins, functions
-├── static                  - Static assets, eg. favicons
-|                             These will be copied directly into /dist
-└── webpack                 - Webpack config
+│       ├── bootstrap        - Bootstraping, uh, Bootstrap
```
yangshun: Bootstraping with double p?

www/src/js/views/layout/Footer.test.jsx
ZhangYiJiang: To make Flow happy, because these are required props. 

I don't think they should be defaultProps, because they don't have good default values. They should probably be tested, but it's not a hugely important component or one that gets changed frequently, so not going to bother too much in this PR.

PR #684, opened 2018-01-11
Stop deleting old dist files

ZhangYiJiang:
Related to #658 

This should hopefully prevent some of the `SyntaxError`s observed due to old dist files being missing.


yangshun:
@taneliang Have a look?

PR #685, opened 2018-01-11
Move PersistGate back under Router

ZhangYiJiang:
Reverts https://github.com/nusmodifications/nusmods/pull/669/commits/2b4264405da6c7f1087b1e3bdc87f987a67cbe2b 

This was preventing route changes from reaching AppShell for some reason.

### DIFF
--- a/www/src/js/App.jsx
+++ b/www/src/js/App.jsx
@@ -22,13 +22,13 @@ export default function App({ store, persistor }: Props) {
   return (
     <AppContainer>
       <Provider store={store}>
-        <Router>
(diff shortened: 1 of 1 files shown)


ZhangYiJiang:
Yeah, not sure why either. I've looked at the PersistGate source (https://github.com/rt2zz/redux-persist/blob/master/src/integration/react.js) and there's nothing that indicates this should be happening, but you can see the bug happening clearly in eg. https://deploy-preview-684--nusmods.netlify.com/

PR #701, opened 2018-01-12
Shim localStorage

### LINKED ISSUE #480: ReferenceError: localStorage is not defined

taneliang:
Fixes #480 and hopefully all the localStorage issues once and for all. Adapted from [the best localStorage polyfill in the world](https://gist.github.com/juliocesar/926500).

Tested on iOS 10 private browsing and Chrome on macOS with cookies disabled.

As this PR only resolves the cases where localStorage doesn't exist, the try/catch blocks in the exported functions are left there to catch all other cases - such as if the localStorage is actually full.

### DIFF
--- a/www/src/js/storage/index.js
+++ b/www/src/js/storage/index.js
@@ -21,10 +21,12 @@ function getLocalStorage() {
     // Resolves https://sentry.io/share/issue/d65da46a7e19406aaee298fb89a635d6/
     if (!storage) throw new Error();
 
-    // Ensure that localStorage can be written to
-    // Next line throws on iOS 10 private browsing
-    storage.setItem('____writetest', 1);
-    storage.removeItem('____writetest');
+    // Ensure that if setItem throws, it's not because of private browsing
+    // If storage is empty AND setItem throws, we're probably in iOS <=10 private browsing
+    if (storage.length === 0) {
+      storage.setItem('____writetest', 1);
+      storage.removeItem('____writetest');
+    }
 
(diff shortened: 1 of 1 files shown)


ZhangYiJiang:
The shim should also be able to handle this case https://sentry.io/nusmods/v3/issues/430208537/

www/src/js/storage/index.js
```diff
@@ -4,12 +4,56 @@ import Raven from 'raven-js';
...
+  } catch (e) {
+    // Shim if we can't use localStorage
+    // Once set, don't override
+    if (!usableLocalStorage) {
+      usableLocalStorage = {
+        privData: {},
+        clear: () => {
+          usableLocalStorage.privData = {};
+        },
+        setItem: (id, val) => {
```
yangshun: `localStorage` stringifies both keys and values. If you really wanna be a true shim, should apply stringification to the value.

Probably a better idea to call them `key`/`val` rather than `id`/`val`.

Issue #677, opened 2018-01-09
Display module title in timetable cell


yangshun:
We received many requests to display the module title in the timetable cell. Seems like we underestimated the number of people who have enough memory space to memorize their module titles.

Design a UI that allows toggling of display of module title within the timetable cells. This setting can possibly be on the timetable UI itself or in the Settings page. For simplicity we probably will not have the module title  show  on vertical mode because the layout will not work with that, and it is not necessary on other pages (for now) where the timetable component is being used.

ZhangYiJiang:
Fixed by #709

PR #709, opened 2018-01-12
Display module title in timetable cell

shanwpf:
Solves issue #677

Added a module title button beside the orientation button.
Had to make changes to the timetable actions button layout to accommodate the new button.
In vertical mode, module titles are not displayed and the button is disabled.

### Horizontal mode single row
image

### Horizontal mode 2 rows
image

### Vertical mode
image


www/src/js/reducers/timetables.js
```diff
@@ -125,10 +127,11 @@ export const defaultTimetableState: TimetablesState = {
   colors: {},
   hidden: {},
   academicYear: config.academicYear,
+  titleDisplay: HIDE,
```
taneliang: Any reason why this isn't a boolean? It seems like in the end it's assigned to a boolean prop in `Timetable`, and the toggling and checking code will be simpler.

ZhangYiJiang: Also, this is the wrong place for this data. Try putting it in `theme` or `settings` reducers instead

www/src/js/views/timetable/TimetableActions.jsx
```diff
@@ -33,6 +35,16 @@ function TimetableActions(props: Props) {
           <Sidebar className={styles.sidebarIcon} />
           {isVerticalOrientation ? 'Horizontal Mode' : 'Vertical Mode'}
         </button>
+
+        <button
+          type="button"
+          className="btn btn-outline-primary btn-svg"
+          onClick={props.toggleTitleDisplay}
+          disabled={isVerticalOrientation}
```
ZhangYiJiang: Better to switch to `outline-secondary` or hide the button entirely when its disabled. Bootstrap's disabled style for outline buttons is not very noticeable

www/src/js/views/timetable/TimetableCell.jsx
```diff
@@ -12,11 +12,15 @@ type Props = {
   lesson: Lesson,
   style: Object,
   isScrolledHorizontally: boolean,
+  showTitle: boolean,
   onModifyCell?: Function,
 };
 
 function TimetableCell(props: Props) {
   const lesson = props.lesson;
+  const moduleCodeWithTitle = `${lesson.ModuleCode} ${lesson.ModuleTitle}`;
```
ZhangYiJiang: Should just inline this - no need for another variable

www/src/js/views/timetable/TimetableCell.jsx
```diff
@@ -39,7 +43,7 @@ function TimetableCell(props: Props) {
       style={props.style}
     >
       <div className={styles.cellContainer}>
-        <div className={styles.moduleCode}>{lesson.ModuleCode}</div>
+        <div className={styles.moduleCode}>{moduleName}</div>
```
ZhangYiJiang: Also update this class name

www/src/js/views/timetable/TimetableActions.jsx
```diff
@@ -33,6 +35,16 @@ function TimetableActions(props: Props) {
...
+        >
+          <Sidebar className={styles.sidebarIcon} />
```
ZhangYiJiang: This icon isn't appropriate. We may be able to get away with abusing the icon for credit cards, or find another one in Font Awesome or Icons8

li-kai: How about the type icon?

www/src/js/reducers/timetables.js
shanwpf: @taneliang I was referring to the orientation toggling code when I wrote this, so I thought it had to have it's own type. I've simplified it to a boolean.

@ZhangYiJiang Moved the data from `timetables` to `theme`

www/src/js/views/timetable/TimetableCell.jsx
shanwpf: Updated

www/src/js/views/timetable/TimetableActions.jsx
shanwpf: I think the eye icon would fit in well. I've changed it to show the eye icon when titles are hidden, and the eye-off icon when titles are shown.

www/src/js/views/timetable/TimetableActions.jsx
shanwpf: Fixed

www/src/js/views/timetable/TimetableCell.jsx
shanwpf: I agree, I wrote this because the linter was going crazy, and I didn't know how to format the ternary operator properly to avoid one long line. Anyway, I figured it out and fixed it.

www/src/js/views/timetable/TimetableActions.jsx
taneliang: We're already using the eye icon to hide modules though, and our horizontal/vertical mode icon is too similar to the credit card icon IMO. I think the type icon is probably the best choice here

shanwpf: Ok, I've changed it to be the type icon.

www/src/styles/utils/themes.scss
```diff
@@ -9,11 +9,11 @@
     $index: 0;
 
     @each $color in $colors {
+
+      $index: $index + 1; /* stylelint-disable-line order/order */
```
taneliang: All the themes are missing their first color. I may be wrong but I think it's because this line was moved

c75cc818-0c87-4638-9cd1-1078388c89c8

shanwpf: I don't remember changing that file. But it seems to be caused by incrementing `$index` before applying the colour at index 0. Fixed it.

ZhangYiJiang:
It's a StyleLint bug. Lines that are marked as ignore are only ignored for linting, not autofix.

taneliang:
2 UI issues:

1. On portrait iPad, the first row buttons get squashed. The new button should have its own row. effc5af6-55bf-4888-ad69-654646f1d92f
1. Looks like the Type icon needs to be shrunk a little as it's making the buttons a bit taller.
781b0d02-1e49-4635-8fcf-a456eb59bce6

PR #715, opened 2018-01-14
Rethrow error in requests middleware

### LINKED ISSUE #704: TypeError: Cannot read property 'History' of undefined
https://sentry.io/nusmods/v3/issues/416311446/

```
TypeError: Cannot read property 'History' of undefined
  at getModuleSemesterData (./js/utils/modules.js:19:17)
  at Object (./js/utils/modules.js:24:16)
  at None (./js/actions/timetables.js:22:23)
```

Need to investigate - this should never happen.

ZhangYiJiang:
Fixes #704 -ish. 

The reason the error is happening is because it depends on `fetchModules` rejecting the promise if the request fails for some reason, but that's not the case right now. This should not break anything else I *think*, and it seems like a more logical design. (We probably will be discarding request middleware in favor of GraphQL soon-ish).

### DIFF
--- a/www/src/js/middlewares/requests-middleware.js
+++ b/www/src/js/middlewares/requests-middleware.js
@@ -61,7 +61,7 @@ const requestMiddleware: Middleware<State, *, *> = () => (next) => (action) => {
           },
         }),
       ),
-    (error) =>
+    (error) => {
       next(
         constructActionWith({
           type: type + FAILURE,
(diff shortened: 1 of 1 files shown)


ZhangYiJiang:
Okay, I manually tested this. The only way I could repro without changing our code is to open the dropdown first with some input, go offline, *then* try to select a module. With this change instead of crashing it just does nothing (there's an uncaught promise rejection logged). Since this whole is async this doesn't actually change anything I think, ie. it's not WSOD, since the original would have done the same, so it's not too urgent. 

Let me find a better fix for this.

ZhangYiJiang:
Superseded by #1092

PR #721, opened 2018-01-15
Redirect all invalid timetable URLs to current timetable

ZhangYiJiang:
Even with the `isV2TimetableUrl` function it seems we're still missing some visitors who have bookmarked the old v2 timetable URLs. I think simply redirecting all visits to `/timetable` which do not appear to be valid may work better.


ZhangYiJiang:
It means any incoming URL that starts with `/timetables` which does not fit `/timetables/:semester/:action?` would get redirected to `/timetables/{activeSemester}`. Not sure what you mean by disappear.

li-kai:
well if they contained any modules, they won't appear in v3

ZhangYiJiang:
Migration is done via `localforage`. The old query string is not relevant.

PR #722, opened 2018-01-15
Fix OpenSearch plugin 

ZhangYiJiang:
Not sure how we missed this, but the OpenSearch plugin was borked for a bit 

- The favicon is broken 
- The `ShortName` is too long (MDN says 16 characters max - works on Chrome, but seems to be broken on Firefox) 

Also switch to the new module search so this is a proper search engine now, yay


ZhangYiJiang:
That took waaay too much effort, grrr: http://www.feedvalidator.org/check.cgi?url=https%3A%2F%2Fdeploy-preview-722--nusmods.netlify.com%2Fopensearch.xml

PR #727, opened 2018-01-15
Add Chrome extensions and local files to Sentry URL blacklist

ZhangYiJiang:
As per title.


www/src/js/bootstrapping/sentry.js
```diff
@@ -17,6 +17,7 @@ if (loadRaven) {
       // Chrome extensions
       /extensions\//i,
```
li-kai: We could just remove the s here...

ZhangYiJiang: I'm actually a little worried that regex is a little too permissive

www/src/js/bootstrapping/sentry.js
```diff
@@ -14,9 +14,11 @@ if (loadRaven) {
       'canvas.contentDocument',
     ],
     ignoreUrls: [
-      // Chrome extensions
-      /extensions\//i,
+      // Local file system
+      /^file:\/\//i,
+      // Chrome and Firefox extensions
```
yangshun: How does this exclude Firefox extensions? They seem Chrome-specific to me.

ZhangYiJiang: `chrome://` I believe is used internally by both Firefox and Chrome. Also added `moz-extension://` to the list.

PR #728, opened 2018-01-15
Venues page bug fixes

ZhangYiJiang:
- Fix `defaultSearchOptions` defaulting to the next day - this is because I forgot JS weeks starts zero-indexed with Sunday at 0. Added unit tests to catch this, and reused timify's util functions 
- Use CSS grid for browsers that support it. It really works so much better than flexbox for this
- Switch to btn-outline-primary for the buttons in the list - much less hacky


www/src/js/views/venues/AvailabilitySearch.jsx
```diff
@@ -20,16 +20,15 @@ const FIRST_CLASS_HOUR = 8;
...
-export function defaultSearchOptions(): VenueSearchOptions {
-  const now = new Date();
-
+export function defaultSearchOptions(
+  now: Date = new Date(), // Used for tests only
+): VenueSearchOptions {
   // Set day of week - if it is not a school day, then set to Monday (0)
-  let day = now.getDay();
-  if (day >= SCHOOLDAYS.length) day = 0;
+  const day = Math.max(getCurrentDayIndex(now), 0); // Minus 1 because JS week starts on Sunday
```
taneliang: Should this comment be in timify.js line 77 instead?

PR #741, opened 2018-01-16
Fix acad week display

yangshun:


yangshun:
Closed in favor of #744

PR #744, opened 2018-01-16
Fixes acad week display

arsalanc-v2:
Fixes #572


www/src/js/utils/weekText.js
```diff
@@ -0,0 +1,36 @@
...
+    parts.push(noBreak(acadWeekInfo.sem));
+  }
+
+  // Do not show the week number if there is only one week, e.g. recess
+  // Hide week if week type is 'Instructional'
+  const type = acadWeekInfo.type === 'Instructional' ? '' : `${acadWeekInfo.type} `;
+  const weekNumber = acadWeekInfo.num || '';
+  parts.push(noBreak(`${type}Week ${weekNumber}`));
+
+  return parts;
```
taneliang: I think it'd be better if you returned `parts.join(', ').trim()` instead. There's not much use for `parts` by itself. Do change line 33 as well of course, and all the tests need to be updated.

www/src/js/utils/weekText.test.js
```diff
@@ -0,0 +1,15 @@
...
+describe('#getWeekText()', () => {
+	test('week type test', () => {
+		expect(getWeekText({ ...defaultAWI, type: 'Instructional' })[2]).not.toContain('Instructional');
+		expect(getWeekText({ ...defaultAWI, type: 'Reading'})[2]).toContain('Reading');
+		expect(getWeekText({ ...defaultAWI, type: 'Examination'})[2]).toContain('Examination');
+		expect(getWeekText({ ...defaultAWI, type: 'Recess'})[2]).toContain('Recess');
+		expect(getWeekText({ ...defaultAWI, type: 'Vacation'})[2]).toContain('Vacation');
+		expect(getWeekText({ ...defaultAWI, type: 'Orientation'})[2]).toContain('Orientation');
+	});
+});
```
taneliang: I think you need to run Prettier on this file to get rid of the space and newline at the end

www/src/js/utils/weekText.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+// Put outside render because this only needs to computed on page load.
```
taneliang: Should probably update this comment to something like "Text computed in an IIFE because this only needs to computed on page load."

www/src/js/utils/weekText.test.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+import { getWeekText } from './weekText';
+
+const defaultAWI = { year: '17/18', sem: 'Semester 2', type: 'Instructional', num: 1 };
+describe('#getWeekText()', () => {
+	test('week type test', () => {
```
taneliang: A better test name might be "should return proper week text" or something similar

www/src/js/utils/weekText.test.js
```diff
@@ -0,0 +1,15 @@
...
+		expect(getWeekText({ ...defaultAWI, type: 'Instructional' })[2]).not.toContain('Instructional');
+		expect(getWeekText({ ...defaultAWI, type: 'Reading'})[2]).toContain('Reading');
+		expect(getWeekText({ ...defaultAWI, type: 'Examination'})[2]).toContain('Examination');
+		expect(getWeekText({ ...defaultAWI, type: 'Recess'})[2]).toContain('Recess');
+		expect(getWeekText({ ...defaultAWI, type: 'Vacation'})[2]).toContain('Vacation');
+		expect(getWeekText({ ...defaultAWI, type: 'Orientation'})[2]).toContain('Orientation');
```
taneliang: 1. As mentioned in the first comment, after changing `getWeekType` to return a string, these tests need to test strings instead of arrays.
1. The tests should test the whole string for clarity, something like `expect(getWeekText({ ...defaultAWI, etc })).toEqual('AY2017/18 Semester 2, Week 2');`
1. You've covered all the week types, but do add tests for null/undefined nums as well. Something like `expect(getWeekText({ ...defaultAWI, type: 'Recess', num: undefined })).toEqual('AY2017/18 Semester 2, Recess Week');`

www/src/js/utils/weekText.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+
```
taneliang: /nit we usually don't leave a blank line after the flow comment

Issue #738, opened 2018-01-15
Timetable layout messed up with module titles


yangshun:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Timetable cells are not aligned in horizontal mode with titles showing.

**If the current behavior is a bug, please provide the steps to reproduce. If possible, please upload a screenshot or link to a video recording (on macOS you can use QuickTime) demonstrating the problem.**

1. Ensure on horizontal mode
2. Turn on titles
3. Open http://modsn.us/hpH2N
4. See that the lessons in Tue is not aligned

![screen shot 2018-01-15 at 11 01 47 am](https://user-images.githubusercontent.com/1315101/34957789-e2df0ecc-f9e3-11e7-912c-2b8581501afe.png)

**What is the expected behavior?**

Lesson cells aligned properly.

**Further information**

- Which version of NUSMods are you currently using? 20180115-e5e55c6
- Did this work in NUSMods BEFORE the R release/redesign? Yes
- Did this work in NUSMods AFTER the R release/redesign? NA. Broke with the introduction of module titles
- Which browser and version are you using? E.g. Chrome 63
- Which device are you using? E.g. MacBook Pro 17
- Which operating system and version are you on? E.g. macOS 10.12.5

ZhangYiJiang:
This actually looks like it might be a duplicate of #519

yangshun:
Probably related 👍

taneliang:
Fixed in #746

Issue #519, opened 2017-12-16
Classes displayed outside timetable if classes are short and packed


taneliang:
E.g. https://latest.nusmods.com/modules/BI3001A
![mod-display](https://user-images.githubusercontent.com/12784593/34069513-f449e334-e28d-11e7-899d-f5ed789f2121.png)

PR #746, opened 2018-01-16
Prevent cells from misaligning when titles are displayed

shanwpf:
Solves issue #738, doesn't solve #519 though


ZhangYiJiang:
Hi, the commit history looks kinda wonky. Would you mind fixing that, maybe by resetting to master then recreating the change since it's just one line?

ZhangYiJiang:
Thanks for the patch!

PR #755, opened 2018-01-19
Exam calendar view

### LINKED ISSUE #585: RFC: New view for exam schedule

ZhangYiJiang:
Fixes #585 

- Displays modules using a table, with each row representing a single time segment (morning, afternoon, evening) - this aligns exams on the same time, so evening exams are pushed down 
- Uses a calendar-like display, which should be familiar with anyone who has used a calendar app 

Stress test (exams occupying the same slot, exams outside the usual date ranges, three exams on one day, etc.)

screenshot from 2018-01-26 15-40-32

A slightly more reasonable case (my actual timetable) 

screenshot from 2018-01-26 15-41-35

### Backend changes 

- Added a `hoverable` class to explicitly enable hover styles for `color-*` styles 
- Moved `daysAfter` function from `util/ical` to `util/timify`


www/src/js/views/timetable/ExamCalendar.jsx
```diff
@@ -0,0 +1,248 @@
...
+
+  return (
+    <Fragment>
+      {headerRow}
+      {moduleRows}
+    </Fragment>
+  );
+}
+
+export default class ExamTimetable extends PureComponent<Props> {
```
taneliang: Should this be renamed to `ExamCalendar` to match the filename?

www/src/js/views/timetable/ExamCalendar.jsx
```diff
@@ -0,0 +1,259 @@
...
+  groupModulesByExams() {
+    const { semester } = this.props;
+    const [firstDayOfExams, weekCount] = this.getExamCalendar();
+
+    // If there are no modules with exams, then we can just stop here
+    if (weekCount === 0) return null;
+
+    // Wrap each module with its exam date info. This means we don't have to recalculate these
+    // every time we need them
+    const modulesWithExams: ModuleWithExamTime[] = [];
```
li-kai: Do you think moving this into reducers/moduleBank.js would be better?
I think `modulesWithExams` should be what is passed into ExamCalendar instead of it being calculated within the component.

www/src/js/views/timetable/ExamCalendar.jsx
```diff
@@ -0,0 +1,259 @@
...
+    const daysWithExams = Math.max(
+      5,
+      ...modulesWithExams.map((module) => examDateStringToDate(module.dateTime).getUTCDay()),
+    );
+
+    // Group modules by their exam date and time
+    const modulesByExamDate = groupBy(modulesWithExams, (module) => module.date);
+    return range(weekCount).map((week) =>
+      // Group by days
+      range(daysWithExams).map((day) => {
```
li-kai: I think moving this to ExamWeek would make the nesting more tolerable.

www/src/js/views/venues/AvailabilitySearch.jsx
```diff
@@ -24,7 +24,7 @@ export function defaultSearchOptions(
   now: Date = new Date(), // Used for tests only
 ): VenueSearchOptions {
   // Set day of week - if it is not a school day, then set to Monday (0)
-  const day = Math.max(getCurrentDayIndex(now), 0);
+  const day = getCurrentDayIndex(now) === 6 ? 0 : getCurrentDayIndex(now);
```
li-kai: I'm being cheeky here 😆 `getCurrentDayIndex(now) % 6`

www/src/js/views/timetable/ExamCalendar.jsx
ZhangYiJiang: I agree that this probably shouldn't be in the view, but I don't think it should go into moduleBank. May be able to extract it to `utils` maybe?

www/src/js/views/timetable/ExamCalendar.jsx
ZhangYiJiang: Don't think that would work. There are three layer of grouping here (by week, by day and then by time).

ZhangYiJiang:
@li-kai I'm not sure about hiding the button, because laying out the buttons is hard (as mentioned previously), and also because it may not be immediately clear to the user why the button isn't available. Disabling it may be a good idea though. 

@taneliang Show/hide title is not really really useful, especially if we assume most users hide title. I want to fix the timetable action buttons, but this requires redesigning the component's styling, and maybe even the code itself to hide/show an arbitrary number of buttons.

PR #756, opened 2018-01-19
Add feature to accomodate missing conjunctions

li-kai:
Well sadly CI not set up to run tests for api yet.

### DIFF
--- a/api/gulp-tasks/local/genReqTree/parseString.js
+++ b/api/gulp-tasks/local/genReqTree/parseString.js
@@ -1,8 +1,4 @@
-import {
-  Token,
-  Lexer,
-  Parser,
-} from 'chevrotain';
+import { Token, Lexer, Parser } from 'chevrotain';
 import R from 'ramda';
 import { OPERATORS, MODULE_REGEX, AND_OR_REGEX } from './constants';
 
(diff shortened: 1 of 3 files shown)


api/gulp-tasks/local/genReqTree/parseString.js
```diff
@@ -215,7 +245,8 @@ function parseString(pre, log) {
   const parser = new ReqTreeParser(tokens);
   const result = parser.parse();
   if (parser.errors.length > 0) {
-    log.error(`'${pre}' encoutered parsing errors:\n${parser.errors}`);
+    console.log(result, parser.errors);
```
ZhangYiJiang: So, why log instead of error?

api/gulp-tasks/local/genReqTree/parseString.js
```diff
@@ -175,7 +180,32 @@ function cleanOperators(tokens) {
...
+  const singularConjunction = removedDuplicates.reduce((acc, token) => {
+    const { image } = token;
+    if (image === 'and' || image === 'or') {
+      if (acc === null) return token;
+      if (image !== acc.image) return false;
+    }
+    return acc;
+  }, null);
+  // Fill in missing values with conjunction found
+  for (let i = 0; i < removedDuplicates.length; i += 1) {
```
ZhangYiJiang: Add `allowForLoopAfterthoughts` and change this to `i++`, so it match the config in the other projects?

api/gulp-tasks/local/genReqTree/parseString.js
```diff
@@ -175,7 +180,32 @@ function cleanOperators(tokens) {
     // then check if each element is different than the one before it
     return !(AND_OR_REGEX.test(currentImage) && AND_OR_REGEX.test(nextImage));
   });
-  return removedDuplicates;
+
+  const moduleTokens = [];
+  // Falsy value if array does not contain unique conjunction
+  const singularConjunction = removedDuplicates.reduce((acc, token) => {
```
ZhangYiJiang: Would it be more readable if you count the number of `and` and `or` make sure it's equal to 1?

ZhangYiJiang: Also best to add a comment explaining why this is necessary?

li-kai: We need the token itself so counting wouldn't really make it more readable.

PR #758, opened 2018-01-19
Fix footer and missing performance object exception

ZhangYiJiang:
Fixes two exceptions reported in Sentry 

- Unsupported time zone specified undefined - this should be a Chrome / V8 bug, but since we aren't using the function for anything important we'll just wrap the thing in a try catch block and ignore the exception. https://sentry.io/nusmods/v3/issues/434084130/
- performance is undefined - the previous fix in #705 was incorrect - sorry. Turns out if you just check `a && a.blah` when a is undefined, this will still crash. You need  `window.a && a.blah`. https://sentry.io/nusmods/v3/issues/433452196/


www/src/js/views/layout/Footer.jsx
```diff
@@ -19,7 +19,16 @@ export function FooterComponent(props: Props) {
...
   const versionStr = process.env.versionStr;
 
-  const lastUpdatedDate = props.lastUpdatedDate;
+  // Try catch because of Chrome crashing on calling toLocaleString with no parameter
+  // See https://sentry.io/nusmods/v3/issues/434084130/
+  let lastUpdatedDate;
+  try {
+    lastUpdatedDate = props.lastUpdatedDate
+      ? props.lastUpdatedDate.toLocaleString()
+      : 'hang on, loading...';
```
taneliang: This is grammatically incorrect when appended to "Data correct as at ". Should we compute the entire sentence in this ternary instead?
- Loading: "Hang on, loading data..."
- Loaded: `Data correct as at ${props.lastUpdatedDate.toLocaleString()}.`

www/src/js/views/layout/Footer.jsx
```diff
@@ -21,13 +21,13 @@ export function FooterComponent(props: Props) {
...
-      ? props.lastUpdatedDate.toLocaleString()
-      : 'hang on, loading...';
-  } catch (e) {
-    // Ignore
+  let lastUpdatedText = 'Loading data...';
+  if (props.lastUpdatedDate) {
+    try {
+      lastUpdatedText = `Data correct as ${props.lastUpdatedDate.toLocaleString()}.`;
+    } catch (e) {
+      lastUpdatedText = `Data correct as ${props.lastUpdatedDate.toString()}.`;
```
taneliang: Small typo. Should be "correct as at"

PR #760, opened 2018-01-20
Retry for ENOTFOUND errors

li-kai:
Apparently I did not RTFD. https://github.com/sindresorhus/got#retries

Uses a function so this retries for ENOTFOUND related errors. Not sure why by default they don't retry, might have a good reason behind it, @ZhangYiJiang do you know why?


ZhangYiJiang:
4xx errors indicate the client is at fault. Presumably Got is assuming the server is correct, and it won't bother to retry fetching a resource that the server says doesn't exist.

Issue #686, opened 2018-01-11
Display warning to users with unsupported browser 


ZhangYiJiang:
Our browser support matrix is 

Desktop 

- Chrome, Safari, Chrome, Edge - latest two versions 
- No version of IE 

Mobile 

- Android: Chrome mobile - latest two versions 
- iOS: Safari 9 and above 

For users on browsers that are not supported, we should show some sort of warning when NUSMods is first launched that they may experience issues. The warning should be dismissible and that preference persisted. It can be useful to include upgrade paths for common unsupported browsers (eg. Chrome/Firefox on Windows 7 and older OS X)  

For best effect it may be included outside the normal React app, since browser issues may prevent the app from running at all. May also want to design two different levels of warnings, one for browsers with known issues (old Safari, IE) and ones which are just slightly outdated (non-recent version of evergreen browsers).

ZhangYiJiang:
Example of error from unsupported browser (Chrome Mobile 34): https://sentry.io/nusmods/v3/issues/432780391/

ZhangYiJiang:
Another one - Safari 8: https://sentry.io/nusmods/v3/issues/432781077/

ZhangYiJiang:
IE11 on Windows 7: https://sentry.io/nusmods/v3/issues/432782926/

liaujianjie:
Can I take up this issue? The most straightforward way to do this is probably by checking against the user agent with static javascript. The alternative would be to detect features instead of browsers.

ZhangYiJiang:
I'm actually already working on this using [Bowser](https://github.com/lancedikson/bowser), which is slightly smaller because it is more limited in scope and has a nice `check` method that makes it quite easy to check browser versions. I'm still working on the UI, so if you want to do that I can hand the branch over to you.

ZhangYiJiang:
@liaujianjie I have pushed my unfinished code up at https://github.com/nusmodifications/nusmods/tree/browser-warning - comment here if you want to take over.

liaujianjie:
@ZhangYiJiang hey thanks, I'd love to take over.

liaujianjie:
How's this implementation UI-wise? I'm not sure if it is a strict requirement to adhere to the MD specs for this feature.

![screen shot 2018-02-04 at 4 49 33 pm](https://user-images.githubusercontent.com/3143132/35775834-9bbc246a-09cb-11e8-9df3-9615c429ba6f.png)

[...]

PR #761, opened 2018-01-20
Improve static pages

taneliang:
A bunch of fixes and improvements.

1. Fix pluralization of contributor commits, specifically "1 commits" -> "1 commit".
1. Overhaul FAQs.
    1. Replace all the rather rude and messy answers with one question and answer that addresses missing data. That was pretty much our entire FAQ.
    1. Add 2 questions from questions we got on Messenger.
1. Replace "Connect with us!" section on the About page with a Future section and action buttons.
1. Create `AnchorBlank` component to replace the inconsistent anchor tags throughout the static pages and footer. All external links in the static pages and footer now open in a new tab.
1. Run all static page text through Grammarly and reword various things for clarity, concision and correctness.

Also not exactly a static page, but I removed the extra line at the bottom of the Settings page.


www/src/js/views/static/FaqContainer.jsx
```diff
@@ -14,114 +13,96 @@ export default function FaqContainer() {
...
+        <p>
+          No, and we have no plans to make this possible. Having said that, if you would like to
+          implement this, we welcome pull requests{' '}
+          <AnchorBlank href={config.contact.githubRepo}>here</AnchorBlank>!
+        </p>
       </div>
 
       <div className={styles.question}>
-        <h5>Why is ... ?</h5>
+        <h5>Why do I only see a white page when I open NUSMods?</h5>
```
ZhangYiJiang: How would they see this if they've already WSOD? 😂

ZhangYiJiang: Anyway #686 should better cover this. We have accidentally WSOD supported browsers in the past too.

www/src/js/views/static/AboutContainer.jsx
```diff
@@ -1,47 +1,75 @@
...
       <p>
-        We would love to hear your feedback and suggestions on how to make NUSMods even better.
-        Please let us know them by leaving a comment on our{' '}
-        <a href={config.contact.facebook}>Facebook page</a>.
+        NUSMods is a fast-evolving project, and there are many things to be done. Help us help you!
       </p>
 
+      <div className={classnames('row no-gutters', styles.actionContainer)}>
+        <div className={classnames('col-lg', styles.btnContainer)}>
+          <button onClick={props.toggleFeedback} className="btn btn-primary btn-block">
```
ZhangYiJiang: Can use `btn-lg` for these I think. Also feel like these buttons are missing something - maybe grab some icons from Icons8 to use here?

www/src/js/views/static/FaqContainer.jsx
```diff
@@ -14,114 +13,96 @@ export default function FaqContainer() {
...
-          from CORS. As soon as the data becomes available, we will reflect it here within a day or
-          two by using the latest blockchain technology.
+          We do not update module data manually. All module data shown in NUSMods is obtained from{' '}
+          <AnchorBlank href="https://myaces.nus.edu.sg/cors/jsp/report/ModuleInfoListing.jsp">
+            CORS
+          </AnchorBlank>{' '}
+          and{' '}
+          <AnchorBlank href="https://ivle.nus.edu.sg/lms/public/search_course_public.aspx">
+            IVLE
+          </AnchorBlank>. Once the data becomes available there, NUSMods will reflect it within a
```
ZhangYiJiang: Maybe add a note here that they can check whether NUS has updated their info by clicking on the links on the module info page

www/src/js/views/components/AnchorBlank.jsx
```diff
@@ -0,0 +1,18 @@
+// @flow
+import React, { type Node } from 'react';
+
+type Props = {
+  ...HTMLAnchorElement,
+  children: Node,
```
ZhangYiJiang: Maybe make `href` a required prop for validation purpose?

www/src/js/views/static/FaqContainer.jsx
```diff
@@ -14,114 +13,96 @@ export default function FaqContainer() {
...
+          NUSMods does not make decisions regarding your curriculum, module availability and module
+          timetable. If you have questions regarding your curriculum, CORS bidding, IVLE, or
+          anything unrelated to NUSMods, it would be better to either contact your faculty and
+          department or just <AnchorBlank href="https://www.google.com/">Google</AnchorBlank> it.
         </p>
       </div>
 
       <div className={styles.question}>
-        <h5>When will semester X data be available?</h5>
+        <h5>There is a mistake! Can you update your system?</h5>
```
ZhangYiJiang: This seems to be *the* most commonly asked question that can be deflected (ha, interning at Zendesk taught me some customer support jargon) by looking at this page, so moving this to the first question may be a good idea

yangshun:
My recursive FAQ... 😭

taneliang:
@yangshun The recursive FAQ was probably more amusing than helpful judging by the number of questions we got on incorrect data 😂

yangshun:
Hmm will nesting all similar questions under one question be better? When
reading an FAQ, I scan through all the questions (which usually looks
structurally different from the answer) that are relevant to the one I want
to ask. If I don't see a question as a heading I'll probably just close the
page instead of reading the answers to other questions.

Could the discoverability of the FAQ be the problem instead of the original
unhelpful structure?

yangshun:
But I do think the newer version is more concise. How about improve the faq
on the discoverability aspect also?

taneliang:
I agree that nesting may not be the best idea because people will just ignore the answers while looking for their question. But I figured that "There's a mistake!" should make it obvious enough. Putting it right at the top as Yi Jiang suggested should also make it discoverable.

Maybe we can have a list of links above the questions? We can list the variations there and they can all link to the "There's a mistake!" question.

www/src/js/views/static/FaqContainer.jsx
taneliang: Hahaha yeah I thought about that. I figured that if they have another browser they'd be able to see it, and that'll still let us deflect (is that how you use that jargon) some of these questions.

I'll add more information about what's supported so that we don't deflect reports from people who are on supported browsers as well

yangshun:
Sounds good. Hope the number of requests decrease soon. (Actually it will
already go down as the semester starts hah)

www/src/js/views/static/AboutContainer.jsx
taneliang: I think the buttons look slightly better full width though. Do you prefer them this way?

localhost_8080_about

Icons sound good. I'll add them

ZhangYiJiang: `btn-lg` does not preclude them from being full-width

taneliang: Right. `btn-lg` is still big though.

localhost_8080_about iphone 5

taneliang: I didn't include `btn-lg`. Let me know if you really want me to add that haha

PR #762, opened 2018-01-20
Refactor timetables

li-kai:
Takes a different approach from the other branch.

We colocate redux actions with their components, this reduces the burden of TimetableContent being a parent of 4 different components. In the future we must take care to move responsibilities to a new component as needed instead of piling on.

Also the z-index PR is made much simpler by just raising z-index of the day name instead of the cells.


www/src/styles/constants.scss
```diff
@@ -180,7 +180,7 @@ $side-menu-overlay-z-index: 690;
 $cors-notification-z-index: 500;
 $timetable-selected-cell-z-index: 110;
 $timetable-timing-z-index: 95;
+$timetable-scolled-day-z-index: 120;
```
ZhangYiJiang: /nit Order these by the z-index value

www/src/js/views/timetable/TimetableCell.test.jsx
```diff
@@ -22,13 +22,7 @@ describe('<TimetableCell />', () => {
...
       const onButtonClick = jest.fn();
       const wrapper = shallow(
-        <TimetableCell
-          onClick={onButtonClick}
-          lesson={DEFAULT_LESSON}
-          style={{}}
-          isScrolledHorizontally
-          showTitle
-        />,
+        <TimetableCell onClick={onButtonClick} lesson={DEFAULT_LESSON} style={{}} showTitle />,
```
ZhangYiJiang: Can `style` be given a `defaultProp` value of empty object?

ZhangYiJiang: Also, it may be worth extracting these props into a variable if they are not relevant to the test

www/src/js/views/timetable/ModulesSelect.jsx
```diff
@@ -15,8 +14,9 @@ import CloseButton from 'views/components/CloseButton';
 import styles from './ModulesSelect.scss';
 
 type Props = {
-  moduleList: ModuleSelectList,
+  getFilteredModules: Function,
```
ZhangYiJiang: Should this be typed?

PR #764, opened 2018-01-22
Silence replaceState spam error on Safari

### LINKED ISSUE #763: SecurityError: Attempt to use history.replaceState() more than 100 times per 30.000000 seconds
If you spam the search box with more than 100 chars in 30 seconds in Safari on both macOS and iOS, this error is thrown and the error page is shown. https://sentry.io/nusmods/v3/issues/438101592/

```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30.000000 seconds
  at replaceState ([native code])
  at None ([native code])
  at push (./js/utils/HistoryDebouncer.js:34:20)
  at updateURL (./js/views/venues/VenuesContainer.jsx:134:13)
  at componentDidUpdate (./js/views/venues/VenuesContainer.jsx:103:12)
...
(14 additional frame(s) were not displayed)
```

taneliang:
Fixes #763.

Add a try/catch block around call to `this.history.replace` in `HistoryDebounce`. We only catch the exact error that Safari is throwing so that we don't inadvertently catch actual errors. Other errors will still be thrown.

Also confirmed that Safari still throws errors with that English message even if the phone's language is different.

### DIFF
--- a/www/src/js/utils/HistoryDebouncer.js
+++ b/www/src/js/utils/HistoryDebouncer.js
@@ -31,7 +31,17 @@ export default class HistoryDebouncer {
     if (Date.now() - this.lastPush > this.wait) {
       this.history.push(path, state);
     } else {
-      this.history.replace(path, state);
+      try {
+        this.history.replace(path, state);
+      } catch (e) {
+        if (
+          e.message ===
+          'Attempt to use history.replaceState() more than 100 times per 30.000000 seconds'
+        ) {
+          return;
+        }
+        throw e;
+      }
     }
 
     this.lastPush = Date.now();


www/src/js/utils/HistoryDebouncer.js
```diff
@@ -31,7 +31,17 @@ export default class HistoryDebouncer {
     if (Date.now() - this.lastPush > this.wait) {
       this.history.push(path, state);
     } else {
-      this.history.replace(path, state);
+      try {
+        this.history.replace(path, state);
+      } catch (e) {
+        if (
+          e.message ===
+          'Attempt to use history.replaceState() more than 100 times per 30.000000 seconds'
```
yangshun: I recommend adding the exception class to the check and making the error message matching less strict in case for some reason the number changes in future:

```js
if (e instanceof SecurityError && 
  e.message.includes('Attempt to use history.replaceState()') {
  ...
}
```

taneliang: I can only match the name of the error, as apparently `SecurityError` is just the `DOMError`/`DOMException`'s name. I could do `e instanceof DOMException` but Flow yells at me hahaha. Any idea how to import the `DOMException` type? Otherwise we'll have to stick with this ugly try/catch-ception, or remove the `SecurityError` check.

yangshun: In that case, I think just remove the `instanceof` check.

yangshun:
Oh what you have done now looks good. I thought `SecurityError` was the class name 😅

taneliang:
The try/catch-ception is fugly though. I'll add a comment

yangshun:
Why is there a need for a nested `try`/`catch`?

taneliang:
Good point. Somehow I thought accessing a non-existent prop will throw an error.

PR #766, opened 2018-01-22
Enable HMR over network

taneliang:
This PR enables HMR on non-local browsers such as phones, the iOS Simulator, other computers and even over the Internet.

- Fixed the `devServer.listen` call. We were passing in a callback where the host was expected. 
- This reenables the callback - the browser now automatically opens `http://0.0.0.0:8080` when `yarn start` is run.
- `yarn start:export` also opens `http://0.0.0.0:8081` in the browser.
- Export server mostly tested. The example config has been altered for consistency.

### TODO

- [x] Decide if we should disable the browser launch, as it may be annoying. "Hell naw" - not @yangshun 
- [x] Check if host and port has to be provided twice, because that bugs me
- [x] Test if https://github.com/nusmodifications/nusmods/pull/766/commits/d617f055850284a2211142d504f2f57e5ac8bf0c broke export


yangshun:
> This reenables the callback - the browser now automatically opens http://0.0.0.0:8080 when yarn start is run. This may be annoying - should I remove it?

No, this is useful. I wanted this behavior back.

www/scripts/start.js
```diff
@@ -58,7 +56,7 @@ function runDevServer(host, port, protocol) {
 
       // Proxy export endpoints to the local version for development
       '/export': {
-        target: 'http://localhost:3000',
+        target: 'http://0.0.0.0:3000',
```
yangshun: Hmm should the host here be using `DEFAULT_HOST`? Maybe port should be extracted out as a constant similar to the `DEFAULT_PORT`.

taneliang:
@ZhangYiJiang Can you help me check if I broke export, especially in the second last commit? Puppeteer is a pita and suddenly refused to work on my computer

ZhangYiJiang:
export server still works, although the auto opened browser is slightly annoying since *we* don't need to see the export page

taneliang:
Thanks! Yeah lol. I'll add a flag for that.

PR #779, opened 2018-01-24
Update ical-generator to our fork to fix macOS calendar bug 

### LINKED ISSUE #772: Can't sync the timetable from NUSMods into my iCalendar

Bug

> Hi can someone help me out? I can't sync the timetable from NUSMods into my iCalendar. Seems like it can't differentiate odd and even weeks. 

From [NUSWhispers](https://www.facebook.com/nuswhispers/posts/1772078842862569).

Not sure if this bug report is too ambiguous. Feel free to close it if it is.

ZhangYiJiang:
Fixes #772 

This uses a new EXDATE field for each excluded date, which should hopefully make it work for macOS. Unfortunately none of *us* can repro the bug. The other possibility is that the file they've imported may be corrupted somehow (and the file they sent us is a new copy, which is somehow uncorrupted), but in that case it seems unlikely all of the events still would show up. 

For reference, the patch to ical-generator: https://github.com/ZhangYiJiang/ical-generator/commit/ed6928fecce6789db4d99f856413074de7eb5a1e

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -105,7 +105,7 @@
     "classnames": "^2.2.5",
     "core-js": "^2.5.1",
     "downshift": "^1.22.3",
-    "ical-generator": "^0.2.10",
+    "ical-generator": "https://github.com/ZhangYiJiang/ical-generator.git#ed6928fe",
     "immutability-helper": "^2.3.1",
     "json2mq": "^0.2.0",
     "localforage": "^1.5.5",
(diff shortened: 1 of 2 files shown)


ZhangYiJiang:
Can confirm through manual testing that other clients (GCal in this case) still accepts the file. Oh, and this does make the file slightly bigger than before I guess, but it's not very significant.

PR #780, opened 2018-01-24
Upgrade dev dependencies

taneliang:
### TODO

- [x] `eslint-config-airbnb` `eslint-plugin-jsx-a11y` - new rules
- [x] `react-dev-utils` - breaks build
- [x] `babel-core` - breaks test
- [x] Remove CSS block https://github.com/nusmodifications/nusmods/blob/master/www/src/styles/bootstrap/style-override.scss#L18

Not updating `flow-bin` in this PR as it requires many changes to the libdefs and our code.

### Possible known issue

`no-scroll` class is not removed from `body` when closing the side menu on mobile.

1. Go to a module info page, e.g. https://deploy-preview-780--nusmods.netlify.com/modules/ACC1002/financial-accounting
1. Open the browser dev tools and shrink the viewport until the FAB appears onscreen.
1. Observe that the `body` has class `theme-eighties`.
1. Click the FAB to open the side menu.
1. Observe that `body` has class `theme-eighties no-scroll`. This is expected.
1. Click the FAB to close the side menu.
1. Observe that `body` class hasn't changed.

For some reason, I can't repro this in the deploy preview.


www/.stylelintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   rules: {
     'at-rule-empty-line-before': null,
     'at-rule-no-unknown': null,
+    'no-descending-specificity': null,
```
li-kai: Although the commit message is great, could we have a comment here too? 😄

taneliang:
I think we should upgrade `flow-bin` in a separate PR, because it introduces a bunch of new errors, some of which have to be fixed by updating the libdefs, which then introduces more errors.

So this PR is ready for review. Could you guys try to reproduce that possible known issue? I only see it on localhost (in both Safari and Chrome) and not in the deploy preview.

www/src/js/views/components/SideMenu.jsx
```diff
@@ -46,11 +46,17 @@ export class SideMenuComponent extends PureComponent<Props> {
 
     return (
       <Fragment>
-        <Fab className={styles.fab} onClick={() => toggleMenu(!isOpen)}>
+        <Fab
+          className={styles.fab}
+          onClick={() => toggleMenu(!isOpen)}
+          onKeyUp={(e) => e.keyCode === 27 && toggleMenu(false)} // Close menu when Esc is pressed
```
ZhangYiJiang: Don't think this works as intended. Since it is a keyboard event handler on the FAB, it will only fire if the focus is on the FAB.

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -273,6 +273,7 @@ class TimetableContent extends Component<Props, State> {
           verticalMode: isVerticalOrientation,
         })}
         onClick={this.cancelModifyLesson}
+        onKeyUp={(e) => e.keyCode === 27 && this.cancelModifyLesson()} // Close menu when Esc is pressed
```
ZhangYiJiang: /nit Not really a menu

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -330,7 +331,11 @@ class TimetableContent extends Component<Props, State> {
 
 function mapStateToProps(state, ownProps) {
   const { semester, timetable } = ownProps;
-  const modules = state.moduleBank.modules;
+  const {
```
ZhangYiJiang: I feel like this is less readable than the original

taneliang: I actually think it's more readable, - we can immediately see what exactly we're taking from `state`, and we can also see where everything comes from.

@li-kai @yangshun what do you think?

www/src/js/views/components/SideMenu.jsx
taneliang: Oops. I don't know how to get it to work so I'm just removing it since it's not important

www/src/js/views/timetable/TimetableContent.jsx
taneliang: I think we should go on a majority vote for this one since it's an opinion thing

yangshun: I prefer the previous approach. Destructuring works fine if you're not accessing nested stuff and want to reduce the number of times you have to write `props` and `state`.  In this case, the destructured property is only being used once. Also now, I have to look at two places to figure out the hierarchy. For e.g., look at L347 and wonder where `activeLesson` is from, then look at L336 to realize that `activeLesson` is a property on `app`.

taneliang: Okay! I'll undo this 👍

ZhangYiJiang:
The issue you described above is pretty unlikely to be caused by this changeset. I would expect it to be caused by the previous change I made to use `<Helmet>` for body classes, but I can't repro it

taneliang:
Alright. We'll fix it in another PR then

PR #781, opened 2018-01-25
Update module mock data

ZhangYiJiang:
Cherry picked from #755 to keep that PR small, and also because I need this in another branch. 

Since mock config returns AY2017/18 S1, I've updated the data to those in AY2017/18 (although they are from S2, I think this is reasonable). Unfortunately BFS1001 is not offered in 17/18 anymore. I've kept it because it's too much work to fix the tests, but we may want to do something about that later.


ZhangYiJiang:
config is already mocked. That is actually why I'm updating these to match config. We won't need to do this again next year.

PR #782, opened 2018-01-25
Fix Invariant Violation in Routes

ZhangYiJiang:
Stupid mistake. It doesn't really affect functionality since the redirection still works, although the "oops" page will appear briefly before the user is redirected since the error bubbled up to AppShell. 

Side note: Why did Flow not catch this? https://github.com/nusmodifications/nusmods/blob/375634dac1c480ce28d843e1cbd0cf5a41b3fb0b/www/flow-typed/npm/react-router-dom_v4.x.x.js#L132 indicates that it should right?

### DIFF
--- a/www/src/js/views/routes/Routes.jsx
+++ b/www/src/js/views/routes/Routes.jsx
@@ -1,5 +1,5 @@
 // @flow
-import React from 'react';
+import React, { type Node } from 'react';
 import { Switch, Route, Redirect } from 'react-router-dom';
 
 import TimetableContainer from 'views/timetable/TimetableContainer';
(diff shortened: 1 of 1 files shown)


ZhangYiJiang:
Sentry report https://sentry.io/nusmods/v3/issues/439860161/

ZhangYiJiang:
`React$Node` is an internal type defined in the react libdef which makes it available globally. It should be functionally equivalent to `Node`. I'm surprised `void` is in there though, because it would mean  `render()` *can* don't render anything, but that's invalid.

taneliang:
Yeah. Should probably file an issue. https://reactjs.org/docs/react-component.html#render doesn't say anything about returning `undefined`

PR #783, opened 2018-01-25
Fix removed tutorials disappearing from timetable 

### LINKED ISSUE #778: Bug: no tutorial slot shown for a mod for which a no-longer-exist tut slot was selected

a *bug*

**What is the current behavior
The tutorial slots for MA2104 are gone on the webpage. However, once I deleted the module and added it back (not by undo, if undo it the tutorial slots are still absent), the tutorial slots are back there again. 

Steps to reproduce: 
1. Before the start of the sem, I selected a tutorial slot which was then removed from cors.
    My designated tut slot: 
image

   However, during the bidding, I realized that this tutorial slot does not exist. I do not have a screenshot of this, but as can see from the screenshot of nusmods schedule below, there isnt a tutorial slot called tut[1]. 
image

  (and unfortunately I didnt manage to get any tutorial slot for this mod. So) after the bidding, (presumably after nusmods has alr synced the new timetable from the cors), I opened nusmods again, and tutorial[1] is gone, and no other tutorial slots appeared.
image
<ANSWER HERE>

To me it makes more sense to have another random tutorial slot in place instead of the complete disappearance of all tutorial slots after realizing the selected tutorial slots are deleted from the cors timetable. 

**Further information**

- Which version of NUSMods are you currently using? 
NUSMods R version 20180113-3682003
- Did this work in NUSMods BEFORE the R release/redesign? 
 You mean the bug? Unsure
- Did this work in NUSMods AFTER the R release/redesign? 
 You mean the bug? Yes it is in this R version. 
- Which browser and version are you using? 
 Firefox Version 57.0.1 (64bit)
- Which device are you using?
 Laptop
- Which operating system and version are you on? 
 Windows 10

ZhangYiJiang:
Fixes #778 

It's a pretty quick and not very clean fix. The bigger problem of #606 remains. I think maybe moving all of the validation stuff out to a middleware and coordinating from there may be a better idea.

### DIFF
--- a/www/src/js/utils/timetables.js
+++ b/www/src/js/utils/timetables.js
@@ -249,7 +249,8 @@ export function validateTimetableModules(
 
 /**
  * Validates the lesson config for a specific module. It replaces all lessons
- * which invalid class number with random class numbers, and
+ * which invalid class number with the first available class numbers, and
+ * removes lessons that are no longer valid
  * @param semester
  * @param lessonConfig
  * @param module
@@ -277,7 +278,7 @@ export function validateModuleLessons(
     //
     // If a lesson type is removed, then it simply won't be copied over
     if (!lessons.some((lesson) => lesson.ClassNo === classNo)) {
-      validatedLessonConfig[lessonType] = _.sample(lessons).ClassNo;
+      validatedLessonConfig[lessonType] = lessons[0].ClassNo;
       updatedLessonTypes.push(lessonType);
     } else {
       validatedLessonConfig[lessonType] = classNo;
--- a/www/src/js/utils/timetables.js
+++ b/www/src/js/utils/timetables.js
@@ -227,6 +227,15 @@ export function findExamClashes(
   return _.omitBy(groupedModules, (mods) => mods.length === 1); // Remove non-clashing mods
 }
 
+/**
+ * Validates the modules in a timetable. It removes all modules which do not exist in
+ * the provided module code map from the timetable and returns that as the first item
+ * in the tuple, and the module code of all removed modules as the second item.
+ *
+ * @param timetable
+ * @param moduleCodes
+ * @returns {[SemTimetableConfig, ModuleCode[]]}
+ */
 export function validateTimetableModules(
   timetable: SemTimetableConfig,
   moduleCodes: ModuleCodeMap,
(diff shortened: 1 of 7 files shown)


ZhangYiJiang:
Manually confirm fixed - 

1. Go to https://deploy-preview-783--nusmods.netlify.com/timetable/sem-1/share?CS1010S=LEC:1,REC:7,TUT:22&CS4243=LAB:4,LEC:2 - notice the invalid `LEC:2` on CS4243. 
2. Import the module - notice the invalid module does not appear 
3. Refresh - the invalid lesson will be fixed a few seconds afterwards 

This is inelegant, but it works, so let's just get this out first and make it better in another PR

www/src/js/utils/timetables.js
```diff
@@ -238,6 +247,48 @@ export function validateTimetableModules(
...
+    // Check that the lesson exists and is valid. If it is not, insert a random
+    // valid lesson. This covers both
+    //
+    // - lesson type is not in the original timetable (ie. a new lesson type was introduced)
+    //   in which case classNo is undefined and thus would not match
+    // - classNo is not valid anymore (ie. the class was removed)
+    //
+    // If a lesson type is removed, then it simply won't be copied over
+    if (!lessons.some((lesson) => lesson.ClassNo === classNo)) {
+      validatedLessonConfig[lessonType] = _.sample(lessons).ClassNo;
```
taneliang: Hmm. Will it be simpler to just take the first lesson?

www/src/js/utils/timetables.js
```diff
@@ -238,6 +247,48 @@ export function validateTimetableModules(
   return [_.pick(timetable, valid), invalid];
 }
 
+/**
+ * Validates the lesson config for a specific module. It replaces all lessons
+ * which invalid class number with random class numbers, and
```
taneliang: The sentence is incomplete

www/src/js/utils/timetables.test.js
```diff
@@ -499,3 +500,61 @@ describe('validateTimetableModules', () => {
...
+        semester,
+        {
+          Tutorial: '10',
+        },
+        cs1010s,
+      ),
+    ).toEqual([
+      {
+        Lecture: '1',
+        Recitation: expect.any(String),
```
taneliang: A deterministic lesson choice will also let us test that a proper lesson was picked

www/src/js/utils/timetables.js
ZhangYiJiang: A better idea would be to put it in a place that doesn't conflict with the existing timetable, but that TODO has been open for a long time now. Yeah, I think we'll just pick the first item (it won't necessarily be the first or earliest class though - I'm not sure how the classes are ordered in the data)

PR #784, opened 2018-01-27
Fix timetable deserialization crash

ZhangYiJiang:
So erm, you know [that last time](https://github.com/nusmodifications/nusmods/pull/707) I promised I'll read the docs properly? Turns out I didn't *properly* properly read it. I have done so now. I think. 

If this function has another bug due to be not reading the docs properly I promise I'll buy everyone on the team pizza (except @yangshun he's too far away)


yangshun:
Almost requested changes because I was excluded from the pizza.

ZhangYiJiang:
It's only hypothetical pizza

PR #794, opened 2018-01-30
Add warning for macOS users about calendar bug

### LINKED ISSUE #772: Can't sync the timetable from NUSMods into my iCalendar

ZhangYiJiang:
Fixes #772 

Well, sort of. We can't fix Apple's horrible calendar app, but we can warn users not to use it (and give them some alternatives in the FAQ).


www/src/js/views/static/FaqContainer.jsx
```diff
@@ -94,6 +100,36 @@ export default function FaqContainer() {
...
+          <ExternalLink href="https://zapier.com/blog/best-calendar-apps/">
+            a list of 18 apps for macOS
+          </ExternalLink>.
+        </p>
+
+        <p>
+          You can also{' '}
+          <ExternalLink href="https://nusit.nus.edu.sg/qat4/wp-content/uploads/2014/03/Apple-Mail-IMAP-Setup-Guide-Student.pdf">
+            link your NUSNet email account with the Mail app
+          </ExternalLink>, which gives you the option to sync your school calendar automatically.
```
taneliang: I don't have a Mac with me right now to confirm, but doesn't this only set up syncing for mail and not calendars?

taneliang: Okay derp didn't see the telegram chat before commenting. This still sounds suspicious

taneliang: I just tested this on a new user account, and it's really outdated. You can't follow those instructions exactly as the setup flow has changed. There's an option to set up Exchange accounts now so most of the steps are not necessary anyway. This may be more relevant: https://support.apple.com/en-sg/HT201951. Step 6 in Apple's guide is unnecessary.

Although it sets up calendar sync, but as mentioned in the previous comment, macOS calendar doesn't sync properly with this. Importing the calendar directly through the web interface doesn't remove the excluded events either (iirc).

www/src/js/views/timetable/ExportMenu.jsx
```diff
@@ -88,6 +115,25 @@ export class ExportMenuComponent extends PureComponent<Props> {
             )}
           </div>
         )}
+
+        <Modal
+          isOpen={this.state.isMacWarningOpen}
+          className="macOSWarning"
```
taneliang: Is this `className` necessary?

www/src/js/views/timetable/ExportMenu.jsx
```diff
@@ -88,6 +115,25 @@ export class ExportMenuComponent extends PureComponent<Props> {
             )}
           </div>
         )}
+
+        <Modal
```
taneliang: Should this `Modal` be pulled out into its own `WarningModal` component? It's not exactly relevant to the export menu's functionality

ZhangYiJiang: I think it's fine - the amount of code is very small, and we can always extract it if we really want to reuse this type of modal elsewhere

PR #797, opened 2018-02-02
Update libdefs

taneliang:
Update libdefs and fix errors:

1. Redux `createStore`: Even though our previous code should work (the null/undefined `defaultState` case is handled [here](https://github.com/reactjs/redux/blob/master/src/createStore.js#L32)), [the definition](https://github.com/nusmodifications/nusmods/compare/update-libdef?expand=1#diff-7fb5be65b98680e43621a033bd2efd26R46) says otherwise.
1. Lodash `keyBy`: The change doesn't sound very legit to me, but let's handle it anyway. See https://github.com/flowtype/flow-typed/pull/681


ZhangYiJiang:
Should revert the lodash and redux libdef changes. We patched previous version of the libdefs so those two uses will not raise errors - don't change our code to compensate for mistakes in the libdefs.

taneliang:
Alright. I added comments so that it'll be clear that they are patches.

taneliang:
Hopefully the next person who updates this will notice them. I spent quite a while trying to figure out when the libdefs were changed to understand why those changes were made, when actually they were just patches made by us.

To future person: here's a medal for you if you find them🥇

Issue #284, opened 2016-07-26
Doc on what to do whenever there is a new sem


ngzhian:
E.g.
1.  we need to update `academicCalendar.json`,
2. Add cors timetable
3. uncomment the html for CORS alerts

First suggested https://github.com/nusmodifications/nusmods/issues/282#issuecomment-235173209

yangshun:
Come up with a reference document on the maintenance stuff that we need to do every semester / year so that new members of the team can help in maintaining.

@ngzhian you so fast, i was halfway creating the issue lol

ngzhian:
@yangshun LOL sorry i sitting at home watching tv that's why got time to do this HAHA

yangshun:
I don't think I have the capacity and context to take on this task. @ZhangYiJiang and @taneliang could you all work on this together with @li-kai ?

PR #798, opened 2018-02-02
Update CORS URL to current acad year

taneliang:
The CORS link on the module info page was pointing to the AY2016/2017 page on CORS. I wonder how we didn't catch this earlier.

We'll need to add this to #284.

### DIFF
--- a/www/src/js/views/modules/ModulePageContent.jsx
+++ b/www/src/js/views/modules/ModulePageContent.jsx
@@ -65,6 +65,15 @@ export class ModulePageContentComponent extends Component<Props, State> {
     const pageTitle = `${ModuleCode} ${ModuleTitle}`;
     const semesters = getSemestersOffered(module);
 
+    const modSem = semesters.includes(config.semester) ? config.semester : semesters[0]; // Pick a sem if mod isn't available this sem
+    const ivleUrl = config.ivleUrl.replace('<ModuleCode>', ModuleCode);
+    const corsUrl =
+      semesters.length > 0 && // CORS URL only exists if mod is held in >=1 sem
+      config.corsUrl
+        .replace('<ModuleCode>', ModuleCode)
+        .replace('<AcademicYear>', module.AcadYear)
+        .replace('<Semester>', modSem.toString());
+
     return (
       <div className={classnames('page-container', styles.moduleInfoPage)}>
         <Title>{pageTitle}</Title>
@@ -173,10 +182,10 @@ export class ModulePageContentComponent extends Component<Props, State> {
                     <h3 className={styles.descriptionHeading}>Official Links</h3>
                     {intersperse(
                       [
-                        <a key="ivle" href={config.ivleUrl.replace('<ModuleCode>', ModuleCode)}>
+                        <a key="ivle" href={ivleUrl}>
                           IVLE
                         </a>,
-                        <a key="cors" href={config.corsUrl + ModuleCode}>
+                        <a key="cors" href={corsUrl}>
                           CORS
                         </a>,
                       ],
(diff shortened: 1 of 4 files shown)


ZhangYiJiang:
Can we use `config.acadYear` instead?

taneliang:
Yeah good idea. I'll inject that in in `config/index.js`

ZhangYiJiang:
`<semester>` needs to be injected at the call site - otherwise the link to modules that are not available in the current semester will 404

taneliang:
Makes sense. Changed it to use the modules's `AcadYear` as well. Some mods still have missing data though, e.g. https://deploy-preview-798--nusmods.netlify.com/modules/GEH1015

I also enabled the sem picker when there's only one sem, so that it's clear which sem's timetable is being shown.

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -65,6 +65,15 @@ export class ModulePageContentComponent extends Component<Props, State> {
     const pageTitle = `${ModuleCode} ${ModuleTitle}`;
     const semesters = getSemestersOffered(module);
 
+    const modSem = semesters.includes(config.semester) ? config.semester : semesters[0]; // Pick a sem if mod isn't available this sem
```
ZhangYiJiang: Would it be better to show one link for each semester?

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -65,6 +65,15 @@ export class ModulePageContentComponent extends Component<Props, State> {
...
+    const ivleUrl = config.ivleUrl.replace('<ModuleCode>', ModuleCode);
+    const corsUrl =
+      semesters.length > 0 && // CORS URL only exists if mod is held in >=1 sem
```
ZhangYiJiang: I don't think this will ever be false

www/src/js/views/modules/ModulePageContent.jsx
taneliang: So the official links will show this?
* IVLE
* CORS Semester 1
* CORS Semester 2
* CORS Special Term I
* CORS Special Term II

We can hide most of them in a dropdown. Should we implement this in another PR though? I think the current implementation is okay, and it can be another good first issue

ZhangYiJiang:
I've made the change to make it list out all CORS links for all semesters. I don't know if CORS links work for special semester modules though. The effect is also a little ugly when the line wraps because CSS doesn't give us the proper tools to typeset inline bullet points (like https://tex.stackexchange.com/questions/372662/creating-a-ragged-right-inline-list-separated-by-bullets-or-line-breaks), but I think it is acceptable for now

taneliang:
Actually does CORS even respect the `sem_c` and `acad_y` params? These URLs show the same info. Both params should be valid as the first URL was obtained from https://myaces.nus.edu.sg/cors/jsp/report/ModuleInfoListing.jsp. It seems like they only show info for the current AY and sem.

- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2017/2018&sem_c=2&mod_c=MA1301
- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2017/2018&sem_c=1&mod_c=MA1301
- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2017/2018&sem_c=4&mod_c=MA1301
- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2017/2018&sem_c=ILOVECORS&mod_c=MA1301
- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?acad_y=2017/2018&mod_c=MA1301
- https://myaces.nus.edu.sg/cors/jsp/report/ModuleDetailedInfo.jsp?mod_c=MA1301

ZhangYiJiang:
... so the query string does nothing except change the text in the heading 

godamnit CORS 

 

In that case it's kind of useless to bother making multiple links. I guess we can revert to the previous commit.

taneliang:
CORS 10/10. Anyway the CORS URL is now generated using the app config. That's probably better than using the module's info since CORS will just show the data for the current AY and sem.

PR #799, opened 2018-02-05
Fix modal glyph display

taneliang:
The nice big glyphs in these two modals were not centered because of the close button.

This PR defines `AbsCloseButton`, a new absolute positioned `CloseButton` wrapped in a relative positioned `div`. It can be used as a drop-in replacement for `CloseButton` when needed. As `CloseButton` is untouched, it will not affect the other components which use it.

Tested in Chrome 56, Safari 10.1, Edge 15, Firefox 50 (although other CSS elements are broken on that version).

Before:
nusmods com_timetable_sem-2
nusmods com_timetable_sem-2 1

After:
localhost_8080_ 1
localhost_8080_

### DIFF
--- a/www/src/js/views/components/CloseButton.jsx
+++ b/www/src/js/views/components/CloseButton.jsx
@@ -4,12 +4,14 @@ import React from 'react';
 import classnames from 'classnames';
 import { Close } from 'views/components/icons';
 
+import styles from './CloseButton.scss';
+
 type Props = {
   onClick: Function,
   className?: string,
 };
 
-export default function({ onClick, className }: Props) {
+export default function CloseButton({ onClick, className }: Props) {
   return (
     <button
       className={classnames('close', className)}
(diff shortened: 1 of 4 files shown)


ZhangYiJiang:
Can we make it a property of the `<CloseButton>` or a classname instead? It doesn't seem to require defining a new component.

taneliang:
A classname wouldn’t work because the actual button has to be wrapped in the relative positioned div, but a prop will sounds good

taneliang:
Hmm it's used in 2 places now, and we'll probably use it in a login modal too, so I thought making it reusable would be a good idea. Defining a new component like in my original approach feels less hacky, but the final approach is more semantic so I guess it's better

ZhangYiJiang:
Oh, I mean the AP-RP wrapper - I think it would be more flexible to just let the user decide which parent has RP. But that does mean the component "leaks" out into the parent, which can be annoying.

PR #807, opened 2018-02-08
Warn users using outdated browsers

### LINKED ISSUE #686: Display warning to users with unsupported browser

liaujianjie:
Fixes #686.

## Implementation

* Use bowser to check for outdated browsers outside of the scope of React
* Saves flag to localStorage if user clicks "Don't show again"

### Firefox 51.0.1

screen shot 2018-02-08 at 4 25 18 pm

### Firefox 32.0.3

screen shot 2018-02-08 at 4 24 05 pm

### Firefox 12.0.0

screen shot 2018-02-08 at 4 24 30 pm

Tested on macOS High Sierra


www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import bowser from 'bowser';
+
+const LOCAL_STORAGE_KEY = 'dismissedBrowserWarning';
+const browserSupportsLocalStorage = (() => {
```
taneliang: Could we use [this existing implementation to check for localStorage support](https://github.com/nusmodifications/nusmods/blob/master/www/src/js/storage/localStorage.js#L30-L44) instead? It can be extracted into a function that can be used here.

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,66 @@
...
+      `;
+    const container = document.createElement('div');
+    container.className = 'browser-warning';
+    container.innerHTML = template;
+    const body = document.body;
+    if (body) body.appendChild(container);
+
+    const addDismissListener = (elementId, composedFunction) => {
+      const element = document.getElementById(elementId);
+      if (element)
```
taneliang: /nit can we have {} around this multi-line block? It'll be easier to read IMO

ZhangYiJiang:
Ah, it works even when its bundled! I was worried we have to split it into another bundle or transpile it separately because the browser may not execute it if it chokes on a syntax error (due to not supporting newer syntax) and stuff, but it looks like it works fine. 

@liaujianjie We have a 5 seat subscription sponsored by BrowserStack, so we can test it on older browsers. 

Mobile Safari 7 

screenshot from 2018-02-08 21-49-41

IE 11 

screenshot from 2018-02-08 21-54-28

Looks like it works pretty well. The heading font size should be a smaller though (`.h3` maybe?). Also since bundling works, we can use CSS modules instead of using global class names.

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,60 @@
...
+) {
+  if (
+    (browserSupportsLocalStorage && !localStorage.getItem(LOCAL_STORAGE_KEY)) ||
+    !browserSupportsLocalStorage
+  ) {
+    const template = `
+      <div class="browser-warning__overlay">
+      <div class="browser-warning__modal">
+      <h1>Your web browser is outdated or unsupported</h1>
+      <p>NUSMods may not work or may work poorly. Please consider upgrading your web browser.</p>
```
taneliang: The second sentence isn't appropriate when the browser is up-to-date but just unsupported. Safari users on older versions of macOS and iOS also will not be able to upgrade Safari easily.

How about "NUSMods may not work properly. Please consider updating your web browser or switching to Google Chrome or Mozilla Firefox.", with links to their download pages.

If you want to go all out, you could suggest different browsers for iOS (e.g. if not Safari/Chrome, switch to Safari/Chrome. If Safari/Chrome, upgrade OS if possible), Android and desktop OSes. That might be best because if they're using an outdated browser they probably won't know how to upgrade it lol

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,60 @@
...
+      <button class="btn btn-primary" id="browser-warning-continue">Continue to NUSMods</button>
+      ${
+        // Show "don't show again" only if the browser supports localStorage
+        browserSupportsLocalStorage
+          ? '<button class="btn" id="browser-warning-ignore">Don\'t show again</button>'
```
taneliang: I don't really like the behavior of this button. It works as advertised, and is perfect when NUSMods mostly works but is just a little funky looking. But if we're crashing on startup and if the user presses this button (perhaps by accident), they'll be looking at a WSOD with no further explanation.

I don't really have a solution for this. Getting rid of this hide forever behavior will be really annoying for people on browsers which mostly work. It'll be best if we could detect a WSOD and show this warning, but I'm not sure how to do that. Alternatively, maybe we could do something dumb like showing this dialog again if we detect 3 refreshes in 30 seconds.

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,60 @@
...
+    container.className = 'browser-warning';
+    container.innerHTML = template;
+    const body = document.body;
+    if (body) body.appendChild(container);
+
+    const addDismissListener = (elementId, composedFunction) => {
+      const element = document.getElementById(elementId);
+      if (element) {
+        element.addEventListener('click', (event) => {
+          event.preventDefault();
```
yangshun: What are you preventing here?

www/src/styles/components/browser.scss
```diff
@@ -0,0 +1,24 @@
...
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 9999;
+  float: left;
+  overflow: hidden;
+
+  &__overlay {
```
yangshun: We don't use BEM. Let's stick with Bootstrap-style class naming

www/src/styles/components/browser.scss
```diff
@@ -0,0 +1,24 @@
...
+    height: 100%;
+    background: rgba(0, 0, 0, 0.5);
+  }
+
+  &__modal {
+    max-width: 48em;
+    padding: 2em;
+    margin-right: auto;
+    margin-left: auto;
+    background: white;
```
yangshun: Nit: `background: #fff;`

www/src/js/bootstrapping/browser.js
liaujianjie: > How about "NUSMods may not work properly. Please consider updating your web browser or switching to Google Chrome or Mozilla Firefox.", with links to their download pages.

Sounds good. Then should use use anchor links or add it to the row of buttons?

> If you want to go all out, you could suggest different browsers for iOS (e.g. if not Safari/Chrome, switch to Safari/Chrome. If Safari/Chrome, upgrade OS if possible), Android and desktop OSes. That might be best because if they're using an outdated browser they probably won't know how to upgrade it lol

Sure, that's fairly easy with bowser.

www/src/js/bootstrapping/browser.js
liaujianjie: Flow throws me a compile error when I try to build if I do not check for null on possibly-null values:

```
Error: src/js/bootstrapping/browser.js:47
           v---------------------------------------------
 47:       element.addEventListener('click', (event) => {
 48:         event.preventDefault();
 49:         if (body) body.removeChild(container);
 50:         if (composedFunction && typeof composedFunction === 'function') composedFunction();
 51:       });
           -^ call of method `addEventListener`. Method cannot be called on possibly null value
 47:       element.addEventListener('click', (event) => {
           ^^^^^^^ null
```

AFAIK, Flow doesn't support optional chaining (yet?). Is there a better way to go around this?

www/src/js/bootstrapping/browser.js
liaujianjie: > It'll be best if we could detect a WSOD and show this warning, but I'm not sure how to do that.

Hmmm I'll try playing around

www/src/js/bootstrapping/browser.js
yangshun: I'm referring to the line `event.preventDefault();`

liaujianjie: I didn't write it, but I also didn't remove it because I assumed @ZhangYiJiang wrote it to prevent some unwanted artefacts.

Should we remove it?

yangshun: Oh I see. It's probably there for a good reason then. Let's leave it there unless @ZhangYiJiang chimes in and gives the green light that it is safe to remove.

ZhangYiJiang: Force of habit lol. It should be safe to remove, though the HTML should be updated to `<button type="button">` so there's no default action associated with the event.

www/src/js/bootstrapping/browser.js
liaujianjie: I tried adding a callback to the `ReactDOM.render` call of the React root node but its called only once when no child components have been rendered into the DOM yet. So we can't check for WSOD with that. Another alternative is to manually test out which browser versions gives a WSOD and hard code those values (yuck). We can then display different dialogue accordingly. 

Anyway to be frank, I don't see a merit in catering for this edge case for 3 reasons:

1. (UX) The unhighlighted colour of the "Don't show again" button already suggests its is a less ideal option than the "Continue" button.
2. The whole point of an overlay dialogue is to focus the entire user's attention on the dialog. If a user misclicks, is that something we really want to cater to? A PC user won't be able to retrieve his/her files after emptying his trash.
3. What are the odds of a user misclicking anyway? That low likelihood of misclicking intersecting with the low number of users (my assumption) using browsers that show WSOD makes it a lot less appealing to introduce additional code complexity.

Perhaps we should consider other alternatives to this problem:

* Use a checkbox instead of a second button.
* Introduce a colour that suggests destructiveness in the design language of this app. If we had that, we could have easily added a destructive styling to the button.

www/src/js/bootstrapping/browser.js
taneliang: Anchor links would be better

www/src/js/bootstrapping/browser.js
taneliang: I like the checkbox idea. It forces people to tap twice to dismiss the dialog forever, which I think is good enough for WSOD cases and not too inconvenient for users of working but unsupported browsers.

Over the past month, Sentry counted 400+ "users" who had 1.5k "'Symbol' is undefined" errors (one of the current WSOD causes). GA counted 65k total users over the past month.

liaujianjie: Wow, that's a lot more users than I expected. Checkbox it is then.

www/src/js/bootstrapping/browser.js
```diff
@@ -20,41 +24,53 @@ if (
     (browserSupportsLocalStorage && !localStorage.getItem(LOCAL_STORAGE_KEY)) ||
     !browserSupportsLocalStorage
   ) {
+    const promptText = (() => {
+      if (bowser.ios && bowser.safari)
+        return `NUSMods may not work properly. Please consider updating your OS or switching to the latest version of <a href="${LINK_FOR_CHROME}">Google Chrome</a></a>.`;
+      if (bowser.android && bowser.chrome)
+        return `NUSMods may not work properly. Please consider updating your web browser.`;
+      return `NUSMods may not work properly. Please consider updating your web browser or switching to the latest version of <a href="${LINK_FOR_CHROME}">Google Chrome</a> or <a href=${LINK_FOR_FIREFOX}>Mozilla Firefox</a>.`;
```
taneliang: Should the links open in a new window/tab? `<a target="_blank" rel="noopener noreferrer" href="...">`

www/src/js/bootstrapping/browser.js
```diff
@@ -20,41 +24,53 @@ if (
...
+          <button class="btn btn-primary" id="browserWarning-continue" type="button">Continue to NUSMods</button>
+        </div>
+        ${
+          // Show "don't show again" only if the browser supports localStorage
+          browserSupportsLocalStorage
+            ? `
+              <div class="col-auto ${styles.checkboxContainer}">
+                <div class="form-check form-check-inline">
+                  <input class="form-check-input" id="browserWarning-ignore" type="checkbox">
+                  <label class="form-check-label" for="browserWarning-ignore">Don't show this dialog again</label>
```
taneliang: "Dialog" is a bit technical. How about "message"

www/src/js/bootstrapping/browser.js
```diff
@@ -20,41 +24,53 @@ if (
     (browserSupportsLocalStorage && !localStorage.getItem(LOCAL_STORAGE_KEY)) ||
     !browserSupportsLocalStorage
   ) {
+    const promptText = (() => {
+      if (bowser.ios && bowser.safari)
+        return `NUSMods may not work properly. Please consider updating your OS or switching to the latest version of <a href="${LINK_FOR_CHROME}">Google Chrome</a></a>.`;
```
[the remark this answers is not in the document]
ZhangYiJiang: All browsers on iOS use Webkit under the hood, yes

www/src/js/storage/localStorage.js
```diff
@@ -20,37 +20,40 @@ export function createLocalStorageShim() {
   return storage;
 }
 
+export function isLocalStorageLike(storage) {
+  try {
+    // Ensure that localStorage isn't null
+    // Resolves https://sentry.io/share/issue/d65da46a7e19406aaee298fb89a635d6/
+    if (!storage) throw new Error();
```
ZhangYiJiang: Should just `return false` - no need to throw, catch then return

taneliang: The previous implementation had to shim local storage in its catch block. There's no need to do that now so `return false` here would work.

Now that we don't have to do that, I think it'll be even clearer if this check was moved above the try block and returned false. So that we know that only the `setItem` and `removeItem` calls might throw.

www/src/js/bootstrapping/browser.js
```diff
@@ -15,7 +15,7 @@ const composeAnchorText = (innerHTML, href) =>
 const linkForChrome = composeAnchorText('Google Chrome', 'https://www.google.com/chrome/');
 const linkForFirefox = composeAnchorText('Mozilla Firefox', 'https://www.mozilla.org/en-US/');
 
-const browserSupportsLocalStorage = checkBrowserSupportsLocalStorage();
+const browserSupportsLocalStorage = isLocalStorageLike(window.localStorage);
```
taneliang: This line should throw on Chrome with cookies disabled. See https://github.com/nusmodifications/nusmods/pull/807/commits/d5f173f3ca03f0a1227d822b108cf6bbcda87d75#diff-a912dc4928d4e196610cdfbf510c7a55L26. I think we should revert this change and use `checkBrowserSupportsLocalStorage`. You could rename it if you think it's too verbose though

www/src/js/storage/localStorage.js
liaujianjie: Ah okay thanks, will change it then!

taneliang: Actually scratch that. I forgot that accessing `window.localStorage` can throw, so we'll still need to put lines 25-28 back in this try block. And that means that we can't do this check outside the try block.

I think throwing the error is cleaner than `return false`, so that the catch block is the only place that handles the case where the browser doesn't support local storage.

www/src/js/bootstrapping/browser.js
liaujianjie: I changed it only because I won't able to test it in the previous form. Should I just not test it then?

taneliang: Is there a way to attach a mock `localStorage` object to a mock `window`? If there isn't then yeah it probably can't be tested then. I found this though, might be useful but it's a bit old: http://booksyoudontread.com/2016/10/10/mocking-window-localstorage-in-jest/

liaujianjie: TIL you can set `window.localStorage` but not `localStorage` lol

taneliang:
@liaujianjie could you update bowser to 1.9.2 too

www/src/js/storage/localStorage.test.js
```diff
@@ -42,3 +42,54 @@ describe('#createLocalStorageShim', () => {
...
+  });
+
+  test('should return true if localStorage is localStorage-like object', () => {
+    window.localStorage = createLocalStorageShim();
+    expect(canUseBrowserLocalStorage()).toEqual(true);
+  });
+});
+
+describe('#getLocalStorage', () => {
+  test('should get usable localStorage-like object', () => {
```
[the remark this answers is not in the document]
liaujianjie: Oh yes how did I not see that lol

www/src/js/storage/localStorage.test.js
```diff
@@ -42,3 +42,54 @@ describe('#createLocalStorageShim', () => {
...
+    // getLocalStorage returns either the actual browser's localStorage or a shim of it and MUST
+    // always return an object that behaves like localStorage
+    window.localStorage = getLocalStorage();
+    expect(canUseBrowserLocalStorage()).toEqual(true);
+  });
+
+  test("should return the actual browser's localStorage if the browser can use localStorage", () => {
+    expect(getLocalStorage()).toBe(window.localStorage);
```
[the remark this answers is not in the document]
liaujianjie: The test on line 47 already tests for this.

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,85 @@
...
+  )
+) {
+  if (
+    (browserCanUseLocalStorage && !localStorage.getItem(LOCAL_STORAGE_KEY)) ||
+    !browserCanUseLocalStorage
+  ) {
+    const promptText = (() => {
+      // Users can only update Safari by updating the OS in iOS
+      if (bowser.ios && bowser.safari)
+        return `NUSMods may not work properly. Please consider updating your OS or switching to the latest version of ${linkForChrome}.`;
```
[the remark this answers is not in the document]
liaujianjie: While I'm add it, should I change the recommendation to "Please consider updating to iOS 11 or higher" instead?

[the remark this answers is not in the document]
taneliang: Yeah sure

www/src/js/storage/localStorage.test.js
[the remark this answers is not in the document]
taneliang: Eh okay my reasoning was wrong. I thought that we should put that extra check in to ensure that window.localStorage exists before running our test, as like a kind of precondition assertion to make diagnosing easier when this test fails. But okay this shouldn't be a big issue, so just ignore me

www/src/js/storage/localStorage.test.js
```diff
@@ -42,3 +42,48 @@ describe('#createLocalStorageShim', () => {
...
+describe('#getLocalStorage', () => {
+  test("should return the actual browser's localStorage if the browser can use localStorage", () => {
+    expect(canUseBrowserLocalStorage()).toEqual(true);
+    expect(getLocalStorage()).toBe(window.localStorage);
+  });
+
+  test('should return a shim if browser cannot use localStorage', () => {
+    window.localStorage = undefined;
+    expect(canUseBrowserLocalStorage()).toEqual(false);
+    expect(getLocalStorage()).toEqual(
```
ZhangYiJiang: Use `.toMatchObject` instead of `.toEqual(expect.objectContaining())`

www/src/js/storage/localStorage.test.js
```diff
@@ -42,3 +42,48 @@ describe('#createLocalStorageShim', () => {
...
+      expect.objectContaining({
+        privData: expect.any(Object),
```
ZhangYiJiang: What's this for?

www/src/js/bootstrapping/browser.scss
```diff
@@ -0,0 +1,29 @@
+.browserWarning {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 9999;
+  float: left;
```
ZhangYiJiang: What's this for?

www/src/js/bootstrapping/browser.scss
```diff
@@ -0,0 +1,29 @@
+.browserWarning {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 9999;
```
ZhangYiJiang: Should move this into `constants.scss`

www/src/js/bootstrapping/browser.scss
```diff
@@ -0,0 +1,29 @@
+.browserWarning {
+  position: fixed;
```
ZhangYiJiang: If you're applying `position: fixed` here I think you can get rid of one layer of `div`s. It's usually the overlay element that's position fixed.

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,85 @@
...
+  if (
+    (browserCanUseLocalStorage && !localStorage.getItem(LOCAL_STORAGE_KEY)) ||
+    !browserCanUseLocalStorage
+  ) {
+    const promptText = (() => {
+      // Users can only update Safari by updating the OS in iOS
+      if (bowser.ios)
+        return `NUSMods may not work properly. Please consider updating your device to iOS 11 or higher.`;
+      if (bowser.android && bowser.chrome)
+        return `NUSMods may not work properly. Please consider updating your web browser.`;
```
ZhangYiJiang: If you really want to you can link to Chrome using a Play store link https://developer.android.com/distribute/marketing-tools/linking-to-google-play.html

www/src/js/bootstrapping/browser.js
```diff
@@ -0,0 +1,85 @@
+// @flow
+
+// This script checks for browser compatibility and is executed outside React's scope. If a browser
+// is incompatible or not optimal for using the app, a pop-up dialog is appended into the DOM body.
+// This is so that in cases where the React app completely fails to render anything at all, the
+// user will at least be able to see the dialog warning them of the browser incompatibility.
+
+import bowser from 'bowser';
+import { canUseBrowserLocalStorage } from '../storage/localStorage';
```
ZhangYiJiang: /nit Can drop the `../` from the import path

ZhangYiJiang:
Oh, and thanks for putting up with the slightly ridiculous amount of review comments from all of us 😅

www/src/js/bootstrapping/browser.scss
liaujianjie: Good point, thanks.

www/src/js/bootstrapping/browser.scss
liaujianjie: I thought it needed to be there, oops.

www/src/js/storage/localStorage.test.js
liaujianjie: `getLocalStorage` returns an object with a newly-instantiated function as a prop and so it behaves kind of like a HOF. I did this only because I couldn't find a way to test HOFs with Jest. Tests will fail if you try to match 2 different objects (containing functions) instantiated from the same function with `.toMatchObject`.

The shape of this object was copied from [`createLocalStorageShim`](https://github.com/nusmodifications/nusmods/pull/807/files#diff-a912dc4928d4e196610cdfbf510c7a55L5).

www/src/js/bootstrapping/browser.js
taneliang: If we don't tell them how to update their browser they probably won't know how to do it lol. I think it's fine to throw them into the store if our message is clear enough, like "Please consider updating your web browser - [this link will take you to Chrome on the Google Play Store](https://developer.android.com/distribute/marketing-tools/linking-to-google-play.html)."

www/src/js/storage/localStorage.test.js
taneliang: `privData` will identify this object as the shim since the browser's localStorage won't have this prop. I don't think we should test for the presence of this prop though, since it's an implementation detail and we don't have the browser localStorage to differentiate our shim from.

www/src/js/storage/localStorage.test.js
```diff
@@ -42,3 +42,48 @@ describe('#createLocalStorageShim', () => {
...
+  test('should return false if localStorage throws when writing on iOS <=10 on private browsing', () => {
+    window.localStorage = {
+      ...createLocalStorageShim(),
+      // the length is set here because canUseBrowserLocalStorage uses a hack to detect private browsing
+      length: 0,
+      setItem: () => {
+        throw new Error();
+      },
+    };
+    expect(canUseBrowserLocalStorage()).toEqual(false);
```
liaujianjie: Should I use `.toBeFalsey()`/`.toBeTruthy()` instead of `.toEqual(false)`/`.toEqual(true)` for all of such matchers?

ZhangYiJiang: No, we prefer `toBe(false/true)` because it is more precise - `.toBeFalsey/Truthy` also matches `0` and `''` (empty string), which we don't want if we know the value will be a boolean

liaujianjie: Wah you're quick! Makes sense, got it.

liaujianjie:
I think this commit should be golden. Thanks all for your patience with me haha.

Issue #796, opened 2018-02-01
User accounts


taneliang:
An issue describing the general approach for user accounts. **TL;DR**: Let's just use Firebase Cloud Firestore and pay a few dollars every peak period.

Feel free to edit and add on to this, or shoot this down.

## Benefits

- No more hacky QR code syncing
- Just about everyone uses multiple devices - no more timetables stuck on your home computer
- Other developers can integrate better with student timetables without resorting to hacky methods like using the share URL. This will allow an ecosystem to be built around NUSMods, e.g. timetable planners, common free time finders, etc.

### Core

- A sync server should be implemented just to handle sync. It should not be integrated with the API (data) server as they don't need to share data, and this will allow us to work on them separately.
- The server should store only preferences and the user's timetable, along with their names and optionally profile pics.
- The server should sync in real-time so that we won't have inconsistent states.
- The server should be ready for multiple timetables per user
- The sync service should be able to continue operating for years without maintainance

### Optional

- The server can support normal REST/GraphQL endpoints for third-party developers and applications which don't require real-time syncing.
- API keys for third-party developers for access control.
- Users should be able to modify their timetable when offline, and have those changes synced.
- Sync clashes should be handled automatically

### Authentication

I think it'll be easiest if we outsource authentication so that we won't have to deal with all the complexities.

If we use Firebase, we can just use their authentication feature as well. A [prebuilt UI component](https://firebase.google.com/docs/auth/web/firebaseui) and [official React component](https://www.npmjs.com/package/react-firebaseui) are available. ezpz.

Alternatively, Auth0 seems like a good choice, and they have an [Open Source Program](https://auth0.com/pricing) which lets us use it for free. They host email/password accounts, support multiple authentication providers like FB and Google, and provide a unique user ID per user. Authentication [can be done client-side](https://auth0.com/docs/quickstart/spa/react/01-login) using their login UI component. The client will then send an [access_token JWT](https://auth0.com/docs/quickstart/spa/react/03-calling-an-api#send-authenticated-http-requests) to our server when performing requests. This may be vulnerable to replay attacks though.

### Good ideas

- Firebase: basically our ideal solution, but we need to pay a bit. We'll probably be only using [Firestore](https://firebase.google.com/docs/firestore/) and Authentication, which can be cheap/free.
[...]

PR #808, opened 2018-02-10
[WIP] Account and sync frontend backend

taneliang:
This PR is a partial implementation of #796 - it implements the backend for NUSMods syncing on the frontend. It's ready for review as I only need to make some minor changes.

### Implementation

- The sync functionality is implemented using 2 middlewares and a reducer enhancer.
	- The sync middleware fires actions when data is received from Firestore, and also intercepts interesting actions and sends reducer states to the server.
	- The reducer enhancer only applies server-sent changes (passed to it through the action fired by the middleware) to the local reducer.
	- The sync completer middleware performs any post-sync-download actions that have to be performed. This is currently only used to fetch remotely added modules that are not in the module bank.
- Every user has one document where document ID === user ID. This ID check can and should be used to implement access control.
- A user's document is a subset of the Redux store.

#### Configuration

- Firebase configuration should be set *somewhere, TBC*
- To set which reducers are to be synced, change the configuration in `bootstrapping/configure-sync.js`. 
- Reducers to be synced should also export a `syncConfig` object that specifies the actions to trigger a sync, and the key paths in the reducer to be synced.
- Sync completers should be added to `middlewares/sync-completer.js`.

### Misc notes

- NUSMods should work normally while signed out, so we can deploy this to production if we want to.
- The bundle size is now uuge.
- The UI implemented is minimal, as it's outside the scope of this PR. I've put a temporary sign in button it the footer so that we can test this more easily, but it should be removed before merging.
- When a user logs in for the first time, the server has no data for the user and does not overwrite any of the user's data. Data is sent only when a reducer's data is changed. On subsequent logins, the local reducer states which have data in Firestore will be nuked in favor of the server's.

### TODO (in this PR)

- [ ] Use `process.env` variables to configure Firebase, so that we can have separate dev and production Firebase projects.
- [ ] Fix occasional error thrown when logging in
- [x] Fix bug where adding a module in one browser session causes all the other connected browsers to show the spinner. This happens only if the added module is not in the module bank.
- [x] Settle login merge conflicts. 
- [x] Store last used NUSMods version in document.
- [ ] Offline support is currently disabled by default. We could enable it and test and fix any bugs
- [ ] Before merging, remove the temporary sign in button.

### TODO (after this PR)

- On Firebase, we need to add the security rules to validate and restrict access to the documents. I've already tried implementing the access restrictions and it seems to be working splendidly.
- Figure out how to display sync status, login status, etc.
- Implement a UI for login/logout and account management. Turns out the login component provided by `firebaseui` may not be the best for this, so we may have to implement our own in the end. We'll see.
- Replace timetable spinner with module card spinners so that the page doesn't flash when another device adds a module that's not in the local moduleBank.


ZhangYiJiang:
As a proof of concept this is interesting. It's too rough to deploy though

- Deps are severely out of date, and it seems currently it is recommended to use `firebase` as the dependency, instead of installing each of the subpackages individually 
- Possibly related to the above, Netlify builds are crashing (it works locally and on CI though) 
- The way this is written makes it pretty much impossible to code split. I don't even know if you *can* hot reload Redux middleware, and I definitely don't think it's a good idea. 
- The file sizes may be acceptable if everything else works. Right now though it more than doubled the vendor bundle size. [Boing!](https://www.youtube.com/watch?v=5RIOyCFmbUE) 

There's also a bunch of data safety concerns 

- Academic year is not taken into account. Last year's data will overwrite this year's data 
- Data is unconditionally synchronized. Logging in will wipe out whatever data exists locally. 
- How is split state / offline handled? ie. I make changes on two different devices, one of which is offline. What happens when it comes back online again? 
- I don't know if I can trust Firebase security. It let me use "password" as a password, and there's now no way to change this password 

Also 

- Use `text-transform` to capitalize text

ravern:

PR #837, opened 2018-02-17
Upgrade nusmoderator deps and update npm package info

taneliang:
Upgraded all nusmoderator dependencies. All tests are passing. Strange that the `yarn.lock` for this repo has ballooned too.

Also updated information in `package.json` to point to this repo instead of the old, archived one.

If after upgrading, `yarn build` prints "Error: Plugin 'jsx' not found", just `rm -rf node_modules` and `yarn build` should be working again. More details at https://github.com/developit/microbundle/issues/74.

After this PR is merged we can enable Dependabot on this repo.


taneliang:
@ZhangYiJiang Thanks! Build size actually decreased. Interesting.

| File         | Old build size | New build size |
|--------------|----------------|----------------|
| index.js     | 860B           | 851B           |
| index.es.js  | 1.23kB         | 849B           |
| index.umd.js | 919B           | 919B           |

taneliang:
Ah `index.es.js` shrunk by so much because it's now minified. The other 2 files were minified all along.

PR #843, opened 2018-02-18
Add failsafe for toast disappearing

### LINKED ISSUE #723: Toast disappears after spamming a keyboard shortcut

Bug

**What is the current behavior?**

Pressing on a keyboard shortcut shows a toast message at the bottom.

Keep spamming the keyboard shortcut and the toast eventually disappears, never to be seen again with further triggering of the shortcut.

I recommend just replacing the text if a toast is already shown, rather than hiding and showing again.

The toast should still appear.

**Further information**

- Which version of NUSMods are you currently using? 20180115-357758b
- Did this work in NUSMods BEFORE the R release/redesign? NA
- Did this work in NUSMods AFTER the R release/redesign? NA
- Which browser and version are you using? Chrome 63.0.3239.132
- Which device are you using? E.g. MacBook Pro 2017
- Which operating system and version are you on? macOS 10.13.2

Rinder5:
kinda fixes #723 

Issue seems to be that `transitioning` isn't turning to false sometimes, so the loop gets stuck in the middle of a transition and nothing can change after that.

I think it has something to do with `onTransitionEnd` potentially not running, but running `onTransitionCancel` didn't seem to fix it for me, so I put in a timer to automatically turn `transitioning` false.

Didn't fix the overscrolling issue, as I don't have a mac to test it on.


taneliang:
Thanks for the PR! I don't think this completely fixes the bug though - if you press and hold a key (e.g. the Z key) for about 1 sec, the toast will stop getting displayed for a while

www/src/js/views/components/Notification.jsx
```diff
@@ -82,13 +83,21 @@ export class NotificationComponent extends Component<Props, State> {
         this.closeSnackbar();
       } else if (!this.transitioning) {
         this.openSnackbar();
+      } else {
+        // onTransitionEnd can be cancelled, causing transitioning to never
+        // turn false and notifications to stop showing up, so we set a timer
+        // and turn transitioning false when the timer is up
+        this.timeoutId2 = setTimeout(() => this.clearTransition(), DEFAULT_TIMEOUT);
```
ZhangYiJiang: I don't think this timeout value is valid. Also, this may start more than one timers while you're only cancelling the most recent one in `onTransitionEnd`.

Rinder5: actually, I think setting a ontransitioncancel event might work. I think my problem earlier was that I didn't know how to call it properly, so it was being ignored

Rinder5: well, I actually still am not sure how to add the listener properly

Rinder5: basically, I think all that needs to be done is add a 'onTransitionCancel' event which points to the 'onTransitionEnd' method, however I am still trying to add the listener to test the theory.

ZhangYiJiang: Ah I see. Can't you add it here together with the transitionEnd event listener? https://github.com/Rinder5/nusmods/blob/6f021f369cc975c1d59297ff6d01926848a23807/www/src/js/views/components/Notification.jsx#L147

Rinder5: When i do that, I get "Warning: Unknown event handler property `onTransitionCancel`. It will be ignored." in console.

Rinder5: When I do that, I get "Warning: Unknown event handler property `onTransitionCancel`. It will be ignored." in console.

ZhangYiJiang: We may still need a `setTimeout` fallback though - MDN isn't very clear about support for `ontransitioncancel` - unlike `ontransitionend` it may not be supported by every browser that supports CSS transitions

Rinder5:
I made the changes which were commented on.
Didn't end up using 'ontransitioncancel' because I couldn't get it to work anywhere or in any format lol.

Rinder5:
According to https://developer.mozilla.org/en-US/docs/Web/Events/transitioncancel, only firefox supports transitioncancel as of last year.

taneliang:
CI is failing. Run `yarn lint` and `yarn test` before pushing, they'll let you catch these errors quicker :)

Rinder5:
my bad, was running 'yarn test' but was only relying on the warnings from 'yarn start' for linting. fixed! =)

taneliang:
Yeah lol sometimes I get surprised when `yarn lint` throws me a bunch of errors when `yarn start` didn't. Probably something to do with the caching

PR #918, opened 2018-03-05
Switch data server port to 3600 to match server config

taneliang:
The previous default port 3000 is taken by NUSMods Launchpad. The server has been set up to serve GraphQL data from port 3600, and this PR changes its config in the repo to match.


ZhangYiJiang:
Should probably document somewhere all of the ports used by these projects. From memory, export uses 3300 for production and 3301 for staging.

PR #926, opened 2018-03-08
Upgrade scraper bundler to Webpack 4 and other dev deps

taneliang:
- Upgrade Webpack to Webpack 4.1.1
- Update webpack config:
	- Add `mode`. Should always be "production" since we're asserting `NODE_ENV === "production"`.
	- Remove `babili-webpack-plugin` since Webpack 4 minifies everything when `mode === "production"`
	- Remove ModuleConcatenationPlugin since Webpack 4 does this optimization by default when `mode === "production"`
- Upgrade `babel` and `babel-*`. `babel-loader` >= v7.1.3 needed for Webpack 4 support.
- Upgrade `jest`
- Upgrade `eslint` and `eslint-*`
- Upgrade `prettier` and ran it on everything (no changes).

`yarn ci` passes. `yarn build:scraper`, `yarn scrape` and `yarn scrape:dev` all work when dead NUS APIs are disabled like in #917.

The compiled `gulpfile.compiled.js` is about 1.5 kB smaller than before (38.8 kB -> 37.3 kB).


ZhangYiJiang:
Not really sure why the scraper code was being minified in the first place (Webpack does need a minifier for tree shaking, but I'm not sure if the benefits are substantial on server-side). Probably no harm in leaving it since it's default in Webpack 4.

PR #936, opened 2018-03-11
Add CI config for data server, nusmoderator and scrapers

taneliang:
We're currently only running CI on /www. This PR adds jobs to lint/test/build /api/data, /packages/nusmoderator and /scrapers/nus, and a workflow to run them all. I wanted to add /export as well but it has no lint, test and build scripts so I just left it out.

We need to find a way to only build the changed packages so that we don't spend too much time building packages that weren't changed. That's not a big problem now since most PRs change /www and that requires the most time to build.

### TODO just before merging

- [x] [Change](https://github.com/nusmodifications/nusmods/settings/branches/master) required status checks for `master`. The `ci/circleci` check will never pass according to [the docs](https://circleci.com/docs/2.0/workflows/#workflows-waiting-for-status-in-github).


taneliang:
Looks like Buildsize doesn't wait for all the tests to pass before declaring that the project has shrunk by >10000%. I've increased its comment threshold to 4MB so it should be silent now. We might want to consider switching to another bot - https://github.com/siddharthkp/bundlesize looks pretty good

taneliang:
@li-kai All 4 jobs run in parallel after the `checkout` job, but the containers are still at 1x parallelism so the builds still take the same amount of time.

PR #948, opened 2018-03-19
Add meta description tags to headers on module and venue pages

ZhangYiJiang:
Since Google can scrape NUSMods even though it is a SPA, I think we can improve our search result listing by using module description. 

screenshot from 2018-03-19 16-01-19

Currently it uses the generic description text that is the same for all pages, but that's not very good.


www/src/js/views/components/Title.jsx
```diff
@@ -6,14 +6,25 @@ import config from 'config';
...
 };
 
-export default function Title(props: Props) {
+function Title(props: Props) {
   // We use defer=false to allow Google Analytics autotrack to send the correct
   // page title. See bootstrapping/google-analytics.js
   return (
     <Helmet titleTemplate={`%s - ${config.brandName}`} defer={false}>
       <title>{props.children}</title>
+      <meta property="description" content={props.children} />
```
taneliang: Should `props.children` be `props.description` instead?

ZhangYiJiang: Yes, I have no idea why I wrote that :sweat_smile:

PR #949, opened 2018-03-19
Fix duplicate React Helmet meta tags

ZhangYiJiang:
In #948 I used React Helmet to add `<meta>` description tags to the pages, but didn't notice that Helmet duplicated the tag when it inserted its own set of tags. According to https://github.com/nfl/react-helmet/issues/341 you can work around this by adding the `data-react-helmet="true"` attribute. This will cause React Helmet to recognize the meta tag and replace it. A bit of a hack, but short of an official solution this seems to work fine. Having two sets of tags isn't a dealbreaker anyway - it's just slightly ugly and we don't really know the SEO effect of it.

On the bright side, this means we can now legitimately claim our app is "server-side rendered" :wink:

### DIFF
--- a/www/src/index.html
+++ b/www/src/index.html
@@ -10,21 +10,21 @@
     <link href="${htmlWebpackPlugin.files.publicPath}manifest.json" rel="manifest">
     <link href="${htmlWebpackPlugin.options.moduleListUrl}" rel="preload" as="fetch">
     <link href="${htmlWebpackPlugin.options.venuesUrl}" rel="prefetch">
-    <title>NUSMods</title>
-    <meta name="description" content="NUSMods is a timetable builder and knowledge platform, providing students with a better way to plan their school timetable and useful module-related information that are community-driven.">
+    <title>${htmlWebpackPlugin.options.brandName}</title>
+    <meta name="description" content="${htmlWebpackPlugin.options.description}">
     <!-- Open Graph general (Facebook, Pinterest & Google+) -->
-    <meta name="og:title" content="NUSMods">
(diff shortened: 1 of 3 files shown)


taneliang:
Should we add "NUSMods" to the `og:title` prop, like we do to the page `title`? I'm not sure how different services handle `og:title`, but if it's used in place of `title` I think it'll look strange.

Also, can confirm that the meta tags get overridden.

screenshot_20180319_214936

ZhangYiJiang:
http://ogp.me/#metadata says specifically not to include brand in og:title

electic:
This does not seem to work in the latest version. For example if I put 

```<link rel="canonical" href="https://test.com" />

in my index.html file and try to override it in the base pages, I end up getting a duplicate. This will cause Google to ignore the page entirely since the link is invalid.

**Comment by ZhangYiJiang** [comment]
@electic You need the `data-react-helmet` attribute as mentioned in the issue and pull request.

**Comment by petervavro** [comment]
Does this solution has a side effect on page score ?

**Comment by bdombro** [comment]
I added the following code to my app to manually remove original header elements:

```
let originalDescriptionElements = document.querySelectorAll('meta[name=description]');
    originalDescriptionElements.forEach(e => {
      // @ts-ignore: ignore html.dataset being unknown
      if (!e.dataset) throw new Error("Helmet: html.dataset is not available in this browser.");
      // @ts-ignore: ignore html.dataset being unknown
      else if (!e.dataset.reactHelmet)
        e.parentNode.removeChild(e);
    });
```

ZhangYiJiang:
This is not the right place to ask for help with react-helmet. If you have a question, ask it on Stack Overflow or read their documentation. I'm not sure why this PR has so much SEO...

PR #950, opened 2018-03-19
Show number of Disqus comments in module info page side menu

### LINKED ISSUE #683: Show number of Disqus comments in module info page side menu
From user feedback 

> Could we have the number of reviews shown with the Reviews tab? It could save one click to the reviews section if no one comments on the module!

This should be relatively simple - create a new `<CommentCount>` component (reference https://github.com/ryanvalentin/disqus-react/blob/master/src/CommentCount.jsx), and with the common code from `<DisqusComment>` extracted out that would display the number of comments. 

The side menu items (https://github.com/nusmodifications/nusmods/blob/master/www/src/js/views/modules/ModulePageContent.jsx#L268) will probably need to be converted back to static items from a for-loop for this to work.

wenmogu:
Fixes #683 

**Outline of solution**
- copied here and there from the [example given](https://github.com/disqus/disqus-react/blob/master/src/CommentCount.jsx) and the `DisqusComments.jsx`
- added style file for that count comments icon at side medu
- written test for an extracted function used both in the `CommentCount.jsx` and `DisqusComments.jsx` (not sure if it is necessary)
- retained the `map` function which produces the side-menu at `ModulePageContents.jsx` by adding a if-else condition to add the count comments icon
And now it looks like this:
image

**Area in which i need help**
- currently there is a visible lag in getting the number of review comments at the side menu (i.e. get the number on page load). I need help on how to eliminate that lag ><


www/src/js/views/components/CommentCount.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+type Props = {
+  children: Node,
+  url: string,
+  identifier: string,
+  title: string,
+  commentIcon: Node,
+};
+
+type Option = {
```
ZhangYiJiang: This is probably unnecessary - just inline the options object

www/src/js/views/components/CommentCount.jsx
```diff
@@ -0,0 +1,70 @@
...
+  reset: boolean,
+};
+
+const SCRIPT_ID = 'dsq-count-scr';
+
+export const MESSAGE_SQUARE_LABEL = 'Open menu';
```
ZhangYiJiang: As above - just inline this. Also, I don't think the text 'Open menu' is correct?

www/src/js/views/components/CommentCount.jsx
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      insertScript(`https://${config.disqusShortname}.disqus.com/count.js`, SCRIPT_ID, true);
+    }
+  }
+
+  option: Option = {
+    reset: true,
+  };
+
+  queueResetCount() {
```
ZhangYiJiang: Also inline this

www/src/js/views/components/CommentCount.scss
```diff
@@ -0,0 +1,11 @@
+.comment {
+  padding-left: 1rem;
+}
+
+.icon {
+  svg {
+    width: 12px;
```
ZhangYiJiang: Use `rem` instead of px - `0.75rem` should be equivalent (12/16 = 0.75)

www/src/js/views/components/CommentCount.jsx
```diff
@@ -0,0 +1,70 @@
+// @flow
+import React, { PureComponent } from 'react';
+import config from 'config';
+import insertScript from 'utils/insertScript';
+import { MessageSquare } from 'views/components/icons';
+import styles from './CommentCount.scss';
+
+type Props = {
+  children: Node,
+  url: string,
```
ZhangYiJiang: Not terribly important, but url, identifier and title can be extracted

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -261,13 +262,34 @@ export class ModulePageContentComponent extends Component<Props, State> {
...
                 >
-                  {map(SIDE_MENU_LABELS, (label, key) => (
-                    <li key={label}>
-                      <a onClick={() => this.toggleMenu(false)} href={`#${SIDE_MENU_ITEMS[key]}`}>
-                        {label}
-                      </a>
-                    </li>
-                  ))}
+                  {map(SIDE_MENU_LABELS, (label, key) => {
+                    if (label.valueOf() === 'Reviews') {
```
ZhangYiJiang: This `if` should be inside the `<a>` so that the code outside is not duplicated (use a `&&`). Also I don't know why `.valueOf()` is needed

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -261,13 +262,34 @@ export class ModulePageContentComponent extends Component<Props, State> {
...
+                            onClick={() => this.toggleMenu(false)}
+                            href={`#${SIDE_MENU_ITEMS[key]}`}
+                          >
+                            {label}
+                            <CommentCount
+                              url={`https://nusmods.com/modules/${ModuleCode}/reviews`}
+                              identifier={ModuleCode}
+                              title={pageTitle}
+                            >
+                              -1
```
ZhangYiJiang: Don't think this placeholder is very useful. Just leave it empty.

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -261,13 +262,34 @@ export class ModulePageContentComponent extends Component<Props, State> {
...
+                    if (label.valueOf() === 'Reviews') {
+                      return (
+                        <li key={label}>
+                          <a
+                            onClick={() => this.toggleMenu(false)}
+                            href={`#${SIDE_MENU_ITEMS[key]}`}
+                          >
+                            {label}
+                            <CommentCount
+                              url={`https://nusmods.com/modules/${ModuleCode}/reviews`}
```
ZhangYiJiang: These three props can be deduplicated with the `<DisqusComments>` below

www/src/js/views/components/CommentCount.jsx
wenmogu: errr hi can i ask what do u mean by "inline"

ZhangYiJiang: Inline is the opposite of extract - put the content of the variable where they are used so we declare one less variable. In this case a lot of the variables declared don't seem to be reused anywhere else and are not likely to change, so the code can be made simpler by not creating another variable.

wenmogu: emmm I think I get what u mean. Thx!

www/src/js/views/modules/ModulePageContent.jsx
wenmogu: Emmm do u mean by extracting them into variables and reuse them at `<DisqusComments>`?

ZhangYiJiang: Yes. You can save the properties as an object and use object spread to apply them to both components

wenmogu: Ok thx! I hope the update code is implementing what you are talking about

ZhangYiJiang:
Looks good, should be mergable on CI green. Will probably get someone else to take a second look on this since I don't trust my code at 2am in the morning.

wenmogu:
Thx for helping so much 🙏

taneliang:
@ZhangYiJiang :laughing: 

image

Cause: https://github.com/wenmogu/nusmods/commit/3698e9c8423fb2dc07a1e76b5c40b81c47fb32a5#diff-7bc976c46d86acc14c2bde248f73a140R271

www/src/js/views/components/disqus/CommentCount.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+const SCRIPT_ID = 'dsq-count-scr';
+
+export default class CommentCount extends PureComponent<Props> {
+  static defaultProps = {
+    commentIcon: <MessageSquare aria-label="Comment count" />,
+  };
+
+  static loadInstance() {
+    if (window.document.getElementById('dsq-count-scr')) {
```
taneliang: I think you meant to use `SCRIPT_ID` here instead of this string literal

www/src/js/views/components/disqus/CommentCount.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = DisqusConfig & {
+  commentIcon: Node,
+};
+
+const SCRIPT_ID = 'dsq-count-scr';
+
+export default class CommentCount extends PureComponent<Props> {
+  static defaultProps = {
+    commentIcon: <MessageSquare aria-label="Comment count" />,
```
taneliang: Hmm I don't think we need this to be a prop - should just put it in `render`.

www/src/js/views/components/disqus/CommentCount.jsx
```diff
@@ -0,0 +1,55 @@
...
+  componentDidUpdate() {
+    CommentCount.loadInstance();
+  }
+
+  render() {
+    const { identifier, url, commentIcon } = this.props;
+
+    return (
+      <span className={styles.comment}>
+        <span className={styles.icon}> {commentIcon} </span>
```
taneliang: /nit Should remove the spaces before and after `{commentIcon}`

www/src/js/views/modules/ModulePageContent.jsx
```diff
@@ -265,6 +268,7 @@ export class ModulePageContentComponent extends Component<Props, State> {
                     <li key={label}>
                       <a onClick={() => this.toggleMenu(false)} href={`#${SIDE_MENU_ITEMS[key]}`}>
                         {label}
+                        {'Reviews' && <CommentCount {...disqusConfig} />}
```
taneliang: As mentioned in the earlier comment, this needs to be fixed haha

ZhangYiJiang: This is why I shouldn't code at 2am

ZhangYiJiang:
Don't think it's possible for the icon to not be shown, but I'm not very familiar with the Disqus embed API.

www/src/js/views/components/disqus/CommentCount.jsx
ZhangYiJiang: I think this is fine because that's where a space would naturally appear.

wenmogu:
I just pushed an alternative to adding a default prop of icon to `CommentCount`. 
As for loading the icon together with the number, I only managed to find one place that achieves it, which is disqus' own website ([something like this](https://disqus.com/home/forum/celebritytoob/)) and I have no idea how they do it :/

PR #952, opened 2018-03-25
Implement dbfy script in data server

taneliang:
This PR implements a new `dbfy` script for the data server, which populates the SQLite database with existing JSON data (i.e. dbfying our JSON data).

It is meant to help us move quicker to GraphQL. While replacing the scrapers with new ones that save data directly into a database is a good idea, I think it is taking too long and will possibly introduce more bugs. And that's holding up our adoption of GraphQL.

These temporary scripts will help us transition to a db, allowing work on the GraphQL server and the client to proceed while we decide on what to do with the scrapers. As such, the dbfy code is intended to be simple and standalone, and can be easily nuked in favor of better approaches when the time comes.

## Work done

- Add new tables to database: `modules`, `classes`, and `corsBiddingStats`.
- Implement JS scripts in /api/data/src/dbfy which clears and then populates a database with venues and module data from the existing JSON data files.
- Add depth param to `walkJsonDir` so that we can work with per-sem data for every acad year. This function still loads all the JSON data into RAM at once.
- Implement unused `walkJsonDirCallback` function, which executes a callback for every file. Intended to be a memory-efficient alternative to `walkJsonDir`. Not used because we actually want to group all data for a sem into one big object before storing it, but this function will execute the callback for every *mod's* data.
- Rename fields in `schools` table - `abbreviation` was a little long.
- Fix some typos in uni seed names.
- Replace Jest test names with function names where possible.

## TODO

- [x] Add actual `dbfy` script that doesn't watch code for changes (as `dbfy:watch` does)
- [x] Update READMEs to include docs for `dbfy` as well as `db:init` script.

## Next steps

- Deployment: Hope that loading all the venue and module data into RAM at once won't fill the server's RAM completely :pray:. If it does, we should replace `walkJsonDir` with `walkJsonDirCallback`.
- Replace in-memory data with database in data server
- Add queries for venues to data server

### DIFF
--- a/api/data/migrations/20170101000000_initial.js
+++ b/api/data/migrations/20170101000000_initial.js
@@ -45,7 +45,106 @@ exports.up = (knex, Promise) => {
     table.unique(['school_id', 'name']);
   });
 
-  return Promise.all([schoolsTable, departmentsTable, venuesTable]);
+  const modulesTable = knex.schema.createTable('modules', (table) => {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
+    table
+      .integer('school_id')
+      .notNullable()
+      .references('id')
+      .inTable('schools')
+      .onDelete('CASCADE')
+      .onUpdate('CASCADE');
+    table.string('acadYear').notNullable();
+    table.integer('semester').notNullable();
+    table.string('code').notNullable();
+    table.string('title').notNullable();
+    table.string('description');
+    table.string('credit').notNullable();
+    table.integer('workload');
+    table.dateTime('examDate');
+    table.string('department');
+    table.string('lecturePeriods');
+    table.string('tutorialPeriods');
+    table.string('types');
+    table.string('prerequisite');
+    table.string('preclusion');
+    table.string('corequisite');
+    table.unique(['school_id', 'code', 'acadYear', 'semester']);
+  });
+
+  const classTable = knex.schema.createTable('classes', (table) => {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
+    table
+      .integer('module_id')
+      .notNullable()
+      .references('id')
+      .inTable('modules')
+      .onDelete('CASCADE')
+      .onUpdate('CASCADE');
+    table
+      .integer('venue_id')
+      .notNullable()
+      .references('id')
+      .inTable('venues')
+      .onDelete('CASCADE')
+      .onUpdate('CASCADE');
+    table.string('classNo').notNullable();
+    table.string('type').notNullable();
+    table.string('weekFrequency').notNullable();
+    table.string('day').notNullable();
+    table.integer('startTime').notNullable();
+    table.integer('endTime').notNullable();
+    table.unique(['module_id', 'classNo', 'type', 'day', 'startTime', 'endTime']);
+  });
+
+  const corsBiddingStatsTable = knex.schema.createTable('corsBiddingStats', (table) => {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
+    table
+      .integer('module_id')
+      .notNullable()
+      .references('id')
+      .inTable('modules')
+      .onDelete('CASCADE')
+      .onUpdate('CASCADE');
+    table.string('acadYear').notNullable();
+    table.integer('semester').notNullable();
+    table.string('round').notNullable();
+    table.string('group').notNullable();
+    table.string('faculty').notNullable();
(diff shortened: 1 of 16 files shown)


api/data/src/util/walkDir.js
```diff
@@ -1,35 +1,80 @@
...
+
+/**
+ * Sync version of walkJsonDir.
+ * Only traverses one layer.
  * @example walkJsonDir('test', 'test.json') will find 'test/1/test.json'
  * @param {string} folderPath folder path to traverse
  * @param {string} destFileName file name to match
  */
-function walkJsonDirSync(folderPath, destFileName) {
+export function walkJsonDirSync(folderPath: string, destFileName: string) {
```
ZhangYiJiang: I wonder if we can't just use `glob` (which should probably come with our existing deps) or some other libs

api/data/migrations/20170101000000_initial.js
```diff
@@ -45,11 +45,108 @@ exports.up = (knex, Promise) => {
...
+      .notNullable()
+      .primary();
+    table
+      .integer('school_id')
+      .notNullable()
+      .references('id')
+      .inTable('schools')
+      .onDelete('CASCADE')
+      .onUpdate('CASCADE');
+    table.string('acadYear').notNullable();
```
ZhangYiJiang: Should set the length parameter, otherwise Knex will always use VARCHAR(255), which is suboptimal for most of these

api/data/src/dbfy/index.js
```diff
@@ -0,0 +1,24 @@
...
+  // Clear tables
+  const tables = ['venues', 'modules', 'classes', 'corsBiddingStats'];
+  await Promise.map(tables, async (table) =>
+    db
+      .select()
+      .table(table)
+      .del(),
+  );
+
+  await populateVenues(db);
```
ZhangYiJiang: Can these two tasks be run in parallel?

api/data/src/dbfy/modules.js
```diff
@@ -0,0 +1,169 @@
...
+  const venueIdMap: IdMap = R.map(R.prop('id'), R.indexBy(R.prop('name'), venueRows));
+
+  await Promise.all(
+    map(rawData, async (ayData, acadYear) => {
+      // Only read data from valid acadYears
+      if (!/^[0-9]{4}-[0-9]{4}$/.test(acadYear)) return;
+      await Promise.all(
+        map(ayData, async (semData, sem) => {
+          // Only read data from valid sems
+          if (isNaN(sem)) return; // eslint-disable-line no-restricted-globals
```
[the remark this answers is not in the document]
taneliang: `Number.isNaN` returns false for text. Is there a better way to do this? I think this is simple enough and works.

```js
console.log(isNaN('notasem')) // Returns true, which we want
console.log(Number.isNaN('notasem')) // Returns false
```

[Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN):

> In comparison to the global isNaN() function, Number.isNaN() doesn't suffer the problem of forcefully converting the parameter to a number. This means it is now safe to pass values that would normally convert to NaN, but aren't actually the same value as NaN. This also means that only values of the type number, that are also NaN, return true.

api/data/src/dbfy/index.js
taneliang: Not in this implementation. `populateModules` fetches all the venues in order to reference the venue IDs.

api/data/src/util/walkDir.js
```diff
@@ -1,35 +1,80 @@
...
-async function walkJsonDir(folderPath, destFileName) {
-  const folders = await fs.readdir(folderPath);
-  const folderToJsonMap = {};
-  await Promise.all(
+export async function walkJsonDirCallback(
+  folderPath: string,
+  destFileName: string,
+  depth: number,
+  shouldReadFile: (path: string) => boolean,
+  dataCallback: (data: Object, filePath: string) => void,
```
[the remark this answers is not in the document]
taneliang: Hmm I'm not sure how since we don't know the shape of the data that's read.

Anyway all the Flow annotations are a bit of a scam. This project doesn't have `flow-bin` as a dependency so the types aren't being checked.

li-kai (commented):
Hoo boy. Somehow I completely missed this PR many times. This is **really useful** now that we finally got #1482 and #1578 going.

ravern:

PR #997, opened 2018-04-03
Upgrade www dev dep

ZhangYiJiang:
This just merges all of the www dev dependency Dependabot PRs.


taneliang:
We aren't updating stylelint-webpack-plugin to v0.10.4? It's been updated with Webpack 4 support and it's should still support Webpack 3

ZhangYiJiang:
@taneliang Missed that one for some reason. I should have all of them now.

PR #1007, opened 2018-04-29
Adopt Yarn Workspaces

taneliang:
This PR configures Yarn workspaces on this repo.

Dependencies shared among the different projects are now hoisted to a root node_modules directory, saving some disk space. In our case, this only decreases the repo size by 200MB (about 15%) so... the savings weren't as dramatic as I was expecting. On the bright side, the combined, repo-level yarn.lock is much smaller than the multiple project specific yarn.locks.

One other benefit is that we can run `yarn` once in the repo root and have the dependencies for all packages installed. Not sure if running `yarn` in the respective project roots will install other projects' dependencies.

Prettier was also upgraded because Yarn was resolving to this newer version which was causing new linting errors.


ZhangYiJiang:
Weird, the a11y eslint plugin is flagging every single `<a>` as having no content?

www/webpack/webpack.parts.js
```diff
@@ -20,6 +20,7 @@ const PATHS = {
   // Using an absolute path will cause transient dependencies to be resolved to OUR
   // version of the same module, so this is kept relative
   node: 'node_modules',
+  nodeHoisted: path.join(ROOT, '..', 'node_modules'),
```
taneliang: Should be changed to `path.join(__dirname, 'node_modules')` since yarn always hoists shared deps to the repo root.

li-kai:
If cotton/renovate supports monorepos this solves a lot of the initial problem of consistently updating all the package versions. But I also held this off because I didn't know if we wanted to run commands in the sub repos or at the repo root. 

### Repo root
Everything takes longer to run but one place to keep track of.

### Sub repos:
Deep nested paths to run npm command.
Jest can't give a complete code coverage %.

taneliang:
Since we control Cotton we can make it support Yarn workspaces. Shouldn't take too much effort to support this.

I think we can leave the commands in the subrepos, since the individual subrepos have very different scripts.

To be honest I don't think migrating to Yarn workspaces has that many benefits for us because the projects in this repo are so different. It may make more sense to retain our current setup.

li-kai:
After significant thoughts about this, I don't think this is going to work. 
Dockerfiles will require the yarn lock files to be in their own directory unless we either copy the entire repo, or drop usage of yarn.lock.

@taneliang Do you have alternative solutions to the docker problem? If not, perhaps it might be best to close this PR.

taneliang:
@li-kai Why should we not copy the entire repo? It doesn't feel particularly clean but it may be fine? But then again I don't really have any strong opinions on this or of adopting Yarn workspaces, so I'll be happy to close this PR.

yangshun:
Moved to pnpm, closing this.

PR #1009, opened 2018-05-07
Cleanup issue templates

yangshun:
Now that I have a better idea of what the templates do, here's some cleaning up. We might want to consider adding a PR template as well, although they won't be as usual because external PRs are rare!


yangshun:
@taneliang Good point. Have updated CONTRIBUTING

PR #1011, opened 2018-05-17
Fix 6.30pm exam timing

### LINKED ISSUE #1005: Error: Unrecognized exam time: 6:30 PM
https://sentry.io/nusmods/v3/issues/524507037/

So, erm, how did these modules get an exam time of 6:30 PM? 

```
Error: Unrecognized exam time: 6:30 PM
  at getTimeSegment (./js/views/timetable/ExamCalendar.jsx:57:13)
  at None (./js/views/timetable/ExamCalendar.jsx:195:22)
  at Array.forEach (<anonymous>)
  at modulesWithExamDate (./js/views/timetable/ExamCalendar.jsx:183:30)
  at render (./js/views/timetable/ExamCalendar.jsx:216:35)
...
(11 additional frame(s) were not displayed)
```

ZhangYiJiang:
Fixes #1005 

Adds 6.30pm as an exam timing. This does mean that if a student adds two conflicting modules with exam at 5pm and another at 6.30pm the label would be inaccurate, but that's very unlikely. As far as I can tell it exists only for some level 5000 modules, so this is very unlikely to affect actual users.

### DIFF
--- a/www/src/js/views/timetable/ExamCalendar.jsx
+++ b/www/src/js/views/timetable/ExamCalendar.jsx
@@ -52,6 +52,7 @@ function getTimeSegment(time: string): TimeSegment {
     case '2:30 PM':
       return 'Afternoon';
     case '5:00 PM':
+    case '6:30 PM':
       return 'Evening';
     default:
       throw new Error(`Unrecognized exam time: ${time}`);


yangshun:
What a weird case 😂

ZhangYiJiang:
Yeah, I wanted to see if I can make the code more general, but it doesn't seem worth it

PR #1016, opened 2018-05-19
Fix issue template line ending and heading 

ZhangYiJiang:
- Switch line ending to use LF instead of CRLF 
- Switch from inline bold to using headings. This is more semantically correct and probably looks better

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -12,8 +12,8 @@ A clear and concise description of what the bug is.
 Steps to reproduce the behavior:
 
 1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
+2. Click on '...'
+3. Scroll down to '...'
 4. See error
 
 ### Expected behavior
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/Feature_request.md
```diff
@@ -1,20 +1,20 @@
...
-
-Add any other context or screenshots about the feature request here.
+---
+name: Feature Request
+about: Suggest an idea for this project
+---
+
+### Is your feature request related to a problem? Please describe.
+
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
taneliang: /nit could we change the "Ex." to "E.g."? Somehow that got reverted in the last PR

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -1,42 +1,42 @@
...
+
+A clear and concise description of what the bug is.
+
+### To Reproduce
+
+Steps to reproduce the behavior:
+
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
```
taneliang: /nit Could you change these 4-dot ellipses to 3 dot ones?

PR #1034, opened 2018-07-04
Fix typo in announcement

ZhangYiJiang:
Always remember to proofread. We aren't in the future (yet)


yangshun:
No information for AY19/10 is not wrong...

PR #1036, opened 2018-07-05
Archive old timetables with new academic year 

ZhangYiJiang:
This PR adds a new `archive` key to the `timetables` state and clears the `lessons` state if the new AY does not match the old persisted one. This will prevent the student's old timetable from being used in the new AY, while keeping the old timetable in case we want to develop new features for it in the future. 

The code is a bit rough around the ages. For how rehydration and migration work, refer to https://github.com/rt2zz/redux-persist/#state-reconciler and https://github.com/rt2zz/redux-persist/blob/master/docs/migrations.md

### DIFF
--- a/www/src/js/reducers/timetables.js
+++ b/www/src/js/reducers/timetables.js
@@ -20,8 +20,22 @@ import {
 } from 'actions/timetables';
 import { SET_EXPORTED_DATA } from 'actions/export';
 import { getNewColor } from 'utils/colors';
+import { createMigrate, REHYDRATE } from 'redux-persist';
 
 const EMPTY_OBJECT = {};
+
+export const PERSIST_KEY = 'timetables';
+export const persistConfig = {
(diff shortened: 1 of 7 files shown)


www/src/js/reducers/timetables.js
```diff
@@ -20,9 +20,23 @@ import {
 } from 'actions/timetables';
 import { SET_EXPORTED_DATA } from 'actions/export';
 import { getNewColor } from 'utils/colors';
+import { createMigrate, REHYDRATE } from 'redux-persist';
 
 const EMPTY_OBJECT = {};
 
+export const PERSIST_KEY = 'timetables';
+export const persistConfig = {
+  /* eslint-disable no-useless-computed-key */
```
taneliang: Can't use `eslint-disable-next-line` instead?

ZhangYiJiang: Not unless you want to add that line for every migration

PR #1040, opened 2018-07-06
Fix lessons sharing hovered state when their lessonTypes are different

### LINKED ISSUE #1038: Lesson type not taken into account when highlighting same ClassNo
peek 2018-07-06 09-52

I believe the bug is caused by that PR, will investigate

ZhangYiJiang:
Fixes #1038 which was caused by changes in #1019. We now differentiate by LessonType too since a module can have lectures and tutorials of the same class no. but are not selected together


ZhangYiJiang:
Created a helper function to extract the HoverLesson object and use `_.isEqual` to do the checking instead

PR #1041, opened 2018-07-06
Add announcement for AY18/19 module data

ZhangYiJiang:
Simple dismissable announcement for new semester data

### DIFF
--- a/www/src/js/views/components/Announcements.jsx
+++ b/www/src/js/views/components/Announcements.jsx
@@ -52,7 +52,7 @@ class Announcements extends PureComponent<Props, State> {
 
         <div className={styles.body}>
           <h3>AY2018/19 modules now available</h3>
-          <p>NUSMods now have AY2018/19 module information available. Happy planning!</p>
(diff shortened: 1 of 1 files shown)


www/src/js/views/components/Announcements.jsx
```diff
@@ -45,24 +47,12 @@ class Announcements extends PureComponent<Props, State> {
...
-              rel="noopener noreferrer"
-            >
-              Check out what&apos;s new
-            </a>, and{' '}
-            <button className="btn btn-inline" type="button" onClick={this.props.toggleFeedback}>
-              tell us what you think
-            </button>.
-          </p>
+          <h3>AY2018/19 modules now available</h3>
+          <p>NUSMods now have AY2018/19 module information available. Happy planning!</p>
```
yangshun: Has or have? 🤔

ZhangYiJiang: Should be "has". I cannot English today

PR #1043, opened 2018-07-06
Fix timetable crashing on invalid timings

ZhangYiJiang:
NUSMods is currently crashing when AY1130 is added to the timetable because it has a lesson that starts at 0800 and ends at 0801 (duration: 1 minute). We fix this temporarily by improving the robustness of the time parsing method, although a long term solution needs to be found for weirdly timed lessons.


ZhangYiJiang:
Sentry issue: [V3-R2](https://sentry.io/nusmods/v3/issues/600869199/)

PR #1047, opened 2018-07-15
Add service worker upgrade notice

ZhangYiJiang:
Mostly following this recipe on Workbox's guide https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offer_a_page_reload_for_users, which itself copies from https://redfin.engineering/how-to-fix-the-refresh-button-when-using-service-workers-a8e27af6df68. The code feels very verbose but the Redfin article mentions this is about as short as you can get. Note that this does not implement point 4 in the article, which I feel is more trouble than its worth. Hopefully when service worker gets more mature this will get better. 

## Changes 

- Split off service worker bootstrapping in to its own file
- Add a PROMPT_REFRESH action and a boolean state into app 
- When the state changes to true, the `<RefreshPrompt>` component will display a alert on the top of the page in the same places as `<Announcement>` asking the user to update the app

### DIFF
--- a/www/src/js/views/components/RefreshPrompt.jsx
+++ b/www/src/js/views/components/RefreshPrompt.jsx
@@ -1,5 +1,4 @@
 // @flow
-
 import React, { PureComponent } from 'react';
 import classnames from 'classnames';
 import { connect, type MapStateToProps } from 'react-redux';
--- a/www/src/js/views/components/RefreshPrompt.jsx
+++ b/www/src/js/views/components/RefreshPrompt.jsx
@@ -14,9 +14,9 @@ type Props = {
 
 class RefreshPrompt extends PureComponent<Props> {
   render() {
-    // if (!this.props.showPrompt) {
-    //   return null;
-    // }
+    if (!this.props.showPrompt) {
+      return null;
+    }
 
     return (
       <div className={classnames('alert alert-success', styles.announcement)}>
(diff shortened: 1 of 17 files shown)


ZhangYiJiang:
Confirmed working with manual testing 

peek 2018-07-15 11-12

www/src/js/views/components/RefreshPrompt.jsx
```diff
@@ -0,0 +1,54 @@
...
+    if (!this.props.showPrompt) {
+      return null;
+    }
+
+    return (
+      <div className={classnames('alert alert-success', styles.announcement)}>
+        <Refresh className={styles.backgroundIcon} />
+
+        <div className={classnames()}>
+          <h3>NUSMods has been updated</h3>
```
taneliang: I have no good solutions for this, but when non-technical people say "NUSMods has been updated", they typically mean that new data is available. So this may end up a bit misleading/confusing

ZhangYiJiang: "A new version of NUSMods is available" - like what WhatsApp Web does?

taneliang: Ah yes that sounds better

www/src/js/views/components/RefreshPrompt.jsx
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    if (!this.props.showPrompt) {
+      return null;
+    }
+
+    return (
+      <div className={classnames('alert alert-success', styles.announcement)}>
+        <Refresh className={styles.backgroundIcon} />
+
+        <div className={classnames()}>
```
li-kai: Nit: empty className

PR #1048, opened 2018-07-15
Stop redux logger diffing large actions

ZhangYiJiang:
On Firefox Redux Logger makes starting the app really slow because it tries to diff actions that insert lots of data into the store. This PR filters those out so that Firefox doesn't lag anymore when opening the app.


ZhangYiJiang:
I'm not sure seeing more than 5000 modules added to the module bank item by item is very useful in any browser

PR #1049, opened 2018-07-15
Integrate Bundlesize

ZhangYiJiang:
Because Buildsize... frankly speaking has issues. 

- Isn't accurate 
- Bot comments too verbose 
- Doesn't account for gzip 

Hopefully this will work better since it's CI integrated rather than living on a different server.


ZhangYiJiang:
I think this looks much better (although having a limit of NaN because we didn't specify the limit is kind of funny) 

In merge check 

screenshot from 2018-07-15 14-31-39

View details 

screenshot from 2018-07-15 14-32-11

li-kai:
300kb sounds reasonable? Once we exceed it means we should code split.

ZhangYiJiang:
I don't feel like setting a limit. Code splitting wouldn't help with the size calculation since I'm grouping all JS file as one tracked "section" (see the changes to package.json). Unless we explicitly name split sections, there isn't a good way to track them (maybe we should set import names?) since the tracking is done via file names. Actually now I'm curious if the compare size against master function will work since our filenames change with every release.

PR #1052, opened 2018-07-22
Update to Webpack 4 

### LINKED ISSUE #874: Webpack 4 plugin support tracker
Webpack 4 has been released, breaking many loaders and plugins. This issue tracks the status of Webpack 4 support in our Webpack-related dependencies.

### /www

| Plugin/Loader                      | Support for Webpack 4                    | Installed in /www                        |
| :--------------------------------- | :--------------------------------------- | ---------------------------------------- |
| autodll-webpack-plugin             | ✘ https://github.com/asfktz/autodll-webpack-plugin/pull/106 | ✘                                        |
| babel-loader                       | ✔ v7.1.3                                 | ✘                                        |
| cache-loader                       | Supported by `next` branch, but looks stale | ✘                                        |
| clean-webpack-plugin               | Unsure. No issues, PRs, or `peerDependencies` package.json entry. |                                          |
| copy-webpack-plugin                | ✔ v4.4.0                                 | [✔](https://github.com/nusmodifications/nusmods/pull/859) |
| css-loader                         | ✔ v0.28.10                               | [✔](https://github.com/nusmodifications/nusmods/pull/860) |
| extract-text-webpack-plugin        | ✘ https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/701. Support added in v4.0.0 but is still in alpha. | ✘                                        |
| file-loader                        | [✔](https://github.com/webpack-contrib/file-loader/commit/3a6a7a191b47d359dd91f1571e45d8e2139e9c95) but not released | ✘                                        |
| flow-status-webpack-plugin         | Unsure. No issues, PRs, or `peerDependencies` package.json entry. |                                          |
| html-webpack-inline-source-plugin  | ✘ Issue reported for errors when used with Webpack 4. Package still has Webpack 1 as a dev dependency. | ✘                                        |
| html-webpack-plugin                | ✔ Support added in this fork: https://github.com/webpack-contrib/html-webpack-plugin. To be installed as `webpack-contrib/html-webpack-plugin` | ✘                                        |
| lodash-webpack-plugin              | Unsure. No issues, loaderPRs, or `peerDependencies` package.json entry. |                                          |
| optimize-css-assets-webpack-plugin | [Unlikely](https://github.com/NMFR/optimize-css-assets-webpack-plugin/issues/34) | ✘                                        |
| postcss-loader                     | Unsure. No issues, PRs, or `peerDependencies` package.json entry. |                                          |
| sass-loader                        | ✘ https://github.com/webpack-contrib/sass-loader/pull/514 | ✘                                        |
| script-ext-html-webpack-plugin     | Unsure. No issues, PRs, or `peerDependencies` package.json entry. |                                          |
| style-loader                       | ✘ Support added in stale PR https://github.com/webpack-contrib/style-loader/pull/282 | ✘                                        |
| stylelint-webpack-plugin           | Unsure. Not in versions listed in `peerDependencies`, but no issues or PRs filed. |                                          |
| uglifyjs-webpack-plugin            | ✔ v1.2.2                                 | [✔](https://github.com/nusmodifications/nusmods/pull/855) |
| url-loader                         | ✔ Likely added in [v1.0.0-beta.0](https://github.com/webpack-contrib/url-loader/releases/tag/v1.0.0-beta.0). Supported according to `peerDependencies`, but that entry was added in December, before Webpack 4 was frozen. | ✘                                        |
| webpack-dev-server                 | ✔ [v3.0.0](https://github.com/webpack/webpack-dev-server/releases/tag/v3.0.0). Supports *only* Webpack 4. | ✘ Must be upgraded with Webpack          |
| webpack-merge                      | Unsure. No issues, PRs, or `peerDependencies` package.json entry. |                                          |
| workbox-webpack-plugin             | ✘ Support [added](https://github.com/GoogleChrome/workbox/pull/1275) after v3.0.0-beta.0, and is thus not released. | ✘                                        |

### /api

| Plugin/Loader          | Support for Webpack 4                    | Installed in /api |
| :--------------------- | :--------------------------------------- | ----------------- |
| babel-loader           | ✔ v7.1.3                                 | ✘                 |
| babili-webpack-plugin  | ✘ Support added in stale PR https://github.com/webpack-contrib/babel-minify-webpack-plugin/pull/70 | ✘                 |
| webpack-node-externals | Unlikely. Last updated May 8, 2017.      |                   |

ZhangYiJiang:
Closes #874 

### Upgrade notes 

- Cache loader has been removed for HardSourcePlugin. The result should actually be faster, but please help test with your computers. Focus is on hot reloading performance first and prod build performance second 
- AutoDllPlugin has been removed. It's not compatible for some reason. 
- LodashWebpackPlugin has been removed. 
- OptimizeCSSAssetsPlugin is removed. For some reason the plugin is stripping sourcemaps. cssnano is used directly, so it is now configured in postcssrc.js 
- The main entry chunk is now inlined directly into index.html. This should improve performance slightly 
- Webpack's new chunking algorithm means our split chunks have more files before since they're split between vendor and app code, whereas before they were together. Performance and size should still be roughly the same.


ZhangYiJiang:
Hmmm, the file size is slightly bigger. Will investigate.

taneliang:
Found it. I'll remove it and push

ZhangYiJiang:
Anyway I forgot to post an update here, but the reason why I've left this branch for a while is because this upgrade isn't giving us better filesizes or build times (better in dev, worse in prod), so all in all it looks like a regression. I'm sure we're doing something wrong, but I can't figure out where - probably because we're not using OptimizeCSSAssetsPlugin, but like I said before it's not playing well with sourcemaps

www/.postcssrc.js
```diff
@@ -10,3 +10,21 @@ module.exports = {
...
+
+// Use CSSNano in production
+if (process.env.NODE_ENV === 'production') {
+  config.plugins.push(
+    require('cssnano')({
+      preset: [
+        'default',
+        {
+          // mergeLonghand produces incorrect transformations with border
+          // https://github.com/cssnano/cssnano/issues/557
```
li-kai: This was fixed!

www/.postcssrc.js
```diff
@@ -10,3 +10,21 @@ module.exports = {
     require('./scripts/postcss-single-root'),
   ],
 };
+
+// Use CSSNano in production
+if (process.env.NODE_ENV === 'production') {
```
li-kai: Should we run this in dev too? It's easier to identify errors that way.

www/webpack/webpack.parts.js
```diff
@@ -45,8 +41,6 @@ const VENDOR = [
   'history', // History module used by router
   'fbjs', // facebook deps
   'prop-types', // gone but not forgotten
-  'loader', // style loader fallbacks
-  'equal', // various comparison libs used by deps
 ];
 
 const DLL = {
```
li-kai: We can drop this now we don't have dll.

www/webpack/webpack.config.prod.js
```diff
@@ -17,8 +19,9 @@ const config = require('../src/js/config/app-config.json');
  * @see https://webpack.js.org/guides/code-splitting-css/
  * @see https://survivejs.com/webpack/styling/separating-css/
  */
-const extractTextPlugin = new ExtractTextPlugin('[name].[chunkhash].css', {
-  allChunks: true,
+const cssExtractPlugin = new MiniCssExtractPlugin({
+  filename: '[name].[chunkhash:8].css',
```
li-kai: We should use content hash instead: https://medium.com/@sahilkkrazy/hash-vs-chunkhash-vs-contenthash-e94d38a32208

www/.postcssrc.js
ZhangYiJiang: May slow it down though. I don't know what's the overhead but I assume it's not nothing

www/webpack/webpack.config.prod.js
ZhangYiJiang: I don't think that's right? Article says content hash is from CommonChunksPlugin which doesn't exist anymore

li-kai: It's replaced by MiniCssExtractPlugin, which should work the same way

ZhangYiJiang:
I'm still hesitant about merging this in. Currently 

- Build size is slightly bigger (5kb) and split into more files. More files is probably not a problem, and 5kb is pretty negligible, but it is no better than before 
- Dev times (both hot and cold) are comparable as far as I can tell 
- CI build times are significantly slower (1.47 vs. 2.49). There will be some variance due to CI resources, but this is pretty bad - it's about 50% more.

ZhangYiJiang:
- Removed lint phase from prod build as suggested by @liaujianjie (thanks!) - this is acceptable because in CI lint is already done separately, so this is just redundant
-  UglifyWebpackPlugin -> TerserWebpackPlugin - uglify-es was forked since the original maintainer abandoned it 

I think the branch is ready to merge now. Just need to run through the usual checks to make sure upgrading the build toolchain didn't break anything

ZhangYiJiang:
Remaining issues (non-blockers) 

- FlowStatusWebpackPlugin is not updated, so you get a DeprecationNotice. The plugin looks unmaintained, but it should be easy to use a fork which fixes the issue 
- Babel optimize-react-preset is not updated, and also looks unmaintained. A preset is just a function that returns a bunch of plugins though, so it's easy to fix this too.

PR #1053, opened 2018-07-23
Fix Safari 9 regression due to query-string

ZhangYiJiang:
query-string@6 does not support anything other than the latest version of evergreen browsers. Downgrading to v5 fixes our current issue with Safari 9. 

We should look into setting up some sort of automated smoke test. Maybe using Browserstack.


ZhangYiJiang:
Confirm fixed with BrowserStack :+1:

PR #1056, opened 2018-07-25
Add basic e2e tests

ZhangYiJiang:
Using Nightwatch and Browserstack, I set up some very basic smoke tests which should make sure we don't make any silly mistakes that break NUSMods for older browsers. Current testing matrix is 

- Safari desktop 9.1 
- Safari mobile 10.3
- Firefox ESR latest 
- Edge lastest minus 2


ZhangYiJiang:
Smoke tests have been made manual. Browserstack local is not very reliable (or maybe I'm using it wrong, who knows?). Instead the plan going forward is to add a button to Launchpad to run this as part of the deploy process. 

The command also takes in environment variable that allows us to point Browserstack at any remote URL eg. Netlify deploy previews and run the tests that way. It's not as convenient as running the tests as part of CI, but it makes the CI tests run faster since these e2e tests are quite slow.

PR #1058, opened 2018-07-26
Make numeric week text more readable

ZhangYiJiang:
Currently numeric week numbers on lessons can overflow if they include too many weeks. They're also not very user friendly because there's no context - it's just a bunch of numbers. This PR fixes both issues by adding the word 'Week(s)' in front of the number and combining consecutive ranges to reduce the amount of text displayed 

Before

screenshot from 2018-07-26 23-42-46

After 

screenshot from 2018-07-26 23-40-15


ZhangYiJiang:
Also a bit strange that CORS (I think?) is reporting weeks 3-13 for tutorials since I think these are the usual tutorial dates anyway

PR #1064, opened 2018-07-29
Add module table footer

ZhangYiJiang:
The footer now contains 

- Total MCs this semester 
- Module sort order on the table 

The sort order is persisted in settings and defaults to exam date. 

Desktop 

screenshot from 2018-07-29 13-31-52

Mobile 

screenshot from 2018-07-29 13-33-09


www/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -15,23 +17,46 @@ import {
...
   readOnly: boolean,
 };
 
+function renderMCs(moduleCredits) {
+  return `${moduleCredits}${NBSP}${moduleCredits === 1 ? 'MC' : 'MCs'}`;
+}
+
 class TimetableModulesTable extends Component<Props> {
+  totalMCs() {
+    return sum(this.props.modules.map((module) => parseInt(module.ModuleCredit, 10)));
```
li-kai: How about `sumBy`?

PR #1076, opened 2018-07-30
Add support for 8k modules

ZhangYiJiang:
Because [they're a thing now](https://nusmods.com/modules/PLS8001/cultivating-collaboration). Yeah. I don't know why they skipped 7k, but ¯\\\_(ツ)\_/¯


ZhangYiJiang:
NUS power level needs to be higher

PR #1077, opened 2018-08-02
Fix some errors 

ZhangYiJiang:
- Hopefully reduce the amount of service worker errors by only updating if the browser is online 
- Fix the unknown timesegment bug once and for all by parsing the time string instead of using a `switch`

### DIFF
--- a/www/src/js/views/timetable/ExamCalendar.test.jsx
+++ b/www/src/js/views/timetable/ExamCalendar.test.jsx
@@ -11,7 +11,7 @@ import mockModules from '__mocks__/modules';
 /** @vars {Module} */
 import GER1000 from '__mocks__/modules/GER1000.json';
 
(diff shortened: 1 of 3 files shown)


www/src/js/views/timetable/ExamCalendar.test.jsx
```diff
@@ -95,3 +95,21 @@ describe('ExamCalendar', () => {
     expect(wrapper.find(Link)).toHaveLength(3);
   });
 });
+
+describe(getTimeSegment, () => {
+  it.each([
```
taneliang: Oh cool new feature. Anyway, should this be `test.each` instead?

ZhangYiJiang: They are aliases, so it's just a matter of style

PR #1092, opened 2018-08-13
Refactor request middleware and improve API error handling 

ZhangYiJiang:
This PR deals with long standing issues with how we handle requests. It supersedes #715 and adds a bunch of other stuff. 

First, we formalize our API request handling strategy. We have two different data sources - 

- Fetch from API, which is async
- Loaded from Redux, which is a cache for previously loaded data 

To load data from the API

1. Component dispatch a Request action. Each request to an unique API endpoint uses a different string key (previously the action type is used, but a single type of action may call different endpoints, so we add a unique key instead). The dispatch function returns a Promise (via Redux Thunk) to the component. 
2. Request middleware intercepts the action and does an actual fetch request. 
   - It also immediately fires a REQUEST action 
   - When the fetch is done, it fires a SUCCESS action with the payload 
   - If the fetch fails, it fires a FAILURE action with the error as the payload 
3. Request reducer stores the request action result (REQUEST, SUCCESS, FAILURE) keyed by the request key. The state is for when components other than the one which initiated the request wants to display eg. errors or statuses - components should if possible consume the Promise instead, since that's much simpler. 
4. Other reducers in module and venue bank consumes SUCCESS actions and update their cache accordingly. 
5. Meanwhile, the component directly pulls the data it needs from module and venue bank states. (This is essentially the stale-while-revalidate strategy Workbox uses) 
    - If this is the first time the data is loaded, then the component displays a spinner 
    - Otherwise the component displays the cached data immediately. 99% of the time the request result matches the state in the cache, so they don't notice anything.
    - If the data is different, then the rendered component also updates. Again this is fine for the most part because we expect the changes to be small 
        - Exception: Sometimes modules remove or change lessons. We already automatically validate and force the local timetable state into a valid state, but we have yet to notify the user when this happens 
    - If the request fails, the promise returned to the component is rejected. We can then pop a notification informing the user they're seeing stale data, or if the data has never been loaded, show an error component 
   
Phew, okay, so please help me check if I missed anything in the strategy above.  

Now here are the actual changes in this PR 

- Refactor request middleware, state and reducer to allow for the above to be implemented 
- Remove existing helper and request actions, since they're never used anywhere 
- Add error handling to AppShell and ModulePageContainer 

Future todos: 

- Show notification in case of timetable validation issues 
- Add retry options for API load errors 
- Add network warning when navigator.online is false for API load errors


ZhangYiJiang:
I actually just realized it would be a good idea to use actions even when the payload is not cached (eg. for module finder's module information data) since it'll still be a good idea to have a record of the request status in the store so that other components can query it if necessary

PR #1094, opened 2018-08-17
Remove usage of deprecated apis

li-kai:
As per title, removes `componentWillX` with a more suitable API for React 16.3x onwards.


ZhangYiJiang:
I don't think this is correct usage of `getSnapshotBeforeUpdate`. It may be better to force a two-stage render instead (ie. side effect in `didMount`, then `setState` which triggers the rendering of the actual modal).

Issue #745, opened 2018-01-16
Venue location enhancement roadmap


taneliang:
This is a meta issue tracking the enhancement of the venues page with location data, and possibly bus routing data. Follows up on #654.

### API

- [ ] Finalize GraphQL API schema for enhanced venues
- [ ] Integrate the prototype [venue location data scraper](https://github.com/taneliang/NUSVenues) with existing venue scraper
- [ ] Find a way to store user inputs from the venue location feedback box (see below)
- [ ] Implement some sort of NUSMods venue/NUS room matcher override. Reasons:
    - NUS's fuzzy matcher can produce bad results, e.g. [LT10 is matched to a female toilet](https://github.com/taneliang/NUSVenues/blob/81a96585550b5a2ea1a3edd2a7f6a9d24bd7d42a/results/finalVenues.json#L366)
    - Some venue names are so dissimilar from their room codes that the fuzzy matcher can't match them
- [ ] Implement a custom location datasource, so that we can use use user-provided locations. These locations won't need to be matched.

### Client

- [ ] Display room names and actual NUS room codes
- [ ] Display map in venue detail view
- [ ] Beta notice and feedback box to allow users to correct wrong venue names/room codes/locations
- [ ] A way to select venues on a map instead of scrolling through a list. We need a way to handle venues with no GPS coords in this case.

### Bus feature

This concept has to be further baked before implementation. It may even be better implemented as a separate mobile app instead of integrating it with NUSMods proper. Bus routes and timings can be obtained from a [Comfort API](https://nextbus.comfortdelgro.com.sg).

- [ ] Nearest bus stop feature
- [ ] Bus routing between venues. https://gothere.sg might be useful
- [ ] Directions to the next class (perfect for a Today view)

PR #1096, opened 2018-08-19
[WIP] Implement data and sync Rails server

### LINKED ISSUE #716: GraphQL Roadmap
Meta issue tracking all tasks we need to do to move to GraphQL.

### Infra and decisions

- [x] Monorepo (good to have but not essential)
- [ ] We need some way to run the server. `PM2`
- [ ] Do we want to run it on the same droplet or another droplet?
- [x] Do we want to use Docker? Will take about 10 min for @li-kai to set up. We can limit memory usage, and if we were extremely rich we could set up auto scaling.

### GraphQL server code

- [ ] Add school name to schema and queries
- [ ] Add query param to be able to filter data by semester
- [ ] Venue support (basically only the new location schema)

Keep in mind other schools' requirements while we're building this schema.

### Actual migration

- [ ] Install Apollo and Apollo GraphQL loaders
- [ ] Migrate venues client to GraphQL backend
- [ ] Migrate modules on client to GraphQL backend
- [ ] Finally migrate module search, which is the hardest problem.
- [ ] Rip out REST related code

taneliang:
Creating a WIP PR so that we can track the implementation of this server, and so that we can decide whether using Rails is a good idea.

screenshot_20180819_235421

Implements #796 (among other things) and closes #716. If merged, this PR will supersede #808.

As this PR is already gigantic, it only implements the backend and does not include any frontend changes. As this is Rails, much of the code here is generated by the various setup scripts. Almost all of the code I've written is in app/graphql and app/models, and a lot of those were generated and modified. It should be easy to review if you are familiar with Rails.

Setup may be a bit difficult (can't remember why), so you might want to hold off on trying this for yourself.

### Implemented Features

- Data server to replace our JSON API. If we decide to use this Rails implementation, we should do some trial data imports, possibly after this PR is merged. I know CORS bidding stats are not stored 
- User accounts
    - Login/forgot password/reset password pages using Devise. Sadly, we may still need to implement our own UI in our React app because the Devise screens only support session auth and not token auth, which we need for SPAs and third party apps. Either that or we could try implementing token login pages in Rails.
- Support for enhanced venue data (#745)
- Support for multiple schools
- Real time updates for user data through GraphQL Subscriptions over Action Cable
- Seed data (except venues and faculties)
- SQLite database for portability and easy setup. It supports JSON columns as well.
- Decent looking admin interface using Rails Admin (http://localhost:3000/admin)
    screenshot_20180819_234857
    screenshot_20180819_234919
- Email viewing in dev using the Letter Opener gem (http://localhost:3000/letter_opener)
    screenshot_20180819_235023

### TODO

- [x] Token auth using Devise Token Auth gem
- [ ] Timetable data
- [ ] Decide if we want to use Rails and accept this PR
- [ ] User roles using the Rolify gem (`user`, `admin`, `scraper`)
- [ ] GraphQL mutations for scraper input
- [ ] Data validation
- [ ] Make seed data pass validation, and add seed data for user roles, faculties and venues
- [ ] Solve remaining N+1 queries
- [ ] Setup/usage docs
- [ ] Rename `sync` folder to something else, since this isn't a sync server.
- [ ] Somehow get subscriptions working in GraphiQL. Subscriptions aren't working there as Action Cable isn't included in that gem
- [ ] Remove unused sample files

### TODO outside this PR

As this PR is huge and we aren't even sure if we want to use Rails, frontend work and some other nice to have things can be left to future PRs.

- Integrate Apollo Client and associated dependencies (including Action Cable) into our web app
- Import our legacy JSON data
- Tests using RSpec
- Code documentation using YARD
- Fix remaining lint failures
- Set up CI
- Set up deployment scripts (probably using Capistrano)
- Create a transition plan. We could do daily JSON imports until scrapers have been converted to use the GraphQL migrations
- Convert scrapers to add data using our GraphQL migrations

### So, Rails?

I think we should use Rails for our server.

There's not much custom code as Rails and other gems do the heavy lifting, which means that this should be maintainable in the long run. The gems I've added to this project are also well maintained.

The biggest con is that it isn't JS, and it'll be harder for new people to onboard onto our codebase. I don't think this is a huge issue, as Rails is widely used. And actually a pure JS codebase also makes it difficult for Rails devs to onboard, so... I guess our choice of language will always leave some people out.

I don't think we can implement anything like this as quickly in any other languages. Node may be simpler but too many things are DIY. My Elixir experiment was a failure as that ecosystem is too immature and there are too many bugs and not enough support. Django might be an option but I'm not familiar with it.

I suppose the only other main competitor to this is something serverless, like AppSync or Firebase. We could definitely explore those.


ravern:

PR #1100, opened 2018-08-25
Update www dev dep

ZhangYiJiang:
Rolls all of the dev dep upgrades into one. Doesn't include Webpack related stuff


ZhangYiJiang:
Note that at next deploy we will need to double check the Autoprefixer upgrade didn't break older browsers (we can also do it now, but we'd have to do it before deploy anyway)

PR #1103, opened 2018-08-25
Add /help and /api redirects 

ZhangYiJiang:
Add redirects for v2 route /help, and /api route to api.nusmods.com. Also split off the external redirect logic into its own component.

### DIFF
--- a/www/src/js/views/routes/Routes.jsx
+++ b/www/src/js/views/routes/Routes.jsx
@@ -1,6 +1,6 @@
 // @flow
-import React, { type Node } from 'react';
-import { Switch, Route, Redirect } from 'react-router-dom';
+import React from 'react';
+import { Redirect, Route, Switch } from 'react-router-dom';
 
 import TimetableContainer from 'views/timetable/TimetableContainer';
 import ModulePageContainer from 'views/modules/ModulePageContainer';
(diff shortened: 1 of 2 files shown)


www/src/js/views/routes/Routes.jsx
```diff
@@ -34,20 +35,11 @@ export default function Routes() {
...
-        render={(): Node => {
-          window.location = 'https://www.facebook.com/bareNUS';
-          return null;
-        }}
-      />
+      <Redirect from="/contact" to="/faq" />
+      <Redirect from="/help" to="/faq" />
+      <ExternalRedirect path="/news/nusdiscount" to="https://www.facebook.com/nusdiscount/" />
+      <ExternalRedirect path="/news/bareNUS" to="https://www.facebook.com/bareNUS" />
+      <ExternalRedirect path="/api" to="https://api.nusmods.com" appendPath />
```
taneliang: Should this redirect to 
https://github.com/nusmodifications/nusmods/tree/master/api instead? Our API link in the footer links to that since it contains our API docs.

ZhangYiJiang: If you look at the Sentry logs you can see that people aren't visiting `/api`, they're trying to access the actual JSON file (for whatever reason). Redirecting them back to the API README is unlikely to be helpful

Issue #51, opened 2014-07-11
Geocoding of NUS locations


yeesian:
related: https://github.com/Leventhan/ModCrasher

ZhangYiJiang:
Done-ish. Data is about ~60% complete, half crowdsourced, half from NUS geography/CSIT data.

PR #1106, opened 2018-08-25
Venue page location MVP 

ZhangYiJiang:
Venue geocoding is from https://github.com/taneliang/NUSVenues/, with the data manually cleaned up and extracted into `venues.json`. The file is currently bundled statically because there's no process to automatically update the data yet, and the data won't change much (I hope). 

Tech stack is Leaflet and React Leaflet + standard Tile Map Server. The whole shebang is loaded async, like `ModulePageContent` because Leaflet is pretty big. 

Currently the tileserver uses OSM tile server. See https://nusmods.com/venues/COM1-B112 for a preview. This may cause issues because I think all of NUS goes behind one IP address, so if usage is heavy we could get the IP banned. Here are some alternatives 

**OneMap** - LTA data, not as detailed (no walkway, shop or toilet data). Buildings are slightly 3D. Has very generous free usage limits, which is much better than any other providers

screenshot from 2018-08-26 15-06-17

**MapBox streets** - default styles. Places of Interest labels are gone, which is a dealbreaker. I can build our own map style which has it enabled, but then we can't use Leaflet and must use Mapbox GL instead (which is built on top of Leaflet, but the React bindings aren't as good https://github.com/alex3165/react-mapbox-gl). 

screenshot from 2018-08-26 15-09-44

**HERE map** - proprietary data (I think). Missing PoI markers, not sure if the data doesn't exist or if it's just not shown. Roads are too big at the zoom level we want. I don't think this is very good. 

screenshot from 2018-08-26 15-30-57

Also checking if we can proxy or cache OSM's tileserver. 

## Todo / future plans 

- Add crowdsourcing instructions - can link GitHub issues API with Google Spreadsheet to solicit updated venue info 
- Venue photos - find some way to get users to take pictures and upload photos which makes it even easier to identify the rooms 
- Integrate travel info - maybe using NUS Maps


li-kai:
This fixes  #51

PR #1110, opened 2018-08-26
Add animation to modal component

ZhangYiJiang:
Currently some (well, one) modals are animated when they open. This PR adds the corresponding close animation and allows for it to be configured using an `animate` boolean prop.


www/src/styles/utils/animations.scss
```diff
@@ -34,6 +34,21 @@
   }
 }
 
+@keyframes zoomOut {
+  from {
+    opacity: 1;
+  }
+
+  50% {
```
li-kai: Is this supposed to be here? It feels like closing animation is way too fast as a result.

ZhangYiJiang: Ah, I made a mistake mirroring the zoomIn animation. Let me fix that

www/src/js/views/components/Modal.scss
```diff
@@ -11,6 +11,21 @@
...
+  :global {
+    animation: zoomIn 0.2s;
+  }
+}
+
+:global(.ReactModal__Overlay--before-close) {
+  .animated {
+    :global {
+      // Update closeTimeoutMS prop in Modal.jsx if this duration is changed
+      animation: zoomOut 0.15s forwards;
```
li-kai: You should indeed change Modal.jsx 😆

ZhangYiJiang: Ooof.

Issue #1104, opened 2018-08-25
[Feature] API server using Node


li-kai:
As dictated by the progenitor @yangshun (and through much discussion) I think we might still go with node.
To summarize, the benefits are

1. Newcomers know Javascript, courtesy of CS1101S
2. Our entire stack is already Javascript
3. We may be able to share code

The cons are

1. More code to write

Thankfully, I don't think there's too much code. The following are steps needed to get out api mvp out:

- [ ] Remove previous data code that reads files
- [ ] Upgrade to apollo-server 2 
- [ ] Add new schemas for a fully functional server
- [ ] Add queries to fetch data from db
- [ ] Add mutations to add data from db
- [ ] Add validations to mutations

taneliang:
Actually, I think a significant part of the complexity is in the auth/sync parts. That doesn't really matter if we're going to split the 2 servers, but if we're implementing them together then we need to try those early so that we get an accurate picture of what's needed and potential problems. That'll allow us to better evaluate Rails vs Node vs whatever else

PR #1111, opened 2018-08-26
Add initial server schema

li-kai:
Fixes 3/n of #1104 

Uses knex to add the schema. I changed `acadYears` + `sem` into a single table called `version`. 

Not sure if this is preferable, as this decreases the amount of joining we need, as well as be more flexible should other school have other ways of classifying courses.

untitled

### DIFF
--- a/api/data/migrations/20170101000000_initial.js
+++ b/api/data/migrations/20170101000000_initial.js
@@ -1,55 +1,137 @@
+const TABLE_NAMES = {
+  schools: 'schools',
+  terms: 'terms',
+  departments: 'departments',
+  venues: 'venues',
+  courses: 'courses',
+  lessons: 'lessons',
+};
+const ID_NAMES = {
+  schools: 'school_id',
+  terms: 'term_id',
+  departments: 'department_id',
+  venues: 'venue_id',
+  courses: 'course_id',
+  lessons: 'lesson_id',
+};
+const DAYS = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
+
 exports.up = (knex, Promise) => {
-  const schoolsTable = knex.schema.createTable('schools', (table) => {
+  function idPrimaryKey(table) {
     table
       .increments('id')
       .notNullable()
       .primary();
+  }
+
+  function idForeignKey(table, foreignTableName, notNullable) {
+    const tableIdName = ID_NAMES[foreignTableName];
+    if (notNullable) {
+      table
+        .integer(tableIdName)
+        .notNullable()
+        .references('id')
+        .inTable(foreignTableName)
+        .onDelete('CASCADE')
+        .onUpdate('CASCADE');
+    } else {
+      table
+        .integer(tableIdName)
+        .references('id')
+        .inTable(foreignTableName)
+        .onDelete('RESTRICT')
+        .onUpdate('CASCADE');
+    }
+  }
+
+  const schoolsTable = knex.schema.createTable(TABLE_NAMES.schools, (table) => {
+    idPrimaryKey(table);
+    table.timestamps();
     table
-      .string('name')
+      .string('long_name')
       .notNullable()
       .unique();
-    table.string('abbreviation', 32);
+    table.string('short_name', 32);
   });
 
-  const departmentsTable = knex.schema.createTable('departments', (table) => {
-    table
-      .increments('id')
-      .notNullable()
-      .primary();
-    table
-      .integer('school_id')
-      .notNullable()
-      .references('id')
-      .inTable('schools')
-      .onDelete('CASCADE')
-      .onUpdate('CASCADE');
+  const termsTable = knex.schema.createTable(TABLE_NAMES.terms, (table) => {
+    idPrimaryKey(table);
+    idForeignKey(table, TABLE_NAMES.schools, true);
+    table.timestamps();
+    table.dateTime('starts_at').notNullable();
+    table.dateTime('ends_at').notNullable();
+    table.string('name', 32).notNullable();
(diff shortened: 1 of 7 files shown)


api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+    const versionId = idForeignKey(table, NAMES.versions, true);
+    idForeignKey(table, NAMES.departments, false);
+    table.timestamps();
+    table.string('code').notNullable();
+    table.string('name').notNullable();
+    table.string('floor');
+    table.decimal('lat', 10, 7);
+    table.decimal('long', 10, 7);
+    table.decimal('altitude', 10, 7);
+    table.unique([versionId, 'name']);
```
taneliang: I think it may be better to make the code unique instead, since room names could be non-unique (e.g. "SEMINAR ROOM 1"s https://github.com/taneliang/NUSVenues/blob/master/results/finalVenues.json#L2586 and https://github.com/taneliang/NUSVenues/blob/master/results/finalVenues.json#L5166)

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+    return tableIdName;
+  }
+
+  const schoolsTable = knex.schema.createTable(NAMES.schools, (table) => {
+    idPrimaryKey(table);
     table
       .string('name')
       .notNullable()
       .unique();
     table.string('abbreviation', 32);
```
taneliang: I think we should rename the `name` and `abbreviation` columns to `long_name` and `short_name`. Much easier to spell and less prone to typos. Can cherry pick and lightly modify this commit https://github.com/nusmodifications/nusmods/pull/952/commits/32857ade206d7afb8e69bafc6b86a758aa149740

ZhangYiJiang:
Which of these columns are NOT NULL? Also, do you have indexes set up?

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+  const coursesTable = knex.schema.createTable(NAMES.courses, (table) => {
+    idPrimaryKey(table);
+    const versionId = idForeignKey(table, NAMES.versions, true);
+    idForeignKey(table, NAMES.departments, false);
+    table.timestamps();
+    table.string('code').notNullable();
+    table.string('title').notNullable();
     table
-      .integer('school_id')
+      .string('description')
```
taneliang: Does this need to be `.text` instead? Strings are limited to 255 chars according to https://knexjs.org/#Schema-text. Might want to consider changing title, *requisite and preclusion as well

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+  function idPrimaryKey(table) {
     table
       .increments('id')
       .notNullable()
       .primary();
+  }
+
+  function idForeignKey(table, foreignTableName, notNullable) {
+    // Simple logic that turns names -> name
+    const singularTableName = foreignTableName.replace(/s$/, '');
```
taneliang: Oh man this is gonna cause some bugs that will be really hard to trace in the future if someone copy pastes this function without checking

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+    table.decimal('long', 10, 7);
+    table.decimal('altitude', 10, 7);
+    table.unique([versionId, 'name']);
+  });
+
+  const coursesTable = knex.schema.createTable(NAMES.courses, (table) => {
+    idPrimaryKey(table);
+    const versionId = idForeignKey(table, NAMES.versions, true);
+    idForeignKey(table, NAMES.departments, false);
+    table.timestamps();
```
taneliang: Nice! I think this should be on the schools and lessons tables as well

api/data/migrations/20170101000000_initial.js
ZhangYiJiang: Maybe use https://www.npmjs.com/package/pluralize instead?

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,130 @@
...
+    table.string('code').notNullable();
+    table.string('name').notNullable();
+    table.string('floor');
+    table.decimal('lat', 10, 7);
+    table.decimal('long', 10, 7);
+    table.decimal('altitude', 10, 7);
```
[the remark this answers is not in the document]
li-kai: Is it an integer?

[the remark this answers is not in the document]
ZhangYiJiang: Don't think so, can't tell with the data, but it's probably best to use decimal. To be honest I don't think the z in the venue data is accurate anyway, and even if it is I'm not sure if it'll be useful

api/data/migrations/20170101000000_initial.js
li-kai: I'm missing something because I don't see the problem here...

api/data/migrations/20170101000000_initial.js
```diff
@@ -1,55 +1,137 @@
...
+    idPrimaryKey(table);
+    idForeignKey(table, TABLE_NAMES.terms, true);
+    idForeignKey(table, TABLE_NAMES.departments, false);
+    table.timestamps();
+    table.string('code').notNullable();
+    table.string('name').notNullable();
+    table.string('floor');
+    table.decimal('lat', 10, 7);
+    table.decimal('lng', 10, 7);
+    table.integer('altitude');
```
taneliang: I don't think this should be an int, but whatever. It fits our current data so we can just change it in the future if necessary

PR #1131, opened 2018-08-30
Add venues location update script 

ZhangYiJiang:
Add a script to automatically pull data from NUSVenues. It tries to merge in venues which only exists in the API, and for any venue that already exists in `venues.json`, it'll only update latlng data. We currently have 46% (259 / 561) coverage on location data. 

I've also removed `z` altitude and `nusRoomCode` from `venues.json`, because neither are very useful. 

Also took this opportunity to fix a bug in the sitemap script and update the sitemap.


ZhangYiJiang:
> CORS's room codes (and therefore the ones in our API) are mangled frequently

Do you have an example? The only difference I see is that CORS's room codes are missing the `-` between unit number and floor. This could be useful for rooms like `COM1-VCRM`, but the Maps API don't have data on these (presumably because they only have the room code), so it's not very useful either.

taneliang:
That was what I found before I moved to the ArcGIS APIs in this commit https://github.com/taneliang/NUSVenues/commit/81a96585550b5a2ea1a3edd2a7f6a9d24bd7d42a. Previously, I had to manually transform our venue names in order to get them to match exactly with the venue names on their API, e.g. the SDE E-studios. The new code doesn't match any SDE venues, presumably because the CORS venue names are too different from the NUS ones.

So yeah, while it's true that the current venues are largely just missing the `-`, once we figure out how to scrape the other venues the `nusRoomCode` will be useful.

But having said that, it may be fine to remove it for now. We can always add it back if and when it becomes useful.

ZhangYiJiang:
Ah, yeah, it might be useful for the scraper, but for the frontend it's not terribly useful

taneliang:
For now. I have a feeling the room codes taken from the ArcGIS APIs are the legit room codes on the doors. So using the CORS codes might cause some confusion as people may not be able to find the rooms IRL.

ZhangYiJiang:
Ah, but in that case it's just CORS using an abbreviation. The data is still correct, and I'm not sure if E-Studio is significantly more usable than ES (assuming the student knows what it stands for, of course)

taneliang:
True. Let's just remove it then

ZhangYiJiang:
I suspect the "Staff Room"s showing up in the venue data are data errors. The question is whether the error is on CORS side or GIS side (I suspect GIS matching the wrong rooms, or having outdated data, because otherwise students would end up going to staff rooms for their lessons)

taneliang:
Yeah those are likely caused by GIS's fuzzy search

PR #1135, opened 2018-08-30
Fix CloseButton click event in ModuleSelect modal 

### LINKED ISSUE #1097: Closing add module modal (and possibly others) on mobile also navigates to settings
### Describe the bug

The close button for the add module modal when pressed causes the app to navigate to the settings page because the close button is on top of where the settings button is 

### To Reproduce

1. On mobile phone, click on 'Add module to timetable' button 
2. Tap the X to close the modal 
3. Observe that in addition to closing the modal, the app has also navigated to the settings page 

### Expected behavior

The settings page shouldn't open 

### Technical Info 

Reproduced on Android Chrome Mobile latest and also in Browserstack Safari Mobile

ZhangYiJiang:
Closes #1097. 

The cause of this is the onBlur handler on the form input. When the close button is clicked, the blur event handler fires first, which causes the modal to close. *Then* when the click event fires, it fires on the element underneath the close button since the modal is now closed. 

The current fix is to add a 0ms timeout to the blur event handler, so the event is effectively pushed to fires after any UI events that's occurring at the same time. I don't really like it, so if you guys have any better suggestions I'm open to it.

### DIFF
--- a/www/src/js/views/timetable/ModulesSelect.jsx
+++ b/www/src/js/views/timetable/ModulesSelect.jsx
@@ -46,9 +46,15 @@ class ModulesSelect extends Component<Props, State> {
   };
 
   onBlur = () => {
-    if (!this.state.inputValue && this.state.isModalOpen) {
-      this.closeSelect();
-    }
+    // A hack to force the onBlur event to be handled after other UI events.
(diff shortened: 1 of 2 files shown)


ZhangYiJiang:
Closed in favor of #1152

PR #1149, opened 2018-09-01
Update venues

ZhangYiJiang:
Incorporates a bunch of crowdsourced data (well, all but one from me, but hey, it's still data)


ZhangYiJiang:
I went to the Bukit Timah campus on Friday, and from what I saw this looks correct

PR #1150, opened 2018-09-01
Database schema update

li-kai:
1. `table.timestamps()` unfortunately isn't non-null and we have to do them manually.
2. Add seed for some tables
3. Fixes some logging


ZhangYiJiang:
Preferably this should be done in a new migration, but this isn't deployed into production yet, so it's fine. 

By the way the util functions should be split into their own files so they can be reused in future migrations

li-kai:
@ZhangYiJiang Did you mean to approve this pr?

Yeah util function to their own files eventually when we have more than one migration file.

api/data/migrations/20170101000000_initial.js
```diff
@@ -44,9 +44,20 @@ exports.up = (knex, Promise) => {
...
 
+  function createTimestamps(table) {
+    table
+      .timestamp('created_at')
+      .notNullable()
+      .defaultTo(knex.fn.now());
+    table
+      .timestamp('updated_at')
+      .notNullable()
+      .defaultTo(knex.fn.now());
```
taneliang: Will your custom `updated_at` be automatically updated when the row is updated?

li-kai: Don't think so...might have to investigate a solution in the future.

PR #1151, opened 2018-09-02
Graphql db

li-kai:
Fixes 4/n of #1104

SQLite stores date in "2018-09-01 09:14:20" format but Javascript reads that date in local time, so we need something like "2018-09-01T09:14:20.000Z". Although appending a Z is fine too, I've found nothing about allowing the "T" separator as space.


api/data/src/graphql/customScalars.test.js
```diff
@@ -0,0 +1,29 @@
+import customScalars from './customScalars';
+
+const dateString = '2018-09-01 09:14:20';
+const dateUTCString = '2018-09-01T09:14:20.000Z';
+const dateInt = 1535793260000;
+const dateDate = new Date(dateInt);
+
+const { DateScalarType } = customScalars;
+
+describe('Date scalar', () => {
```
ZhangYiJiang: Use `DateScalarType` directly here perhaps

api/data/src/graphql/customScalars.js
```diff
@@ -0,0 +1,25 @@
+const { GraphQLScalarType } = require('graphql');
```
ZhangYiJiang: Shouldn't this be using `import`?

api/data/src/graphql/customScalars.js
```diff
@@ -0,0 +1,25 @@
...
+    if (ast.kind === Kind.INT) {
+      return parseInt(ast.value, 10); // ast value is always in string format
+    } else if (ast.kind === Kind.STRING) {
+      return ast.value;
+    }
+    return null;
+  },
+});
+
+export default {
```
ZhangYiJiang: You can just `export` the `const` directly (eslint-disable the import rule if necessary)

api/data/src/graphql/index.js
```diff
@@ -56,29 +75,64 @@ const typeDefs = gql`
...
 const resolvers = {
   Query: {
-    modules(root, { acadYear, first, offset }) {
-      const yearData = jsonData[acadYear];
-      if (yearData == null) {
-        return [];
-      }
-      const modules = Object.values(yearData);
-      return modules.slice(offset, offset ? offset + first : first);
+    schools(root, { first, offset }) {
```
ZhangYiJiang: May want to avoid using `limit` + `offset` for the scolling cursor. Use `id > {last}` should be better for performance.

li-kai:
Gonna close this in favour of new #1578

PR #1152, opened 2018-09-02
Fix module select

li-kai:
Behavior changes:
1. Esc no longer blurs
2. 'Done' on iOS closes if input is empty
3. 'Done' on iOS remains open otherwise
Dev:
1. Normalises 'isOpen' and 'isModalOpen'
2. Removed a bunch of duplicated listeners
3. Add some tests

### DIFF
--- a/www/src/js/views/timetable/ModulesSelect.jsx
+++ b/www/src/js/views/timetable/ModulesSelect.jsx
@@ -1,7 +1,7 @@
 // @flow
 import React, { Component } from 'react';
 import { has } from 'lodash';
-import Downshift, { type ChildrenFunction } from 'downshift';
+import Downshift, { type ChildrenFunction, DownshiftState, StateChangeOptions } from 'downshift';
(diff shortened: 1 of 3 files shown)


li-kai:
Side note: stateReducer is really powerful and could let us remove our own state... but I've no idea how to test the changes. Perhaps mounting and doing manual 'onChange' and 'focus'? I'd like them to use shallow as far as possible to avoid incurring costs :sweat:

ZhangYiJiang:
There’s a problem on the iPad - not sure if it’s Safari’s fault or our code’s. If you tap on the input, the keyboard opens, and if you then try to scroll the page, the menu closes, but the keyboard is still open and there’s no way to get it back unless you close the keyboard and tap on the input again.

li-kai:
@ZhangYiJiang Since it's a bug that can only be repro-ed on iPad and exists on both versions, let's merge this first and work on it later? 😄

PR #1173, opened 2018-09-18
Integrate Leaflet.GestureHandling into map components

### LINKED ISSUE #1165: Integrate Leaflet.GestureHandling into our map components
Currently our map used on venue pages uses the default Leaflet touch behaviors. This means that they use single finger scroll, which makes it hard to scroll down the venue pages to see the timetable on phones and tablets. The solution is to use two finger instead of one to scroll the map - this will let us preserve the single finger scroll for scrolling the page.

Fortunately there is a plugin for this already- https://github.com/elmarquis/Leaflet.GestureHandling. Unfortunately we use *React* Leaflet, so compatibility is unknown. Your mission, should you choose to accept it, is to attempt to make the plugin work, or if that is not possible, port the behavior for React Leaflet using the original plugin code.

JingYenLoh:
This PR resolves #1165 and integrates [Leaflet.GestureHandling](https://github.com/elmarquis/Leaflet.GestureHandling) into the Leaflet maps. On desktop devices, this will ignore mouse wheel scroll. On mobile, this overrides the single finger dragging, and prompts the user to use 2 fingers to pan the map on load.

With advice from @ZhangYiJiang , I extracted the CSS from the Leaflet.GestureHandling library into its own scss file. Woud like to thank him for guiding me along!


JingYenLoh:
@ZhangYiJiang , would you like me to lock the version of Leaflet.GestureHandling?

ZhangYiJiang:
Oh yes, please do. Thanks!

www/src/styles/leaflet-gesture-handling.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.leaflet-gesture-handling-touch-warning::after,
+.leaflet-gesture-handling-scroll-warning::after {
+  -webkit-animation: leaflet-gestures-fadein 0.8s forwards;
+  animation: leaflet-gestures-fadein 0.8s forwards;
+}
+
+.leaflet-gesture-handling-touch-warning::after {
+  content: attr(data-gesture-handling-touch-content);
```
ZhangYiJiang: This is pretty cute.

www/src/styles/leaflet-gesture-handling.scss
```diff
@@ -0,0 +1,59 @@
+@-webkit-keyframes leaflet-gestures-fadein {
```
ZhangYiJiang: These non-standard props can be removed. We use autoprefixer, so these will be automatically if needed. This includes pretty much all of the prefixed stuff below.

www/src/styles/leaflet-gesture-handling.scss
```diff
@@ -0,0 +1,59 @@
...
+  left: 0;
+  z-index: 461;
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: 15px;
+  font-size: 22px;
+  font-family: 'Roboto', Arial, sans-serif;
```
ZhangYiJiang: Remove this. We have our own font stack.

www/flow-typed/package-dep-libdefs.js
```diff
@@ -8,6 +8,14 @@ declare module '@material/snackbar' {
   declare module.exports: any;
 }
 
+declare module 'downshift' {
```
ZhangYiJiang: If it's not too much trouble, can you remove this module declaration for Downshift? I know the script annoyingly automatically generates it, but Downshift comes with typedefs which are not uploaded to flow-types annoyingly.

ZhangYiJiang:
Hmmm, also noticed that on the iPad two finger scrolling produces a lot of jitter as the map interprets the movement as zoom in addition to movement. I'm not sure what can be done about that though.

JingYenLoh:
I've integrated most of the changes requested.

Regarding overriding of the language content, there's an [issue](https://github.com/elmarquis/Leaflet.GestureHandling/issues/17) upstream to load the language content via ajax. I had a look at the source and the main module directly imports the LanguageContent file, so I'm not sure what can be done about that.

PR #1187, opened 2018-10-02
Change 0.0.0.0 to localhost to fix Windows dev

taneliang:
0.0.0.0 doesn't seem to work on Windows. As nobody *really* needs to develop things over the network by default (and I managed to achieve that on my setup through an Nginx reverse proxy), we can partially revert https://github.com/nusmodifications/nusmods/pull/766. Setting the `HOST` env var will still allow you to override the default `localhost`.

Untested on Windows for now.

- WDS now defaults to `localhost` instead of `0.0.0.0`.
- The browser now opens http://localhost:8080 when `yarn start` is run.
- `yarn start:export` now opens http://localhost:8081 in the browser.

### TODO

- [x] Test this on Windows


ZhangYiJiang:
Ping me if you need help testing this. I have a Windows box at home.

taneliang:
Wow thanks guys! Y'all are the best @li-kai @ZhangYiJiang :tada:

PR #1188, opened 2018-10-02
Fix settings page styles

taneliang:
Minor improvements to Settings page styles.

* Improve toggle row styles
    * Right align toggles on all display sizes
    * Give text more space on larger displays
    * Reduce whitespace between toggle and text
* Reduce h4 top padding as it's too big
* Restyle page header to be consistent with our static pages

#### Mobile (iPhone 5)

| Before | After |
|-|-|
| nusmods com_settings iphone 5_se | localhost_8080_settings iphone 5_se |

#### Laptop

| Before | After |
|-|-|
| nusmods com_settings laptop with hidpi screen | localhost_8080_settings laptop with hidpi screen |


ZhangYiJiang:
I don't understand the `<div class="row"><div class="col">`

One of the reasons we used narrow cols for the text is to keep text at a comfortable reading width. That's not the case anymore on your laptop screenshot

taneliang:
I've always thought the toggle buttons were floating in a really weird location though, but I see what you mean about the text. I can tweak this further over the next few days.

ZhangYiJiang:
For the toggles I think right aligning them would fix that. So I think we can just add `max-width` to the description.

taneliang:
2 screenshots showing what it looks like now. The page feels a bit too wide, with a bit too much space between the toggle descriptions and the toggles. It looks better if the settings page's width is limited to 50em (it's now 60em), but then the timeline preview will start scrolling horizontally.

localhost_8080_settings
localhost_8080_settings 1

PR #1189, opened 2018-10-02
Fix issue #1130

### LINKED ISSUE #1130: <select> hard to read in night mode using  Safari Mobile
`<select>` elements cannot have their background colors changed in Mobile Safari, but their font color can. In night mode this makes them difficult to read as the font color do not have enough contrast with the background. This does not reproduce in Safari Desktop or Chrome Mobile 

index

Two options  

1. Use Bowser (which we already use for legacy browser detection, see https://github.com/nusmodifications/nusmods/blob/master/www/src/js/bootstrapping/browser.js) to add a global class to body which we can then use to override `<select>` font color in Safari Mobile only 
2. Use `-webkit-appearance: none` or some other proprietary property, which should turn off native styling. Unfortunately this may also remove the down arrow on the `<select>` and make it harder to recognize, so use this with care 
    - https://stackoverflow.com/questions/21400182/safari-css-font-color-issue
    - https://stackoverflow.com/questions/42157894/style-html-select-tag-on-ios-safari-iphone

fllprbt:
Fixes #1130.

I have not tested it extensively on other ios devices with safari. I am saying this because there have been issues with old versions of bowser. 

Moreover I am open to any feedback/proposals.

### DIFF
--- a/www/src/js/utils/css.js
+++ b/www/src/js/utils/css.js
@@ -3,6 +3,7 @@
 // Define media breakpoints
 import json2mq from 'json2mq';
 import { entries } from 'lodash';
+import bowser from 'bowser';
 
 export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 export type QueryObject = { [string]: string | number | boolean } | QueryObject[];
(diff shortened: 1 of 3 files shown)


taneliang:
Thanks for your contribution! The text color is too light outside of dark mode though, which makes it look disabled.

img_6beacf1f38a3-1

@li-kai The CI runner somehow ran out of RAM. I reran it and it passed.

fllprbt:
@taneliang I will look through it and submit a solution on of these days.

li-kai:
Thanks @fllprbt for contributing!

PR #1195, opened 2018-10-04
Attempt to fix test times

li-kai:
Follows https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
       - run: yarn
       - save_cache: *save_node_deps
       - run: yarn lint
-      - run: yarn test
+      - run: yarn test --runInBand
 
   nus-scrapers:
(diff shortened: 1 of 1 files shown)


li-kai:
Tests take 55s instead of 4min now.

taneliang:
It doesn't look like anything changed though. The test part was previously 46 seconds on [this job](https://circleci.com/gh/nusmodifications/nusmods/9029#build-timing/containers/0).

ZhangYiJiang:
is-ci is unnecessary to set explicitly - jest already sets it when it detects CI environment

li-kai:
@taneliang At least the red bars from each individual test taking too long is removed.

Hoping this would solve random circleci problems we're seing.

PR #1196, opened 2018-10-04
Attempt to fix test times

li-kai:
Follows https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server but enable 4 workers to use 2 cores on circleci


li-kai:
Not faster than #1195

PR #1200, opened 2018-10-05
Fix select font color under Safari Mobile

ZhangYiJiang:
Fixes the issue with the previous PR #1189. I would have merged this into master already, but I want to double check, and the easiest way to get a preview build that my iPad can access is through Netlify


ZhangYiJiang:
@li-kai Can help check with iPhone if arrow is better aligned now?

PR #1210, opened 2018-10-09
Feature/tombstone

mrassili:
(WIP) It was brought to my attention by the maintainers that I should store the tombstone state in the component itself instead of Redux store. So I will need to refactor this solution in the next few days


ZhangYiJiang:
Yes, that's one issue - the data won't need to be so complicated. I think all that's necessary would be to store any deleted module as state.

PR #1215, opened 2018-10-11
Reload Disqus when night mode is activated/deactivated

ZhangYiJiang:
Disqus autodetect page font color and sets its own font colors appropriately. However this fails if the page suddenly changes colors, which happens if the user switches on/off night mode. This PR adds a bit of logic that forces the comment widget to reload if that happens. Apparently this is the only reliable way I can find - even though this page https://help.disqus.com/installation/disqus-appearance-tweaks claims the script uses CSS font color to determine the widget theme, I can't get it to work. 

Instead, the component just uses a constant 2s delay, which is found through experimentation. A bit dirty, and may not work well in all cases, but it seems good enough for now.

### DIFF
--- a/www/src/js/views/components/disqus/DisqusComments.jsx
+++ b/www/src/js/views/components/disqus/DisqusComments.jsx
@@ -9,12 +9,13 @@ import type { State } from 'reducers';
 import config from 'config';
 import insertScript from 'utils/insertScript';
 
-type Props = DisqusConfig & {
+type Props = {|
+  ...DisqusConfig,
   // Disqus autodetects page background color so that its own font color has
   // enough contrast to be read, but only when the widget is loaded, so we use
   // this to force the widget after night mode is activated or deactivated
   mode: Mode,
-};
+|};
 
 const SCRIPT_ID = 'dsq-embed-scr';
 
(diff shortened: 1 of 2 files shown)


www/src/js/views/components/disqus/DisqusComments.jsx
```diff
@@ -2,23 +2,37 @@
...
 import React, { PureComponent } from 'react';
+import { connect } from 'react-redux';
+
+import type { Mode } from 'types/settings';
+import type { State } from 'reducers';
 import config from 'config';
 import insertScript from 'utils/insertScript';
 
-type Props = DisqusConfig;
+type Props = DisqusConfig & {
```
li-kai: Use `{| ...DisqusConfig, mode: Mode |}` instead

www/src/js/views/components/disqus/DisqusComments.jsx
```diff
@@ -2,23 +2,37 @@
...
   componentDidMount() {
     this.loadInstance();
   }
 
-  componentDidUpdate() {
-    this.loadInstance();
+  componentDidUpdate(prevProps: Props) {
+    // Wait a bit for the page colors to change before reloading instance
+    if (prevProps.mode !== this.props.mode) {
+      setTimeout(this.loadInstance, 2000);
```
li-kai: We should leave a comment on why this is set to 2000ms

PR #1216, opened 2018-10-14
Issue#1212 add fullscreen button for venue maps

schalkneethling:
@ZhangYiJiang r?

### DIFF
--- a/www/src/js/views/venues/VenueLocation/VenueLocation.jsx
+++ b/www/src/js/views/venues/VenueLocation/VenueLocation.jsx
@@ -25,6 +25,7 @@ type Props = {
 
 type State = {
   isFeedbackModalOpen: boolean,
+  isFullscreen: boolean,
 };
 
 LeafletMap.addInitHook('addHandler', 'gestureHandling', GestureHandling);
@@ -34,7 +35,7 @@ function renderMap(position: LatLngTuple) {
   const googleMapQuery = encodeURIComponent(position.join(','));
 
   return (
-    <div className={styles.mapWrapper}>
+    <div id="venue-map" className={styles.mapWrapper}>
       <ExternalLink
         href={`https://www.google.com/maps/search/?api=1&query=${googleMapQuery}`}
         className={classnames('btn btn-sm btn-primary', styles.gmapBtn)}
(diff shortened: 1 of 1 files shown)


schalkneethling:
@ZhangYiJiang I just noticed that you want to use it on the venue data modal as well. I wonder if it would then make more sense to turn this into its own component? Let me know.

ZhangYiJiang:
@schalkneethling Thanks for the PR! Unfortunately I there has been a bit of a misunderstanding.

We're not looking to use the browser's fullscreen API - I forgot that such an API exists and it's completely my fault - but rather for the map to expand out and cover the whole page, since in the current implementation the map is a bit small. 

The CSS doesn't look sufficient - I think you would at least need to use `position: fixed`, and the CSS should be applied as a class on the component rather than through inline styles, unless that doesn't work. 

Also there seems to be a number of Flow error that's causing the CI to fail. Those should go away if the fullscreen toggle and DOM manipulation code is removed though I think. 

Again, sorry about this!

schalkneethling:
@ZhangYiJiang ok, updated. There is one problem now though. The map is not being redrawn and thus does not fill its available space. I thought doing something like `this.forceUpdate()` might do the trick, but no go.

I also tried the old CSS trick of hiding and showing the div, but still not luck.

Any ideas?

ZhangYiJiang:
@schalkneethling That's not a React issue, but a Leaflet issue. You need to call [`invalidateSIze`](https://leafletjs.com/reference-1.3.4.html#map-invalidatesize) if the map size changes. To get the map object you need to access React Leaflet's context. To make this easier the toggle button can be made a child of the `<Map>` instance. 

Also try using React `className` instead of DOM `classList` if possible. Raw DOM access need to have null checks, otherwise Flow will complain about possible null values

schalkneethling:
@ZhangYiJiang ok, I have updated this but, it is still not working 100%. I am going to make some comments in the code, to explain where I am stuck.

www/src/js/views/venues/VenueLocation/VenueLocation.jsx
```diff
@@ -29,25 +29,55 @@ type State = {
...
+      />
+      <Marker position={position} icon={icon} />
+    </Map>
+  );
+}
+
 function renderMap(position: LatLngTuple) {
   // Query param for https://developers.google.com/maps/documentation/urls/guide#search-action
   const googleMapQuery = encodeURIComponent(position.join(','));
+  const map = getMap(position);
```
schalkneethling: Attempting to get the `Map` object here so I can get access to `invalidateSize`. This does not work :(

www/src/js/views/venues/VenueLocation/VenueLocation.jsx
```diff
@@ -29,25 +29,55 @@ type State = {
...
 
+/**
+ * Switches the map between in content, and expanded
+ */
+function toggleMapSize() {
+  const mapWrapper = document.querySelector(`.${styles.mapWrapper}`);
+  const closeButton = mapWrapper.querySelector('.close');
+  const leafletControlContainer = document.querySelector('.leaflet-control-container');
+
+  mapWrapper.classList.toggle(styles.expanded);
```
schalkneethling: As these are elements I get from the DOM using standard DOM functions, these will not have `className` available. You mentioned something about `null` checks in order to make Flow happy. How would I go about implementing this? Thanks!

schalkneethling:
@ZhangYiJiang Thanks for pulling this all together. It was definitely more involved that I had envisioned, also still new to the world of React, and have never used Leaflet. This was a great learning experience. Happy to have contributed in a small way. 🎉

ZhangYiJiang:
GestureHandling plug-in needs to be disabled when map is expanded, otherwise using the map on phones when expanded will be awkward. There’s also a weird iOS issue where the map goes behind the other UI elements even though it has a higher z-index. 

May have to use portals to get around this.

ZhangYiJiang:
Found the cause. `.scrollable-y` is on one of the parent elements which uses `overflow-x: hidden`, which clips the map. Now I wonder why the other browsers *don't* have this issue...

Issue #1219, opened 2018-10-15
Find free room should use the current time when the box is opened, rather than when the page is loaded 


ZhangYiJiang:
On the Venue page, there's a button that allows students to find free classrooms. The box tries to be intelligent and use the current time (or closest future time if it is currently outside office hours). 

However, currently the box always uses the time when the page is loaded, rather than when the box is opened. This is because the current implementation calls `defaultSearchOptions()` when the component `state` is initialized. If the user has already been on the page for a while, then the time/date may be outdated. A good way to fix this is to update the search options when the button is clicked. Do take note not to clear any existing search options, eg. from query string. 

Component: https://github.com/nusmodifications/nusmods/blob/master/www/src/js/views/venues/VenuesContainer.jsx

ParkourKarthik:
I'll try to contribute on this one :)

ParkourKarthik:
@xit4 That's ok. I hope you got it done already :smile:

PR #1220, opened 2018-10-15
Find venue button uses the current time when clicked

xit4:
### Few things about my PR

- I chose to implement my function as a separate named function 'cause I believe it would have been _too long_ to be a arrow function
- ~I chose to implement a proper property function instead of an arrow function because me and some [other people](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) believe it is _better_~ apparently I can't bind class methods so I declared it as an arrow function instead.

### User flows
- User enters the page `/venues` with no parameters 
    -> clicks the "Find free rooms" button right away: search options set to current time (time he clicked the button)
    -> waits a few hours then clicks "Find free rooms": search options set to current time (time he clicked the button
    -> clicks the "Find free rooms" button right away waits a few hours then clicks it again: both calls are made with the search options set to the time the user clicked the button

If the user changes manually the search options, from that moment on, the search options will always be set to those chosen by the user. So unless he changes the options manually or refreshes the page back to `venues` he won't lose his precious parameters.
- User enters the page `/venues?day=0&duration=1&time=12` with parameters already set

whenever he clicks on the "Find free rooms" button the search options are going to be set to the parameters unless he manually changes the options when the box is open

Hopefully I did not confuse the reviewer with these "use cases". Anyway, I'm available for changes and clarifications.

this PR takes care of #1219

### DIFF
--- a/www/src/js/views/venues/VenuesContainer.jsx
+++ b/www/src/js/views/venues/VenuesContainer.jsx
@@ -76,7 +76,10 @@ export class VenuesContainerComponent extends Component<Props, State> {
       loading: true,
       venues: null,
       searchTerm: params.q || '',
+      pristineSearchOptions: isAvailabilityEnabled,
     };
+
+    this.onFindFreeRoomsClicked = this.onFindFreeRoomsClicked.bind(this);
   }
 
   componentDidMount() {
@@ -95,6 +98,16 @@ export class VenuesContainerComponent extends Component<Props, State> {
     }
   }
 
+  onFindFreeRoomsClicked() {
+    const { pristineSearchOptions, searchOptions, isAvailabilityEnabled } = this.state;
+    const newSearchOptions = pristineSearchOptions ? defaultSearchOptions() : searchOptions;
+    const stateUpdate = {
+      isAvailabilityEnabled: !isAvailabilityEnabled,
+      searchOptions: newSearchOptions,
+    };
+    this.setState({ ...stateUpdate });
+  }
+
   onClearVenueSelect = () =>
     this.props.history.push({
       ...this.props.history.location,
@@ -109,7 +122,10 @@ export class VenuesContainerComponent extends Component<Props, State> {
 
   onAvailabilityUpdate = (searchOptions: VenueSearchOptions) => {
     if (!isEqual(searchOptions, this.state.searchOptions)) {
-      this.setState({ searchOptions });
+      this.setState({
+        searchOptions,
+        pristineSearchOptions: false, // user changed searchOptions
+      });
     }
   };
 
@@ -174,7 +190,7 @@ export class VenuesContainerComponent extends Component<Props, State> {
             styles.availabilityToggle,
             isAvailabilityEnabled ? 'btn-primary' : 'btn-outline-primary',
           )}
-          onClick={() => this.setState({ isAvailabilityEnabled: !isAvailabilityEnabled })}
+          onClick={this.onFindFreeRoomsClicked}
         >
           <Clock className="svg" /> Find free rooms
         </button>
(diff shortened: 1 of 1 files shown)


www/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -95,6 +97,16 @@ export class VenuesContainerComponent extends Component<Props, State> {
...
   }
 
+  onFindFreeRoomsClicked = () => {
+    const { pristineSearchOptions, searchOptions, isAvailabilityEnabled } = this.state;
+    const newSearchOptions = pristineSearchOptions ? defaultSearchOptions() : searchOptions;
+    const stateUpdate = {
+      isAvailabilityEnabled: !isAvailabilityEnabled,
+      searchOptions: newSearchOptions,
+    };
+    this.setState({ ...stateUpdate });
```
ZhangYiJiang: This spread operator seems unnecessary

www/src/js/views/venues/VenuesContainer.jsx
```diff
@@ -95,6 +97,16 @@ export class VenuesContainerComponent extends Component<Props, State> {
     }
   }
 
+  onFindFreeRoomsClicked = () => {
+    const { pristineSearchOptions, searchOptions, isAvailabilityEnabled } = this.state;
+    const newSearchOptions = pristineSearchOptions ? defaultSearchOptions() : searchOptions;
```
ZhangYiJiang: Personally I feel this is clearer if it is written imperatively, ie. 

``` js
const update = { isAvailabilityEnabled: !isAvailabilityEnabled };

if (pristineSearchOptions && !isAvailabilityEnabled) {
    update.searchOptions = defaultSearchOptions();
}
```

(also does not try to reset the options when closing the box, but that's a micro-optimization) 

@li-kai  What do you think?

ZhangYiJiang:
For method binding, we have Stage 3 class properties enabled, so that's unnecessary. Using arrow functions is correct :)

www/src/js/views/venues/VenuesContainer.jsx
xit4: It totally is. It's a leftover from a previous implementation attempt (I'll fix that right away)

www/src/js/views/venues/VenuesContainer.jsx
xit4: I do agree the imperative way looks way cleaner. I can change it right away if you agree.

ZhangYiJiang: Yes, I think that would be good. Thanks!

ZhangYiJiang:
@xit4 Ah, there's a Flow error because Flow doesn't allow additional undeclared properties on object literals. You can declare the type of `stateUpdate` to be 

``` js
const stateUpdate: $Shape<State>
``` 

Also would be a good idea to add a comment about why `searchOption` don't need to be reset if `!isAvailabilityEnabled`. Sorry about this!

xit4:
> @xit4 Ah, there's a Flow error because Flow doesn't allow additional undeclared properties on object literals. You can declare the type of `stateUpdate` to be
> 
> ```js
> const stateUpdate: $Shape<State>
> ```
> 
> Also would be a good idea to add a comment about why `searchOption` don't need to be reset if `!isAvailabilityEnabled`. Sorry about this!

I was just about to tackle that. Thanks for the tip

ZhangYiJiang:
@xit4 Actually I just realized I can push to your branch, since I've already fixed it I'll just push it up and merge. You don't need to do anything. Thanks again!

xit4:
Beat me to it 😛 

Thank you for the review!

Issue #1136, opened 2018-08-30
Use a tombstone to represent removed modules


ZhangYiJiang:
Currently when you remove a module from the timetable, we use the snackbar to give the user an 'Undo' button to restore the module, in case they made a mistake. This works well, except 

1. The snackbar disappears after a while 
2. On Safari Mobile, we can't use Snackbar with action because they require two taps, so we use an alert instead, but this is less user friendly 

An alternative is to replace the module with a tombstone - a box that remains at where the removed module is, with a button that allows the user to undo. The tombstone doesn't stay there forever - it's gone the next time you navigate to the timetable, or if you press the close button, but before that it stays on screen, giving the user as much time as they need to undo their action. An example of this is what you get when you dismiss a sidebar ad on YouTube (although YouTube doesn't allow you to dismiss the tombstone). 

Here's where you can start - https://github.com/nusmodifications/nusmods/blob/master/www/src/js/views/timetable/TimetableModulesTable.jsx . The tombstone may need to be stored in the Redux store, as adding a module or deleting another module should clear the tombstone. Alternatively using local component State may also work, since the timetable table can check when timetable modules change.

jiholim:
Mockup for reference - 

![banner](https://user-images.githubusercontent.com/25242767/44856825-034ead00-aca1-11e8-839a-4150c8775850.png)

mrassili:
Hi, Marouane here. I'll be working on this issue if you allow me

ZhangYiJiang:
@mrassili  Sure, please go ahead

PR #1231, opened 2018-10-17
Add a tombstone component to the modules table (instead of snackbar)

mrassili:
Pull Request for #1136 

Done : 
• Create a ModuleTombstone component that receives a `tombstone` prop from TimetableContent > TimetableModulesTable
• Make it `connected` to Redux in order to gain access to the `undo` action
• Made changes to the CSS to match the provided mockup
• Used `setTimeout` to clear the tombstone after a certain time (12s)
• Write unit tests for the tombstone component

Please help me with your suggestions. And if there's any buggy behavior.


www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -76,14 +76,29 @@ type Props = {
...
   state: State = {
     isScrolledHorizontally: false,
     showExamCalendar: false,
+    tombstone: {
+      moduleCode: '',
+      semester: null,
+    },
   };
 
+  componentDidUpdate(prevProps, prevState) {
```
ZhangYiJiang: This should not be necessary. One of the benefits of using tombstones is to avoid the problem of the notification disappearing after 12s

ZhangYiJiang: In addition, even if we do want it to disappear after 12s, it would be better to write this timeout with the `setState` code instead of here

www/src/js/views/timetable/TimetableContent.jsx
```diff
@@ -76,14 +76,29 @@ type Props = {
...
   isScrolledHorizontally: boolean,
   showExamCalendar: boolean,
+  tombstone: Tombstone,
 };
 
 class TimetableContent extends Component<Props, State> {
   state: State = {
     isScrolledHorizontally: false,
     showExamCalendar: false,
+    tombstone: {
```
ZhangYiJiang: I think it would be better to make this nullable, and the default value be null

www/src/js/views/timetable/TimetableModulesTable.jsx
```diff
@@ -99,51 +102,81 @@ class TimetableModulesTable extends Component<Props> {
...
+        </div>
+      </div>
+    );
+  }
+
   render() {
+    const { semester, horizontalOrientation, tombstone } = this.props;
+
     if (!this.props.modules.length) {
+      if (tombstone.moduleCode !== '') {
```
ZhangYiJiang: There's a lot of code duplication here. There should be a easier way to write this without duplicating the tombstone code

www/src/js/views/timetable/ModuleTombstone.scss
```diff
@@ -0,0 +1,47 @@
+@import '~styles/utils/modules-entry.scss';
```
ZhangYiJiang: This style file does not appear to be imported anywhere

www/src/js/views/timetable/ModuleTombstone.jsx
```diff
@@ -0,0 +1,60 @@
...
+              onClick={() => {
+                props.undo();
+                props.resetTombstone();
+              }}
+            >
+              Undo
+            </button>
+          </div>
+        </div>
+        <span>{`Removed ${tombstone.moduleCode}`}</span>
```
ZhangYiJiang:
```suggestion
        <span>Removed {tombstone.moduleCode}</span>
```

www/src/js/views/timetable/ModuleTombstone.jsx
```diff
@@ -0,0 +1,60 @@
...
+  horizontalOrientation: boolean,
+  undo: () => void,
+  resetTombstone: Function,
+};
+
+function ModuleTombstone(props: Props) {
+  const { tombstone, horizontalOrientation } = props;
+
+  return (
+    <div
```
ZhangYiJiang: This wrapper and the `moduleTableRow` below should not be necessary.

www/src/js/views/timetable/ModuleTombstone.jsx
```diff
@@ -0,0 +1,60 @@
...
+      className={classnames(styles.modulesTableRow, 'col-sm-6', {
+        'col-lg-4': horizontalOrientation,
+        'col-md-12': !horizontalOrientation,
+      })}
+    >
+      <div className={styles.moduleInfo}>
+        <div className={styles.moduleActionButtons}>
+          <div className="btn-group">
```
ZhangYiJiang: Can use two `btn-link` instead of two buttons in a button group

www/src/js/views/timetable/ModuleTombstone.jsx
```diff
@@ -0,0 +1,60 @@
...
+            <button
+              type="button"
+              className={classnames('btn btn-outline-secondary btn-svg', styles.moduleAction)}
```
ZhangYiJiang: `.btn-svg` should only be used for SVG buttons

www/src/js/views/timetable/TimetableContent.jsx
mrassili: Removed

www/src/js/views/timetable/TimetableContent.jsx
mrassili: Changed the tombstone state to be nullable (maybe type) and tweaked the logic to make it less verbose

www/src/js/views/timetable/ModuleTombstone.jsx
mrassili: I wanted to make the rows consistent so I opted for the provided wrappers. I can't think of a way to accomplish this otherwise, any suggestions?

www/src/js/views/timetable/ModuleTombstone.jsx
mrassili: Replaced `.btn-svg` with `.btn-link` (for the next commit)

www/src/js/views/timetable/ModuleTombstone.scss
mrassili: I have no idea where this came from?!

www/src/js/views/timetable/ModuleTombstone.jsx
mrassili: Replaced

ZhangYiJiang: I don't see your changes here. Did you forget to push your changes?

mrassili: Yes sorry, I just pushed the changes

ZhangYiJiang:
@mrassili I took another look at the code, and came up with a better approach - instead of using a special Tombstone object, we just save the module data itself. In `<TimetableModuleTable>` we insert the tombstone back into the array of modules so the tombstone is sorted back into the correct position, then when rendering we check and display the tombstone component if the item is the tombstone element. 

This preserves the original position of the module when it was deleted, which is quite nice UI wise. I've also touched up on the styling and markup a bit to make it look nicer. Note that in Bootstrap `btn-link` and `btn-secondary-outline` shouldn't be applied at the same time, otherwise the result is quite weird. Also the `title` and `aria-label` attributes are unnecessary because the buttons already have text in them.

mrassili:
Hi @ZhangYiJiang that's clever, I thought of adding the tombstone module to the modules array but they had props that aren't relevant to a tombstone (like color)
Is there anything I should do on my part now?

ZhangYiJiang:
@mrassili You can help test the feature and look through the code

ZhangYiJiang:
@li-kai  Hahaha, the duplicate footer bug is my fault, and it's there since that footer feature was merged in. I can't believe I only found it now

ZhangYiJiang:
Okay, here's the current solution - 

- Footer is extracted to its own component. This is how it should have been written, but I got lazy. This stops the footer from appearing in every section 
- Tombstone will now only appear in the non-conflict modules section. This means that if you have a three-way conflict and delete one of the mods, the tombstone will appear in the "wrong" section, but this scenario is rate enough that I think we can accept this. Otherwise if you have a two-way conflict, deleting one of the modules will resolve the conflict anyway, so this is fine.

Issue #1214, opened 2018-10-11
Upgrade flow types to use exact types


li-kai:
Currently many of our [types](https://github.com/nusmodifications/nusmods/tree/master/www/src/js/types) look like the following:

```
export type ModuleSearch = {
  term: string,
  tokens: string[],
};
```

We should, however, change them to use [exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types) whenever possible:
```
export type ModuleSearch = {|
  term: string,
  tokens: string[],
|};
```

Then, to further make it explicit that these types should be read-only, set them to be [covariant](https://medium.com/@forbeslindesay/covariance-and-contravariance-c3b43d805611):
```
export type ModuleSearch = {|
  +term: string,
  +tokens: string[],
|};
```

This issue is extremely broad. We will accept PRs that fixes any amount types, but please do not split them up just for hacktoberfest contribution.

nirmalks:
I would like to work on this

li-kai:
@nirmalks All yours for now!

nirmalks:
Hi, 
Any pointers on what sort of tests should be made? Should I include test cases?

li-kai:
@nirmalks As long as `yarn flow` command passes, you're good.

nirmalks:
@li-kai Ok, Thanks. Will work on it and get back.

PR #1232, opened 2018-10-17
Upgrade flow types to exact object types

nirmalks:
* Change flow types to exact object types
* Add covariance to properties
* Fix related to https://github.com/nusmodifications/nusmods/issues/1214

### DIFF
--- a/www/src/js/types/modules.js
+++ b/www/src/js/types/modules.js
@@ -55,20 +55,20 @@ export type RawLesson = {
 };
 
 // Semester-specific information of a module.
-export type SemesterData = {
-  ExamDate?: string,
-  LecturePeriods: Array<string>,
-  Semester: Semester,
-  Timetable: Array<RawLesson>,
-  TutorialPeriods?: Array<string>,
-};
+export type SemesterData = {|
+  +ExamDate?: string,
+  +LecturePeriods: Array<string>,
+  +Semester: Semester,
+  +Timetable: Array<RawLesson>,
+  +TutorialPeriods?: Array<string>,
+|};
 
 // Recursive definition for walking a module tree
-export type TreeFragment = {
-  name: string,
+export type TreeFragment = {|
+  +name: string,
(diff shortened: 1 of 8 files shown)


www/src/js/types/modules.js
```diff
@@ -97,11 +97,11 @@ export type ModuleWithColor = Module & {
 };
 
 // This format is returned from the module list endpoint.
-export type ModuleCondensed = {
+export type ModuleCondensed = {|
   ModuleCode: ModuleCode,
```
ZhangYiJiang: Just curious, why were these not changed to read only?

nirmalks: I noticed some errors, for example. 

Cannot return object literal because:
 • property ModuleCode is read-only in ModuleCondensed [1] but writable in
   object type [2] in array element of property modules.

Any suggestions on how to approach this?

ZhangYiJiang: Can you comment with the file and line so we can take a look? We were probably constructing ModuleCondensed objects somewhere, and Flow doesn't like that

nirmalks: https://github.com/nusmodifications/nusmods/blob/050b3f34164766d3280f1d307872fea850a67053/www/src/js/views/layout/GlobalSearchContainer.jsx#L83:10

ZhangYiJiang: Okay, I know what's wrong. `moduleSearch`'s signature is 

```
sortModules<T: { ModuleCode: ModuleCode, ModuleTitle: ModuleTitle }>(
  searchTerm: string,
  modules: T[],
)
```

Because the properties in `T` are not declared covariant, Flow doesn't trust the function to not change them (even though it doesn't). Just had to update the function signature - already pushed the fix to master

nirmalks: Ok, I wasn't aware about it. Thank you.

Issue #1074, opened 2018-07-30
Replace Google Analytics with self hosted solutions 


ZhangYiJiang:
Google Analytics is great, but we don't really use a lot of its features. We mostly use it to track user browser (to update our support matrix), and to track events (to see if students are using certain featuees). In addition, using a third-party to collect user data is somewhat sketchy privacy wise, especially since Google explicitly uses GA data to support its own advertising by tracking users across multiple sites. In addition with more and more students using adblockers, it is possible we aren't collecting good data due to GA being blocked (I know my adblocker does block GA). 

We can try self hosting our analytics instead to improve privacy and address the concerns above. Here are the candidates I've found - feel free to edit this issue to add more 

## [Fathom](https://github.com/usefathom/fathom) 

Pros

- Lightweight - written in Go, so should be fast and reasonably simple to set up 
- Uses MySQL - no additional set up needed since we already have a MySQL installation 

Cons 

- No explicit SPA support - tracking code has no documentation, but from reading the source there's no explicit way to send pageviews I think 
- New - not 1.0 yet. May be unstable 

## [Mamoto (formerly Piwik)](https://matomo.org/) 

Pros 

- Old, well known - Mamoto is really old. That makes it pretty stable
- Uses MySQL - no additional set up needed since we already have a MySQL installation 
- SPA support - not particularly easy to use, but shouldn't be too hard either 

Cons 

- No Docker - it's a PHP app, so updating probably won't be too problematic, but still worth taking note 
- No npm package - analytics JS has to be served separately. Not a huge deal, and probably preferable from a performance point of view anyway

ZhangYiJiang:
Fathom doesn't work because we need browser / device data, and they don't have that. Testing Mamoto now.

PR #1234, opened 2018-10-18
Enable Mamoto analytics

ZhangYiJiang:
Part of #1074 

Mamoto (formerly Piwik) is a self-hosted analytics service that is more privacy friendly, allowing us better control over user data while also making it more open. The service is currently running at https://analytics.nusmods.com/. 

This PR adds Mamoto tracking, and we'll see if it works as expected. If it does, we'll probably keep GA for one release, then remove it in the next to ensure we're not missing anything.

### DIFF
--- a/www/src/js/utils/insertScript.js
+++ b/www/src/js/utils/insertScript.js
@@ -1,12 +1,25 @@
 // @flow
 
-export default function insertScript(scriptSrc: string, scriptId: string, scriptAsync: boolean) {
-  const script = window.document.createElement('script');
-  script.src = scriptSrc;
-  script.id = scriptId;
-  script.async = scriptAsync;
+type ScriptOptions = {
+  id?: string,
+  async?: boolean,
+  defer?: boolean,
+};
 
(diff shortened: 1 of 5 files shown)


www/src/js/utils/insertScript.js
```diff
@@ -1,12 +1,25 @@
...
-export default function insertScript(scriptSrc: string, scriptId: string, scriptAsync: boolean) {
-  const script = window.document.createElement('script');
-  script.src = scriptSrc;
-  script.id = scriptId;
-  script.async = scriptAsync;
-
-  if (window.document.body) {
-    window.document.body.appendChild(script);
-  }
+type ScriptOptions = {
```
li-kai:
```suggestion
type ScriptOptions = {|
  +id?: string,
  +async?: boolean,
  +defer?: boolean,
|};
```

ZhangYiJiang: I tried that, but it doesn't seem compatible with optional props. Flow will complain that the prop doesn't exist

PR #1236, opened 2018-10-19
Beta test toggle and initial today page implementation

ZhangYiJiang:
Initial, incomplete implementation of the today page. Implemented the second design here 

layout1 layout2

Feature is currently hidden behind a toggle on the settings page. Code is a bit messy right now. 

Todo

- [x] Implement current time 
- [x] Refactor code for handling current day and fix "no lesson" message 
- [x] Handle odd / even / numbered weeks
- [ ] Add venue map


www/src/js/utils/timify.js
```diff
@@ -82,7 +97,7 @@ export function getCurrentMinutes(
 }
 
 // Monday = 0, Friday = 4, Sunday = 6
-export function getCurrentDayIndex(now: Date = new Date()): number {
+export function getDayIndex(now: Date = new Date()): number {
```
li-kai: Logic can be replaced with https://date-fns.org/v1.29.0/docs/getISODay

www/src/js/views/today/TodayContainer.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+  renderBeforeNextLessonCard(nextLesson: Lesson, marker: Node) {
+    const nextLessonDate = getStartTimeAsDate(nextLesson);
+    const hoursTillNextLesson = differenceInHours(nextLessonDate, new Date());
+
+    let comment = null;
+    if (hoursTillNextLesson <= 1) {
+      comment = <p>Better get a move on to your next class!</p>;
+    } else if (new Date().getHours() < 7) {
+      // Why are you up right now?
```
li-kai: Sleep is for the weak /s

www/src/js/reducers/settings.js
```diff
@@ -91,6 +93,12 @@ function settings(state: SettingsState = defaultSettingsState, action: FSA): Set
         moduleTableOrder: action.payload,
       };
 
+    case TOGGLE_BETA_TESTING_STATUS:
+      return {
+        ...state,
+        beta: !(state.beta || false),
```
li-kai: Not sure why you need the conditional, the default is already set.
```suggestion
        beta: !state.beta,
```

ZhangYiJiang: Let me double check, but I think this was to make Flow happy

ZhangYiJiang: Ah no, this works. Nice!

PR #1247, opened 2018-10-23
Truncate contributor usernames that are long

liaujianjie:
### Before

image

### After

image

### DIFF
--- a/www/src/js/views/static/ContributorsContainer.jsx
+++ b/www/src/js/views/static/ContributorsContainer.jsx
@@ -7,6 +7,7 @@ import Loader from 'views/components/LoadingSpinner';
 import ExternalLink from 'views/components/ExternalLink';
 
 import StaticPage from './StaticPage';
+import styles from './ContributorsContainer.scss';
 
 const CONTRIBUTORS_URL =
   'https://api.github.com/repos/NUSModifications/NUSMods/contributors?per_page=100';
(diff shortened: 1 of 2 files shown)


ZhangYiJiang:
The PR looks good, although I would tweak it using the patch below (I would push it to this branch, but I don't seem to have permission) 

The changes to the HTML just removes some redundant HTML since we're now using custom CSS  

``` patch
@@ -97,20 +97,14 @@
           <div className="row">
             {this.state.contributors.map((contributor) => (
               <div className="col-md-3 col-6 text-center" key={contributor.id}>
-                <div>
-                  <ExternalLink href={contributor.html_url}>
-                    <img
-                      src={contributor.avatar_url}
-                      alt={`${contributor.login} thumbnail`}
-                      className="rounded-circle img-fluid img-thumbnail"
-                    />
-                  </ExternalLink>
-                </div>
-                <div className="font-weight-bold">
-                  <ExternalLink className={styles.contributorUsername} href={contributor.html_url}>
-                    {contributor.login}
-                  </ExternalLink>
-                </div>
+                <ExternalLink href={contributor.html_url}>
+                  <img
+                    src={contributor.avatar_url}
+                    alt={`${contributor.login} thumbnail`}
+                    className="img-fluid img-thumbnail"
+                  />
+                  <span className={styles.contributorUsername}>{contributor.login}</span>
+                </ExternalLink>
                 <p>
                   <ExternalLink
                     className="text-muted"
```

Using inline-block + max-width here instead so that to allow implementing in the future https://stackoverflow.com/questions/44024946/add-a-title-property-to-elements-with-text-overflow-ellipsis#44025308, though this is somewhat tough since we're using React. 

``` patch
@@ -1,7 +1,8 @@
 .contributorUsername {
-  display: block;
-  width: 100%;
+  display: inline-block;
+  max-width: 100%;
+  margin: 0.8rem 0 -0.4rem;
   overflow: hidden;
-  white-space: nowrap;
+  font-weight: bold;
   text-overflow: ellipsis;
 }
```

PR #1252, opened 2018-10-25
Basic Today page tests and bug fixes 

ZhangYiJiang:
Fix a rather bad bug introduced in the last commit on the today page PR, and set up some basic tests to help with development. The tests are very simple right now, but the page is still under construction so I don't want to write too much. 

Also implement simple error handling and validation for the weather API, and extract interval timer into a HOC, which makes testing components that use timers much simpler since we can mock time now.

### DIFF
--- a/www/src/js/views/hocs/withTimer.jsx
+++ b/www/src/js/views/hocs/withTimer.jsx
@@ -2,15 +2,15 @@
 import React, { Component } from 'react';
 import { wrapComponentName } from 'utils/react';
 
-type State = {|
+export type TimerData = {|
   currentTime: Date,
 |};
 
 function withTimer<Props: {}>(
   WrappedComponent: ComponentType<Props>,
   intervalInMs: number = 60 * 1000,
-): ComponentType<$Diff<Props, State>> {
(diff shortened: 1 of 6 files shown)


www/src/js/views/hocs/withTimer.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+export type TimerData = {|
+  currentTime: Date,
+|};
+
+function withTimer<Props: {}>(
+  WrappedComponent: ComponentType<Props>,
+  intervalInMs: number = 60 * 1000,
+): ComponentType<$Diff<Props, TimerData>> {
+  return class extends Component<Props, TimerData> {
```
li-kai: Could we implement https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API here as well? This prevents nusmods from showing the “something changed” glow in Firefox.

ZhangYiJiang: I thought that glow comes from changing the page `<title>`. Do you mean you want to force a refresh when the user goes back to the tab after the tab has been inactive for a while?

PR #1259, opened 2018-10-27
Add support for customizing height width in exporter (#1227)

### LINKED ISSUE #1227: Add support for customizing width and height to exporter service
As part of #1003, before we can build the frontend for this, we need first a way for the service to return images of different sizes. To do that, we can let the service accept some additional query params and pass that to the render method. 

To get started, take a look at https://github.com/nusmodifications/nusmods/tree/master/export for the instructions to setting up the service. If you have any question, just @ me since I built this.

JingYenLoh:
Resolves #1227 

Not sure sure if this is the right way, because I ignore the device scale factor if there's height and width in query params


export/app.js
```diff
@@ -42,12 +42,17 @@ const router = new Router();
...
+    const { height, width } = ctx.query;
 
     // Validate options
-    const options = {
+    let options = {
       pixelRatio: _.clamp((Number(ctx.query.pixelRatio) || 1), 1, 3),
     };
 
+    if (height && width) {
+      options = { ...options, height, width };
```
ZhangYiJiang: Casting to number should be done here. There should also be a check here to make sure the height and width are numbers and positive. Remember that this service is exposed publicly, and the service should not crash on invalid input.

export/render.js
```diff
@@ -78,8 +78,11 @@ async function injectData(page, data) {
 }
 
 async function image(page, data, options = {}) {
-  if (options.pixelRatio) {
-    await setViewport(page, Number(options.pixelRatio));
+  if (options.height && options.width) {
```
ZhangYiJiang: width and height shouldn't override pixelRatio. Try changing `setViewport` to use an option object instead

PR #1260, opened 2018-10-28
Implement cache eviction logic for moduleBank

### LINKED ISSUE #1245: Implement cache eviction logic for moduleBank
Currently when modules are loaded, they are stored and persisted in the moduleBank reducer state (https://github.com/nusmodifications/nusmods/blob/master/www/src/js/reducers/moduleBank.js#L48) to improve load performance and for offline use. We currently use localStorage for all client-side persistence. It is in theory possible to fill up the space if the user keeps opening new modules, but in practice this almost never happens. That said, I recently noticed a quota exceeded error on Sentry which indicates that someone (or a bot) really did hit 5MiB, so we should consider implementing this. 

We should be able to set a cap on the size of the cache (either by number of modules or the length of the `JSON.stringifiy` string). Cache eviction can be done using least recently used (LRU) logic, though that takes some additional work as we don't record when the cache was refreshed for each module. This information can also be used in the future to improve our data loading logic. An easier alternative is to do random evictions, though that's less user friendly.

raiden-x:
Fix #1245


www/src/js/reducers/moduleBank.js
```diff
@@ -45,21 +77,47 @@ function moduleBank(state: ModuleBank = defaultModuleBankState, action: FSA): Mo
...
     case SET_EXPORTED_DATA:
       return {
         ...state,
         modules: keyBy(action.payload.modules, (module: Module) => module.ModuleCode),
       };
 
+    case FETCH_MODULE + FAILURE: {
+      const moduleCode = get(action, ['meta', API_REQUEST], '').split('_')[2];
+      // Update the timestamp to keep track of when the module was last accessed even when the call fails if it is in state/cache
+      if (state.modules[moduleCode]) {
```
ZhangYiJiang: This should be inverted so that the golden path is less nested, ie. `if (!state.modules[moduleCode]) return state`

www/src/js/reducers/moduleBank.js
```diff
@@ -45,21 +77,47 @@ function moduleBank(state: ModuleBank = defaultModuleBankState, action: FSA): Mo
...
+    }
 
     case SET_EXPORTED_DATA:
       return {
         ...state,
         modules: keyBy(action.payload.modules, (module: Module) => module.ModuleCode),
       };
 
+    case FETCH_MODULE + FAILURE: {
+      const moduleCode = get(action, ['meta', API_REQUEST], '').split('_')[2];
```
ZhangYiJiang: There's probably a better way to get the module code. Maybe add it to the `meta` in `fetchModule`? Or maybe create a new action for updating the timestamp, so the logic isn't duplicated

raiden-x:
Thanks for reviewing and i'll work on the issues that you have mentioned.

raiden-x:
I have made some changes by creating new actions for updating the timestamp and trimming the modules. A test is failing since the fetchModule has been rewritten to return a thunk instead. I will fix them if these changes are acceptable. Thanks for helping out with thunk.

www/src/js/actions/moduleBank.js
```diff
@@ -17,8 +18,62 @@ export function fetchModuleRequest(moduleCode: ModuleCode) {
...
+  return {
+    type: REMOVE_LRU_MODULE,
+    payload: moduleCode,
+  };
+}
+
+function getLRUModule(modules, lessons, currentModule) {
+  // Pull all the modules in all the timetables
+  // used parseInt since object.keys returns array of strings even if the key is a integer
+  const timeTableModules = Object.keys(lessons).reduce(
```
ZhangYiJiang: If I understand this code correctly, it’s flattening nested object keys? In that case lodash flatMap is probably easier to read

www/src/js/actions/moduleBank.js
```diff
@@ -17,8 +18,62 @@ export function fetchModuleRequest(moduleCode: ModuleCode) {
...
+    (moduleList, semester) => [...moduleList, ...Object.keys(lessons[parseInt(semester, 10)])],
+    [],
+  );
+  const sortedModules = Object.keys(modules).sort(
```
ZhangYiJiang: As mentioned previously, this can be replaced with `minBy`, although I'd prefer if this returned an array of the `n` least recently used modules

www/src/js/actions/moduleBank.js
```diff
@@ -17,8 +18,62 @@ export function fetchModuleRequest(moduleCode: ModuleCode) {
...
+const MAX_MODULE_LIMIT: number = 100;
+export function fetchModule(moduleCode: ModuleCode) {
+  return (dispatch: Function, getState: GetState) =>
+    dispatch(
+      requestAction(fetchModuleRequest(moduleCode), FETCH_MODULE, {
+        url: NUSModsApi.moduleDetailsUrl(moduleCode),
+      }),
+    ).finally(() => {
+      // Update the timestamp of the accessed module if it is in the store.
+      if (getState().moduleBank.modules[moduleCode]) {
```
ZhangYiJiang: Need to double check if this will run before or after the moduleBank actions - if this runs before, then the timestamp will not be correctly set when the module is loaded for the first time

raiden-x: This does get called after the newly loaded module is in the store and the timestamp did get updated for the case of success and failure also.

www/src/js/actions/moduleBank.js
raiden-x: Right now sortedModules returns the sorted array based on the condition and i'll replace it with the appropriate lodash function. So it does return `n` recently used modules n being the entire size here.
Maybe we can remove more than 1 module at the time so that this removeLRUmodules function runs with less frequency.

raiden-x:
i tried adding a test module for the getLRUModule in the actions test file. I'm unsure as to how to go about mocking the entire module and the timetableconfig data structure.

ZhangYiJiang:
A `TimetableConfig` is just a mapping of lessons to modules to semesters. Since in this case we don't care about lessons, and also the contents of the `Module`, you can force Flow to accept the code by typing the objects as `any`. This is of course unsafe, but we don't care about that as much for test code. 

```js
  /* eslint-disable no-useless-computed-key */
  const timetableConfig: TimetableConfig = {
    [1]: {
      CS1010S: ({}: any),
    },
  };
  /* eslint-enable */
  const modules = {
    CS1010S: ('CS1010S': any),
  };
```

PR #1309, opened 2018-12-04
Fix venue name capitalizations

### LINKED ISSUE #1213: Correct venue names to use Title Case instead of ALL CAPS
ALL OF THE- sorry, my caps lock got stuck. All of the venues pulled from the NUS venue APIs currently have ALL CAPS venue names. This may work for them, but it looks pretty ugly for us. All caps text is also harder to read. 

It should be possible to automatically convert names like "SEMINAR ROOM 2" to "Seminar Room 2". 
Currently we use lodash to capitalize the text on the frontend, and the result is not optimal because things like abbreviations are incorrectly converted to lowercase eg. https://nusmods.com/venues/WS2-01-30. 

A good way to fix this is to write a script to clean up the data in venues.json https://github.com/nusmodifications/nusmods/blob/master/www/src/js/data/venues.json. Doing this offline allows us to use a proper dictionary to correctly determine which words need to be capitalized. Do note that some venue names should not be capitalized, like the i3 building.

ZhangYiJiang:
Fixes #1213 

General approach is 

- Each word is capitalized if the word is in the dictionary 
- Otherwise we leave the word alone 
- Special exceptions are made for words like 'i3' 
- Dictionary is the standard Unix words file, supplemented with a few jargon like *biomolecular* and some British English spelling like *honours*

### DIFF
--- a/www/scripts/fix-venue-name.js
+++ b/www/scripts/fix-venue-name.js
@@ -1,9 +1,25 @@
+/**
+ * Simple script to correctly capitalize venue names coming from
+ * NUS venues API, which only has names in ALL CAPS. This makes the
+ * venue names more human readable.
+ *
+ * It tries to match works in the name with dictionary words, and
+ * assume anything that's not to be an acronym or abbreviation and
+ * leaves them untouched.
+ */
 const venueFile = '../src/js/data/venues.json';
 
 const venues = require(venueFile); // eslint-disable-line import/no-dynamic-require
 const _ = require('lodash');
 const fs = require('fs-extra');
 const path = require('path');
+
+if (process.platform === 'win32') {
+  console.warn(
+    'This script will probably not work on Windows as ' +
+      'it relies on the Unix words dictionary file',
+  );
+}
 
 // List of special names, mapping all lower version of the name to
 // the properly cased version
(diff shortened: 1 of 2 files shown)


www/scripts/fix-venue-name.js
```diff
@@ -0,0 +1,74 @@
...
+let dict;
+dictLocations.forEach((dictionary) => {
+  try {
+    dict = new Set(fs.readFileSync(dictionary, 'utf-8').split('\n'));
+  } catch (e) {
+    // Swallow error - will throw at the end of loop
+  }
+});
+
+if (!dict) throw new Error('Cannot find words file');
```
li-kai: Might want to specify that the script only runs on a unix system

www/scripts/fix-venue-name.js
```diff
@@ -0,0 +1,74 @@
...
+additionalWords.forEach((word) => dict.add(word.toLowerCase()));
+
+_.each(venues, (venue) => {
+  if (!venue.roomName) return;
+  const newRoomName = venue.roomName.replace(/\b\w+\b/g, (word, index) => {
+    const lowercaseWord = word.toLowerCase();
+
+    if (specialNames[lowercaseWord]) {
+      return specialNames[lowercaseWord];
+    } else if (index !== 0 && articles.has(lowercaseWord)) {
```
li-kai: Could you move this into the next else if? Blank else if statements are hard to understand.

PR #1310, opened 2018-12-05
Remove flow-scripts to use flow-typed instead

ZhangYiJiang:
Flow scripts is nice, but now that we have several dependencies that we don't want stubs automagically generated for, and since our dependencies don't change all that often, I think we can try moving to manual libdef management using flow-typed instead. Thoughts?


li-kai:
Might be a bit hard to onboard people if we have to do manual dependency maintenance.

ZhangYiJiang:
I feel it's better than the current situation where you have to tell everybody who uses `yarn install` to remove the unnecessary stub that's always autogenerated

PR #1348, opened 2018-12-26
Much animate such wow

ZhangYiJiang:
Experimenting with animation using [React Flip Toolkit](https://github.com/aholachek/react-flip-toolkit)


taneliang:
When you first open a venue on mobile, the timetable immediately animates from horizontal to vertical. So, 2 things:

1. It may be best to remove the horizontal to vertical animation entirely? On the timetable page, it's a little strange to have the hours and the modules in the module list flying across the timetable, and it also causes a perceivable lag on my Android phone.
2. I don't think the timetable should animate as you move between venues on the venue page, because the lessons aren't animating and it's weird to have lessons appearing with the timeline moving/shrinking below them. Best to have everything just appear so that it feels snappy.

ZhangYiJiang:
I'm not sure why @li-kai  approved this :sweat_smile: . This is just a proof of concept, and there are too many problems to move forward. 

The big issue is that modules are positioned using margins. This makes them very hard to tween, because the animation has to take into account the new element being inserted as well as the margin, and as far as I can tell there's nothing that can do this correctly. Using AP would solve a lot of problem, but introduce a ton of others, such as how we can get the parent container to have the correct size given that the child elements now take up no space in the layout (invisible elements could work, but that's still pretty silly).

li-kai:
Ah. There wasn’t any WIP or bugs that I could find...
I’d admit the experience isn’t the best but it’s no worse than what we have now. If we rough out the edges that’ll be ideal. We can keep working on this on future PRs. Let’s merge this and keep it under beta?

ZhangYiJiang:
Closed because the code is out of date and the prototype too buggy to be released

PR #1351, opened 2018-12-26
Test CSS scroll snap

ZhangYiJiang:
Scroll snap is an interesting new CSS property. Let's test it out on the timetable to see if this makes it better


taneliang:
Nah it isn't better in its current state. It's very strange to have the timetable only snap to lessons on the left edge. It may be better to be able to snap to every hour, but I don't think it'll be any better than just not snapping at all.

PR #1362, opened 2018-12-31
[#1289] Add support for detecting and matching OS dark mode

### LINKED ISSUE #1289: Add support for detecting OS dark mode
Safari 12 TP 68 on macOS Mojave [allows websites to detect OS dark mode](https://kevinchen.co/blog/support-macos-mojave-dark-mode-on-websites/) using the [proposed CSS Media Query level 5 `prefers-color-scheme` media query](https://drafts.csswg.org/mediaqueries-5/#descdef-media-prefers-color-scheme).  

We can look into supporting this. Here's my proposal: 

- For browsers that support it, we add a third, default option to the night mode toggle - prefer OS default. The other two option will force the site to always use light or dark mode 
- We can set the site's mode when the page is first loaded, and use a media query listener to toggle if the user updates their preferences while the page is loaded

Calyeefornia:
Fixes #1289 
Browser support to detect MOJAVE's light/dark preference.
Allows Toggling between NightMode Settings.

### DIFF
--- a/www/src/js/views/AppShell.jsx
+++ b/www/src/js/views/AppShell.jsx
@@ -45,7 +45,6 @@ type Props = {
   theme: string,
   mode: Mode,
   activeSemester: Semester,
-  osEnabled: boolean,
 
   // From Redux actions
   fetchModuleList: () => Promise<*>,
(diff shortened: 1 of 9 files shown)


www/src/js/views/AppShell.jsx
```diff
@@ -101,6 +113,18 @@ export class AppShellComponent extends Component<Props, State> {
         });
       });
   };
+  checkFirstLoad = () => localStorage.getItem('persist:settings') == null;
```
ZhangYiJiang: This is never set, so this will never be false. Also extract the key to `storage/keys.js`

www/src/js/views/AppShell.jsx
```diff
@@ -62,7 +65,16 @@ export class AppShellComponent extends Component<Props, State> {
 
   componentDidMount() {
     const { timetables } = this.props;
-
+    // Checks initial load
+    if (this.checkFirstLoad()) this.enableOSMode();
+    const mqlDark = matchMedia('(prefers-color-scheme: dark)');
+    mqlDark.addListener((e) => {
```
ZhangYiJiang: Move this to bootstrapping instead, so this will only ever get called once

www/src/js/views/settings/ModeSelect.jsx
```diff
@@ -20,26 +22,50 @@ const MODES: Array<ModeOption> = [
     label: 'Off',
     value: LIGHT_MODE,
   },
+  {
+    label: 'OS DEFAULT',
```
ZhangYiJiang: This is not clear to non-technical users. Also do not use all caps.

www/src/js/views/AppShell.jsx
Calyeefornia: Hey Yijiang, isn't "persist:settings" being stored as a key after the initial load?

www/src/js/views/settings/ModeSelect.jsx
Calyeefornia: Would the label "OS mode" be better?

www/src/js/views/AppShell.jsx
ZhangYiJiang: Oh, that's a key used by Redux Persist. It's not a good idea to use that because you're creating indirect and unnecessary dependencies that are not clearly expressed in the code

li-kai:
@Calyeefornia We really appreciate your effort on this task. Would you still be interested in continuing development? If not, could you close the PR and let others take a shot at this? Thanks!

Calyeefornia:
@li-kai 
Apologies as I thought i've already fixed it and waiting for review. 
But sure, i'll close it if it is deemed not fixed.

PR #1383, opened 2019-01-07
Prototype module planner

ZhangYiJiang:
Module planner prototype based on [this suggestion](https://github.com/nusmodifications/nusmods/files/2735549/nusmods.idea.pdf)

UI is based on kanban board similar to Trello. Uses [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd), which works really well, but isn't very customizable. 

### Demo

Adding multiple modules at once 

peek 2019-01-08 15-28

Checking prerequisites and deleting module 

peek 2019-01-08 15-29

### Features 

- Drag and drop to shift modules around 
- Supports special term
- Supports adding modules not in our data 
- Supports exemptions and a 'plan to take' list for modules to be taken in some unspecified time in the future 
- Can add multiple modules at once
- Detect prerequisites and highlight modules with unfulfilled prerequisites 

### Technical 

Planner state is saved as a new substate in Redux with this shape 

```js
export type PlannerState = {|
  +minYear: string,
  +maxYear: string,

  +modules: {
    +[ModuleCode]: ModuleTime,
  },
|};

// The year, semester the module will be taken in, and the order
// it appears on the list for the semester
export type ModuleTime = [string, Semester, number];
```

This is then transformed into this shape which matches closely the shape of the UI elements 

```js
export type AcadYearModules = {
  +[string]: {| +[Semester]: ModuleCode[] |},
};
```

### Other changes 

- `node_modules` is removed from flow ignore list, since modules like Downshift and Beautiful DND have their own types bundled
- Testing out immer for complex reducers. The code looks better than with `immutability-helper`, though it adds a few kb to the bundle. 

### Todo

- [ ] Highlight modules not offered in semester 
- [ ] Highlight modules with exam clashes in current year 
- [ ] Allow users to fill in module credits for mods not in our data 
- [ ] Allow adding from module info page
- [ ] Sync with timetable 
- [ ] Add onboarding / wizard to pre-fill exemptions and other common modules 
- [ ] Preload with faculty suggested plans 
- [ ] Polish UI


www/src/js/views/planner/PlannerModule.jsx
```diff
@@ -0,0 +1,149 @@
...
+          <div
+            ref={provided.innerRef}
+            className={classnames(styles.module, {
+              [styles.warning]: conflicts,
+              [styles.isDragging]: snapshot.isDragging,
+            })}
+            {...provided.draggableProps}
+            {...provided.dragHandleProps}
+          >
+            <Downshift onChange={this.onMenuSelect}>{this.renderMenu}</Downshift>
```
li-kai: Suspect there may be a perf cost. Draggable produces a new function on render, so downshift will also rerender the entire menu on every input. Might be wrong, check with dev tools.

www/src/js/views/planner/PlannerModule.scss
```diff
@@ -0,0 +1,93 @@
...
+  background: lighten(theme-color('success'), 3);
+
+  &.warning {
+    color: darken(theme-color('warning'), 40);
+    border-color: darken(theme-color('warning'), 10);
+    background: theme-color('warning');
+
+    .menuBtn {
+      :global(.close) {
+        color: #000;
```
li-kai: A little too much nesting for my taste :X

www/src/js/views/planner/PlannerModule.jsx
ZhangYiJiang: Extracted out the entire Menu into a SFC wrapped in React.memo. I honestly have no idea if this is better - in theory it should be since `removeModule` prop is never changed

www/src/js/views/planner/PlannerModule.scss
ZhangYiJiang: Removed the extraneous layer of `{  }`. The resultant selector can't be made shorter though

PR #1385, opened 2019-01-07
Remove tombstone when new module is added

### LINKED ISSUE #1374: Adding a module that was previously removed causes duplicate "Undo" message
Adding a mod that was removed without dismissing the "Undo" message will result in a second "Undo"message as seen in the screenshot below.

image

To Replicate:
1. Add a module
2. Delete the module without dismissing the undo message
3. Add the same module again

This bug has no effect on functionality. The module still appears on the timetable correctly.

wn:
Fixes #1374 

Remove tombstone when a new module is added. Another solution can be found https://github.com/nusmodifications/nusmods/pull/1386

### DIFF
--- a/www/src/js/views/timetable/ModulesSelectContainer.jsx
+++ b/www/src/js/views/timetable/ModulesSelectContainer.jsx
@@ -6,7 +6,6 @@ import type { ModuleSelectList } from 'types/reducers';
 import type { ModuleCode, Semester } from 'types/modules';
 
 import Online from 'views/components/Online';
-import { addModule } from 'actions/timetables';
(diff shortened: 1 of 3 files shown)


ZhangYiJiang:
I've added a slightly different solution - lifted `addModule` out of the module select component. Also keyed TimetableContent so that state is reset when semester changes

wn:
Have you considered the other solution, where I check if the added module is the same as the tombstone? To add on to this solution, we can have a simple checker in `addModule`

wn:
Ahhh, sounds good. Any other changes I should make?

PR #1386, opened 2019-01-07
Remove tombstone when new module is added - Solution 2

### LINKED ISSUE #1374: Adding a module that was previously removed causes duplicate "Undo" message

wn:
Fixes #1374 

Remove tombstone **only** when the same module is added. Another solution can be found https://github.com/nusmodifications/nusmods/pull/1385


ZhangYiJiang:
Closed in favor of #1385

PR #1395, opened 2019-01-10
Downshift retain values

### LINKED ISSUE #1350: Retain input value after blur on Downshift component
[Downshift](https://github.com/paypal/downshift) is currently used in several places, including the add module select dropdown and global search on the top right of every page. The default Downshift behavior is to clear the input on blur, but this doesn't make for great user experience because the user could potentially lose some input. 

According to the docs, there doesn't seem to be any parameter for this, so I think we may have to tap into stateReducer, or make inputValue a controlled property.

juxd:
Fixes #1350 

`stateReducer`s blur seems to be overridden by `onOuterClick` or something to that effect, so I took the approach of making `inputValue` controlled.

I realise that the React documents recommend using an updater function to `setState` if it involves the current state, but none of my fixes seemed to work when I used it.


ZhangYiJiang:
@juxd We're doing a deploy soon-ish, so if you don't have the iOS bugfix ready then can you open a new PR instead since I'm merging this in? Thanks!

PR #1425, opened 2019-01-14
Use server GitHub proxy and refactor contributor loading logic 

ZhangYiJiang:
We frequently hit GitHub rate limits since our frontend makes unauthenticated calls to the API, which is IP rate limited to ~100 calls per hour, and all of NUS shares one IP address. I've used nginx to set up a simple proxy at https://github.nusmods.com/repo which proxies https://api.github.com/repos/nusmodifications/nusmods, and this PR migrates our code to use that. 

I've also pushed the logic for loading contributors down into the ContributorList component using `Loadable.Map`. This cuts quite a bit of code.


www/src/js/views/contribute/ContributorList/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+import ApiError from 'views/errors/ApiError';
+import LoadingSpinner from 'views/components/LoadingSpinner';
+import ContributorList from './ContributorList';
+
+type Props = {
+  size?: number,
+};
+
+// Wrapper around ContributorList that loads contributor data
+const AsyncContributorList = Loadable.Map<Props, *>({
```
taneliang: Having this component in an index.jsx seems weird. Is the actual 'ContributorList' component used anywhere else? Otherwise maybe you could rename that to 'ContributorListComponent' or something and call this wrapper 'ContributorList'.

www/src/js/views/contribute/ContributorList.jsx
```diff
@@ -39,3 +39,34 @@ export default function ContributorList(props: Props) {
...
+    if (props.error) {
+      return <ApiError dataName="venue locations" retry={props.retry} />;
+    } else if (props.pastDelay) {
+      return <LoadingSpinner />;
+    }
+
+    return null;
+  },
+
+  // This is not a proper render function, so prop validation doesn't work
```
taneliang: Why shouldn't this be a proper render function? Maybe you could put a link to a reference for this.

ZhangYiJiang: The lint rule expects the first param to be `props`, whereas here the second param is `props`. `Loadable.Map` is a HOC, not a component, so this is just a false positive on the lint rule

PR #1429, opened 2019-01-15
Optimize selectors and memoize icons

ZhangYiJiang:
- Use reselect to memoize the `getModuleCondensed` selector 
- Upgrade react-redux to 5.1.1 so it supports `React.memo`
- Wrap icons in `React.memo` since currently they are SFCs, and unnecessarily re-render every time their parent component re-renders 

Any other low hanging fruits we can do here to improve performance?


ZhangYiJiang:
Hmmm, to be very honest I'm not sure what SVGR is doing that React Feather isn't. Quite sure the prop spread will trigger the babel-constant-elements deopt, so it won't matter

ZhangYiJiang:
@li-kai Reverted the icons commit. I'll be adding SVGR with the logo PR instead, where it makes more sense

PR #1436, opened 2019-01-16
Add Zoho to supported by section of README

taneliang:


taneliang:
https://github.com/nusmodifications/nusmods/blob/zoho/README.md

image

PR #1448, opened 2019-01-18
Add privacy controls for analytics and Disqus 

ZhangYiJiang:
Add options to explicitly opt out of Matomo tracking on NUSMods (it's arguable whether what we're doing is tracking, although the script does maintain a session across multiple page views, so I guess it is?).

Also add option to make Disqus load manually instead of automatically, since Disqus loads a ton of unwanted junk onto the page, which some users may not want.


taneliang:
I don't think this feature is necessary. If people are so concerned about their privacy they'd already have DNT enabled, which I assume also blocks Disqus's trackers together with Matomo. If they are not (e.g. arts students, according to what I've learned in CS3216), they wouldn't care about (or understand) this setting.

But if we do want to keep this, then I have some comments.

The description text is too technical and long winded for non-technical users. In fact, I think this setting (especially the Disqus one) may be too complicated. I suspect people will read it and be like "k, sounds scary, let's turn everything off." We should explain in simple terms – without jargon – what data we are collecting, why it's so cool and essential that we do, and reassure them that we aren't doing anything nefarious. I also don't think anyone really cares that we're using Matomo instead of GA, and people may even trust GA more because they've heard of Google. Having a link to analytics.nusmods.com is a great start in reassuring people and letting them see and know what we collect, but we have to build on that instead of just putting the link without explanation.

In other news, text and buttons don't display well on 375px and below devices.
deploy-preview-1448--nusmods netlify com_settings

ZhangYiJiang:
DNT is not a user visible option in Chrome and Safari, and also in the mobile versions of the two. I'll try to work on the phrasing and layout issues.

www/src/js/views/settings/SettingsContainer.tsx
```diff
@@ -104,6 +143,66 @@ class SettingsContainer extends React.Component<Props> {
...
           toggleStates={this.props.toggleBetaTesting}
         />
+
+        <h4 id="privacy">Privacy</h4>
+
+        <div className="row">
+          <div className="col-md-8">
+            <p>
+              We collect anonymous, aggregated usage information on NUSMods - think of it as a
+              survey to tells us which browsers to support and what features are popular. Note that
```
li-kai:
```suggestion
              survey to tells us which browsers to support and what features are popular.
```

www/src/js/views/settings/SettingsContainer.tsx
```diff
@@ -104,6 +143,66 @@ class SettingsContainer extends React.Component<Props> {
...
+              if you opt out, we may remove features that you use since we won&apos;t know if anyone
```
li-kai:
```suggestion
              if you opt out, we could end up removing features that you use since we won&apos;t know if anyone
```

ZhangYiJiang:
@taneliang Do you want to re-review or can I just merge this in?

PR #1449, opened 2019-01-19
Modtris - Tetris in NUSMods 

ZhangYiJiang:
As everyone knows, in addition to module planning, finding free classrooms to sneak into, and locating where your next class is, NUSMods is also an excellent game engine. To prove that I have built Tetris in using NUSMods as a framework. No existing code except for the router has been modified. 

- [X] Keyboard controls 
- [X] Scoring 
- [X] Pausing 
- [x] Game over 
- [ ] Gamepad support 
- [ ] Touch support
- [ ] Animation and special effects 
- [ ] T-spin 
- [ ] Sound and music 

peek 2019-01-20 11-21

Play here: https://deploy-preview-1449--nusmods.netlify.com/tetris


www/src/js/views/routes/Routes.jsx
```diff
@@ -30,6 +31,7 @@ export default function Routes() {
       <Route path="/venues/:venue?" component={VenuesContainer} />
       <Route path="/today" component={TodayContainer} />
       <Route path="/planner" component={PlannerContainer} />
+      <Route path="/tetris" component={TetrisContainer} />
```
li-kai: Could we use suspense here and lazy load the tetris container? 🔥

ZhangYiJiang: Yes, that's in the plan but I did without it to make coding easier during the Hackerthon (dynamically imported components are not properly hard reloaded)

ZhangYiJiang:
- Screens smaller than 1080p are supported by resizing the entire `TetrisGame` element down using CSS scale transform 
- Hold now works correctly according to Tetris guidelines 
- Preview is now implemented 
- Lazy loading is implemented 

I think this is ready for merging

PR #1463, opened 2019-01-22
Switch feedback form to use Google Groups email and add private email address

ZhangYiJiang:
Makes Google Groups the default email address for feedback, and mods@ as the backup for private communication (eg. for security issues)


www/src/js/views/components/FeedbackModal.jsx
```diff
@@ -20,19 +20,27 @@ type Props = {
...
-      </ExternalLink>
-      <ExternalLink className={styles.github} href={config.contact.githubRepo}>
-        <GitHub />
-        GitHub
-      </ExternalLink>
-      <ExternalLink className={styles.form} href="mailto:mods@nusmods.com">
-        <Mail />
-        Email
-      </ExternalLink>
+    <div>
```
taneliang: Any reason why this is a `div` and not a `Fragment`?

www/src/js/views/components/FeedbackModal.jsx
```diff
@@ -20,19 +20,27 @@ type Props = {
...
+          <GitHub />
+          GitHub
+        </ExternalLink>
+        <ExternalLink className={styles.email} href={`mailto:${config.contact.email}`}>
+          <Mail />
+          Email
+        </ExternalLink>
+      </div>
+      <p className="text-muted">
+        To contact us privately, such as for security issues, please email{' '}
```
taneliang: There should be a period at the end of this sentence

taneliang: Actually this is a bit confusing. Wouldn't Messenger be private too?

www/src/js/views/components/FeedbackModal.jsx
ZhangYiJiang: It's probably easier to handle components that return a single element than two in most situations I feel

www/src/js/views/components/FeedbackModal.jsx
ZhangYiJiang: I wouldn't describe Messenger as private, haha. Email for security related issue is pretty standard

taneliang: Hahah that opinion is likely to be confusing for most people. I think this should still be reworded

taneliang: I don't have any good ideas either. But this may work: "To email us privately, such as for security issues, ..."

PR #1466, opened 2019-01-22
New logo and add SVGR loader 

ZhangYiJiang:
- Replace logo with new v4 logo 
- Add SVGR loader and update all existing SVG usage 
- Add `?url` resource query to allow files to be loaded as URL instead of React component 
- Add typing for Flow 

screenshot from 2019-01-22 23-10-21

screenshot from 2019-01-22 23-10-41

New logo is slightly thicker visually than the old one. 

Performance impact seems minimal, and aside from having to mess around a bit manually with some of the SVG file's width and height attributes, SVGR works pretty well out of the box. It does add 15MB to our `node_modules` though, unfortunately.


taneliang:
@jiholim Should the logo be made a bit smaller to give it more breathing room, especially on mobile? I think it's a bit too big, but it could just be because I'm not used to the new logo.

jiholim:
Yeap, agreed. We need to shrink the size, esp on mobile.

ZhangYiJiang:
Reduced the size slightly, also made the entire top left hand corner on desktop and the entire top bar on mobile tappable, maximizing 

screenshot from 2019-01-23 11-07-55
screenshot from 2019-01-23 11-08-08
screenshot from 2019-01-23 11-08-20

www/webpack/webpack.parts.js
```diff
@@ -208,14 +208,40 @@ exports.loadImages = ({ include, exclude, options } = {}) => ({
...
+          },
+          /**
+           * Load SVG as React component
+           * @see https://github.com/smooth-code/svgr
+           */
+          {
+            test: /\.(svg)(\?.*)?$/,
+            use: {
+              loader: '@svgr/webpack',
+              options,
```
li-kai: We should enable the titleProps option, it's sorta undocumented https://github.com/smooth-code/svgr/pull/103. That will solve a bunch of accessibility issues with using svg icons.

ZhangYiJiang: Weird that their docs are not updated. The PR is indeed in master, though for some reason it is not in the changelog

PR #1483, opened 2019-01-25
Add sentry caught exception in service worker #1292

### LINKED ISSUE #1292: Add error handling inside service worker
Currently our service worker code is mostly a black box. We catch on the install registration Promise, but in our Sentry log we only see generic errors like "Failed to register a ServiceWorker: ServiceWorker script evaluation failed" with no stack trace and no insight into what is actually failing. 

To fix this it seems we can add a global error handler *inside* the service worker https://stackoverflow.com/questions/37736322/how-does-global-error-handling-work-in-service-workers#37784386. We'll have to use `importScript` to load RavenJS from inside the service worker, then simply run `captureException` like in the code from this issue https://github.com/getsentry/raven-js/issues/1123

To do this, we can have Workbox load the script from the generated service-worker JS file. See https://github.com/nusmodifications/nusmods/blob/master/www/static/service-worker-notifications.js for an example of this. 

Note that we do still need to test if this is effective - it is possible that since this is loaded asynchronously, the error handler may not attach quickly enough to catch errors like the initialization error above.

emer7:
Fix #1292


www/static/service-worker-notifications.js
```diff
@@ -1,3 +1,6 @@
+self.importScripts('https://browser.sentry-cdn.com/4.5.3/bundle.min.js');
```
ZhangYiJiang: Since importScripts is synchronous, this should be moved below the existing code to not block it

PR #1510, opened 2019-02-03
Add custom planner module data

ZhangYiJiang:
This PR adds UI and data support for adding title and module credit data to the planner to fill in modules which we don't have data for or to override our data. 

peek 2019-02-03 13-53

MCs in headers are updated accordingly, and the module loses the warning about missing data. Currently there's no way to remove the data, but I don't foresee a lot of need for that anyway.


ZhangYiJiang:
Thanks, I had intended to implement the autofill but somehow it slipped by mind. Is it unintuitive that the same custom data shows up even after the module is deleted? In other words, when you fill up the form do you think the custom data is tied to that card? I think it may be useful to keep the data, but not if it is unexpected for the user.

ZhangYiJiang:
- We currently don't have any easy way to check for archived modules, so there's no way to prevent 404 if the user enters nonsense. We may want to implement some sort of API in the future 
- There's button to reset the info in the boxes back to the original module info 
- Form is now prefilled by module info 
- Remove is now below edit, and marked as red (we might want to implement undo at some point)

taneliang:
Nice, the reset button is a great addition. Storing the custom data after removing the module makes more sense now, and I agree that it's useful.

PR #1520, opened 2019-02-05
[nusmoderator] Upgrade deps and add support for Node 10

taneliang:
Changes listed in the commits, but TL;DR: Ran `babel-upgrade`, upgraded all other deps, regenerated docs, add and fix lint, and fix tests. CI now passes on Node 8, 10.14, and 11.8.

Shouldn't affect the built result, so no new release should be necessary.

TODO in the next PR: Add Prettier


packages/nusmoderator/src/index.test.js
```diff
@@ -1,4 +1,4 @@
-import nusmoderator from '../src';
+import nusmoderator from '.';
```
ZhangYiJiang:
```suggestion
import nusmoderator from './index';
```

packages/nusmoderator/.babelrc
```diff
@@ -1,3 +1,5 @@
 {
-  "presets": ["env"]
+  "presets": [
+    "@babel/preset-env"
```
ZhangYiJiang: I think we need to define the output target here. I'm not actually sure what babel env defaults to without that

PR #1523, opened 2019-02-05
Upgrade scraper V1 deps to support Node 10

taneliang:
Minimal upgrades as this scraper will no longer be needed in a few weeks/months.

CI passes on Node 8.15.0 and 10.15.1, but NOT 11.9.0 as fsevents does not build.

@li-kai could you run it and check if this works as wel? I'm getting an error when running `consolidateForSem` locally on Node 10.15.1 using the `scrape:dev` script, but I'm getting the same error on master on Node 8 so I'm not sure what's wrong.


li-kai:
@taneliang probably cause of the lessonTypes.json that you have to copy from api.nusmods.com

PR #1526, opened 2019-02-07
Automate calculation of start date of academic year

### LINKED ISSUE #1434: NUSModerator: Automate calculation of start date of academic year

Right now our start date of acad year has to be updated manually (and it ends this year!). If we could automate this, that'd be great.

The start date of each academic year is defined to be the first Monday of August (Week 0 / Orientation Week). It would be good to automatically calculate it using something like moment.js rather than maintaining our own mapping.

Retain current behavior

### Additional context

https://github.com/nusmodifications/nusmoderator/issues/2

Lessthanfree:
Add a function to calculate the start date of AY when given an AY.
Also add tests for the new function

This adds 2 dependencies to `package.json`: `date-fns` and `nusmoderator`.

Resolves #1434`


packages/nusmoderator/package.json
```diff
@@ -57,5 +57,9 @@
     "microbundle": "0.9.0",
     "prettier": "1.16.4",
     "rimraf": "2.6.3"
+  },
+  "dependencies": {
+    "date-fns": "^2.0.0-alpha.27",
+    "nusmoderator": "^2.1.1"
```
taneliang: I don't think this one is necessary. Try removing it by runnning `yarn remove nusmoderator`

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -1,4 +1,17 @@
...
+// Constant variables.
+const oneWeekDuration = 1000 * 60 * 60 * 24 * 7;
+const sem1 = 'Semester 1';
+const sem2 = 'Semester 2';
+const special1 = 'Special Term I';
+const special2 = 'Special Term II';
+
 /* eslint-disable no-fallthrough, no-console */
+// TODO: OLD STUFF
 export const acadYearStartDates = {
```
taneliang: This object can be removed since (as far as I can tell) no code is using it

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -1,4 +1,17 @@
...
+
+// Constant variables.
+const oneWeekDuration = 1000 * 60 * 60 * 24 * 7;
+const sem1 = 'Semester 1';
+const sem2 = 'Semester 2';
+const special1 = 'Special Term I';
+const special2 = 'Special Term II';
+
 /* eslint-disable no-fallthrough, no-console */
+// TODO: OLD STUFF
```
taneliang: This should be removed I think

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -8,35 +21,55 @@ export const acadYearStartDates = {
...
+ * Returns a Date object of the first weekday of Week 0 of that academic year.
+ * Assumes Week 0 begins on the first Monday of August.
+ * @param acadYear the academic year. E.g. "18/19"
+ * @return {Date} Start date of the academic year
+ */
+export function getAcadYearStartDate(acadYear) {
+  const DAY_MONDAY = 1;
+  const MONTH_AUGUST = '7';
+  const lastTwoDigits = acadYear.split('/')[0];
+  const targetYear = parseInt(20 + lastTwoDigits, 10);
```
ZhangYiJiang: Should be `'20'`, otherwise this will coerce `20` to string, which is unnecessary

packages/nusmoderator/package.json
ZhangYiJiang: I didn't even know Yarn allowed you to add a module as a dependency of itself! :P

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -8,35 +21,55 @@ export const acadYearStartDates = {
...
+  const firstDateOfMonth = new Date(targetYear, MONTH_AUGUST, 1, 0, 0, 0);
+  const firstDay = getDay(firstDateOfMonth);
+  if (firstDay === DAY_MONDAY) {
+    return firstDateOfMonth;
+  }
+  // Finds the next nearest Monday.
+  // % 7 is because if Sunday (0), will return +8
+  const daysDifference = (7 - firstDay + DAY_MONDAY) % 7;
```
ZhangYiJiang: This calculation might be simpler if you get the [start of the week](https://date-fns.org/v2.0.0-alpha.27/docs/startOfWeek) then add one week or seven days

Lessthanfree: Ah yes that would be much cleaner.

packages/nusmoderator/src/academicCalendar.js
ZhangYiJiang: Actually it might be better to tag it as `@deprecated` and leave it, otherwise we'll need to do a major version release. Thoughts @taneliang @li-kai  ?

taneliang: I'd prefer doing a major version release. I don't think we should bother keeping old code around, especially since we're the biggest (and possibly only?) users of it. Plus it's a small change which doesn't require much effort to update to.

packages/nusmoderator/src/academicCalendar.js
Lessthanfree: implemented

packages/nusmoderator/src/academicCalendar.js
Lessthanfree: I changed it to 2000 + parseInt(lastTwoDigits)

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -8,35 +21,55 @@ export const acadYearStartDates = {
...
-    const year = years[i];
-    if (date >= acadYearStartDates[year]) {
-      return {
-        year,
-        startDate: acadYearStartDates[year],
-      };
-    }
+  const dateYear = getYear(date);
+  const firstTwoDigits = dateYear % 100;
+  const lastTwoDigits = firstTwoDigits + 1;
```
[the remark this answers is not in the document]
Lessthanfree: removed

packages/nusmoderator/src/academicCalendar.js
Lessthanfree: The whole chunk has been removed

li-kai: :+1: to following semver

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -1,12 +1,30 @@
+const addDays = require('date-fns/addDays');
```
ZhangYiJiang: Does microbundle support tree shaking? If it does I think you can just write `import { addDays, startOfWeek, isBefore, getYear } from 'date-fns'`

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -16,27 +34,31 @@ const special1 = 'Special Term I';
...
-      };
-    }
+  const dateYear = getYear(date);
+  const firstTwoDigits = dateYear % 100;
+  const potentialAcadYear = `${firstTwoDigits}/${firstTwoDigits + 1}`;
+  const potentialStartDate = getAcadYearStartDate(potentialAcadYear);
+
+  let year;
+  // check if date is before the start of that year's AY start date
+  // Small HACK to workaround isBefore() because it returns true for same day
```
ZhangYiJiang: I don't understand this comment

packages/nusmoderator/package.json
```diff
@@ -58,5 +59,8 @@
     "microbundle": "0.9.0",
     "prettier": "1.16.4",
     "rimraf": "2.6.3"
+  },
+  "dependencies": {
+    "date-fns": "^2.0.0-alpha.27"
```
ZhangYiJiang: Pin the version (remove the "^") since we use Renovate to help manage dependency versions

packages/nusmoderator/src/academicCalendar.js
li-kai: It uses rollup under the hood so it should

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -1,12 +1,30 @@
...
+ */
+export function getAcadYearStartDate(acadYear) {
+  const DAY_MONDAY = 1;
+  const MONTH_AUGUST = '7';
+  const lastTwoDigits = acadYear.split('/')[0];
+  const targetYear = 2000 + parseInt(lastTwoDigits, 10);
+  const firstDateOfMonth = new Date(targetYear, MONTH_AUGUST, 1, 0, 0, 0);
+  const nearestMonday = startOfWeek(firstDateOfMonth, { weekStartsOn: DAY_MONDAY });
+  if (isBefore(nearestMonday, firstDateOfMonth)) {
+    const firstMonday = addDays(nearestMonday, 7);
```
taneliang: nit: `addWeeks` might be easier to understand https://date-fns.org/v2.0.0-alpha.27/docs/addWeeks

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -1,12 +1,30 @@
...
+/**
+ * Returns a Date object of the first weekday of Week 0 of that academic year.
+ * Assumes Week 0 begins on the first Monday of August.
+ * @param acadYear the academic year. E.g. "18/19"
+ * @return {Date} Start date of the academic year
+ */
+export function getAcadYearStartDate(acadYear) {
+  const DAY_MONDAY = 1;
+  const MONTH_AUGUST = '7';
+  const lastTwoDigits = acadYear.split('/')[0];
```
taneliang: nit: this variable name is confusing because you're actually getting the first year in `acadYear` and using those as the last 2 digits of the final year. I don't have any good solutions, but maybe rename it to `firstHalfYearDigits`?

packages/nusmoderator/src/academicCalendar.js
```diff
@@ -16,27 +34,31 @@ const special1 = 'Special Term I';
...
-        startDate: acadYearStartDates[year],
-      };
-    }
+  const dateYear = getYear(date);
+  const firstTwoDigits = dateYear % 100;
+  const potentialAcadYear = `${firstTwoDigits}/${firstTwoDigits + 1}`;
+  const potentialStartDate = getAcadYearStartDate(potentialAcadYear);
+
+  let year;
+  // check if date is before the start of that year's AY start date
```
taneliang: This comment should be expanded to describe what the code below does at a higher level so that it's easier to understand all the date manipulation

packages/nusmoderator/src/academicCalendar.js
ZhangYiJiang: I've changed this to `shortYear`. I think the fact that it's the first year in the AY string should be obvious from the code without having to name it as such

packages/nusmoderator/src/academicCalendar.js
ZhangYiJiang: Going to assume the comment is incorrect

Lessthanfree: Yes this comment is wrong. Forgot to remove it

PR #1529, opened 2019-02-07
Fix weather API code and add docs

ZhangYiJiang:
There's an error that occurs on the Today page caused by the `tomorrow()` endpoint not being able to get the forecast. This fixes that.


ZhangYiJiang:
This PR also forces www CI to run in Asia/Singapore, to avoid functions like `differenceInCalendarDays` returning incorrect results. This is acceptable because we assume all developers and more importantly users are running in SGT (the API that gets the data is literally returning Singapore weather), and the amount of code to make the date math TZ agnostic is... well, a lot.

PR #1530, opened 2019-02-08
Maintain lesson position when viewport size changes

### LINKED ISSUE #1447: Scroll timetable when lesson is clicked so the lesson remains on screen
### Is your feature request related to a problem? 

When the user clicks on a lesson with a lot of options, the current lesson will move to a new location instantly. This will disorientate the user, especially if there are a lot of options. A classic example is GER1000. 

We can scroll the page such that the lesson is still where the mouse is. The mental model for this is that the rest of the timetable is moving, but the lesson remain where the user just clicked/tapped. Currently the new lessons above the selected one pushes it down. To do this we can capture the location of the clicked lesson relative to the viewport, then after next render update scroll such that the same relative scroll from viewport is maintained (the math is left as an exercise to the reader) 

Another way to solve this is to use animation, aka. #1348.

dvrylc:
Fixes #1447 ✨ 

Maintains a lesson's position relative to the viewport when the viewport size changes. Applicable when working with modules with a large amount of options, such as GER1000. This is done by calculating the necessary scroll amount and works in both horizontal and vertical mode. 

Failed approach -
The initial approach was to persist the click event such that it doesn't get nullified by React, then query for the event target's position before and after the render. However, upon re-render, the original lesson moves to a completely different element and `e.currentTarget` now refers to a different lesson. Due to the lack of a common identifier before and after the re-render, I had to attach IDs to every lesson based on the module code, lesson type and class number (e.g. `#GER1000-TUT-D24`) then query for the position based on that.


dvrylc:
Seems like Coveralls isn't happy with the PR. I'm not entirely sure how this should be tested, open to ideas!

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
...
+ * Prevents disorientation while nagivating a module with lots of options (e.g. GER1000)
+ * @param {string} cellID
+ * @param {boolean} isScrolledHorizontally
+ */
+export function maintainLessonPosition(cellID: string, isScrolledHorizontally: boolean): void {
+  let cell;
+
+  // 1. Store timetable's position
+  const timetable = document.querySelector('.scrollable');
+  let timetablePos;
```
ZhangYiJiang: We try to avoid abbreviations. In this case I think there's no reason to shorten the variable names

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
...
+  if (timetable) {
```
ZhangYiJiang: You can use safe navigation operator to write this as `const timetablePosition = timetable?.getBoundingClientRect()` since we have that plugin enabled, although in this case it is probably better to just exit if the check is false, ie. 

``` js 
if (!timetable) return;
const timetablePosition = timetable.getBoundingClientRect();
```

www/src/js/views/timetable/TimetableCell.scss
```diff
@@ -24,6 +24,11 @@ $cell-border-radius: 6px;
     outline: none;
   }
 
+  // Prevent cell's children from intefering with click events
+  > * {
```
ZhangYiJiang: Try to avoid the `*` selector. Is there any reason this can't be on `.cellContainer` ?

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
     lessonType: lesson.LessonType,
   };
 }
+
+/**
+ * Resets the scroll position to (0,0)
+ */
+export function resetScrollPosition() {
```
ZhangYiJiang: This should be in `utils/react` - this file is for operations on the timetable data, and not the DOM

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
...
+ * Maintain a lesson's position when the viewport size changes
+ * Prevents disorientation while nagivating a module with lots of options (e.g. GER1000)
+ * @param {string} cellID
+ * @param {boolean} isScrolledHorizontally
+ */
+export function maintainLessonPosition(cellID: string, isScrolledHorizontally: boolean): void {
+  let cell;
+
+  // 1. Store timetable's position
+  const timetable = document.querySelector('.scrollable');
```
ZhangYiJiang: Try to replace this with a ref. `.scrollable` is a global utility class that is used in a lot of places

ZhangYiJiang:
We don't enforce 100% coverage on new code. I'm fine with testing this manually, especially since we need to check for cross browser consistency

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
...
+    // 5. Perform scroll if new position is beyond limits
+    if (oPos && nPos && (nPos.bottom > maxH || nPos.right > maxW)) {
+      const x = nPos.left - oPos.left + window.scrollX;
+      const y = nPos.top - oPos.top + window.scrollY;
+
+      window.scroll({ top: y, left: 0, behavior: 'smooth' });
+
+      if (!isScrolledHorizontally) {
+        // $FlowFixMe: Remove when typedefs are updated for Element.scroll
+        timetable.scroll({ top: 0, left: x, behavior: 'smooth' });
```
ZhangYiJiang: `element.scroll` is not supported in Edge, so wrap this in a feature detection to make sure it won't crash

www/src/js/utils/timetables.js
```diff
@@ -459,3 +459,61 @@ export function getHoverLesson(lesson: Lesson): HoverLesson {
...
+    if (cell) {
+      nPos = cell.getBoundingClientRect();
+    }
+
+    // 5. Perform scroll if new position is beyond limits
+    if (oPos && nPos && (nPos.bottom > maxH || nPos.right > maxW)) {
+      const x = nPos.left - oPos.left + window.scrollX;
+      const y = nPos.top - oPos.top + window.scrollY;
+
+      window.scroll({ top: y, left: 0, behavior: 'smooth' });
```
ZhangYiJiang: I'm not sure if having the animation is necessary since the rest of the timetable moves instantaneously right now

dvrylc:
> The behavior feels correct, which is great, but this approach is problematic as it relies on direct DOM access (`document.getElementById`) and implicitly working around React's render lifecycle (`setTimeout(..., 0)`). This is fragile and not future proof, especially since async rendering is on React's roadmap already.

Yeah I agree with this, alright will work on it and update the PR

dvrylc:
> Here's another approach to try - store `eventTarget`, which is a DOM element (you can't store the event itself since React recycles synthetic events), as a class property, then on `componentDidUpdate`, which fires after the next render, do the scrolling and unset the property so the scrolling only happens once.

I've given this a shot and I'm still kinda stuck. The problem remains - 

Even though I've stored `eventTarget` as a class property, upon `componentDidUpdate`, the component now holds a different lesson (different timeslot/class number). Computing the difference between the old and new positions is inaccurate since they refer to different lessons. In the picture below, GER1000 D82 was originally in the highlighted button, but moved to a new button after the click. 

screen shot 2019-02-10 at 1 33 48 am

I'm really trying to avoid direct DOM access but I don't see how I can have a reference to both the old and new positions of the same lesson without attaching an ID and querying for the positions based on that. Would appreciate any help! 😅

ZhangYiJiang:
Does adding a `key` to the `<TimetableCell>` component work? I suspect there might still be issue when the cell moves across rows though.

dvrylc:
I actually tried that and the issue persists, there isn't enough information to target and calculate the new position accurately. In fact, this issue is only present when the cell moves to a different row.

dvrylc:
@ZhangYiJiang What do you think? 

Couldn't find a use case for `getSnapshotBeforeUpdate` that makes solving this issue simpler since it only exposes the previous props and state - both of which don't help with the position before or after the update. The current implementation of getting the original position from the event and then querying for the new position after the update seems more straightforward.

ZhangYiJiang:
Yeah, it looks like this is the cleanest approach. Can't use any of the lifecycle methods on the Cell itself because when the new options are rendered the props change, which means the same component and element now represents a different lesson, and using `key` just means the component is destroyed instead of re-rendered since the cell can be in a different row, and there's no way to get rows to share children. 

Changes 

- Lift `maintainScrollPosition` up to `<TimetableContent>` so that the prop doesn't have to be drilled down to the cell. Since we're querying directly for the DOM element anyway it doesn't matter where we're calling `maintainScrollPosition` 
- Move the variable for storing the position and class name to instance property instead of state since we don't want this to trigger a re-render. 
- Move the helper function into `TimetableContent.tsx` because it doesn't look like this function can be reused anywhere else 
- Use `class` instead of `id` because the identifier is not unique - a module can have more than one lecture with the same class numbe - and because there might be more than one timetable on the same page

dvrylc:
> * Use `class` instead of `id` because the identifier is not unique - a module can have more than one lecture with the same class numbe - and because there might be more than one timetable on the same page
> * Switch from `DOMRect` to `ClientRect`. Casting is unsafe, and that cast specifically would have failed on Edge if you tried to use anything in `DOMRect` that is not in `ClientRect`

Thanks for the detailed review! Never knew these issues existed. Everything looks good, I'll just bring this up to date.

PR #1533, opened 2019-02-09
Upgrade dev dependencies

li-kai:
1. Upgrades babel & jest
2. Improve jest performance
3. Improve babel and js performance


www/babel.config.js
```diff
@@ -0,0 +1,57 @@
...
+  ];
+
+  const plugins = [
+    '@babel/plugin-syntax-dynamic-import',
+    // // See https://github.com/facebook/create-react-app/issues/4263
+    ['@babel/plugin-proposal-class-properties', { loose: true }],
+    ['@babel/plugin-proposal-optional-chaining', { loose: true }],
+  ];
+
+  if (IS_DEV || IS_PROD) {
```
ZhangYiJiang: I think we can keep this for tests. Do we need any of the extra lodash functions / features in tests?

www/babel.config.js
```diff
@@ -0,0 +1,57 @@
...
+        exclude: ['transform-typeof-symbol'],
+      },
+    ],
+    ['@babel/preset-react', { development: !IS_PROD }],
+    '@babel/preset-flow',
+  ];
+
+  const plugins = [
+    '@babel/plugin-syntax-dynamic-import',
+    // // See https://github.com/facebook/create-react-app/issues/4263
```
ZhangYiJiang: /nit double comment (what does it mean!?)

ZhangYiJiang:
Code coverage somehow increased 16%. Either you broke the code coverage path config or you fixed it. Can you check which one it is? :P

ZhangYiJiang:
CI seem to be slightly faster, which is nice, and bundle size is smaller by 4kb gzipped, which is a small win

www/package.json
```diff
@@ -64,6 +63,7 @@
     "eslint-plugin-jsx-a11y": "6.0.3",
     "eslint-plugin-prettier": "2.6.0",
     "eslint-plugin-react": "7.7.0",
+    "eslint": "4.19.1",
```
taneliang: Why did `eslint` get moved down?

www/babel.config.js
li-kai: No, it's just that compilation would be slower if we included it

www/package.json
li-kai: Sorted by alphabetical order

www/babel.config.js
li-kai: Added a comment, is that what you were looking for?

taneliang: Don't think so. Line 27 has two `//`s haha (`// // `). The new comment is useful though, let's leave it there

li-kai:
@ZhangYiJiang It was broken because it wasn't covering all the files without a test file beside. Now it is still up, because we ignore test-utils instead of test_utils now :100:

www/babel.config.js
ZhangYiJiang: Probably worth keeping the lodash plugin just to keep test behavior consistent with the other envs

www/jest.config.js
```diff
@@ -1,41 +1,25 @@
...
-    'src/js/test_utils',
-    'src/js/e2e',
-    // Code in this file triggers this bug - https://github.com/istanbuljs/babel-plugin-istanbul/issues/116
-    'src/js/views/modules/ModulePageContent.jsx',
-  ],
+  roots: ['<rootDir>/src/js'],
+  moduleDirectories: ['node_modules', '<rootDir>/src/js'],
+  moduleFileExtensions: ['jsx', 'js'],
+  testPathIgnorePatterns: [],
+  testRegex: 'src/js/.+\\.test\\.jsx?$',
```
ZhangYiJiang: Can this be a regex literal instead?

www/jest.config.js
```diff
@@ -1,41 +1,25 @@
...
+    '\\.(?:css|scss)$': 'identity-obj-proxy',
   },
   // Allow us to directly use enzyme wrappers for snapshotting
   // Usage: expect(enzyme.shallow(<div/>)).toMatchSnapshot();
-  snapshotSerializers: [
-    '<rootDir>/node_modules/enzyme-to-json/serializer',
-  ],
+  snapshotSerializers: ['<rootDir>/node_modules/enzyme-to-json/serializer'],
+  collectCoverageFrom: ['**/*.{js,jsx}'],
+  coveragePathIgnorePatterns: ['/node_modules/', '<rootDir>/src/js/(?:test-utils|e2e)'],
```
ZhangYiJiang: I think the `?:` is unnecessary. It makes it a non-capturing group, which captures intent better, but it makes it a bit harder to read. I can go either way 🤷‍♂️

www/jest.config.js
li-kai: Unfortunately, no.
```
Option "testRegex" must be of type:
    string or array
```

PR #1535, opened 2019-02-09
Switch to scraper v2 to TypeScript

ZhangYiJiang:
Switch the entire scraper v2 codebase to TypeScript. Most of the changes was done automatically, but  some stuff needed hacking by hand. 

### The bad 

Most of the trouble came from a bad TS libdef update to Jest and a bug that prevented `eslint-plugin-import` from playing well with `@typescript-eslint`. Those should be ironed out later though.  

TypeScript inference seems less powerful / generous than Flow's. The most annoying one is where it always assumes an unannotated object is an empty object, which is rather useless. 

There's no option to compile on the fly now (aka. `babel-node`). TS does have a watch option, but if it is anything like Babel's watch option then I'd rather suffer the extra compile time. 

Since typechecking is split from code output, it's not possible to run code without also adding annotations and stuff, which can be annoying if you're just trying to hack something together temporarily, but then again you can just throw a bunch of `as any` and `@ts-ignore` at the compiler. 

### The good 

TypeScript libdefs are stored as separate npm packages. The good thing about this is it is possible to pin the libdef version so we can avoid buggy defs by simply updating the version number. However, this also means it's not possible to patch bugs locally by hacking the libdef directly unlike in Flow. 

I do like that there's an option to mark areas where there are implicit `any`s. Will turn this on later once I've ironed out the other bugs 

IDE integration works much better now. Hard to tell if performance is better because this codebase is tiny, so both TS and Flow runs in under 10 seconds.


li-kai:
There's ts-node, which acts like babel-node and pretty stable from my usage.

scrapers/nus-v2/src/config.ts
```diff
@@ -1,25 +1,23 @@
-// @flow
-
 import path from 'path';
 import * as fs from 'fs-extra';
 
-export type Config = {|
-  +appKey: string,
-  +studentKey: string,
+export interface Config {
+  appKey: string;
```
li-kai: There's `readonly` which is the verbose version of flow's `+`

scrapers/nus-v2/src/types/modules.ts
```diff
@@ -51,96 +49,100 @@ export type WorkloadComponent = 'Lecture' | 'Tutorial' | 'Laboratory' | 'Project
...
-export type RawLesson = {|
-  ClassNo: ClassNo,
-  DayText: DayText,
-  EndTime: EndTime,
-  LessonType: LessonType,
-  StartTime: StartTime,
-  Venue: Venue,
-  Weeks: LessonWeek[],
-|};
+export interface RawLesson {
```
li-kai: These should all be type aliases. The differences [(article)](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) [(table)](https://stackoverflow.com/a/54101543/4819795) mean that interfaces imply that objects contains functions imo.

PR #1536, opened 2019-02-10
Freeze lockfile in CI and update lockfile

ZhangYiJiang:
Add `--frozen-lockfile` option which will cause CI to fail if the lockfile changes and `--non-interactive` to avoid prompts.


.circleci/config.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - attach_workspace:
           at: ~/nusmods
       - restore_cache: *restore_node_deps
-      - run: yarn
+      - run: yarn --frozen-lockfile
```
li-kai: Why not non-interactive here?

ZhangYiJiang: Because copy-pasting is hard :joy:

PR #1546, opened 2019-02-12
Fix leftover things from www TS migration

taneliang:
* Migrate remaining JS/Flow references in README
* Add typecheck to ci script


www/package.json
```diff
@@ -18,7 +18,7 @@
     "test:watch": "jest --watch",
     "e2e": "nightwatch -e default,safariMobile,safariDesktop,edge",
     "cors": "node scripts/cors-crawler",
-    "ci": "cross-env NODE_ENV=production npm run lint && npm test && npm rebuild node-sass && npm run build"
+    "ci": "cross-env NODE_ENV=production npm run lint && npm run typecheck && npm test && npm rebuild node-sass && npm run build"
```
ZhangYiJiang: Weird, I don't think we actually use this (at least not on CircleCI), since we split our CI into separate tasks, and we don't see that annoying node-sass error anymore. We should just delete this.

taneliang: Nah this script is for us to be able to run the CI pipeline conveniently locally. We should keep it

www/README.md
```diff
@@ -22,15 +22,15 @@ Mobile browsers:
 - iOS 10 and above
 - Chrome Mobile last two versions
 
-We try not to break iOS 9 (ie. display white screen of death or use unsupported APIs), but we don't guarantee the page will be appear exactly the same as in more modern browsers.
+We try not to break iOS 10 (ie. display white screen of death or use unsupported APIs), but we don't guarantee the page will be appear exactly the same as in more modern browsers.
```
ZhangYiJiang: This is not correct - our minimum support level is technically at iOS 10, but we try not to break 9 even if it is outside the range because there's still ~1% of users on that.

taneliang: Okay, will revert

www/package.json
ZhangYiJiang: That's fair.

PR #1551, opened 2019-02-12
Upgrade dev dependencies

li-kai:
Went through their changelogs, doesn't seem like much has changed. Mostly bug fixes/features that don't impact us.


ZhangYiJiang:
Actually, since Edge has supported CSS variables since 15, and our usage shows very few users below 16, I think we can remove the CSS variable PostCSS plugin entirely to shave a few kb off our styles

ZhangYiJiang:
Oh yeah, it might be best to avoid updating Jest's typing until https://github.com/DefinitelyTyped/DefinitelyTyped/issues/32890 is resolved

www/.postcssrc.js
```diff
@@ -2,10 +2,6 @@
 const config = {
   plugins: [
     require('autoprefixer'),
-    require('postcss-custom-properties')({
-      preserve: true, // Preserve the original CSS variable declaration
-      warnings: false, // Ignore warnings about variables declared on non-:root selectors
-    }),
     // Custom plugin used to remove extra ':root' rules
     require('./scripts/postcss-single-root'),
```
ZhangYiJiang: This plugin can also be removed

ZhangYiJiang: And basically reverse most of the hacks from 
https://github.com/nusmodifications/nusmods/pull/676

PR #1552, opened 2019-02-12
Update misc configuration

li-kai:
1. Remove travis config
1. Rename /docs to /misc.
    I've repeatedly thought to myself "Huh, since when did we have docs?"
    This makes the intent of the folder more explicit
1. Removes the duplicated nvmrc, telling netlify to use dubnium
1. Updates renovate config to group monorepos and etc together so we have less of a burden
    This is the reason behind the PR title.


renovate.json
```diff
@@ -3,6 +3,8 @@
     "config:base",
     "schedule:earlyMondays",
     "group:linters",
+    "group:monorepos",
+    "group:definitelyTyped",
```
ZhangYiJiang: I don’t think this is a good idea. We want to group the type with their dependents, and not upgrade all types all at once

renovate.json
```diff
@@ -11,21 +12,61 @@
...
+    {
+      "description": "Group stylelint packages together",
+      "packagePatterns": ["stylelint"],
+      "excludePackageNames": ["stylelint-webpack-plugin"],
+      "groupName": "stylelint"
+    },
+    {
+      "description": "Group test-related packages together",
+      "extends": "monorepo:jest",
+      "packagePatterns": ["(jest|enzyme|mock|nightwatch)"],
```
ZhangYiJiang: Nightwatch should probably not be bundled here.

renovate.json
```diff
@@ -11,21 +12,61 @@
...
+      "packageNames": ["identity-obj-proxy"],
+      "groupName": "test"
+    },
+    {
+      "description": "Group test-related packages together",
```
ZhangYiJiang: Copy-paste error? Also if we're extending can we just leave out the description?

renovate.json
```diff
@@ -11,21 +12,61 @@
...
+      "extends": "monorepo:react",
+      "packageNames": ["@types/react", "@types/react-dom"],
+      "groupName": "react"
+    },
+    {
+      "description": "Group test-related packages together",
```
ZhangYiJiang: Also copy paste error?

renovate.json
```diff
@@ -11,21 +12,61 @@
...
+      "groupName": "react-router"
+    },
+    {
+      "description": "Group state management packages together",
+      "packagePatterns": ["(redux|reselect|immer)"],
+      "excludePackagePatterns": ["redux-mock"],
+      "groupName": "state-management"
     }
   ],
-  "ignoreDeps": [
```
ZhangYiJiang: Can we ignore `query-string`, since we can't upgrade past 5.x

renovate.json
ZhangYiJiang: As an aside, 1.x was just released and comes with a new API, so upgrading to that will be a bit troublesome

li-kai: Fixed in #1563

renovate.json
li-kai: Ignored in #1563

PR #1553, opened 2019-02-12
Fixes navbar logo

li-kai:
Tested in MacOS Firefox 66 and Safari 12.0.3. 

out


ZhangYiJiang:
Looks like you need to run Prettier, can merge after that

PR #1554, opened 2019-02-13
Update frontend and export for scraper v2 

ZhangYiJiang:
This PR updates the frontend for scraper v2's data shape. 

- Removed timeslot filters and tables since the API doesn't have those information anymore 
- Update our test fixtures. This accounts for most of the lines removed 
- Remove the workload parser and dual coded venue module merging logic because those are done on the backend now 

Todo

- [x] Update lesson week formatting to display odd, even and non-instruction weeks properly 
- [x] Test timetable export and sharing works 
- [x] Update module filters to use faculty directly from modules 
- [x] Update module page to show module aliases 
- [x] Update module info to show exam duration


ZhangYiJiang:
If you check out and run this branch locally, before switching back to a v1 branch remember to run `delete localStorage['persist:moduleBank']` to clear the cache

ZhangYiJiang:
Finished updating the client code for v2 data changes. 

- Redux Persist migration clears moduleBank cache so that cached data does not crash the client 
- Removed timeslot filter and UI, workload parsing, lesson availability check and venue merging logic 
- Update exam time formatting to use date-fns. This is done by subtracting the machine's local tz offset then adding SGT offset, which causes the date object's local time to be identical to Singapore's 
- Dual coded modules have both module codes displayed: https://deploy-preview-1554--nusmods.netlify.com/modules/GET1025/science-fiction-and-philosophy
- Exam duration is now available in addition to exam timing - https://deploy-preview-1554--nusmods.netlify.com/modules/ 
- Typings are tightened using `ReadonlyArray`
- No changes required for export except pointing the server to the new API endpoint

www/src/js/utils/ical.ts
```diff
@@ -89,37 +78,14 @@ export function datesForAcademicWeeks(start: Date, week: number): Date {
...
-      break;
-
-    // 4. If WeekText is not any of the above, then assume it consists of a list of weeks
-    //    with lessons, so we exclude weeks without lessons
-    default: {
-      const weeksWithClasses = lesson.WeekText.split(',').map((w) => parseInt(w, 10));
-      excludedWeeks = _.union(excludedWeeks, _.difference(ALL_WEEKS, weeksWithClasses));
-    }
-  }
+  // TODO: Make this work with new format
```
li-kai: Does the below not work?

ZhangYiJiang:
> Any ideas how to test for persist migrate? 

Only way is to start with a fresh persisted state on master (add a few modules on master), then switch to the branch and check that the next time the page is loaded it doesn't crash.

www/src/js/utils/ical.ts
ZhangYiJiang: Oops, forgot to update that. Let me fix it.

www/src/js/utils/ical.ts
```diff
@@ -31,97 +36,111 @@ export function getTimeHour(time: string) {
...
-  const d = new Date(date.valueOf());
-  d.setUTCHours(d.getUTCHours() + Math.floor(sgHour), (sgHour % 1) * 60);
-  return d;
+function addLessonOffset(date: Date, hourOffset: number): Date {
+  return addMinutes(date, hourOffset * 60);
 }
 
 export function iCalEventForExam(module: Module, semester: Semester): EventOption | null {
-  const examDateString = _.get(getModuleSemesterData(module, semester), 'ExamDate');
+  // TODO: Account for ExamDuration field
```
li-kai: What does this mean?

www/src/js/utils/ical.ts
```diff
@@ -31,97 +36,111 @@ export function getTimeHour(time: string) {
...
+  const rangeStart = parseDate(weekRange.start);
+  const rangeEnd = parseDate(weekRange.end);
+  const { start, end } = calculateStartEnd(rangeStart, lesson.startTime, lesson.endTime);
+
+  const interval = weekRange.weekInterval || 1;
+  const exclusions = [];
+
+  if (weekRange.weeks) {
+    for (
+      let current = rangeStart, weekNumber = 1;
```
li-kai: This for loop is really complex 😱

ZhangYiJiang: It's just a standard for loop with two variables instead of one. I don't think it's *that* bad. (Using a while loop or a one variable with the index separate will give you the same level of complexity)

www/src/js/utils/ical.ts
ZhangYiJiang: It means I forgot to fix this :laughing:

PR #1557, opened 2019-02-14
Improve typing and reduce lint errors 

ZhangYiJiang:
A lot of cleaning up the usage of `any` in the code. 

- Replace with finer type where possible 
- Replace with `unknown` when the type doesn't really matter 
- Add comment and `eslint-ignore` where it would be too troublesome/verbose to type 

Also cleaned up some other lint errors, some of which are not fixed because they are fixed in other PRs

### DIFF
--- a/www/src/js/utils/react.tsx
+++ b/www/src/js/utils/react.tsx
@@ -63,13 +63,13 @@ export function noBreak(text: string): string {
   return text.replace(/ /g, NBSP);
 }
 
-export function defer(task: () => any) {
+export function defer(task: () => unknown) {
   window.requestAnimationFrame(() => {
     window.requestAnimationFrame(task);
   });
 }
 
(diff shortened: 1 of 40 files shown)


www/src/js/utils/react.tsx
```diff
@@ -63,12 +63,14 @@ export function noBreak(text: string): string {
...
 
-export function defer(task: () => any) {
+export function defer(task: () => unknown) {
   window.requestAnimationFrame(() => {
     window.requestAnimationFrame(task);
   });
 }
 
+// We really don't care about the props here
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
li-kai: Would this work?
```suggestion
export function wrapComponentName<T extends {}>(Component: React.ComponentType<T>, wrapper: string): string {
```

PR #1563, opened 2019-02-15
Update renovate config

li-kai:
According to renovate's docs, their bot should have triggered on #1552 and told us if the config was valid, but it didn't. So here's another attempt to triggering it.

Docs: https://renovatebot.com/docs/reconfigure-renovate/

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,7 @@
 {
   "extends": [
     "config:base",
-    "schedule:earlyMondays",
+    "schedule:weekends",
     "group:babelMonorepo",
     "group:materialMonorepo",
     ":automergePr",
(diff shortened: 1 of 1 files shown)


ZhangYiJiang:
Hmmm, maybe we can try setting `@types` to autoupdate?

li-kai:
@ZhangYiJiang I don't see how we can make `@types` auto merge while keeping them in step with our actual dependencies at the same time.

ZhangYiJiang:
The problem I see now is that we get spammed with a ton of `@types` updates since typedefs can change more frequently than the actual packages, especially for larger packages. It's also relatively safe to merge in libdefs if CI is green, and we can't effectively review the PRs for `@types` because there's no changelog and the source link doesn't even link to the correct subdirectory in the repo. Automerging the `@types` changes is meant to solve this, but there might be other solutions I've not thought of.

li-kai:
I've added a `@types` group which excludes ones that are already grouped. This should automerge types and other groups if all the tests pass.

PR #1566, opened 2019-02-17
Highlight search period

amrut-prabhu:
Part of #538 

Highlight period of time searched for when availability search is on.

Sample output:
2019-02-17_21h28_14

There is an issue with the output when changing the search start time. The searchedPeriod duration (and TimeTable component ) is not updated and so, it ends up looking like this.
2019-02-17_21h24_55
.

### DIFF
--- a/www/src/js/views/venues/VenueDetails.tsx
+++ b/www/src/js/views/venues/VenueDetails.tsx
@@ -18,6 +18,7 @@ import { breakpointDown } from 'utils/css';
 import VenueLocation from './VenueLocation';
 
 import styles from './VenueDetails.scss';
+import { convertIndexToTime, SCHOOLDAYS } from "../../utils/timify";
 
 type Props = RouteComponentProps & {
   readonly venue: Venue;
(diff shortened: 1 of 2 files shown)


www/src/js/views/venues/VenueDetails.tsx
```diff
@@ -18,12 +18,14 @@ import { breakpointDown } from 'utils/css';
 import VenueLocation from './VenueLocation';
 
 import styles from './VenueDetails.scss';
+import { convertIndexToTime, SCHOOLDAYS } from '../../utils/timify';
```
ZhangYiJiang: Use absolute import, like `util/timify` above

www/src/js/views/venues/VenueDetails.tsx
```diff
@@ -36,13 +38,41 @@ export class VenueDetailsComponent extends React.PureComponent<Props> {
...
 
     const coloredLessons = colorLessonsByKey(lessons, 'ModuleCode');
+    if (this.props.searchedPeriod !== undefined) {
+      coloredLessons.push(this.makeSearchedPeriodLesson());
+    }
     // @ts-ignore TODO: Fix this typing
     return arrangeLessonsForWeek(coloredLessons);
   }
 
+  makeSearchedPeriodLesson() {
```
ZhangYiJiang: This isn’t what we’re looking for. We’re trying to highlight the region which the user is searching for. Inserting a lesson there would be confusing. Highlighting a part of the timetable isn’t supported by the current timetable component, so you need to dig into the timetable component to add this.

amrut-prabhu: Noted, will try to update it soon

amrut-prabhu: Just a couple of questions before I proceed further:

I'll need to create a new component like `TimeTableCell`, right?

And should it look like this? Is any text needed?
image

Thanks!

ZhangYiJiang: It doesn’t need to be a component, although it can be. It doesn’t need to display any text, and should go behind the cells visually

PR #1567, opened 2019-02-17
Fix navbar clipping bug

li-kai:
Seems like safari doesn't like it when svgs have a float height. 0.875rem = 14px.

Although I could be wrong, debugging this the logo is really tiny.


taneliang:
Nice! Thanks for fixing it :D could you also replace the SVG with the updated one?

PR #1576, opened 2019-02-18
Switch to Typescript FSA

ZhangYiJiang:
Experimenting with TypeScript FSA. This gives us proper typing on the `payload` prop in the reducer automatically if we use their helpers. 

The downside is that we can't use `switch`, which makes the code a bit uglier. The bigger issue is that the action must always take one parameter which is the shape of the payload. This is problematic because we're breaking abstraction as the action now can't serve as a clean interface between the view and the model (reducers), since changes to the action's internal implementation (payload shape) now leaks into the views. 

On the other hand most of our action creators are simply binding parameters to payload, so it might not be too big an issue? 

Alternatively we can consider [`typesafe-actions`](https://github.com/piotrwitek/typesafe-actions#typesafe-actions) which does offer a `map` method which allows us to map the params to the payload, at the expense of more boilerplate.

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -164,7 +164,9 @@
     "redux": "4.0.1",
     "redux-persist": "5.9.1",
     "redux-thunk": "2.2.0",
-    "reselect": "4.0.0"
+    "reselect": "4.0.0",
+    "typescript-fsa": "3.0.0-beta-2",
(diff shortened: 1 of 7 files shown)


ZhangYiJiang:
Closing in favor of #2280

PR #1578, opened 2019-02-21
Hasura GraphQL backend

li-kai:
- [x] Sets up GraphQL backend
- [x] Adds authentication
- [ ] Add readme for common commands
- [x] Add tests

For now:
```
cp .env.example .env
docker-compose up
```


api/hasura/migrations/1550397235824_init.up.sql
```diff
@@ -0,0 +1,311 @@
...
+--
+
+COMMENT ON TABLE public.lesson IS 'Contains school lesson/recitation/lecture data';
+
+
+--
+-- Name: url; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.url (
```
ZhangYiJiang: We're building our own URL shortener? What's wrong with the current one we're using?

api/hasura/src/ExpiringMap.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+class ExpiringMap<K, V> {
+  private innerMap: Map<K, Value<V>>;
+  private ttl: number;
+  private vacuumHandler: NodeJS.Timeout;
+
+  constructor(ttl: string) {
+    this.innerMap = new Map();
+    this.ttl = ms(ttl);
+    this.vacuumHandler = setInterval(() => this.vacuum(), this.ttl);
```
ZhangYiJiang: Is this necessary? It might be simpler to just run the cleanup function randomly (1 in x chance) when getting or setting values. There will be some "wasted" space, but it probably won't be very large. You can use `setImmediate` if you're worried about the vacuum blocking returning the value

api/hasura/src/token.test.ts
```diff
@@ -0,0 +1,10 @@
+import tokenGenerator from './token';
+
+jest.mock('./config');
+
+describe('getMailToken', () => {
+  it('should generate token', () => {
+    const token = tokenGenerator.getMailToken();
+    expect(typeof token).toBe('string');
```
ZhangYiJiang: I think `.toBeInstanceOf` works better here

api/hasura/src/token.ts
```diff
@@ -0,0 +1,55 @@
+import jwt from 'jsonwebtoken';
+
+import config from './config';
+
+function getAccessToken(): Promise<string> {
+  return new Promise((resolve, reject) => {
+    jwt.sign(
```
ZhangYiJiang: Can we use `utils.promisify` for this?

api/hasura/migrations/1550397235824_init.up.sql
```diff
@@ -0,0 +1,311 @@
...
+
+CREATE TABLE public.course (
+    course_id integer GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    created_at timestamp with time zone DEFAULT now() NOT NULL,
+    updated_at timestamp with time zone DEFAULT now() NOT NULL,
+    code text NOT NULL,
+    title text NOT NULL,
+    description text NOT NULL,
+    value numeric,
+    prerequisite_text text,
```
ZhangYiJiang: The `_text` suffix is redundant

ZhangYiJiang:
Also, how do you feel about using Koa instead of Express? Koa feels more modern because it has proper support for Promise, which allows for using async/await, while Express is stuck on callbacks. 

SendGrid actually has 15k free emails per month for students as part of GitHub education pack, which may be useful for us.

li-kai:
Koa's primary router library had its recent [debacle](https://github.com/ZijianHe/koa-router/issues/494) which is pretty worrying. Given that this is for an authentication server, I am choosing express for its security and maintenance.

Yep, SendGrid seems pretty good. I don't mind paying for the spikes.

api/hasura/migrations/1550397235824_init.up.sql
li-kai: It's additional maintenance overhead. A simple hash function to new url is all we need and allows us to backup data more easily.

api/hasura/src/ExpiringMap.ts
li-kai: What is the main concern here about the design?

ZhangYiJiang: The main issue is the dangling handler that's left behind if you forget to call `cleanup`. On the client side this is fine in some cases because the page can just refresh to clear these (though it's still not a good idea), but for a long runner server process this can be a big source of pain.

api/hasura/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "author": {
+    "name": "NUSModifications",
+    "email": "mods@nusmods.com"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/nusmodifications/nusmods.git"
+  },
+  "licenses": [
```
ZhangYiJiang: Should just be license

api/hasura/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "graphql",
+  "version": "1.0.0",
+  "description": "University data GraphQL API server",
+  "homepage": "http://api.nusmods.com",
```
ZhangYiJiang:
```suggestion
  "homepage": "https://api.nusmods.com",
```

api/hasura/src/utils/ExpiringMap.test.ts
```diff
@@ -0,0 +1,64 @@
+import { advanceBy } from 'jest-date-mock';
+import ExpiringMap from './ExpiringMap';
+
+jest.mock('./config');
+jest.useFakeTimers();
+
+describe('ExpiringMap', () => {
```
ZhangYiJiang: /nit use class and functions directly as names in `describe`

api/hasura/src/utils/configUtils.ts
```diff
@@ -0,0 +1,43 @@
+import assert from 'assert';
```
ZhangYiJiang: Use [strict mode](https://nodejs.org/docs/latest-v10.x/api/assert.html#assert_strict_mode) instead of legacy 

```suggestion
import { strict as assert } from 'assert';
```

api/hasura/README.md
```diff
@@ -0,0 +1,67 @@
...
+### Authentication server
+
+Clients, such as the webpage at [nusmods](https://nusmods.com), will call the few paths below.
+
+The server is written in Javascript. The code runs on [NodeJS](http://nodejs.org/), which provides the basics for running a web server.
+
+### Secondary Goals
+
+1. Account support that stores user information so users no longer have to copy-paste urls.
+1. URL shortening that removes our dependency on [yourls](https://yourls.org/), and hence the MySQL database. (WIP)
```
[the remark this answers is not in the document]
li-kai: 😢 Oh, is that so...

api/hasura/migrations/1550397235824_init.up.sql
```diff
@@ -0,0 +1,339 @@
...
+--
+
+COMMENT ON TABLE public.url IS 'URL shorterner service table';
+
+
+--
+-- Name: days_of_week; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.days_of_week (
```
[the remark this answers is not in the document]
li-kai: When I was working on this, there wasn't enum support yet. Or I could have been blind. Leaning towards the latter.

api/hasura/src/utils/ExpiringMap.test.ts
li-kai: I'd do that but I don't understand _why_ we should. 😕

api/hasura/src/__mocks__/config.ts
```diff
@@ -1,17 +1,26 @@
...
+  },
   // Access Token for obtaining information (aka Hasura Token)
-  hasuraTokenNameSpace: '',
-  hasuraTokenSecretKey: '',
-  hasuraTokenSecretAlgorithm: '',
-  hasuraTokenLifeTime: '1s',
+  accessToken: {
+    nameSpace: '',
+    secretKey: '',
+    secretAlgorithm: '',
```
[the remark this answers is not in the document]
li-kai: I don't fully understand your question. I put this here because these are shared environment variables with hasura. We can hard code it but that means duplicating information.

api/hasura/src/token.ts
li-kai: I tried it but typescript complains and it's harder to fix the typings to just write out own promisfy.

api/hasura/migrations/1550397235824_init.up.sql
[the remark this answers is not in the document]
ZhangYiJiang: My personal preference is to encode enums at the application level, not the database, although for something like day of week that’s probably fine because the week is not getting longer :P

api/hasura/src/utils/ExpiringMap.test.ts
ZhangYiJiang: When renaming the function and class the IDE will rename both. Right now it’s just a string, so the two can become different

api/hasura/src/mailService.ts
```diff
@@ -3,15 +3,15 @@ import config from './config';
...
 
-function sendToken(recipientEmail: string, token: string) {
+function sendPasscode(recipientEmail: string, passcode: string) {
   return sendgrid.send({
     from: config.mailAddress,
     to: recipientEmail,
-    subject: 'Your NUSMods auth token',
-    html: `<h1>Token</h1><p>Your token is: ${token}</p>`,
+    subject: 'Your NUSMods auth passcode',
+    html: `<h1>Passcode</h1><p>Your passcode is: ${passcode}</p>`,
```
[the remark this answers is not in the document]
li-kai: Should it be hardcoded or through a `callbackUrl` variable in the post body?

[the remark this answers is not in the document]
ZhangYiJiang: Usually a `host` environment variable would be available via config

api/hasura/src/Database.ts
```diff
@@ -0,0 +1,166 @@
...
+  RETURNING   account_id
+)
+SELECT account_id FROM ins
+UNION  ALL
+SELECT account_id FROM account
+WHERE  email = $1
+LIMIT  1;
+`.trim();
+
+const INSERT_SESSION = `
```
ZhangYiJiang: Are you sure you don't want to use a query builder? For complex queries like the upsert account query it's not going to be powerful enough, but for simple queries query builders make it much simpler without having to switch to another language.

api/hasura/src/__tests__/utils/dbMigration.ts
```diff
@@ -0,0 +1,50 @@
+import fs from 'fs';
+import path from 'path';
+import { promisify } from 'util';
+import { Client } from 'pg';
+
+const afs = {
```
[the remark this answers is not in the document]
li-kai: fs-extra doesn't have readdir with the `{ withFileTypes: true }` option, and we're definitely only needing fs here for this file, so I think promisifying the two functions should be acceptable.

api/hasura/src/Database.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+      const session = sessions.rows[0];
+      expect(session).toMatchObject({
+        session_id: sessionId,
+        account_id: accountId,
+        user_agent: userAgent,
+      });
+
+      expect(session.expires_at).toEqual(expiresAt);
+      expect(dateUtils.toEpoch(session.last_accessed_at)).toBeLessThanOrEqual(Date.now());
```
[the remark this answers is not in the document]
li-kai: Typescript enforces `toBeLessThanOrEqual` to be numbers, so I've changed this to a <= with `toBeTruthy` check

api/hasura/src/Database.ts
li-kai: This should be the last of it, I'm not sure if we're going to need any more queries. I'll pull in knex if we exceed 10.

api/hasura/src/Database.test.ts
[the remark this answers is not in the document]
ZhangYiJiang: Hmm, it looks like Jest checks that the operands are numbers at runtime too, which is a bit odd. If you're fine with adding a dep, jest-extended adds support for comparing dates https://github.com/jest-community/jest-extended#tobeafterdate

api/hasura/migrations/1550397235824_init.up.sql
li-kai: Ah, I was preparing for us to add `prerequisite` as a json glob. Yay or nay?

api/hasura/src/token.test.ts
li-kai: Jest just fails with `toBeInstanceOf` checks.

li-kai:
defering until #1726 gets merged

PR #1581, opened 2019-02-22
Show map by default on Today and Venues page

### LINKED ISSUE #1525: Display the map by default on the venue page
Since we have bus route and timing info merged in, the map should now be useful enough to be shown by default on the venues page. This will let students check bus timings without having to click through additional steps.

ZhangYiJiang:
Fixes #1525 

Displays the map by default without a position marker on venue and today page. This lets the user look at bus timings and other info without having to open up a venue or class, especially once we start adding more features to the map 

There's also currently no way to unselect a venue, not entirely sure if that's necessary.  Maybe we should split off to a separate site or create a new tab. 

### Todo 

- Add geolocation button so the map can be zoomed to where the user is 
- Allow mobile users to access the map (currently the mobile view don't have a default view). Currently thinking of a pull-tab style UI for opening up the map from the bottom 

### Backend changes 

- Upgrade to React 16.8
- Split LocationSelect off from improve venue form 
- Extract GestureHandling into its own React component


ZhangYiJiang:
Closed because the design doesn't quite work. We'll probably need a major redesign for this to work

PR #1582, opened 2019-02-22
Add tooltips for navtab labels 

ZhangYiJiang:
With this PR, on mobile devices users can now long press the navtabs to see the label which is currently only shown on desktop / tablet sizes

### DIFF
--- a/www/src/js/utils/css.ts
+++ b/www/src/js/utils/css.ts
@@ -5,13 +5,16 @@ import bowser from 'bowser';
 
 export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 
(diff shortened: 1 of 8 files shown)


li-kai:
This is a really nice feature but I question the feasibility. Long holds are practically never used on the web, and by the time the long press triggered, they could have loaded the page. 
I don't think it's worth shipping this, if we account for the additional javascript needed to load the tooltips.

ZhangYiJiang:
Tooltip is lazy loaded, and I don't think the few additional lines of code to wrap the links would make a big difference here.

ZhangYiJiang:
A little too outdated, esp. since Tippy is now two major versions ahead. Going to close this and maybe see if the code can be rewritten for the new version

PR #1583, opened 2019-02-24
Highlight search period in Venues page

amrut-prabhu:
Part of #538

Highlight period of time searched for when availability search is on.

Created a new component `TimetableHighlight.tsx` and added type `timePeriod.ts`


amrut-prabhu:
@ZhangYiJiang Ready for review :D

Not sure why the tests are failing :/
They seem lint related, but I haven't modified those files

www/src/js/types/timePeriod.ts
```diff
@@ -0,0 +1,53 @@
...
+  Day: number; // Day of week (ie. 0 = Monday, 1 = Tuesday etc.)
+  DayText: DayText;
+  StartTime: StartTime;
+  EndTime: EndTime;
+};
+
+/**
+ * Represents a time period with a specified color in the timetable.
+ */
+export type ColoredTimePeriod = TimePeriod & {
```
ZhangYiJiang: This type doesn't make any sense. We have `ColoredLesson` because lessons are colored on timetables. The highlighted period doesn't really need this, especially since it doesn't use theme color

www/src/js/types/timePeriod.ts
```diff
@@ -0,0 +1,53 @@
+import { SCHOOLDAYS } from 'utils/timify';
+
+export type DayText = string; // E.g. "Monday", "Tuesday"
```
ZhangYiJiang: Do not redeclare types

www/src/js/types/timePeriod.ts
```diff
@@ -0,0 +1,53 @@
...
+export type StartTime = string; // E.g. "1400"
+export type EndTime = string; // E.g. "1500"
+
+/**
+ * Represents a time period in the timetable.
+ */
+export type TimePeriod = {
+  Day: number; // Day of week (ie. 0 = Monday, 1 = Tuesday etc.)
```
ZhangYiJiang: Use camelcase. Lesson uses titlecase because of the API, but this type is used purely on the frontend so it should follow the casing for the rest of the codebase

www/src/js/types/timePeriod.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export type ColoredTimePeriod = TimePeriod & {
+  colorIndex: number;
+};
+
+/**
+ * Returns a generic time period with some default parameter values, if they are not specified.
+ */
+export function createGenericTimePeriod(
+  day: number = 0,
```
ZhangYiJiang: I don't think default params are a good idea here. It doesn't make sense to arbitrarily set start and end time like this if one is not provided - eg. if I call `createGenericTimePeriod(1, '0900')`, in the current code this will produce a `TimePeriod` which ends before starting

www/src/js/types/timePeriod.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Represents a time period with a specified color in the timetable.
+ */
+export type ColoredTimePeriod = TimePeriod & {
+  colorIndex: number;
+};
+
+/**
+ * Returns a generic time period with some default parameter values, if they are not specified.
+ */
+export function createGenericTimePeriod(
```
ZhangYiJiang: This can just be `createTimePeriod`

www/src/js/views/timetable/TimetableHighlight.scss
```diff
@@ -0,0 +1,16 @@
+@import './_variables.scss';
+
+.row {
+  position: absolute;
+  display: flex;
```
ZhangYiJiang: `display: flex` and all of the `flex` properties won't do anything on an absolute position element

www/src/js/views/timetable/TimetableHighlight.scss
```diff
@@ -0,0 +1,16 @@
...
+  flex: 1 1 auto;
+  flex-direction: column;
+  min-height: $timetable-row-height;
+  background: rgba(255, 255, 0, 0.7);
+
+  @include vertical-mode {
+    flex-direction: row;
```
ZhangYiJiang: Add `width: 100%`, otherwise the element will collapse to 2px wide in vertical mode

www/src/js/views/timetable/TimetableHighlight.scss
```diff
@@ -0,0 +1,16 @@
...
+    padding: 0 0 0 1px;
+    background: rgba(255, 255, 0, 0.7);
```
ZhangYiJiang: This is unnecessary

www/src/js/views/timetable/TimetableRow.tsx
```diff
@@ -36,8 +39,34 @@ function TimetableRow(props: Props) {
...
+    }
+    return undefined;
+  }
+
+  const highlightStyle = getHighlightStyle(props.highlightPeriod);
+
   return (
     <div className={styles.timetableRow}>
+      {props.highlightPeriod !== undefined ? (
+        <TimetableHighlight
```
ZhangYiJiang: I think this can be lifted to `TimetableDay` since the same period is highlighted for all rows

www/src/js/views/timetable/TimetableRow.tsx
```diff
@@ -36,8 +39,34 @@ function TimetableRow(props: Props) {
   const sizeStyle = verticalMode ? 'height' : 'width';
 
   let lastStartIndex = startingIndex;
+
+  function getHighlightStyle(period?: ColoredTimePeriod) {
+    if (period !== undefined) {
+      const periodStartIndex: number = convertTimeToIndex(period.StartTime);
```
ZhangYiJiang: This logic is nearly the same as the one used for calculating lesson position, and can be extracted to `utils/timetable` since it is now used in two different components if you move `TimetableHighlight` up to `TimetableDay`

ZhangYiJiang: Also it's not necessary to declare types on all variable declarations

www/src/js/views/timetable/TimetableHighlight.scss
```diff
@@ -0,0 +1,16 @@
+@import './_variables.scss';
+
+.row {
```
ZhangYiJiang: The class name `row` doesn't make sense. Try `highlight` instead

www/src/js/views/timetable/TimetableHighlight.tsx
```diff
@@ -0,0 +1,19 @@
...
+type Props = {
+  highlightPeriod: ColoredTimePeriod;
+  style?: React.CSSProperties;
+};
+
+/**
+ * A highlighted time period in the timetable.
+ */
+function TimetableHighlight(props: Props) {
+  return <div className={classnames(styles.row)} style={props.style} />;
```
ZhangYiJiang: Can add `className` as a prop to let parent add additional classes.

www/src/js/views/timetable/TimetableHighlight.tsx
```diff
@@ -0,0 +1,19 @@
+import classnames from 'classnames';
+import * as React from 'react';
+
+import { ColoredTimePeriod } from 'types/timePeriod';
+import styles from './TimetableHighlight.scss';
+
+type Props = {
+  highlightPeriod: ColoredTimePeriod;
```
ZhangYiJiang: This prop does not appear to be used

www/src/js/types/timePeriod.ts
ZhangYiJiang: Also a factory a simple type like this is probably unnecessary, especially since you're not providing getter/setter functions, so this provides no abstraction anyway

www/src/js/views/timetable/TimetableRow.tsx
amrut-prabhu: @ZhangYiJiang 
I initially added `TimetableHighlight` in `TimetableDay`, but changed it to `TimetableRow` in [9959e18](https://github.com/nusmodifications/nusmods/pull/1583/commits/9959e18ecb2cb7aa2cd79d13d15a128ff83fccd7). 

I did so due to some CSS problems caused by the div added [here](https://github.com/nusmodifications/nusmods/pull/1583/commits/9959e18ecb2cb7aa2cd79d13d15a128ff83fccd7#diff-59c4c68e31d50421b7194765079dd72cL109) earlier. The height of some lessons
decreases. I tried to fix it, but couldn't manage to do so.
2019-02-26_19h01_33

www/src/js/types/timePeriod.ts
amrut-prabhu: > Also a factory a simple type like this is probably unnecessary, especially since you're not providing getter/setter functions, so this provides no abstraction anyway
Not sure if I get what you mean. Do you mean that I should move `TimePeriod` to some other file (like `modules.ts`)?

www/src/js/views/timetable/TimetableRow.tsx
amrut-prabhu: I encountered a [problem](https://github.com/nusmodifications/nusmods/pull/1583#discussion_r260233861) when I tried to put it in `TimetableDay`.

amrut-prabhu:
@ZhangYiJiang Necessary changes have been made

www/src/js/types/timePeriod.ts
ZhangYiJiang: No, a function like `createGenericTimePeriod` is a [factory method](https://en.wikipedia.org/wiki/Factory_method_pattern) - it creates TimePeriod objects. This is useful if you want abstraction, that is all access to TimePeriod is done via setter and getters so its internal structure is not revealed. In this case though there's nothing very special about TimePeriod, so there's really no need to create a function just to create a plain object like this.

www/src/js/views/timetable/TimetableRow.tsx
ZhangYiJiang: After taking another look at this the code isn't going to be similar enough, so I left them as they are before. The extracted getElementStyle actually made the code messier because it is accessing mutable state outside the function, which makes the code a lot harder to read. When extracting functions make sure the data which the function is operating over is passed in via parameters.

www/src/js/views/timetable/TimetableRow.tsx
ZhangYiJiang: Fixed this. You might want to read up about how layout using absolute position

amrut-prabhu: Alright, thanks for making the required changes. I'll look into them to understand how to go about it in the future!

www/src/js/views/timetable/Timetable.tsx
```diff
@@ -102,6 +106,12 @@ class Timetable extends React.PureComponent<Props, State> {
                     ? currentTimeIndicatorStyle
                     : nullCurrentTimeIndicatorStyle
                 }
+                highlightPeriod={
+                  this.props.highlightPeriod !== undefined &&
```
li-kai: Could you hoist the highlightPeriod higher so the code reads something like:
```
const { highlightPeriod } = props;
// ...
highlightPeriod={ highlightPeriod && highlightPeriod.day === index ? highlightPeriod : undefined }
```

www/src/js/utils/timify.ts
```diff
@@ -82,13 +84,21 @@ export const DEFAULT_LATEST_TIME: LessonTime = '1800';
...
 ): { startingIndex: number; endingIndex: number } {
   let earliestTime: number = convertTimeToIndex(DEFAULT_EARLIEST_TIME);
   let latestTime: number = convertTimeToIndex(DEFAULT_LATEST_TIME);
   lessons.forEach((lesson) => {
     earliestTime = Math.min(earliestTime, convertTimeToIndex(lesson.StartTime));
     latestTime = Math.max(latestTime, convertTimeToIndex(lesson.EndTime));
   });
+
+  // Consider time range of period, if applicable
+  if (period !== undefined) {
```
li-kai:
```suggestion
  if (period != null) {
```
This handles both `undefined` and `null`, which prevents footguns from other developers.

www/src/js/views/timetable/Timetable.tsx
amrut-prabhu: Alright, done.

Would it be good to do the same for `this.props.lessons` and `this.props.isVerticalOrientation` since they're used more than once?

li-kai: Those are fine since they don't affect readability that much.

li-kai:
@amrut-prabhu Could you run `yarn lint:code --fix`? Seems like there are some fixable linting errors. Sorry for the unreadable eslint output, we're working on it.

amrut-prabhu:
@li-kai Fixed the relevant lint errors.

li-kai:
Thanks for the good work!

amrut-prabhu:
@li-kai @ZhangYiJiang Thanks for you reviews and help btw!

PR #1591, opened 2019-02-26
Regenerate www and export lockfile

ZhangYiJiang:
Rebuilt `yarn.lock` file using the extremely scientific method of deleting the old files and running `yarn` again. 

I *think* this is working because `yarn check` now comes up cleaner, and the lockfile is now much smaller, although `www` still has one error for some reason which doesn't make any sense. Also reduces the number of errors in `yarn audit` because transitive dependencies are updated, although this might break things (though hopefully not) 

Yarn version: 1.13 stable


ZhangYiJiang:
Coverage is a bit flaky because I added a non-deterministic test. I thought the coverage should be stable if I use enough iterations, but apparently not

PR #1593, opened 2019-03-02
Replace use of immutability-helper with immer

### LINKED ISSUE #1534: Replace usage of immutability-helper with immer
As per title, in www/, we're using both immer and immutability-helper. That is not ideal, and since immer is the better replacement, we should remove usages of immutability-helper.

rrtheonlyone:
Resolves #1534 

I have replaced the files that use `immutability-helper` with `immer` and used the test suite to verify that they are working.

Some obstacles I ran into (did not replace in these files):
1. **moduleBank.ts**
Immutability-helper is used as follows:
```
return update(state, {
        moduleArchive: {
          [action.payload.ModuleCode]: {
            $auto: {
              [meta.academicYear]: {
                $auto: { $set: action.payload },
              },
            },
          },
        },
      });
```
The `$auto` keyword implies that it will create an object if needed or update an existing one. However, if I were to use `immer`, how would I translate this in a clean way? I can manually check it using JavaScript but its quite clunky.

2. **FilterGroup.ts**
It uses immutability-helper like this:
```
  const params = qs.parse(query);
  const updater: Spec<FilterGroups> = {};

  each(groups, (group) => {
    const currentQuery = group.toQueryString();
    if (currentQuery === params[group.id] || (!params[group.id] && !currentQuery)) return;
    updater[group.id] = { $set: group.fromQueryString(params[group.id]) };
  });

  if (isEmpty(updater)) return groups;
  return update(groups, updater);
```
For `immer`, I guess we have to replace the `updater` variable. I tried something like this:
```
return produce(groups, (draftGroups) => {
  each(draftGroups, (group) => {
    const currentQuery = group.toQueryString();
    if (currentQuery === params[group.id] || (!params[group.id] && !currentQuery)) return;
    draftGroups[group.id] = group.fromQueryString(params[group.id]);
  });
});
```
But, I am not sure if this is correct though?

Note: `immer` also has support for types! Should we add types into the `produce` functions? This will make tracing errors easier.


ZhangYiJiang:
For autovivification, try using lodash's `get` and `set`. I don't think writing it out in JS would be too cumbersome though. It would just be an additional line to check and add the missing property 

```js
if (!draft.moduleArchive[payload.ModuleCode]) draft.moduleArchive[payload.ModuleCode] = {};
```

Your second piece of code is weird. Did you make a copy-paste error? 

Also the types for `produce` should be applied already. There shouldn't be any need to explicitly declare the types on `draft`

rrtheonlyone:
@ZhangYiJiang I will update the first part now!

Yup sry, in the second part, I was referring to something else! Have updated my comment

ZhangYiJiang:
Yes, I think the second piece of code looks correct. Do remember to remove `immutability-helper` from `package.json` once everything is cleaned up

li-kai:
@rrtheonlyone Do you need assistance in fixing this issue? If you're busy, we can get the existing changes merged first. 😄

rrtheonlyone:
Hey @li-kai so sry, was caught up in school work.

I can push in a final version later tonight for review again. I am unsure of some specific parts but I will point them out when I push in the new code. 

I am actually 90% done (just one change which has been on my waiting list for a long time)

li-kai:
School definitely takes priority, I just wanted an update as PRs that hang around too long tend to accumulate merge conflicts. Take your time, there is no rush.

rrtheonlyone:
I have updated my changes and pushed them in.

There is one part which I am was unable to change (its commented out in the code):
```
    const newUpdated = produce(this, draft => {
       draft.filters[id].enabled = newValue;
    });
```

The above is a replacement to this:
```
    const updated = update(this, {
      // @ts-ignore
      filters: { [id]: { enabled: { $set: newValue } } },
    });
```

However, this seems to mutate the value. I tried playing around with it but was unable to fix it.

I have updated the rest to use `immer`

ZhangYiJiang:
@rrtheonlyone The code doesn't work not because its mutating the value, but because immer will not work with any class that doesn't have the `immerable` symbol set

PR #1614, opened 2019-03-13
Switch scraper casing to camelCase

ZhangYiJiang:
All tests passing, and I have ran the scraper successfully locally. The rationale for doing this is that since the model changes are already backwards incompatible, then we might as well do this. It's surprisingly easy with TS, tests and an IDE that can read TS. 

Also rename `DayText` to `Day`,  `ModuleTitle` to `title` and `ModuleDescription` to `description` 

The client PR can be merged in once the scraper is merged.


scrapers/nus-v2/src/services/requisite-tree/index.ts
```diff
@@ -55,8 +55,8 @@ function parse(data: ModuleWithoutTree[], subLogger: Logger): PrereqTreeMap {
   const results: PrereqTreeMap = {};
 
   for (const module of data) {
-    const moduleCode = module.ModuleCode;
-    const value = module.Prerequisite;
+    const { moduleCode } = module;
```
li-kai:
```suggestion
    const { moduleCode, prerequisite: value } = module;
```

scrapers/nus-v2/src/services/requisite-tree/index.ts
```diff
@@ -55,8 +55,8 @@ function parse(data: ModuleWithoutTree[], subLogger: Logger): PrereqTreeMap {
...
+    const value = module.prerequisite;
```
li-kai:
```suggestion
```

PR #1633, opened 2019-03-21
Add size to timetable lesson

ZhangYiJiang:
Map the `csize` properties from the `classtt` endpoint to `size` properties on lessons in our API. Not entirely sure what to do with these yet though.


ZhangYiJiang:
To be quite honest I'm not entirely sure what the size is for. It's probably not the actual class size, but either an estimate or the max size.

PR #1635, opened 2019-03-23
Implement planner placeholder

### LINKED ISSUE #1468: Support adding placeholder modules to module planner

Study planners from NUS faculties frequently include blanks indicating things like University Level Requirement (ie. GE* modules) or Unrestricted Electives for students to fill in. 

screenshot from 2019-01-22 19-38-26

This can be useful to pad out semesters without having to decide on something immediately, and for us to support using a preset planner during onboarding 

A way to insert these modules into the planner. We also need to consider the data model and UI changes necessary.

ZhangYiJiang:
Fixes #1468 

Peek 2019-03-26 11-39

### User interface 

Adds placeholders for entire categories of modules. Externally I think we can call this 'categories', while internally this feature is referred to as 'placeholders'. This allows users to add categories which will count as MCs without having to enter any specific module code, though they can select from a pre-filtered list if they want to fill in a specific module in the slot. These placeholders otherwise have the same behavior as normal modules in the planner 

A new `PlannerModuleSelect` component is implemented using Downshift to allow users to select modules from a list, or type in the module code they need. This is also now enabled on the normal add module textarea. This new component has fairly complex UI interaction rules which I hope are both intuitive and makes it easy to enter data. 
Placeholders themselves are coded manually in `util/placeholders` and are implemented as 

``` ts
export interface PlannerPlaceholder {
  id: string;
  name: string;
  modules?: Set<ModuleCode>;
  filter?: (module: ModuleCondensed) => boolean;
}
``` 

where `modules` indicate a limited, fixed set of modules while `filter` is a more traditional filter (we might want to swap this for `ModuleInformation` in the future, but for now I'm just using `ModuleCondensed` because its easier to work with). 

### Data changes 

This PR also involves a relatively complicated migration to a new shape for `PlannerState.modules` because the old one is not expressive enough to hold the data for this. The new shape is 

``` typescript 
modules: {
  [id: string]: {
    // The key in PlannerState.modules
    id: string;
  
    // The year, semester and zero-indexed position of the module in the planner
    year: string;
    semester: Semester;
    index: number;
  
    // Technically this should be { moduleCode } | { moduleCode?, placeholderId }
    // that is, it should either have a placeholder ID and maybe a module code
    // or a module code, but because TS handles union types poorly we're just
    // merging the two here
    moduleCode?: ModuleCode;
    placeholderId?: string;
  }
}
```


taneliang:
Something broke and you can't add 2 mods to the planner. If you add a mod to the planner, then add another mod to the planner, the second mod will replace the first mod.

ZhangYiJiang:
@taneliang Fixed. The bug is pretty embarrassing - you can see the commit. https://github.com/nusmodifications/nusmods/pull/1635/commits/9272f44d28ca23c01816c3b32c7d964ae8ef3171

PR #1636, opened 2019-03-23
Clean up cyclic references

### LINKED ISSUE #1556: Clean up cyclic references
Our codebase currently has quite a few cyclic imports. A cyclic import can cause import to fail in unexpected ways, so it is best if we can remove them. A cyclic import happens when module A imports something from module B, and module B also import something from A. To see all cyclic imports in our codebase, uncomment the line `'import/no-cycle': 'off'` in `.eslintrc.js` and run ESLint using `yarn lint:code`. 

The biggest one in our current codebase is around reducers - I think this is because the substate reducers are also exporting their own types, which is then being exported to the main reducer. An easy solution for cyclic import is to lift one of the elements into its own file. 

This is not terribly urgent, because as far as I can tell our code still works, but as I said it can cause imports to fail unexpectedly, so it's best if we fix this sometime.

amrut-prabhu:
Fixes #1556


li-kai:
Wow this is awesome! Thanks for taking on this issue, @amrut-prabhu. In order to help with the reviewing, could you briefly explain your approach? 

Meanwhile, we'll prioritise getting this reviewed and merged so you don't have further merge conflicts.

amrut-prabhu:
@li-kai I'll get back to you tonight. I think I still need to clean up a bit more

amrut-prabhu:
> could you briefly explain your approach?

@li-kai My approach to remove the cycles was a bit haphazard since I'm not completely familiar with where things should go in the codebase. But, here's the end result.

In addition, should we add `'import/no-cycle': 'error'` in eslint to avoid cycles in the future?

---
Changes in this PR:

Created:
- `actions/constants` (constants from `moduleBank`, `undoHistory`, `export`)
- `types/planner`
- `types/state` (`State` from `reducers`)
- `types/vendor/piwik` (from `types/views`)

Moved to `types/views`:
- `ModuleWithColor` from `types/modules`

Moved to `types/reducers`:
- Moved `ModulesMap` from `reducers/moduleBank`
- Moved `ModulesBank` from `reducers/moduleBank`
- Moved `UndoHistoryState` from `reducers/undoHistory`
- Moved `VenueBank` from `reducers/venueBank`
- Moved `ModuleTableOrder` from `types/views`

Moved to `types/planner`:
- Moved `CustomModule` from `types/reducers`
- Moved `CustomModuleData` from `types/reducers`
- Moved `ModuleTime` from `types/reducers`

Moved to `types/timetables`:
- Moved `Lesson` from `types/modules`
- Moved `ColoredLesson` from `types/modules`
- Moved `ModifiableLesson` from `types/modules`
- Moved `ColorIndex` from `types/reducers`

Moved to `types/venues` :
- Moved `BusTiming` from `types/views`
- Moved `TimePeriod` from `types/views`

Moved to `utils/filters/FilterGroup`:
- Moved `FilterGroupId` from `types/views`

ZhangYiJiang:
There's no need to explicitly add that rule because it's already enabled by default (if it wasn't then we wouldn't need to explicitly override it to turn it off before). I've merged most of the files where I don't think a split is necessary. I still don't like `actions/constants`, but that's relatively minor, and we can look at typescript-fsa (#1576) or some other typesafe library to help remove that boilerplate

ZhangYiJiang:
Yeah, that's partly my fault because WebStorm automatically reorders imports in alphabetical order when refactoring

PR #1648, opened 2019-03-27
Hasura firebase

li-kai:


li-kai:
Defering until #1726 gets merged.

PR #1658, opened 2019-04-04
Make NUSMods official 

ZhangYiJiang:
Add a banner announcement, and a button to report errors (which is just a mailto link). I'm not sure if the button is the best solution - we can spend a bit more time to set up a proper form like the one we use for location updates, but I don't know if it's worth the effort.

### DIFF
--- a/www/src/js/views/components/notfications/Announcements.tsx
+++ b/www/src/js/views/components/notfications/Announcements.tsx
@@ -64,7 +64,9 @@ class Announcements extends React.PureComponent<Props, State> {
           <p>
             We&apos;re official! NUSMods now uses data provided directly by NUS Registrar&apos;s
             Office and NUSIT, which means better and more accurate information for you.{' '}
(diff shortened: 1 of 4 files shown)


taneliang:
The nice things about life

Are that they a

www/src/js/views/components/notfications/Announcements.tsx
```diff
@@ -51,10 +54,17 @@ class Announcements extends React.PureComponent<Props, State> {
...
+              ❤️
+            </span>{' '}
+            NUS
+          </h3>
           <p>
-            CORS has been updated with lesson timetables. Please note that these can change at any
-            time during the holidays, so check back often. Happy planning!
+            We&apos;re official! NUSMods now uses data provided directly by NUS Registrar&apos;s
+            Office and NUSIT, which means better and more accurate information for you.{' '}
+            <ExternalLink href={`${config.contact.blog}/title`}>Learn more.</ExternalLink>
```
taneliang: We need to update this link when the blog post is posted

PR #1662, opened 2019-04-11
Improve basic documentation

li-kai:
As per title. I tried to change the structure and language to be more accommodating to beginners.


yangshun:
@li-kai You deserve a medal 🏅

PR #1689, opened 2019-04-14
Ignore tetris folder in test coverage

li-kai:
This prevents coveralls from failing a build, as coverage from the test is pretty inconsistent.


ZhangYiJiang:
Good idea, but I don't think this is working. The test coverage has not gone down by ~8% which is the amount it increased by when I added the test. This is probably because it's just ignoring coverage of code from that folder, and not the test in the folder (or you've just configured it wrong)

li-kai:
Turns out its `/views/tetris` and not `/tetris`. Doh.

li-kai:
@ZhangYiJiang ping

ZhangYiJiang:
@li-kai I think adding `.skip`, so that it doesn't run on CI is a better idea. The coverage skip just ignores coverage of code in the tetris folder, not for coverage originating from tests in the folder, so it doesn't quite achieve what you want.

li-kai:
@ZhangYiJiang That is the intention, keep the code coverage at a constant but the test passing. That way if tetris fails, we know, but won't have PRs get the red check because the coverage decreased.

ZhangYiJiang:
No, the tests are integration tests, so they affect coverage of code outside `/views/tetris`, and because they are fuzzed (ie. randomized) their coverage also changes, so skipping coverage on the tetris folder won't fix this problem

li-kai:
Ah right. Code outside of the path.

li-kai:
Woops didn't see that you skipped the test.

ZhangYiJiang:
Closing this in favor of #1777

PR #1697, opened 2019-04-15
Improve browser support

li-kai:
I'll admit I've sort of dug a hole with this one, but I think its worth it.

This script scrapes our analytics and outputs a support matrix that provides 98% coverage of the student body. It's slightly naive, and takes shortcuts in places, but I think this is accurate enough.

As a result of this new coverage (and usage), our bundlesize has gone up by ~17kb. This is a worthy trade off to support older devices and might help prevent weird bugs from happening.


website/README.md
```diff
@@ -8,21 +8,9 @@ NUSMods R is built using [React][react], [Redux][redux] and [Bootstrap][bootstra
...
-- Last two versions of all evergreen desktop browsers (Chrome, Firefox, Edge, Safari)
-- IE is completely **unsupported**
-
-Mobile browsers:
-
-- iOS 10 and above
-- Chrome Mobile last two versions
-
-We try not to break iOS 9 (ie. display white screen of death or use unsupported APIs), but we don't guarantee the page will be appear exactly the same as in more modern browsers.
+Please refer to our [browserslist config](../packages/browserslist-config-nusmods/index.js).
```
taneliang: I think this should link to the readme instead so that there's an explanation of the data

packages/browserslist-config-nusmods/README.md
```diff
@@ -0,0 +1,19 @@
...
+yarn
+```
+
+## Generating config
+
+```sh
+yarn start
+```
+
+Files `index.js` and `browserslist-stats.json` will be generated.
```
taneliang: Could you add an explanation of what these files contain?

ZhangYiJiang:
Does this need to be a separate package? `website` already have every dependency on this list, so I'm not sure what difference splitting this out makes

packages/browserslist-config-nusmods/scripts/generate.js
```diff
@@ -0,0 +1,101 @@
...
+      continue;
+    }
+    output[browserslistCode][versionMapping] += usage;
+  }
+  return output;
+}
+
+async function output(stats) {
+  const browsers = browserslist('cover 98% in my stats', { stats });
+  const browserslistString = prettier.format(JSON.stringify(browsers), {
```
ZhangYiJiang: What advantage does this have over `JSON.stringify(browsers, null, 2)` ?

packages/browserslist-config-nusmods/scripts/generate.js
```diff
@@ -0,0 +1,101 @@
...
+    ...prettierConfig,
+    parser: 'babel',
+  });
+  console.log(`\nSupported browsers:\n${browsers.join('\n')}`);
+
+  return Promise.all([
+    fs.outputFile(path.resolve('../index.js'), `module.exports = ${browserslistString}`),
```
ZhangYiJiang: `path.resolve` doesn't do anything here. Did you forget `__dirname`?

packages/browserslist-config-nusmods/scripts/generate.js
li-kai: Prettier won't autoformat the file if it is already formatted to its liking

packages/browserslist-config-nusmods/scripts/generate.js
li-kai: path.normalize would be more appropriate...let me change that

packages/browserslist-config-nusmods/scripts/helpers.js
```diff
@@ -0,0 +1,57 @@
...
+
+function getVersionMapping(browserCode, versionString) {
+  const versions = _.get(browserslist, ['data', browserCode, 'versions']);
+  if (!versions) {
+    return undefined;
+  }
+
+  let version;
+  if (browserCode === 'edge') {
+    version = versionString.split('.')[0];
```
ZhangYiJiang: Do we want to do this for all browsers, splitting into major, minor and patch and parse int to normalize them

li-kai: Could you clarify on this?

li-kai:
Chrome was supported, but not version 63. This script will always make sure we're up to date, including old versions of evergreen browsers.

packages/browserslist-config-nusmods/scripts/helpers.js
ZhangYiJiang: Since you're already splitting major and minor version numbers for two different browsers, wouldn't it be better to do this for all of them instead of doing the parsing for just these two

ZhangYiJiang:
But the versions will only be updated when you re-run this script and update this package, so it'll still be the same as having this as a script under `/website`

packages/browserslist-config-nusmods/scripts/helpers.js
li-kai: Ah, that's because Edge's a bit of an edge case, returning long minor versions, but browserslist only needs the major version. iOS needs the `.0` for some reason on the other hand.

li-kai:
We'll have a bot using the CI in the future soon(tm) to trigger this.

packages/browserslist-config-nusmods/scripts/helpers.js
ZhangYiJiang: Edge *edge* case 👀

PR #1701, opened 2019-04-17
Assume the day is available if it is not found in DayAvailability

### LINKED ISSUE #1700: Venues that are not booked the entire day are not shown under 'Find free rooms'
### Describe the bug

Certain venues are available for the entire day on certain days. The 'Find free rooms' feature will not return such venues as a result. 

### To Reproduce

1. Go to 'Venues'
2. Find a venue that is available the entire day. (e.g. COM1-B108 in AY18/19 Sem 2 is not booked for any lessons on Wednesdays)
3. Click on 'Find free rooms' 
4. Select any time range on the fully available day of the venue (e.g. Wednesday, 3PM-4PM for COM1-B108)
5. With the 'Find free rooms' feature active, type the venue name (can be a partial match or complete match) under 'Venue Search'
6. The result of the search will be : 
a) 'No matching venues found' (if complete match of the name is typed), or 
b) it may return other available venues but still without the venue with availability the entire day (if partial match of the venue name is typed)

### Expected behavior

The venue which is available the entire day should be returned as a search result if the time range selected for 'Find free rooms' is within the day when the venue is available the entire day. 

### Screenshots
Screenshot below shows the timetable for COM1-B108 where Wednesday is available the entire day

Screenshot below is a combination of steps 3-5 of reproducing the behavior. Here, a complete match of the venue name is typed and it returns 'No matching venues found'

Screenshot below is also a combination of steps 3-5. However, here a partial match of the venue name is typed and it returns some available venues for the specified time range, but COM1-B108 is not among the options

Additional screenshots: 
These 2 screenshots below changes the time range to be within a day when the venue is not available the entire day (but still has some availability) . As seen in the screenshot, COM1-B108 still appears as expected for both partial match and complete match of venue name.
Complete match 

Partial match

* OS: Windows
* Browser: Firefox and Chrome
* Version: 66 (Firefox) and 73 (Chrome)

### Additional context

The 'Find free rooms' feature seems to work as expected for days which have lessons in the venue. I am not sure if this problem is specific to venues which are free the entire day, but I have tested and confirmed that this unexpected behavior is encountered for other venues of similar nature.

ZhangYiJiang:
## Context

Fixes #1700 

## Implementation

The old code assumed that if the day does not exist, then the room is not available. This is no longer true under API v2, so we flip that around and assume the room *is* available. 

## Other Information

There is a unit test but we should also manually test this to check if my assumption about the cause of the error is correct.

### DIFF
--- a/website/src/utils/venues.test.ts
+++ b/website/src/utils/venues.test.ts
@@ -98,6 +98,20 @@ describe(filterAvailability, () => {
       }),
     ).toEqual(getVenues('CQT/SR0622'));
   });
+
+  test('should return venue which are empty the whole day', () => {
+    const availableVenues = filterAvailability(
+      [['LT1', []]], // Venue has no lessons at all
+      {
+        day: 0,
+        time: 9,
+        duration: 1,
+      },
+    );
+
+    expect(availableVenues).toHaveLength(1);
+    expect(availableVenues[0][0]).toEqual('LT1');
+  });
 });
 
 describe(floorName, () => {
--- a/website/src/utils/venues.ts
+++ b/website/src/utils/venues.ts
@@ -37,7 +37,7 @@ export function filterAvailability(
   return venues.filter(([, venue]) => {
     const start = time * 100;
     const dayAvailability = venue.find((availability) => availability.day === SCHOOLDAYS[day]);
-    if (!dayAvailability) return false;
+    if (!dayAvailability) return true;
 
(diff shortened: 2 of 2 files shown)


ZhangYiJiang:
Verified fixed in https://deploy-preview-1701--nusmods.netlify.com/venues/COM1-B108?day=2&duration=1&q=COM1-B108&time=12

Screen Shot 2019-04-17 at 12 16 49 PM

PR #1725, opened 2019-04-28
Clamp search options duration in AvailabilitySearch and Timetable when displayed

### LINKED ISSUE #1647: Highlighted search period doesn't match with search options in Venues page
### Describe the bug

The search period highlighted in the timetable does not match with the one shown in the search options.

### To Reproduce

1. Go to 'Venues Page' (`/venues`)
2. Click on 'Find free rooms'
3. Indicate a large search period (eg. `From 5pm` - `To 10pm`)
4. Click on one of the search result venues and see the highlighted search period in its timetable.
5. Change the search period to start later. (eg. `From 8pm`)
6. See the error. The searched time period does not match the highlighted period.

### Expected behavior

The searched time period should match the highlighted period in the timetable.

### Screenshots

After step 5, the search period becomes 8-9pm (1 hour), but 8pm-12(4 hours) is highlighted 

2019-03-28_01h19_37

* OS: [Windows]
* Browser [Chrome]

ZhangYiJiang:
## Context

Fixes #1647 

The bug happens because the search options don't respect the closing hours of the school, but the loop used to generate the `select` element's options do, so when the duration exceeds the school's closing time, the select is given an invalid value and goes back to the first option. 

## Implementation

The solution is to clamp the duration before it is rendered. There is a simpler alternative which is to clamp the duration when the new time is selected, but I don't know if this will be confusing to the user.


li-kai:
We're not clamping the duration field in the url.

Also I think there's a bug: I can get stuck easily by selecting a venue by clicking around, e.g. https://deploy-preview-1725--nusmods.netlify.com/venues/AS3-0101?day=0&duration=2&time=13

Screenshot_2019-04-28 AS4-0114 - Venues

ZhangYiJiang:
I've given this another look, and I think clamping when the time is selected would actually make more sense. There's no reason why the user expects the original duration to be kept by the page. The original code was also just buggy af, probably because I wrote it while been sleep deprived

PR #1726, opened 2019-04-28
Graphql API

li-kai:
## Context
This provides a graphql api, which would help us build features faster.

## Implementation
This uses [Hasura](https://hasura.io/) to build a graphql api without much config. This implementation has no authentication flow, which will be worked out later.


api/migrations/1550397235824_init.up.sql
```diff
@@ -0,0 +1,306 @@
...
+
+CREATE TABLE public.venue (
+    venue_id integer GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    created_at timestamp with time zone DEFAULT now() NOT NULL,
+    updated_at timestamp with time zone DEFAULT now() NOT NULL,
+    code text NOT NULL,
+    name text NOT NULL,
+    latitude numeric,
+    longitude numeric,
+    alttitude numeric,
```
taneliang:
```suggestion
    altitude numeric,
```

api/README.md
```diff
@@ -0,0 +1,66 @@
+# GraphQL Server
+
+This project provides school data such as lessons and modules through an [API](https://www.mulesoft.com/resources/api/what-is-an-api).
```
taneliang: Haha I don't think we need to define what an API is. We should put a note saying that this is a WIP though

li-kai: It's helpful for beginners and it'll look like this: [API](https://www.mulesoft.com/resources/api/what-is-an-api), so I think it's fine to leave it. :D

api/migrations/1550397235824_init.up.sql
taneliang: Hold up this also appears 5 times in the .yaml file. I only corrected the .sql file, but I'm not sure if the .yaml file gets its value from this .sql file or vice versa

api/README.md
taneliang: Okay sgtm. But we should still put a WIP note so that potential contributors won't think we're already using this in prod and get confused

taneliang: Okay I've done this, and upgraded Hasura as well

api/migrations/1550397235824_init.up.sql
taneliang: I've fixed the remaining typos

li-kai: .yaml obtains values from .sql

taneliang: Ah. How do you update the .yaml file without doing it manually?

li-kai: https://docs.hasura.io/1.0/graphql/manual/hasura-cli/hasura_metadata_export.html

li-kai:
@ZhangYiJiang @taneliang  Do we want to merge this?

PR #1737, opened 2019-04-30
Use packtracker instead of bundlesize

li-kai:
## Context
Bundlesize requires installing native libraries that slow the installation process.
It also doesn't work as its reporting is somewhat wonky.

## Implementation
Uses https://packtracker.io/ to analyse our bundle instead.
Free for open source.


li-kai:
Trying with the Github App on.

PR #1738, opened 2019-04-30
Use packtracker instead of bundlesize

li-kai:
## Context
Bundlesize requires installing native libraries that slow the installation process.
It also doesn't work as its reporting is somewhat wonky.

## Implementation
Uses https://packtracker.io/ to analyse our bundle instead.
Free for open source.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -33,7 +33,7 @@
     "@babel/preset-react": "7.0.0",
     "@babel/preset-typescript": "7.3.3",
     "@hot-loader/react-dom": "16.8.6",
-    "@packtracker/webpack-plugin": "^2.0.1",
+    "@packtracker/webpack-plugin": "2.0.1",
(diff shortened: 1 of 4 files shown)


ZhangYiJiang:
This is the third bundle size tracking solution we've tried 🤣 let's see if this is better than the last two

website/package.json
```diff
@@ -34,6 +33,7 @@
     "@babel/preset-react": "7.0.0",
     "@babel/preset-typescript": "7.3.3",
     "@hot-loader/react-dom": "16.8.6",
+    "@packtracker/webpack-plugin": "^2.0.1",
```
ZhangYiJiang: Use exact version please

li-kai:
Pretty easy to revert, just remove the app and we'll be good. Bundlesize integration is not removed yet in case we wanna move back.

PR #1739, opened 2019-04-30
Transform nusmoderators to typescript

li-kai:
## Context
NUSModerator was not in typescript and this sort of blocks #1007 as well as making maintenance hard. Oh, also types were wrong which made this is a breaking change.

## Implementation
Update files to use typescript natively.
Use [tsdx](https://github.com/palmerhq/tsdx) to bundle our files 

## Other Information
@ZhangYiJiang could you help fix the eslint config? The /website config is voodoo to me now :sweat_smile:


packages/nusmoderator/src/academicCalendar.ts
```diff
@@ -162,7 +163,9 @@ export function getAcadWeekInfo(date) {
         break;
       }
       acadWeekNumber -= 1;
-      ({ weekType, weekNumber } = getAcadWeekName(acadWeekNumber));
+      const acadWeek = getAcadWeekName(acadWeekNumber)!;
```
ZhangYiJiang: Does this set variables in the global scope, or is this just a style change?

.gitignore
```diff
@@ -10,6 +10,10 @@ build/
 # Cache folders
 cache/
 .cache-loader
+.rts2_cache_cjs
```
ZhangYiJiang: Any way we can configure tsdx to use, say, `node_modules/.cache` instead? I'd prefer the cache for all our projects to be in a somewhat consistent place

packages/nusmoderator/src/academicCalendar.ts
li-kai: I don't get what you mean. I'm just doing destructuring the old way to allow the `!` for non-null assertion.

ZhangYiJiang:
@li-kai The configs look reasonable. We can always copy over configs when we actually hit those issues. Do you want to extend AirBnb style like /website?

li-kai:
The default is pretty good and the code base is so small. I don't have an opinion, feel free to change it :)

PR #1770, opened 2019-05-12
Fix URL not found crash in scraper V2

taneliang:
Fixes a "URL is not defined" crash.

image


ZhangYiJiang:
This shouldn’t be necessary if you’re running Node 10+, since in Node 10 the URL object was made a global

taneliang:
Okay so it turns out that Yarn on NixOS locks itself to Node 8, independent of the version of Node on the system. My bad

Issue #1775, opened 2019-05-17
Allow searching and display of modules not offered in the current AY


taneliang:
[NUS Bulletin's Search Modules page](http://www.nus.edu.sg/nusbulletin/search-modules/) will be removed in July, and NUSMods is to take over this functionality for the university.

![image](https://user-images.githubusercontent.com/12784593/57903531-01fdf880-78a1-11e9-8711-0a42d82f40fd.png)

However, NUSMods currently does not show modules which are not offered in the system, an essential feature. We accomplish this by ignoring all modules which do not have timetables. This was done because there was no need to show non-offered mods. However RO has since told us that:

> The reason why we may need a comprehensive listing of active modules in NUSMods (including those that haven’t been timetabled) is because students may wish to e.g. map a module they have taken while on SEP back to NUS. 

### TODO

- [ ] Modify scraper to process all mods, even if they're not offered in the AY
- [ ] Modify website to search through all 11+k modules. This will require an overhaul of the module search page as the current offline search is not scalable for longer - it currently downloads 5+MB of data (for ~5.5k mods) when the module page is loaded, and the new combined offered+non-offered mods list is over 10MB in size. Also, searching through more modules will require more computational power that weaker devices will not be able to support.

taneliang:
New module search bugs introduced in #1791 that have to be fixed before deploying

- [ ] Pager is too wide on mobile devices and tablets, especially after navigating a few pages.
- [ ] Pager is not styled correctly in dark mode.
- [ ] API error looks bad as it does not take up the whole screen.
    ![localhost_8080_modules_level 0 =8000 mcs 0 =9](https://user-images.githubusercontent.com/12784593/58023350-29272500-7b42-11e9-98db-ab0f422b103e.png)
- [ ] `<X> modules found` text does not display the current page or the current range of displayed mods, which can be confusing.
- [ ] No results found for empty search string is bad. Demo: http://localhost:8080/modules?level[0]=8000&mcs[0]=9 
    ![image](https://user-images.githubusercontent.com/12784593/58022874-047e7d80-7b41-11e9-88ed-f42710c960e1.png)
- [ ] Only show mods offered this AY by default
- [ ] Mods should be sorted by module code if there are no selected filters and no search query

taneliang:
New module search bugs caused by data changes in #1776 

- [ ] Fix display of non-offered mods in the module list (together with the Elasticsearch module search overhaul)
[...]

PR #1776, opened 2019-05-17
Add non-offered modules to scraper

taneliang:
## Context

#1775

## Implementation

- Swagger documentation has been updated
- The newly-deprecated moduleInformation.json will not contain the non-offered mods, so that existing clients will not be overwhelmed by 2x the number of mods. A new moduleInfo.json endpoint replaces it. The only difference between the 2 endpoints is that moduleInfo.json will contain modules which are not available in the AY. These mods can be identified by their empty `semesterData` array.
- moduleList.json will still NOT contain the modules which are not offered this AY.
- Data for modules not offered in a semester will not be saved in that module's folder.

## TODO in future PRs

- Fix display of non-offered mods in the module list (together with the Elasticsearch module search overhaul)
image
- Fix non-functional "Add to" button in the module page 
image
- Add notice to module page to notify user that the module is not offered in the AY


ZhangYiJiang:
The scraper code looks fine, but my main concern is actually on the client, where I'm afraid the appearance of these modules in search might be confusing. I feel like we should clearly demarcate them, or just not show them at all in global search.

taneliang:
@ZhangYiJiang yeah, I'm thinking of putting a "Not offered" label in the global search. This of course won't appear in any older client. Should we also deprecate moduleList.json and replace it like what we're doing with moduleInformation.json?

taneliang:
Let's move fast and fix in prod :1st_place_medal:

PR #1791, opened 2019-05-20
Replace local module search with Elasticsearch

taneliang:
Apologies for the size of the PR - there was no way to split it.

## Context

https://github.com/nusmodifications/nusmods/issues/1775

This PR changes the module search page to use an Elasticsearch search server, which will be added in another PR. 

TODO: explain why Elasticsearch over Algolia

## Features

- Fuzzy search!
    image

- No results suggestions 
    localhost_8080_modules_q=joke sem 0 =2 level 0 =3000 level 1 =5000

- Navigating back/forward in the browser updates the module search box text if the query is changed. Demo: type "a" in search box, wait 1 second, type "bcd" into search box, then press back button. The search box should update.

- Supports search of mods not offered in this AY - deselect all "Available In" filters to show all 11k+ mods. Unfortunately, this is the default behavior. To be fixed in a future PR.

## Implementation

- The Elasticsearch host URL is specified in the app-config.json file, but it can also be overridden by loading the modules page and setting an `eshost` param in the URL, e.g. `http://localhost:8080/modules?eshost=http://localhost:9200/modules`
- Old module filters were removed, but I'm not confident that I managed to remove everything.
- UI components and styling were reused as far as possible.
- Infinite scrolling has been replaced with paging, as that was the out of the box behavior of Searchkit's `Hits` component. It does not support infinite scrolling.
- `SearchBox` is now a controlled component, so that Searchkit can change its value.
- `VenueComponent` had to be modified to support the controlled `SearchBox`.

## Review TODO

An Elasticsearch backend needs to be started in order to run this, along with the new moduleInfo.json file. I'll try to start a staging on a new DO droplet so that you don't have to do this, but I'll take a few days.

```sh
# Generate new moduleInfo.json with scraper updated in https://github.com/nusmodifications/nusmods/pull/1776
# If you are lazy you can just wget the moduleInformation.json file from the NUSMods API, and modify the import.ts script to use that instead. Note that if you use moduleInformation.json, you won't be able to see mods not offered in this AY.
git checkout scraper-non-offered-mods
cd scrapers/nus-v2
scp <nusmods_server>:<location of scraper v2>/env.json . # Get NUS API credentials from our server
yarn dev all # Run new scraper

# Start Elasticsearch server
git checkout search
cd api/search
docker-compose up
# Wait for the containers to start. You'll know when it's ready when you can load http://localhost:5601/app/kibana#/management/kibana/indices/
cp ../../scrapers/nus-v2/data/2018-2019/moduleInfo.json . # Copy moduleInfo.json to the current directory
yarn # Install deps
yarn ts-node import.ts # Import data. Once this is done, the Kibana page linked above should show a modules index

# Start web server for this PR (probably in another shell)
cd website
yarn
yarn start
```

- [x] Ensure that code that is no longer used is removed
- [x] Check that bundle is of a decent size
- [ ] Find UI bugs to be fixed
- [X] Fix bug where the attribute filter hides itself because it's not configured correctly.
- [X] Display loading indicators when loading after the initial load.

## Minor known issues, to be fixed in future PRs

- <del>Module items in list are not highlighted with the search string. We can get highlighting specs from Elasticsearch instead of doing a substring search like the old module search page.</del> The title and description need to be escaped when inserted into ES though, or we can do it on the client side 
- Pager is too wide on mobile devices and tablets, especially after navigating a few pages.
- <del>Pager should scroll to top when page is changed, but doesn't.</del>
- Pager is not styled correctly in dark mode.
- API error looks bad as it does not take up the whole screen.
    localhost_8080_modules_level 0 =8000 mcs 0 =9
- `<X> modules found` text does not display the current page or the current range of displayed mods, which can be confusing.
- No results found for empty search string is bad. Demo: http://localhost:8080/modules?level[0]=8000&mcs[0]=9 
    image
- Only show mods offered this AY by default
- Mods should be sorted by module code if there are no selected filters and no search query


website/src/config/app-config.json
```diff
@@ -3,6 +3,7 @@
   "defaultDescription": "NUSMods is a timetable builder and knowledge platform, providing students with a better way to plan their school timetable and useful module-related information that are community-driven.",
   "academicYear": "2018/2019",
   "apiBaseUrl": "https://nusmods.com/api",
+  "elasticsearchBaseUrl": "http://localhost:9200",
```
taneliang: To be changed to a prod URL once there's a prod ES instance. Should also add an env var override for Docker

website/src/views/components/ModuleFinderItem.tsx
```diff
@@ -87,6 +86,9 @@ export class ModuleFinderItemComponent extends React.PureComponent<Props> {
   }
 }
 
-export default connect((state: State) => ({
-  search: state.moduleFinder.search,
-}))(ModuleFinderItemComponent);
+export function ModuleFinderHitModuleItem(props: HitItemProps) {
```
taneliang: Add a comment explaining that this component is a wrapper around ModuleFinderItem that makes it compatible with Searchkit.

website/src/views/modules/ModuleFinderSidebar.scss
```diff
@@ -1,7 +1,9 @@
-@import "~styles/utils/modules-entry.scss";
+@import "~styles/bootstrap/bootstrap";
```
taneliang: Should be removed

ZhangYiJiang:
The joke/jokes thing is a bit weird because both are English words. Do you know how the spelling correction works?

website/src/views/components/SearchBox.tsx
```diff
@@ -9,13 +9,13 @@ type Props = {
   className?: string;
   throttle: number;
   useInstantSearch: boolean;
-  initialSearchTerm: string | null;
-  placeholder: string;
-  onSearch: (str: string) => void;
+  value: string | null;
```
ZhangYiJiang: Are we making SearchBox a controlled component? If not this might be better named `initialValue`

website/src/views/components/SearchkitSearchBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+  prefixQueryOptions?: Record<string, any>;
+}
+
+type State = {
+  input?: string;
+};
+
+/**
+ * A Searchkit wrapper around our SearchBox.
+ * @see Adapted from <a href="https://github.com/searchkit/searchkit/blob/016c899c97f72ea3ad5afc017345e41c9003172a/packages/searchkit/src/components/search/search-box/SearchBox.tsx">Searchkit's SearchBox component</a>.
```
ZhangYiJiang: Technically JSDoc uses HTML, but this is a bit unwieldy, especially since we don't use any sort of documentation generator that actually ingests these, so I  think we should optimize for human readability instead and maybe use Markdown instead

website/src/views/components/SearchkitSearchBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+  defineAccessor() {
+    const {
+      id,
+      prefixQueryFields,
+      queryFields,
+      queryBuilder,
+      queryOptions,
+      prefixQueryOptions,
+    } = this.props;
+    return new QueryAccessor(id || 'q', {
```
ZhangYiJiang: Maybe extract `'q'` out to a constant

website/src/views/components/SearchkitSearchBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+      prefixQueryFields,
+      prefixQueryOptions: assign({}, prefixQueryOptions),
```
ZhangYiJiang: This is the same as `Object.assign`, just use that or the rest operator

website/src/views/components/SearchkitSearchBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+        if (!this.unmounted && this.state.input) {
+          this.setState({ input: undefined });
+        }
+      },
+    });
+  }
+
+  getValue() {
+    const { input } = this.state;
+    if (typeof input === 'undefined') {
```
ZhangYiJiang: Why not just set state to `this.getAccessorValue` in the first place?

website/src/views/components/filters/CheckboxItem.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { ItemProps } from 'searchkit';
+import classnames from 'classnames';
+import styles from './styles.scss';
+
+type Props = Pick<
```
ZhangYiJiang: Is the `Pick` necessary? Even if they are unused they are still passed in, so it's fine to define them. On the other hand if we want to abstract this away from `searchkit`, then this shouldn't import `searchkit` at all.

website/src/views/components/filters/CheckboxItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <label
+        htmlFor={itemKey}
+        className={classnames('form-check-label', {
+          [styles.enabled]: !disabled,
+        })}
+      >
+        {label}
+        {showCount && typeof count !== 'undefined' && (
+          <>
+            &nbsp;
```
ZhangYiJiang: We have an NBSP constant defined at https://github.com/nusmodifications/nusmods/blob/master/website/src/utils/react.tsx#L5 which might be more readable

website/src/views/components/filters/Checklist.tsx
```diff
@@ -1,45 +1,33 @@
...
-              className="form-check-input"
-              type="checkbox"
-              checked={filter.enabled}
-              onChange={() => onChange(filter)}
-            />
+  const itemComponents = displayItems.map(({ key, doc_count: count, selected }) => (
+    <CheckboxItem
+      key={key}
+      active={selected}
+      count={count || -1}
```
ZhangYiJiang: If this is zero it becomes `-1`?

website/src/views/modules/ModuleFinderSidebar.tsx
```diff
@@ -0,0 +1,131 @@
...
+        </header>
+
+        <RefinementListFilter
+          id="sem"
+          title="Offered In"
+          field="semesterData.semester"
+          operator="OR"
+          orderKey="_term"
+          orderDirection="asc"
+          bucketsTransform={(a: RefinementItem[]) => {
```
ZhangYiJiang: Maybe use a better var name than `a`

website/src/views/modules/ModuleFinderSidebar.tsx
```diff
@@ -0,0 +1,131 @@
...
+          containerComponent={FilterContainer}
+          itemComponent={CheckboxItem}
+        />
+
+        <NumericRefinementListFilter
+          id="mcs"
+          title="MCs"
+          field="moduleCredit"
+          multiselect
+          options={[
```
ZhangYiJiang: Since this is a constant this should be extracted as a constant variable to prevent this from causing re-renders

website/src/views/components/SearchBox.tsx
taneliang: Yup I made `SearchBox` a controlled component so that Searchkit can update the search box's contents when the query is changed (e.g. by navigating back/forward in the page). I've put some steps to demonstrate this in the PR descriptions

website/src/views/modules/ModuleFinderSidebar.scss
```diff
@@ -1,7 +1,9 @@
...
+@import "~styles/utils/modules-entry";
 
 .moduleFilters {
   height: 100%;
+  user-select: none;
```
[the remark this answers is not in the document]
taneliang: I just thought it didn't make sense for the sidebar header and checkbox text to be selectable. Any thoughts? This is a screenshot of what it looked like:

image

taneliang:
@li-kai it's not ready to be merged just yet. I'll fix the review comments by tomorrow and we can merge it, then fix the long list of issues in future PRs. I definitely intend to add Docker, but that'll likely depend on your work on the dockerize branch. Will that be usable in prod?

@ZhangYiJiang I'm not too sure how the text search works, but I think what's happening is that there are some filters selected, and there are no mods with "joke" in their modcode/title/description. Because there are no results, the best it can do is suggest that "jokes" be searched for instead, because "jokes" is present in a mod's title.

website/package.json
```diff
@@ -165,11 +165,11 @@
     "react-redux": "7.0.3",
     "react-router-dom": "5.0.0",
     "react-scrollspy": "3.4.0",
-    "react-waypoint": "9.0.2",
     "redux": "4.0.1",
     "redux-persist": "5.10.0",
     "redux-thunk": "2.3.0",
-    "reselect": "4.0.0"
+    "reselect": "4.0.0",
+    "searchkit": "2.3.1-alpha.9"
```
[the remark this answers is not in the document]
taneliang: Yeah. But even if it isn't, we don't really have much of a choice as we don't have enough time to implement it ourselves, and the only other alternative, ReactiveSearch, is too buggy to use. And Algolia is also too expensive and will likely be a problem in the long run.

website/src/views/components/SearchkitSearchBox.tsx
taneliang: `accessor` is only set when the component is mounted (https://github.com/searchkit/searchkit/blob/016c899c97f72ea3ad5afc017345e41c9003172a/packages/searchkit/src/core/react/SearchkitComponent.ts#L81), so we can't initialize `state.input` to `this.getAccessorValue` in the constructor.

It'll be possible to implement `componentDidMount` and set value there, but we'll need to be careful that any changes to `SearchkitComponent`'s implementation doesn't break it. Doing this will also cause our implementation to diverge further from Searchkit's [`SearchBox`](https://github.com/searchkit/searchkit/blob/016c899c97f72ea3ad5afc017345e41c9003172a/packages/searchkit/src/components/search/search-box/SearchBox.tsx) implementation, which will make it harder to adapt future changes into our component.

I'll implement the `componentDidMount` override and associated changes for now, as the code will be cleaner without a possible `undefined` hanging around.

taneliang: Oh okay I realized that the original implementation sets `input` to undefined when the text box isn't focused. This allows browser back/forward navigation to change the value in the text box as `getValue` will get the accessor's value. I'll follow this instead.

website/src/views/modules/ModuleFinderSidebar.scss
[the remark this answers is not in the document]
ZhangYiJiang: Yeah, I think that makes sense

taneliang:
Merging. Bugs introduced in this PR will be tracked at https://github.com/nusmodifications/nusmods/issues/1775#issuecomment-502332444

PR #1855, opened 2019-06-16
Type all module finder related components with React.FC

taneliang:


taneliang:
@ZhangYiJiang let's see if we can codemod that. I'll look into it next month

ZhangYiJiang:
There's no need to codemod. A simple find and replace will do.

PR #1856, opened 2019-06-16
Implement pager component on module search page

taneliang:
## Context

https://github.com/nusmodifications/nusmods/issues/1775#issuecomment-502332444

The current pagination component uses Bootstrap's default styles, which do not work well in dark mode, and Searchkit also provides us too many buttons, causing the pagination component to be too wide on tablets and phones.

@jiholim has redesigned the pagination component, and this PR implements it.

### Desktop mockup

image

### Mobile mockup

image

## Implementation

I've implemented a replacement for [Searchkit's `Pagination` component](https://github.com/searchkit/searchkit/blob/staging/packages/searchkit/src/components/search/pagination/src/Pagination.tsx) as Searchkit's implementation didn't give us the total number of pages.

#### First page

localhost_8080_modules_p=1067 (3)

localhost_8080_modules_p=1067 (5)

#### Middle page

localhost_8080_modules_p=1067 (1)

#### Last page

localhost_8080_modules_p=1067 (2)

#### Hover over no. 3

image

#### First page

localhost_8080_modules_p=1067 (6)

#### Middle page

localhost_8080_modules_p=1067 (7)

#### Last page

localhost_8080_modules_p=1067 (8)

### Phone (320px)

| First page | Middle page | Last page |
|-|-|-|
|localhost_8080_modules_p=1067(iPhone 5_SE) (2)|localhost_8080_modules_p=1067(iPhone 5_SE) (1)|localhost_8080_modules_p=1067 (9)|

#### Light mode
localhost_8080_modules_p=1067(iPhone 5_SE) (3)

## Design issues

- I've hidden the chevrons when they are disabled, instead of graying them out, as the buttons are gray by default and the disabling it with a lighter gray is too difficult to distinguish. @jiholim what do you think?
- Text wraps on 320px phones. I don't think it's worth fixing.
- Orange text looks like a button - could be misleading


website/src/views/components/ModuleFinderPager.tsx
```diff
@@ -1,33 +1,114 @@
...
+          <span className="sr-only">First</span>
+        </ModuleFinderPagerButton>
+        <ModuleFinderPagerButton
+          disabled={selectedPage === FIRST_PAGE_INDEX}
+          onClick={onGoToPrevious}
+        >
+          <span aria-hidden="true">
+            <ChevronLeft className={styles.svg} />
+          </span>
+          <span className="sr-only">Previous</span>
```
ZhangYiJiang: Making this a tooltip or `title` would be more helpful I think

website/src/views/components/ModuleFinderPagerButton.scss
```diff
@@ -0,0 +1,17 @@
+@import '~styles/utils/modules-entry';
+
+.pagerButton {
+  min-width: 2.3em;
+  border-width: 0;
+  font-weight: bold;
+  color: inherit;
+
+  &.disabled {
+    opacity: 0;
```
ZhangYiJiang: Use `visibility: invisible` if you want something to still take up space but not be seen. `opacity: 0` elements can still be interacted with (although in this case you've also added the disabled property so that's alright). Also you can use the `:disabled` pseudo class to save having to add a class

website/src/views/components/ModuleFinderPagerButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const ModuleFinderPagerButton: React.FC<Props> = ({ disabled, active, onClick, children }) => (
+  <li>
+    <button
+      type="button"
+      className={classnames(
+        'btn',
+        styles.pagerButton,
+        disabled && styles.disabled,
+        active && styles.active,
```
ZhangYiJiang: I feel that an object with `[styles.active]: active` is clearer.

website/src/views/components/ModuleFinderPager.tsx
taneliang: Good idea. I've used tooltips that appear on touch hold or on a delay.

image

website/src/views/components/ModuleFinderPagerButton.scss
taneliang: Cool tricks that I didn't know about. Thanks!

taneliang:
Updated screenshot with updated mobile text and tooltip. Merging first so that we can progress on other tasks; @jiholim can do a final design review if necessary once deployed on staging.

image

PR #1857, opened 2019-06-16
Move module search page loading indicator to search box

taneliang:
## Context

https://github.com/nusmodifications/nusmods/issues/1775#issuecomment-502332444

Search loading indicator is distracting and is not visible when scrolled down
image

This PR moves the search indicator from the stats component into the search box. When a search is loading, the search icon will be replaced by a loading indicator.

The loading indicator doesn't smoothly transition between itself and the search icon, so it's still pretty ugly. @jiholim if you could provide any suggestions that'll be great!

I've also tried replacing the input field clear button with the spinner, but:
1. The code will be pretty ugly as the clear button is added by the browser and has to be hidden when the spinner is shown.
2. Also, the clear button is currently touching the right edge of the screen on `xs`. We cannot fix that by adding right padding, because that'll truncate the placeholder text. This means that if we want to replace the clear button with the spinner on `xs`, it either needs to also touch the edge or it'll jump to a more reasonable location, which is also ugly.

### >= `lg`

localhost_8080_modules

### `sm`
localhost_8080_modules (4)

### `xs`

localhost_8080_modules (3)


li-kai:
I'd leave a bit more padding to the right of the search icon. Right now it's a little too close. Perhaps @jiholim can give a better opinion.

taneliang:
@jiholim when you notice the uglyness of this page pls forgive us :pray:

jiholim:
Oh psps was busy with Mobbin stuff 😥. Usually the loading indicator appears at the top of search result content area (not in the searchfield) but I think our implementation looks fresh haha - agree with @li-kai that more padding to right would look better.

PR #1861, opened 2019-06-18
Update browserlist

li-kai:
## Context
Run the scraper for new browserlist support

## Implementation

Run `yarn start` in packages/browserslist-config-nusmods
Find '9.1' and fix config where applicable.
1. Remove bowser checks with `String.prototype.padStart` which is a fairly new ES6 method which we no longer polyfill.
1. Remove body's `height: auto` fix for iOS 9. Might break older iOS versions.
1. Enable arrow functions minification. Might break older browsers.

## Other Information

Good riddance Safari 9.1.


li-kai:
Closing and reopening to trigger packtracker.

PR #1893, opened 2019-07-03
Update NUSMods for AY19/20

taneliang:


MAINTENANCE.md
```diff
@@ -4,8 +4,15 @@ Create a new issue on GitHub with this checklist after the finals every semester
...
 
-- [ ] Update academic year in `app-config.json`
-- [ ] Update with next year's holiday data and delete old data
+- [ ] On server: update academic year in `scrapers/nus-v2/src/config.ts`
+- [ ] Ensure that scraper is scraping next AY's data
+- [ ] Update academic year and semester in `app-config.json`, and add current academic year to archive years
+- [ ] Update with next year's holiday data from https://github.com/rjchow/singapore_public_holidays and delete old data, then run `scripts/holidays-csv-to-json.js`
+- [ ] Add announcement to website
+- [ ] Update academic year in `api/search/src/index.ts`
+- [ ] On server: Update academic year in export/.env
```
ZhangYiJiang: /nit Can remove some of the older semester stuff

export/.env.example
```diff
@@ -1,5 +1,5 @@
 # For explanation about these values, refer to config.js
-ACADEMIC_YEAR=2017-2018
+ACADEMIC_YEAR=2018-2019
```
ZhangYiJiang: Remember to update this file on the server too

scrapers/nus-v2/src/tasks/GetSemesterExams.ts
```diff
@@ -78,7 +78,11 @@ export default class GetSemesterExams extends BaseTask implements Task<void, Out
         this.logger,
       );
     } catch (e) {
-      throw new TaskError('Cannot get exam data', this, e);
+      this.logger.error(
```
ZhangYiJiang: This will always be logged by Sentry. If this is *expected*, then we shouldn’t log at the error level

website/src/actions/__snapshots__/moduleBank.test.ts.snap
```diff
@@ -26,6 +26,18 @@ Array [
       "type": "FETCH_ARCHIVE_MODULE",
     },
   ],
+  Array [
```
ZhangYiJiang: Should fix this later, I thought we had mocked config, hmmm

scrapers/nus-v2/src/tasks/GetSemesterExams.ts
taneliang: I think it's a gray area though. Although we're expecting the API to fail now, we also want to know if it's failing during normal times? I'm not sure when they're going to finalize and release that data.

export/.env.example
taneliang: Yep, it's in the maintenance checklist

PR #1915, opened 2019-07-07
Fix 16 lint warnings

taneliang:
Down from 44 lint warnings to 28! More will fix themselves once the new search is unreverted -- a few lint warnings are in the nuked code.


website/src/views/components/filters/DropdownListFilters.tsx
```diff
@@ -53,7 +53,7 @@ export class DropdownListFiltersComponent extends React.PureComponent<Props, Sta
   };
 
   onOuterClick = () => {
-    this.setState({ inputValue: this.state.inputValue });
+    this.setState((prevState) => ({ inputValue: prevState.inputValue }));
```
ZhangYiJiang: This was added to preserve the input after search loses focus. Please check that this still works.

ZhangYiJiang: This doesn't preserve the previous text entered when focus is lost, unfortunately. There should be a proper way to do this, but last time I checked when reviewing the PR that added this this was the only way

website/src/views/today/TodayContainer/TodayContainer.tsx
```diff
@@ -130,22 +130,24 @@ export class TodayContainerComponent extends React.PureComponent<Props, State> {
...
       })
       .catch(captureException);
 
     weatherAPI
       .fourDay()
       .then((forecasts) => {
-        this.setState(
-          produce(this.state, (draft) => {
+        this.setState((prevState) => {
+          produce(prevState, (draft) => {
```
ZhangYiJiang: You should be able to remove the first parameter so that `produce` returns a function that updates prevState passed in from setState

website/src/views/components/filters/DropdownListFilters.tsx
taneliang: This function no longer exists in its current form on master (it's been reimplemented with hooks, so it doesn't have this issue)

website/src/views/today/TodayContainer/TodayContainer.tsx
taneliang: Cool trick!

PR #1937, opened 2019-07-15
Deploy using traefik

li-kai:
## Context

Uses traefik to discover endpoints automatically.

Since traefik finds paths automatically, we do not need to manually specify ports.
Adding deploy previews is easy. Just change www.nusmods.com to hash.nusmods.com

Also allows us to do blue-green deploys easily: https://coderbook.com/@marcus/how-to-do-zero-downtime-deployments-of-docker-containers/

## Implementation

- [x]  Use traefik instead of caddy for load balancing
- [x]  Rewriting api.nusmods.com
- [x]  Fix es-importer as project-name breaks it due to network for `elasticsearch:9200` no longer working
- [ ]  Figure a way out for fastcgi support


infra/machine/traefik.toml
```diff
@@ -0,0 +1,52 @@
...
+  compress = true
+    [entryPoints.http.redirect]
+    entryPoint = "https"
+
+  [entryPoints.https]
+  address = ":443"
+    [entryPoints.https.tls]
+
+  [entryPoints.traefik]
+  address = ":5678"
```
ZhangYiJiang: Port numbers are hard

infra/machine/traefik.toml
```diff
@@ -0,0 +1,52 @@
...
+
+# [retry]
```
ZhangYiJiang: Do we still need these?

prod-deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+docker-compose --project-name=green  -f docker-compose.prod.yml up -d
+
+# Build process takes quite a while here
+
+# Build the new blue and restart blue, bringing it to latest
+docker-compose --project-name=blue -f docker-compose.prod.yml build --no-cache
+docker-compose --project-name=blue -f docker-compose.prod.yml down --remove-orphans
+docker-compose --project-name=blue -f docker-compose.prod.yml up -d
+
+sleep 2m
```
ZhangYiJiang: Two minutes?

website/src/config/app-config.json
```diff
@@ -3,7 +3,7 @@
   "defaultDescription": "NUSMods is a timetable builder and knowledge platform, providing students with a better way to plan their school timetable and useful module-related information that are community-driven.",
   "academicYear": "2019/2020",
   "apiBaseUrl": "https://nusmods.com/api",
-  "elasticsearchBaseUrl": "/elasticsearch",
+  "elasticsearchBaseUrl": "https://382c4d616054428291bc86fdf4001a6e.ap-southeast-1.aws.found.io:9243",
```
ZhangYiJiang: Can we configure this to go through a domain we control so we can switch providers if we want to in the future?

infra/machine/traefik.toml
taneliang: @li-kai I'm removing these

prod-deploy.sh
taneliang: It's just an estimate of how long it'll take the services to start up and configure themselves. It used to take about a min or so because of ES, but now without ES it should be much faster. We can just leave it though.

website/src/config/app-config.json
taneliang: We're having some issues with the wildcard SSL cert so I don't think this is worth doing. Plus we also won't be able to stick the port number in DNS so we'd most likely still have to change this config if we switch providers. We could revisit this in the future though.

taneliang:
CI failing because Coveralls is down. Merging

PR #1973, opened 2019-07-19
Convert export service to TypeScript

ZhangYiJiang:
## Context

Convert Export service to TypeScript to improve its maintainability.


export/README.md
```diff
@@ -73,34 +73,26 @@ Returns the HTML content of the page that Puppeteer renders.
 
 ## Deployment
 
-In production the app runs behind forever in addition to nodemon, so the app will automatically restart when its files are changed or if the app crashes.
+In production the app runs behind pm2 in addition to nodemon, so the app will automatically restart when its files are changed or if the app crashes.
```
taneliang, line 76: Actually since we're running the export server in Docker in prod, we're not using PM2 anymore. Should we remove the PM2 stuff and overhaul this section? I can do this.

export/Dockerfile.prod
```diff
@@ -17,13 +17,15 @@ RUN mkdir -p /home/node/Downloads \
 
 WORKDIR /home/node/app/export
 
-COPY ./export/package.json ./export/yarn.lock ./
+COPY ./export/package.json ./export/yarn.lock ./export/tsconfig.json ./
 
-RUN ls -la1; yarn --no-cache --frozen-lockfile --production
+RUN ls -la1; yarn --no-cache --frozen-lockfile
```
taneliang, line 22: So that `tsc` exists for `yarn build`

export/README.md
taneliang: I've just added Docker documentation without removing this as I think it could still be valuable. I didn't test the PM2 setup though.

export/package.json
```diff
@@ -2,15 +2,19 @@
...
-    "dev": "nodemon --no-update-notifier -r dotenv/config ./src",
-    "devtools": "DEVTOOLS=1 nodemon -r dotenv/config ./src",
-    "deploy": "rsync -avu --delete-after . ../../nusmods-export && pm2 restart ecosystem.config.js"
+    "build": "tsc",
+    "nodemon": "nodemon --no-update-notifier -r dotenv/config ./build/index.js",
+    "watch": "tsc --watch",
+    "dev": "run-p nodemon watch",
+    "devtools": "cross-env DEVTOOLS=1 yarn dev",
+    "deploy": "rsync -avu --delete-after . ../../nusmods-export && pm2 restart ecosystem.config.js",
+    "lint": "prettier --write"
```
taneliang, line 17: We'll probably want to add ESLint or change this to `prettier --check` instead, but we can do that in another PR

export/src/app.ts
```diff
@@ -1,15 +1,17 @@
-const Koa = require('koa');
-const Router = require('koa-router');
-const views = require('koa-views');
-const Sentry = require('@sentry/node');
-const _ = require('lodash');
+import Koa from 'koa';
+import Router from 'koa-router';
+import views from 'koa-views';
+import * as Sentry from '@sentry/node';
```
taneliang, line 4: Namespace import needed to fix a startup crash in prod

taneliang (approved):
@ZhangYiJiang I think the TS-fication of the export server is pretty complete apart from ESLint.

PR #1977, opened 2019-07-19
Module issue deflection

ZhangYiJiang:
## Context

Deflect questions using a modal that tries to get students to email the faculties instead. This will hopefully reduce the support email load.

## Implementation

Capture


ZhangYiJiang:
It reads a bit better but the links to the contacts are now further away, so it's a wash IMO.

taneliang:
Yeah I'm least happy that the links are further away. I asked a non-technical friend and she says it makes sense and flows so I guess it's a win...?

PR #2016, opened 2019-07-28
Update service worker sentry version

li-kai:
## Context

Sentry version is out of date

## Implementation

Update sentry version


li-kai:
We could pass it through webpack and import the library but that's more complicated.

PR #2029, opened 2019-08-01
Enforce sentryDsn and academicYear environment variables

li-kai:
## Context

We set a wrong `ACADEMIC_YEAR` value during deployment because default values allowed it to continue.
Forcing `ACADEMIC_YEAR` to be manually set before proceeding would prevent such silent bugs in the future.

Additionally, fallback to api because we are using separate machines for now and there's no MODULE_DATA folder.

## Implementation

1. Check for `ACADEMIC_YEAR` and its form of `<year>-<year> `
1. Use `SENTRY_DSN` because public dsn and raven dsn are the same? (@ZhangYiJiang could you confirm this?)
1. Allow api as fallback instead of local folder.


export/src/data.js
```diff
@@ -5,17 +5,22 @@ const _ = require('lodash');
 const config = require('./config');
 
 async function fetchModule(moduleCode) {
-  let fileName = `${moduleCode}.json`;
+  const fileName = `${moduleCode}.json`;
 
+  let mod;
   if (config.moduleData) {
-    return fs.readJSON(path.join(config.moduleData, fileName));
+    mod = await fs.readJSON(path.join(config.moduleData, fileName)).catch(() => undefined);
```
ZhangYiJiang: Since this is async, a try catch block with an error check for ENOTFOUND would probably be more readable

li-kai: I've changed it, could you please take another look?

PR #2034, opened 2019-08-06
Allow ElasticSearch to be skipped on develop

ZhangYiJiang:
## Context

With the new ES persist code in #2020, it's now pretty difficult to run the code locally without spinning up an ES server. This PR skips the ES config check if the env is not production (ES is still used if it is available, but the code will now fall back to filesystem only if not)


li-kai:
@taneliang Do we want to add back local elasticsearch w/ docker that would make this less painful?

li-kai:
@taneliang Do we want to add back local elasticsearch w/ docker that would make this less painful?

ZhangYiJiang:
Regardless of that we should still make it easy (easier?) to run the scraper without ES

li-kai:
Could we use an environment variable like `SEND_ELASTICSEARCH` instead? By default it will be false so we can run it in dev if we want?

ZhangYiJiang:
Isn't the config the same as using an env variable?

li-kai:
Guess I wasn't looking at this very closely. You're right, it does. Could we highlight this in the README?

PR #2037, opened 2019-08-07
Fix ModReg Schedule

donjar:
As per http://www.nus.edu.sg/registrar/info/modreg/GD_ModRegSchedule.pdf and http://www.nus.edu.sg/registrar/info/modreg/UGCPE_ModRegSchedule.pdf


ZhangYiJiang:
Thanks. I should just tell the lazy ass ModReg people to fix this themselves

PR #2054, opened 2019-08-10
Fix browser warning

ZhangYiJiang:
## Context

Our browser warning doesn't work anymore on IE11 because presumably some incompatible syntax is no longer transpiled by Babel and that's triggering some error that means the entire file doesn't run. 

## Implementation

This PR 

- Moves all entry points into a folder called `/entry` 
- Adds a new entry point and build config called `browser-warning` which has its own Babel config that enables all transpilation options 
- Update production config to take in the path to the new bundle so it can be injected separately into the main HTML file 
- Remove CleanWebpackPlugin - it does nothing fs-extra.remove can't. This way we have better control over when the assets get deleted
- Updates some Webpack config
- Replace bowser with obsolete-web/detective for UA parsing / detection, since it looks like it can understand browserslist and is slightly lighter 

The only downside to this is that production build will now be slightly slower (~6s) and larger since we're bundling Bowser twice, plus the overhead of the new bundle. We mitigate that by using async and defer to load the browser warning bundle.


ZhangYiJiang:
Tested the following 

- On newer browsers outside of browserlist range, the browser warning is NOT shown 
- On non-Safari browsers on iOS, the warning is NOT shown 
- on IE 11, the warning displays despite everything else not working

PR #2055, opened 2019-08-11
Fix downloading of shared timetable

### LINKED ISSUE #2042: Unable to download shared timetable without importing
### Describe the bug

Attempting to download an shared modsn.us timetable without importing results in downloading the saved timetable

### To Reproduce

1. Click on any modsn.us timetable link
2. Click on Download, select any option
3. See that the downloaded timetable is the saved timetable and not the shared timetable

### Expected behavior

Expected to download the shared modsn.us timetable even though it is not imported.

### Additional Context
Can be 'bypassed' by importing the timetable but that replaces the local timetable, and there is no indication that it is a solution

jonchan51:
## Context

Fix #2042 

## Implementation

`ExportMenu` extracts the timetable from the store to send to the export api. I made it accept the timetable props from its parent, `TimetableActions` instead, so that it acts the same way as the share feature. Now, it uses this timetable prop instead of extracting it from the store.


ZhangYiJiang:
Thanks for the PR!

PR #2064, opened 2019-08-13
Update scraper config to scrape hourly

taneliang:
There have been many complaints of outdated data. We were previously
promised in a meeting that the data will only be updated at around 4am
everyday, but this does not appear to be true as ModReg gets data faster
than that. Let's just scrape hourly to ensure that we're up to date as
well.

#accept2ship


ZhangYiJiang:
I've switched to use pm2's built in restart instead of cron, which doesn't work very well (and I gave up on debugging). The restart mechanism is a bit more robust, although it doesn't give us a guaranteed time to restart the thingmajig

PR #2069, opened 2019-08-15
Exam date format

### LINKED ISSUE #2023: Spell out months in date strings
Our dates are currently formatted ambiguously. Although most of the world uses day-month-year, the US uses month-day-year, resulting in 2 possible interpretations for days <=12. We should spell out the months instead.

image

image

image

shimeilee:
## Context

Closes #2023. Minimise ambiguous dates in number format by spelling out the months of dates.

## Implementation

This PR 
- Changes the function `formatExamDate` in **modules.ts** to return a string in the format of dd-MMM-yyyy, which spells out the first three letters of the month. 
E.g. Instead of 05-04-2019, it would reflect 05-Apr-2019
- Test cases for **modules.test.ts** have been modified to check for dd-MMM-yyyy formats instead.

## Other Information
Have checked that `formatExamDate` does not affect other date methods. I did not amend `getExamDate` as this affects other methods which depends on it. 
However, I am not 100% sure that this will fix the exam date formats in NUSMods. Feel free to let me know if I am missing out on critical areas of the code.


website/src/utils/modules.ts
```diff
@@ -43,7 +43,7 @@ export function formatExamDate(examDate: string | null | undefined): string {
   if (!examDate) return 'No Exam';
 
   const localDate = toSingaporeTime(examDate);
-  return format(localDate, 'dd-MM-yyyy p');
+  return format(localDate, 'dd-MMM-yyyy p');
```
taneliang: Should remove the `-` surrounding the months also to match standard SG date formats (see https://www.ibm.com/support/knowledgecenter/en/SSS28S_8.1.0/XFDL/i_xfdl_r_formats_en_SG.html)

taneliang: Looks like the IBM docs I linked to specify SG's date presentation format as `dd-MMM-yyyy`. Weird but okay. Let's just use that then.

PR #2263, opened 2019-10-27
Add no exam filter

ZhangYiJiang:
## Context

Seems like we forgot to bring this over from the old non-ES search? Anyway this is implemented using a custom ES query because SearchKit doesn't implement `exists`

## Implementation

Currently this is just one checkbox filter with a custom query. I can see us implementing a special non-overlapping exam query in the future though.


ZhangYiJiang:
I just noticed the state doesn't sync properly from URL query param. Will fix.

PR #2269, opened 2019-10-31
Minor UI tweaks to better support iPhone X landscape notch

taneliang:
## Context

iPhone X was released years ago and NUSMods has not been updated for it, resulting in really ugly white bars in landscape (see screenshot). This cannot be allow.

### Before

White bars beside nav bar and footer.

image

image

### After

image

The backgrounds of the nav bar and footer now stretch to the edges of the screen.

image

Venues page. Bottom padding was added so that the UI content is safely away from iOS's bar.

image

## Tests

Changes tested on:
* iPhone 11 running iOS 13.1 IRL
* iPhone X running iOS 11 on Browserstack (footer's left padding is messed up) 
image
* iPhone XS Max running iOS 12 on Browserstack

Portrait mode was confirmed to remain unchanged on the above devices.

Also ensured that there were no observable differences on:
* Chrome on Linux locally
* iPhone 7 running iOS 10 on Browserstack (iOS 10 does not support `env`)
* Chrome on Android 9 Pie on a OnePlus 7 on Browserstack
* Chrome on Android 9 Pie on a Google Pixel 3 XL on Browserstack

Android browsers do not display content behind the notch, so they display content like notchless phones.

More testing should be done on Android browsers before merging.

## Other Information

Support matrix for `env()` can be found at https://caniuse.com/#feat=css-env-function.

### DIFF
--- a/website/src/index.html
+++ b/website/src/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width,user-scalable=no,initial-scale=1">
+    <meta name="viewport" content="width=device-width,user-scalable=no,initial-scale=1,viewport-fit=cover">
     <link href="${htmlWebpackPlugin.files.publicPath}apple-icon-180x180.png" rel="apple-touch-icon" sizes="180x180">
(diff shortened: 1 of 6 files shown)


ZhangYiJiang:
Hmm, would it have made more sense to have created a SASS mixin instead? Also a bit worried about code rot if the maintainer stops using a noticed iPhone, and mildly annoyed we have one more device to add to our testing matrix, but oh well

Issue #2139, opened 2019-09-03
Remove added module(s) from the drop down list


iamjackslayer:
### Is your feature request related to a problem? Please describe.

I'm always frustrated when I couldn't remove an added module directly from the drop down module list. Currently user has to close the drop down list and look for bin icon in the modules table. 

### Describe the solution you'd like
Allow user to directly remove an added module from the drop down list.

Simply tap on the module from the drop down list to remove it.

### Describe alternatives you've considered
Add trashcan icon in the each item(row) in the drop down list. 

Tap the trashcan icon in the drop down list to remove the corresponding module.

sciffany:
Will take this issue

ZhangYiJiang:
Fixed by #2271

PR #2271, opened 2019-10-31
Allow deletion of module from ModuleSelect

sciffany:
## Context
#2139

## Implementation
I changed `removeModule` in `TimetableContent.tsx` such that it accepts moduleCode type instead of ModuleWithColor.

With this, `removeModule` can be called with ease from both of its children, `ModulesSelect` and `TimetableModulesTable`

In ModulesSelect, when a module has been added, a Trash icon is added beside the `Added` badge. Clicking this calls `removeModule` and removes the module from the semester.

image

image

Edited interface
image

## Other Information


sciffany:
@taneliang , I've done some changes. Some of the tests don't pass. 
image

Are they my fault?

ZhangYiJiang:
Yeah, that design needs work. 

The ones you screenshotted are just warnings. Open the show more button and look for the actual errors, or just run `yarn lint:code` locally to see the errors

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -182,16 +182,15 @@ class TimetableContent extends React.Component<Props, State> {
...
 
-  removeModule = (module: ModuleWithColor) => {
+  removeModule = (moduleCode: ModuleCode) => {
     // Save the index of the module before removal so the tombstone can be inserted into
     // the correct position
-    const index = this.addedModules().findIndex(
-      ({ moduleCode }) => moduleCode === module.moduleCode,
-    );
-    this.props.removeModule(this.props.semester, module.moduleCode);
+    const index = this.addedModules().findIndex(({ moduleCode }) => moduleCode === moduleCode);
```
taneliang: This line is causing lint to fail 
image

ZhangYiJiang: By the way @sciffany you should try to enable ESLint integration or checks when developing, it'll flag these errors for you as you code. Feel free to talk to us on Telegram if you have problem setting them up

sciffany:
image

ZhangYiJiang:
Moved the negative margins to the container of the button instead of on the button itself, which seems to work better. Also it's generally a bad idea to not use units. The lint rule you're triggering is about not putting units on `0` specifically, not about removing units everywhere. 

Also while tabbing technically works, the tab order is not very intuitive (the first thing you select when you tab is the dropdown menu itself, and it doesn't matter which item you've selected), but fixing that is probably not worth the effort

website/src/views/timetable/ModulesSelect.tsx
```diff
@@ -155,22 +155,25 @@ export class ModulesSelectComponent extends React.Component<Props, State> {
...
-                        onClick={() => {
-                          this.props.onRemoveModule(module.moduleCode);
-                        }}
-                      >
-                        <Trash className={styles.actionIcon} />{' '}
-                      </button>
-                    </Tooltip>
-                    <span className="badge badge-info">Added</span>
-                  </div>
+                  <>
```
taneliang: Is this fragment necessary? It only contains one div

PR #2380, opened 2019-12-24
Scraper: Ignore lessons with no start/end time

taneliang:
Resolves #2379.

Modifies scraper to ignore lessons without start/end times. The affected lessons' module codes are logged to Sentry; see [example](https://sentry.io/organizations/nusmods/issues/1404161671/events/3c0a95501acf4d56a034ac1c09a13e78) (internal only).

This branch is currently checked out on our scraper prod server.

### TODO after merging

On scraper prod server:
- `git checkout master`
- `git pull`
- `git branch -D dammit-neverending-lesson`

PR #2384, opened 2019-12-25
Scraper: Move lesson validation logic to standard validations

taneliang:
Improve implementation of #2380 by moving validation logic to shared checks in validations.ts, to allow number of valid/invalid lessons to be counted correctly.

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterTimetable.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterTimetable.ts
@@ -99,7 +99,7 @@ export function mapLessonWeeks(dates: string[], semester: number, logger: Logger
   return weekRange;
 }
 
-export function mapTimetableLesson(lesson: TimetableLesson, logger: Logger): TempRawLesson | null {
+export function mapTimetableLesson(lesson: TimetableLesson, logger: Logger): TempRawLesson {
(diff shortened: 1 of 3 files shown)


scrapers/nus-v2/src/tasks/GetSemesterTimetable.ts
```diff
@@ -183,13 +183,12 @@ export default class GetSemesterTimetable extends BaseTask implements Task<Input
         }
 
         // Report serious error to Sentry
-        if (!lesson.start_time || !lesson.end_time) {
+        if (!lesson.start_time || !lesson.end_time || lesson.start_time === lesson.end_time) {
           const { start_time, end_time, module } = lesson;
           this.logger.error(
             { moduleCode: module, end_time, start_time },
             'Lesson has no start and/or end time',
```
taneliang: This error message is misleading now if start time === end time. I can fix

taneliang: But are we sure we want to ignore such lessons? I think it’ll be clearer to have a weird 0 length lesson appear just so that it’ll be clear to whoever has to fix the data that there actually is a problem with the data, not because NUSMods is out of sync.

PR #2397, opened 2019-12-31
Improve module error reporting form 

ZhangYiJiang:
## Context

We can't actually do anything about these module errors, so we taichi them back to the school automatically now 

## Implementation

Implemented in two parts. Since both depend on each other - the form on NUSMods, and a serverless script that connects to the modules@nusmods.com email account via SMTP that actually sends out the email - this PR is going in first, but do not deploy since the form won't work until the serverless script is completed. 

Both the script and the UI reads from `data/facultyEmails.json`. We won't send the email address from the form directly so as to not let users spam random people from the form. 

The form collects the user's name, email and message. We try to automatically match the module to an email. The form won't submit unless the user actually picks a faculty to send email to. The user's name and email are persisted to `localStorage` so they won't have to enter it again in the unlikely event they need to send multiple emails. 

Capture

After submit: 

Success


website/src/utils/modules.ts
```diff
@@ -118,3 +125,7 @@ export function getYearsBetween(minYear: string, maxYear: string): string[] {
   years.push(maxYear);
   return years;
 }
+
+export function isGraduateModule(module: { moduleCode: ModuleCode }): boolean {
+  return Boolean(/\w+(5|6)\d+/.test(module.moduleCode));
```
taneliang: Maybe put a condition instead? There are some 7k and 8k mods

website/src/views/modules/ReportError.tsx
```diff
@@ -29,50 +154,146 @@ const ReportError: React.FC<Props> = ({ moduleCode }) => {
...
+          NUSMods updates its information from the Registrar's Office every night. Please wait up to
+          48 hours for information to be updated before reporting any issues.
         </p>
-
         <p>
-          For most issues related to modules and timetables, it will be faster to email your faculty
-          and CC <a href="mailto:modules@nusmods.com">modules@nusmods.com</a>. Examples of questions
-          that will be better answered by the faculties are:
+          This form will send an email about this module to the faculty. If you think the issue is a
+          bug in NUSMods, please email <a href="mailto:modules@nusmods.com">bugs@nusmods.com</a>{' '}
```
taneliang: The href needs to be updated as well

website/src/utils/modules.ts
ZhangYiJiang: That's the point - those aren't graduate level mods

PR #2401, opened 2020-01-01
Fix Navtabs and SideMenu layout issues on mobile

### LINKED ISSUE #2398: FAB modal page layout bugs
### Describe the bug

FAB modal things aren’t displayed correctly (see screenshots)

### To Reproduce

1. Go to https://nusmods.com/modules/ACC1002/financial-accounting or https://nusmods.com/modules on a phone
2. Click on the FAB

### Screenshots

856DF446-9792-4DD9-9C57-E5C27F19CEE6
955FC0D8-06F5-4F23-B1B8-F7E55C0A1D1F

* Device: iPhone 11
* OS: iOS 13.3
* Browser: Safari

shimeilee:
## Context

Closes #2398 where navbar does not fill up the entire width for smaller screens. 

## Implementation

Screenshot 2020-01-01 at 3 58 37 PM

Setting right: 0, left: 0 forces the navbar to take up the entire width. 

## Other Information

Qn: Is there any benefits of doing `position: fixed` over `position: sticky` in this particular scenario? I was testing locally and saw that sticky (default for the navbar on other views) had the same behaviour.

- [ ] Works on build


taneliang:
Hmm why isn’t our deploy bot commenting? @zhangyijiang any idea?

We prefer to use fixed for better browser support. IE and some Chinese browsers still don’t support position sticky. To be clear, we don’t officially support them, but since there’s an easy alternative we can just use it so that it won’t be so broken on those browsers. https://caniuse.com/#search=Position%20sticky

ZhangYiJiang:
The bot seems to have difficulty with PRs coming in from forks. I can take a look.

ZhangYiJiang:
> We prefer to use fixed for better browser support.  

To be clear, we would only if both do the exact same thing which is the case here - you only need sticky if the element is not going to be fixed relative to the viewport or if the element needs an offset before it becomes sticky.

taneliang:
Thanks for rushing this out, and sorry for taking such a long time to review.

Nav bar looks good, but the panel on the module page shouldn't take up the whole screen. Could you quickly fix it? It's probably caused by the same PR

|This PR|What it should look like|
|-|-|
|image|image|

website/src/views/components/SideMenu.scss
```diff
@@ -34,24 +26,47 @@ $animation-duration: 0.3s;
...
+
+  display: flex;
+  pointer-events: none;
+}
+
+.sideMenu {
+  composes: scrollable from global;
+  max-height: calc(100vh - #{$offset-top});
+  overscroll-behavior: contain; // stylelint-disable-line property-no-unknown
+  pointer-events: all;
```
ZhangYiJiang, line 41: `all` is not a valid `pointer-events` value for HTML elements - you want `auto` https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

taneliang: Oh TIL, thanks!

website/src/views/components/SideMenu.scss
```diff
@@ -34,24 +26,47 @@ $animation-duration: 0.3s;
...
+  position: sticky;
+  top: $offset-top;
+
+  display: flex;
+  pointer-events: none;
+}
+
+.sideMenu {
+  composes: scrollable from global;
+  max-height: calc(100vh - #{$offset-top});
```
ZhangYiJiang, line 39: I assume you've tested this on iOS, but it is possible for 100vh to be larger than the viewport - https://caniuse.com/#feat=viewport-units (see known issues) due to the address bar resizing with scrolling. I think that should be okay in this case though

taneliang: Yeah our previous implementation had issues because 100vh is too tall, so the side menu's header often got hidden behind our search field.

With the current implementation, I don't have this issue on iPhone anymore. Not too sure anymore why it's fixed, but I suspect it has something to do with the container being a flexbox, which may limit the menu's height?

On iPad, the bottom of the side menu goes offscreen due to the address bar resizing, but it's not a big deal I think. I suspect the `bottom` may just be offscreen, if my flexbox hypothesis above is correct.

taneliang: But yes 100vh on iOS seems to be the height of the entire screen, edge to edge, past any persistent UI chrome like the telco, status bar, and Safari's address bar

ZhangYiJiang: That's... awful. It seems to completely miss the point of vh. But I guess we can't expect better from Safari

taneliang:
Turns out overscroll-behavior is supported by Stylelint. I've removed it.

Also fixed 2 more issues:

* Drop shadow from menu appears at the bottom of the screen (bug from master). The shadow is now removed when the menu is closed.
	image

* Menu appears behind toolbar while scrolling to dismiss header bar (bug caused by my work in this PR). The menu is now translated further down when closed.
	image
	image

I'll merge once CI passes

PR #2404, opened 2020-01-03
Update error reporting form

ZhangYiJiang:
- Fix regex so that it'll only check the first numeric digit of the module code 
- Fix submitting state in the form 
- Add a small amount of padding below the submit button 
- Update API link to the error reporting form 

Once this PR is merged, we can flip the switch on and deploy


taneliang:
Oh no, Volatile.wtf is no more 😢 

image

I guess one possible replacement is to store a file in our API, like https://api.nusmods.com/killswitches/error_reporting_emails. We can define a successful request for that file as an activated killswitch

ZhangYiJiang:
:ohno: yeah let's just put a file there

PR #2427, opened 2020-01-14
Change modgrp to classNo transform to remove only one activity

### LINKED ISSUE #2426: Scraper trims too many characters off a class's modgrp
### Describe the bug

If a class timetable has a class with `activity` "T" and `modgrp` "TTE1", the scraper trims T multiple times from the start of the `modgrp` string to produce a `classNo` of "E1" in the NUSMods API instead of the correct "TE1".

### To Reproduce

Query the NUS class timetable API for module GEQ1000H in term 1920. Core team members only; you need the API details.

https://latest.nusmods.com/modules/GEQ1000H/asking-questions

### Additional context

This line is the cause: https://github.com/nusmodifications/nusmods/blob/5b5034dbce8a227f611bc33e634a813f24e8a9b0/scrapers/nus-v2/src/tasks/GetSemesterTimetable.ts#L115

Reported by Ching Mun (NUS); refer to internal emails.

taneliang:
Fixes #2426. Tests written and passing, but a scrape hasn't been run (but I'm running one now). How the data is affected also needs to be checked; I'll try to do a git diff on the scraped data to find any differences.

Please review the code first and I'll merge once I've tested it.

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterTimetable.test.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterTimetable.test.ts
@@ -5,9 +5,21 @@ import CN4205ETimetable from './fixtures/api-timetable/CN4205E.json';
 import CS1010XTimetable from './fixtures/api-timetable/CS1010X.json';
 import RE5001Timetable from './fixtures/api-timetable/RE5001.json';
 
-import GetSemesterTimetable from './GetSemesterTimetable';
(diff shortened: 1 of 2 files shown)


taneliang:
@ZhangYiJiang added a second check to follow NUS's advice; using a ternary will be uglier now imo. It's ready to merge if the code looks good to you

Issue #2183, opened 2019-10-02
Module not found error message persists after searching for a module


taneliang:
### Describe the bug

Module not found error message persists even after searching for a module.

### To Reproduce

1. Go to https://nusmods.com/modules/abc. You should see a module ABC not found message.
2. Use NUSMods' global search to search for a module. 
![image](https://user-images.githubusercontent.com/12784593/66017886-3ea85480-e50f-11e9-8b03-176b7be44756.png)
3. Select a module.
4. The module not found error page still remains, even though it is updated with the selected mod's data. 
![image](https://user-images.githubusercontent.com/12784593/66017916-70b9b680-e50f-11e9-95be-27c15b804922.png)

### Expected behavior

The module data should be displayed instead of the error page.

![image](https://user-images.githubusercontent.com/12784593/66017923-7dd6a580-e50f-11e9-86da-c0dbbcaa154a.png)

RisingGeek:
I would like to take up this issue.

taneliang:
Thanks @RisingGeek! Feel free to reach out if you have any issues and we'll be happy to help.

PR #2450, opened 2020-01-19
Fix persisting module not found error

jonchan51:
## Context

Fixes https://github.com/nusmodifications/nusmods/issues/2183

## Implementation

Unsure whether this is a proper way to do it, if it isn't do give feedback or close the PR. I realised the error persists because when we retrieve a module through the search bar, it simply reloads the `moduleCode` and calls `fetchModule()` to update the `ModulePageContainer`. However, as `fetchModule()` only adds `ModulePageContent` to the state, the `error` key in the state remains when it loads a new module. As the error is still present, `render()` loads the `ModuleNotFoundPage` component instead.

Hence, I saw to chain a `.then()` to clear the error from state on a successful `fetchModule()`


ZhangYiJiang:
Good attempt for your first PR, but the PR is not quite right. You are right that the issue is that the `error` state is never reset, but that should actually happen before the fetching starts, otherwise on slow connections the user will see briefly a flash of the 404 error for the _new_ module since the error is still present while the new module is being loaded. 

Module info is actually retrieved from Redux, and `fetchModule()` is actually dispatching a Redux action that fetches the module info from the API. This is not terribly obvious, but that's unfortunately how Redux is. 

Also a good practice is to always create a new branch, even on forks, since on GitHub you usually have the option to let maintainers push new commits to the branch checked, you want a new branch for your PRs, otherwise my changes here will also affect your master.

PR #2538, opened 2020-03-03
Update CI to use Node 12

taneliang:
Our Docker setup already uses Node 12 LTS. Once this PR is merged, I'll upgrade Node on the 1. server from Node 10 to Node 12.


scrapers/nus-v2/src/services/nus-api.test.ts
```diff
@@ -109,19 +109,23 @@ describe(NusApi, () => {
...
     expect(mockedAxios.post).toBeCalledTimes(2);
     expect(api.queue.getPendingLength()).toEqual(2);
+    expect(api.queue.getQueueLength()).toEqual(2);
 
     await p1;
     await p2;
 
+    // Expect remaining 2 requests to have started.
     expect(mockedAxios.post).toBeCalledTimes(4);
-    expect(api.queue.getPendingLength()).toEqual(0);
```
taneliang: @ZhangYiJiang any reason this should be 0? `getPendingLength` returns 2 in Node 12 but returns 0 in Node 10.

I think a pending length of 2 may make sense since p3 and p4 may have already been dequeued (and thus started) after the values are resolved. The relevant library code is at https://github.com/promise-queue/promise-queue/blob/fd3ea37d56def993da2b34dc1a87f66e92b9e808/lib/index.js#L153-L160.

Either way, I think the remaining 2 checks on lines 121-122 should be enough to ensure that all the requests fire at some point. Lines 113-115 now ensure that we hit the concurrency limit without exceeding it.

ZhangYiJiang: I think you're right, I might have mistaken getPendingLength for the number of buffered tasks

taneliang: Understandable, both `getPendingLength` and `getQueueLength` have extremely confusing names and comments. I had to read the source to figure out what they actually referred to

PR #2547, opened 2020-03-09
Fix Visual Bug for Lessons with Zero Duration

### LINKED ISSUE #2383: Lesson with 0 time causes lessons to its right to shift
### Describe the bug

If a lesson starts and ends at the same time, its timetable cell takes up a non-zero horizontal width and causes the lessons to its right to get displaced.

An example will make this clearer: Tutorials 4-7 in the screenshot below start and end on the hour (1pm, 2pm, etc.) but are shifted right.

B3A32600-BAB1-4A30-8127-277D8C740BD7

### To Reproduce

1. Go to https://nusmods.com/modules/ST2131/probability
2. If the data has not been fixed, you should see the above screenshot.

In dev, it may be possible to hardcode the lesson to repro easily without depending on data.

### Expected behavior

0 time lessons should somehow be visible but not shift other lessons.

### Additional context

This issue may be present for extremely short lessons as well.

This issue is of very low priority since lessons should obviously not start/end at the same time. We could potentially ignore such lessons in the scraper, but I am against this as it's confusing for a lesson to disappear entirely.

tanyonghe:
## Context
Closes #2383 where lessons with zero duration causes a visual bug.

## Implementation
There was an additional offset previously because lessons with zero duration had a minimum cell size allocated to it visually even though its width was set with a value of 0px (or below).
By setting a predefined cell size for lessons with zero duration, we can reliably account for the offset thus fixing the visual bug.
image

## Other Information
This is my first contribution to NUSMods so please let me know if I could have done something better! >.<


taneliang:
Hey, thanks so much for this PR! Sorry for the late review, our response time is pretty terrible.

Anyway, I think your approach is to hide the issue, but unfortunately it isn't completely hidden.

image

If you're still up for it, it may be better to set 0 padding, 0 width and 0 gap for lessons that have 0 duration. It's fine that it looks weird, since we don't expect any instantaneous lessons.

image

tanyonghe:
> 
> 
> Hey, thanks so much for this PR! Sorry for the late review, our response time is pretty terrible.
> 
> Anyway, I think your approach is to hide the issue, but unfortunately it isn't completely hidden.
> 
> image
> 
> If you're still up for it, it may be better to set 0 padding, 0 width and 0 gap for lessons that have 0 duration. It's fine that it looks weird, since we don't expect any instantaneous lessons.
> 
> image

I made the background transparent if the duration is zero so users can still use it like any other lesson slot (if we don't want this, we can do a filter prior to mapping the lessons to remove zero-duration lessons instead?).

website/src/views/timetable/TimetableCell.scss
```diff
@@ -30,6 +30,32 @@ $cell-border-radius: 6px;
...
 }
 
+.transparentCell {
+  // Flex to ensure cell content sticks to top, because browsers tend to
+  // vertical align middle button content
+  z-index: $timetable-cell-z-index;
+  display: flex;
+  flex-direction: column;
+  padding: $cell-padding;
+  border-width: 0 0 0 0;
```
li-kai, line 40: Border-width and background-color seems to be the only difference between `.cell` and `.transparentCell`. Could we move all of the common styles into a `.baseCell` class?

So something like
```css
.baseCell {
  z-index: $timetable-cell-z-index;
  display: flex;
  ...
}

.cell {
  border-width: 0 0 0.2rem;
  border-style: solid;
}

.transparentCell {
  border: 0;
  background-color: transparent;
}
```

tanyonghe: Thanks for the suggestion! I made a new `.baseCell` class to hold these common styles.

taneliang (approved):
Thank you!

I've merged master and rearranged some code.

Really sorry for taking more than half a year to get this PR in. We really appreciate the work :)

taneliang:
Closing and reopening to kick CI

PR #2560, opened 2020-03-22
Add prerequisite tree warning

li-kai:
## Context

Refer to the email from the school requesting that we at least inform students that the prereq tree may not be accurate.

## Implementation

This adds a warning box below the prereq tree.
image


li-kai:
I can't seem to come up with a design that makes alert look good...

PR #2629, opened 2020-05-09
Clean up website ESLint rules

taneliang:
Clean up `eslint-disable` and ESLint rules.

Depends on #2628.

## Implementation

* Removed as many `eslint-disable`s as I could.
* Made some file-level eslint-disables more specific by changing them to `eslint-disable(-next)?-line`
* In eslintrc, Un-disable rules that no longer cause any problems. Most of them just don't cause any more errors. Some of them were redundantly disabled, i.e. they were already disabled in our presets. I fixed the errors for 1 or 2.
* Add the awesome and sorely needed `@typescript-eslint/switch-exhaustiveness-check`, because turns out TypeScript doesn't check for switch case exhaustiveness.

### DIFF
--- a/website/src/utils/elasticSearch.ts
+++ b/website/src/utils/elasticSearch.ts
@@ -6,11 +6,11 @@ const POST_TAG = '</mark>';
 const PRE_TAG_REGEX = new RegExp(escapeRegExp(PRE_TAG), 'gi');
 const POST_TAG_REGEX = new RegExp(escapeRegExp(POST_TAG), 'gi');
 
-/* eslint-disable @typescript-eslint/camelcase, no-underscore-dangle */
-
 // For options, see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#highlighting-settings
 export const HIGHLIGHT_OPTIONS = {
+  // eslint-disable-next-line @typescript-eslint/camelcase
   pre_tags: [PRE_TAG],
+  // eslint-disable-next-line @typescript-eslint/camelcase
   post_tags: [POST_TAG],
   fields: {
     moduleCode: {},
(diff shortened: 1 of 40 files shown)


website/src/utils/elasticSearch.ts
```diff
@@ -6,11 +6,11 @@ const POST_TAG = '</mark>';
 const PRE_TAG_REGEX = new RegExp(escapeRegExp(PRE_TAG), 'gi');
 const POST_TAG_REGEX = new RegExp(escapeRegExp(POST_TAG), 'gi');
 
-/* eslint-disable @typescript-eslint/camelcase, no-underscore-dangle */
```
ZhangYiJiang, line 9: The reason this was a whole file disable is that ES uses these names as part of its API so we have no choice in using these names, and manually disabling it every where we interact with ES is noisy. But up to you - it seems we only use this in two places which is tolerable

website/src/views/tetris/board.ts
```diff
@@ -266,7 +267,9 @@ export function removeCompleteRows(board: Board) {
     // TODO: Optimize based on where the piece has landed
     let row = ROWS - 1;
     while (row >= 0) {
+      // eslint-disable-next-line no-loop-func
```
ZhangYiJiang, line 270: Is no-loop-func really useful?

website/src/utils/elasticSearch.ts
taneliang: Yep, I understand. I figured it'll still be best to disable it in places where we use it since it will prevent us from using it elsewhere unintentionally. But yeah if at some point it gets too noisy we can move it back to a file level disable.

website/src/views/tetris/board.ts
taneliang: Looks like it warns of a potentially subtle bug? I was a bit lazy to rewrite this code to get around the lint warning hahaha

taneliang:
Rebased all changes on master and force pushed, as this PR was built on top of #2628 that was just merged.

PR #2631, opened 2020-05-10
Fix and improve nusmoderator docs

taneliang:
## Context

As mentioned in #2630, it looks like nusmoderator's doc generation was broken by our TypeScript migration. documentation.js doesn't automatically pick up TS files.

* New docs: https://github.com/nusmodifications/nusmods/blob/fix-moderator-docs/packages/nusmoderator/README.md
* Old docs: https://github.com/nusmodifications/nusmods/blob/master/packages/nusmoderator/README.md

## Implementation

* Make documentation.js parse TS files.
* As documentation.js's lint command doesn't allow us to specify that TS files should be parsed, I've changed the lint script to regenerate docs and use `git diff` to ensure that there are no changes.
* Removed types from our JSDoc docstrings, so that docs will use our TS types.
* Added docstrings for our TS types so that they'll show up in the generated docs with links.

## Other Information

#accept2ship

### DIFF
--- a/packages/nusmoderator/README.md
+++ b/packages/nusmoderator/README.md
@@ -38,7 +38,7 @@ Assumes Week 0 begins on the first Monday of August.
 
 #### Parameters
 
--   `acadYear`  the academic year. E.g. "18/19"
+-   `acadYear` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** the academic year. E.g. "18/19"
 
 Returns **[Date](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Date)** Start date of the academic year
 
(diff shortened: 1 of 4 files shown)


packages/nusmoderator/src/types.ts
```diff
@@ -6,22 +9,89 @@ export type WeekType =
...
+
+  /**
+   * This `AcadWeek` is the `weekNumber`-th week with this academic week type
+   * in the semester. `null` if the semester has exactly 1 week of this type.
+   * Examples below:
+   *
+   * - Instructional week 1: `weekNumber` = 1
+   * - Instructional week 13: `weekNumber` = 13
+   * - Examination week 2: `weekNumber` = 2
+   * - Examination week 2: `weekNumber` = 2
```
ZhangYiJiang, line 49: This is the same as the line above - is this suppose to be vacation?

packages/nusmoderator/README.md
```diff
@@ -84,30 +97,124 @@ Expects a week number of a semester.
...
+
+#### type
+
+Type of this academic week.
+
+Type: ([WeekType](#weektype) | null)
+
+#### num
+
+-   **See: getAcadWeekInfo
```
ZhangYiJiang, line 202: This new line is inconvenient, although that seems like a bug in Documentation

packages/nusmoderator/src/types.ts
```diff
@@ -6,22 +9,89 @@ export type WeekType =
...
+
+  /**
+   * Type of this academic week.
+   */
   type: WeekType | null;
+
+  /**
+   * This `AcadWeek` is the `weekNumber`-th week with this academic week type
+   * in the semester. `null` if the semester has exactly 1 week of this type.
+   * Examples below:
```
ZhangYiJiang, line 80: A bit weird that this has more examples than the original docs on `AcadWeek.weekNumber`

ravern:

PR #2709, opened 2020-07-10
Replace ugly webkit search clear button with nicer button

chrisgzf:
## Context
In any webkit-based browsers (e.g. Chrome/Safari), there is a clear search button that is rendered for any `<input>` elements where `type="search"`.

This causes our modules search page to have an ugly blue cross on Chrome:
image

On Safari, there is a okay-ish button that doesn't quite fit our design-language:
image

On mobile chrome it looks hilarious:

Also, nothing appears on Firefox, so this behaviour is inconsistent across browsers.

This button is very annoying to me, so I decided to:
1. Remove webkit's search button
2. Implement our own clear search input button because I think it's a nice feature to have

## Implementation

1. Remove webkit's search button by adding CSS
2. Add our own clear search input button with react-feather's `X` icon

On desktop viewports:
Desktop1

On mobile viewports (320px):
Mobile2

Incidentally, this also fixes an issue where the placeholder text isn't fully displayed on 320px viewports:

## Other Information
🤪


website/src/views/components/SearchBox.scss
```diff
@@ -43,10 +61,15 @@
...
   .searchWrapper {
     input {
-      padding: 0.8rem 0 0.8rem 2.2rem;
+      padding: 0.8rem 1.6rem 0.8rem 2.2rem;
       border: 0;
       font-size: 1rem;
     }
+
+    input::placeholder {
+      // For placeholder text to fully show on 320px viewports
```
taneliang, line 70: Thanks for testing on 320px width viewports! 😍

PR #2735, opened 2020-07-11
Add matric number field and email validation to the form 

teikjun:
## Context

Add matric number form field and validation for email field

## Implementation

Added a new field, and used regex pattern (thanks li kai :D)

### DIFF
--- a/website/src/views/modules/ReportError.tsx
+++ b/website/src/views/modules/ReportError.tsx
@@ -169,7 +169,7 @@ const ReportError = React.memo<Props>(({ module }) => {
         Report errors
       </button>
 
-      <Modal isOpen={isOpen} onRequestClose={() => setIsOpen(false)} animate>
+      <Modal isOpen={isOpen} onRequestClose={() => setIsOpen(false)} shouldCloseOnOverlayClick={false} animate>
         <CloseButton onClick={() => setIsOpen(false)} />
         <h2 className={styles.heading}>Reporting an issue with {module.moduleCode}</h2>
         <p>
@@ -182,7 +182,7 @@ const ReportError = React.memo<Props>(({ module }) => {
           instead.
         </p>
 
-        {debug && <div className="alert alert-warning"><strong>Debug mode</strong> - submitting this form will send an email to modules@nusmods.com instead</div>}
+        {debug && <div className="alert alert-warning"><strong>Debug mode</strong> - this form will email modules@nusmods.com instead</div>}
 
         {formState.type === 'error' && (
           <div className="alert alert-danger" role="alert">
@@ -239,6 +239,7 @@ const FormContent: React.FC<FormContentProps> = ({
         <input
           className="form-control"
           id="report-error-name"
+          placeholder="Tan Ah Lee"
           value={formData.name}
           onChange={updateFormValue('name')}
           required
@@ -261,11 +262,13 @@ const FormContent: React.FC<FormContentProps> = ({
             </option>
           ))}
         </select>
+
         {selectedContact && (
           <p className="form-text text-muted">
             This will email <a href={`mailto:${selectedContact.email}`}>{selectedContact.email}</a>
           </p>
         )}
+
         <p className="form-text text-muted">
           If the department or faculty for this module cannot be found on this list, please refer to
           ModReg's contact list for{' '}
@@ -288,6 +291,7 @@ const FormContent: React.FC<FormContentProps> = ({
           className="form-control"
           pattern=".+@.+nus.+"
           value={formData.replyTo}
+          placeholder="e0012345@u.nus.edu"
           onChange={updateFormValue('replyTo')}
           required
         />
@@ -300,6 +304,7 @@ const FormContent: React.FC<FormContentProps> = ({
           className="form-control"
           value={formData.matricNumber}
           onChange={updateFormValue('matricNumber')}
+          placeholder="A01234567B"
           required
         />
       </div>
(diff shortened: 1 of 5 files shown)


website/src/views/modules/ReportError.tsx
```diff
@@ -268,17 +273,29 @@ const FormContent: React.FC<FormContentProps> = ({
...
           className="form-control"
+          pattern=".+@.+nus.+"
           value={formData.replyTo}
           onChange={updateFormValue('replyTo')}
           required
         />
       </div>
 
+      <div className="form-group col-sm-12">
+        <label htmlFor="report-error-email">Your Matriculation Number</label>
```
ZhangYiJiang, line 289: This `htmlFor` is incorrect, it should be linked to the ID

teikjun: ah thanks for catching it, I'll fix this now!

teikjun:
Form submission is now only enabled in production mode or debug mode of development.
To enable debug mode, you can go to the developer tool's console and run `window.SET_ERROR_REPORTING_DEBUG(true)`

website/src/views/modules/ReportError.tsx
```diff
@@ -268,12 +299,14 @@ const FormContent: React.FC<FormContentProps> = ({
...
       <div className="form-group col-sm-12">
-        <label htmlFor="report-error-email">Your Email</label>
+        <label htmlFor="report-error-email">Your School Email</label>
         <input
           type="email"
           id="report-error-email"
           className="form-control"
+          pattern=".+@.+nus.+"
           value={formData.replyTo}
+          placeholder="e0012345@u.nus.edu"
```
ZhangYiJiang, line 309: Help check this is the correct format for u.nus email

website/src/views/modules/ReportError.tsx
```diff
@@ -222,16 +235,32 @@ const FormContent: React.FC<FormContentProps> = ({
...
+        <label htmlFor="report-error-matric-number">Your Matriculation Number</label>
+        <input
+          id="report-error-matric-number"
+          className="form-control"
+          value={formData.matricNumber}
+          onChange={updateFormValue('matricNumber')}
+          placeholder="A1234567B"
+          minLength={9}
+          maxLength={9}
+          pattern="[aA][0-9]{7}[a-zA-Z]"
```
ZhangYiJiang, line 259: Check this is correct

website/src/views/modules/ReportError.tsx
```diff
@@ -106,6 +109,11 @@ const ReportError = React.memo<Props>(({ module }) => {
   const [isOpen, setIsOpen] = React.useState(false);
   const [formState, setFormState] = React.useState<FormState>({ type: 'unsubmitted' });
 
+  // Causes the error reporting function to email modules@nusmods.com instead.
+  // In production, use SET_ERROR_REPORTING_DEBUG(true) to enable debug mode
+  const [debug, setDebug] = React.useState(process.env.NODE_ENV !== 'production');
+  window.SET_ERROR_REPORTING_DEBUG = setDebug;
```
taneliang, line 115: Interesting approach! We should do this more often. We could probably migrate some of our existing debug switches from URL params to this

website/src/views/modules/ReportError.tsx
```diff
@@ -222,16 +235,32 @@ const FormContent: React.FC<FormContentProps> = ({
       }}
     >
       <div className="form-group col-sm-12">
-        <label htmlFor="report-error-name">Your Name</label>
+        <label htmlFor="report-error-name">Your Full Name</label>
         <input
           className="form-control"
           id="report-error-name"
+          placeholder="Tan Ah Lee"
```
taneliang, line 251: This is a little exclusionary as it assumes our users will have Chinese names, even though it's good that this name sounds gender neutral. I think this field should be clear enough that we can avoid any offence and controversy by just not putting a placeholder


```suggestion
```

website/src/views/modules/ReportError.tsx
```diff
@@ -158,7 +169,7 @@ const ReportError = React.memo<Props>(({ module }) => {
         Report errors
       </button>
 
-      <Modal isOpen={isOpen} onRequestClose={() => setIsOpen(false)} animate>
+      <Modal isOpen={isOpen} onRequestClose={() => setIsOpen(false)} shouldCloseOnOverlayClick={false} animate>
```
taneliang: Not caused by this PR exactly but we should wrap this callback in a `useCallback` so that we don't cause Modal to render unnecessarily (if it's optimised)

website/src/views/modules/ReportError.tsx
ZhangYiJiang: It's actually a bit annoying we need to do this, I had the impression the dev tool allowed live editing of hook state, but that seems to only work for class components.

website/src/views/modules/ReportError.tsx
ZhangYiJiang: Fair, but Modal is pretty simple and by itself doesn't seem worth optimizing.

taneliang: Yeah, was just about to say this after looking up useMemo again. Cool then

website/src/views/modules/ReportError.tsx
taneliang: It'll be less annoying if we made a `useGlobalDebugSwitch` hook hahah. But I think it may be a better solution than using URL params as iirc some of them interfere with other code that need URL params

ZhangYiJiang: Yeah, I think you're right there. The hook should be simple enough to build

taneliang:
Automerge seems to be dead. I'll just merge it manually

PR #2736, opened 2020-07-11
Show E-Learning venues as "E-Learning"

chrisgzf:
## Context
Before:
| Venues Page | Timetable Cell | Today Page |
| ------------- | -------------- | ------------ |
|image|image|image|

After:
| Venues Page | Timetable Cell | Today Page |
| ------------- | -------------- | ------------ |
|image|image|image|

`E-Learn_*` venues are hidden from the Venues page, but they are still accessible directly via permalink (e.g. if the user visits https://nusmods.com/venues/E-Learn_C directly), though that shouldn't be a big problem.


ZhangYiJiang:
Hmm, can this made a bit more generic? It seems to be duplicating this same logic between all places where we display venue codes

PR #2743, opened 2020-07-12
Fix polyfills working incorrectly on Edge

chrisgzf:
## Context
Users who use the non-Chromium Edge were reporting that they were seeing a white-screen of death when loading NUSMods, caused by:
image

Which is caused by:
image

`PersistGate` from one of our deps, `redux-persist` uses `Promise.prototype.finally` which somehow isn't supported on Edge.

But no, the story doesn't end here.

`Promise.prototype.finally` actually IS SUPPORTED by Edge. Except that there is some oddity with our polyfill library `core-js` or with Edge that causes it to [fail Promise feature detection](https://github.com/zloirock/core-js/issues/579#issuecomment-504325213), and Promise is entirely polyfilled incorrectly by `core-js` on Edge, causing the native implementation of `finally` to be gone.

To resolve this, we polyfill `finally`.

## Implementation

Tried 3 methods

1. Add `require.resolve('redux-persist/integration/react')` to the list of files transpiled by Babel. This should make a lot of sense, but somehow `core-js` has [issues with adding polyfills to deps](https://github.com/babel/babel/issues/9419). ❌
1. Change `useBuiltIns: usage` to `useBuiltIns: entry` in Babel's config, so that it adds *every single stable polyfill from `core-js` to every file*. Works great, but greatly increases the bundle size (+13 KB), see https://github.com/nusmodifications/nusmods/pull/2741. ✅
1. Just import the polyfill directly. Works great! Minimal bundle size increase! (In fact, it decreased??) ✅ ✅ ✅
Screenshot 2020-07-13 at 3 25 44 AM

Tested on Edge 44.17763.1.0 (which is from donkey years ago).

## Other Information
Congrats to the 70 people who use Edge! 🥴 🤢 🤮 You can now use NUSMods again!

image

### DIFF
--- a/website/src/entry/main.tsx
+++ b/website/src/entry/main.tsx
@@ -1,5 +1,10 @@
 // Import Sentry earliest to capture exceptions
 import 'bootstrapping/sentry';
+// core-js has issues with Promise feature detection on Edge, and hence
+// polyfills Promise incorrectly. Importing this polyfill directly resolves that.
+// This is necessary as PersistGate used in ./App uses `Promise.prototype.finally`.
+// See: https://github.com/zloirock/core-js/issues/579#issuecomment-504325213
+import 'core-js/es/promise/finally';
 
 import React from 'react';
 import ReactDOM from 'react-dom';


website/src/entry/App.tsx
```diff
@@ -7,6 +7,11 @@ import { hot } from 'react-hot-loader/root';
 import { BrowserRouter as Router } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
+import 'core-js/es/promise/finally';
```
ZhangYiJiang, line 10: Is this order correct? I think the polyfill should go as early as possible, since you are importing for side-effect here. Technically from the stacktrace it looks like the error would only manifest on mount, but for safety polyfills should be loaded as early as possible since the other modules can also have side effect / call the polyfilled methods during init

ZhangYiJiang:
Also good job sleuthing out this solution!

ZhangYiJiang:
Also I think Packtracker has the same issue as the other GitHub apps in that it has issue with PRs from forks, which is rather annoying :( Thanks for manually validating the size issue

website/src/entry/App.tsx
ZhangYiJiang: Also another nit - comments usually go above the line they are commenting on, not below

taneliang:
Wow great work @chrisgzf!

website/src/entry/App.tsx
chrisgzf: Fixed in e28803a!

ZhangYiJiang:
Sorry, one last issue - the polyfill added in the wrong file - App.tsx is not the entry point. It should be `main.ts`. You can put it right after where we import Sentry, here - https://github.com/nusmodifications/nusmods/blob/master/website/src/entry/main.tsx#L3

chrisgzf:
@ZhangYiJiang fixed in 09ca440!

PR #2783, opened 2020-07-25
Restrict node and yarn version

ZhangYiJiang:
## Context

This ensures all developers use the same node and yarn versions to ensure that builds and lock files generated are more stable 

## Implementation

- Restrict node version using the the `engines` field in `package.json`
- Ensure everyone use the same version of yarn using `yarn policies` command (https://classic.yarnpkg.com/en/docs/cli/policies) and symlink the `.yarn` and `.yarnrc` files from the root


taneliang:
Interesting! I didn't know you could do this.

Would this mean that we'll need to periodically make PRs to update the repo's copy of Yarn?

Unless I'm missing something, I'm not sure if that's worth it since I don't think the lockfile has caused any problems, and it doesn't seem like any problems it'll cause will be significant enough to justify the hassle of keeping Yarn updated in the repo.

li-kai:
We should redo this from scratch, or move to pnpm.

Issue #2748, opened 2020-07-13
Implement Covid zoning as indexable lesson / module metadata 


ZhangYiJiang:
### Is your feature request related to a problem? Please describe.

It should be possible to combine our crowdsourced venue latlon with NUS Covid zoning https://emergency.nus.edu.sg/downloads/NUS_Campus_Zonal_Map.pdf so that students can see which zone a lesson is in, and search by zones on the module page 

### Additional context

The scraper will need to load venue data to access the coordinates. We'll also need to manually draw out a polygon enclosing the different zones.

agnesnatasya:
Hii, can I clarify about this? The desired feature is to add a filtering component in ModuleFinder according to zoning, and it uses the data from `data/venues.json` to categorise them?  For the 'manual draw out of the polygon, does it mean that we need to hardcode the latitude and longitude borders of the different zones?

Thank you!!

ZhangYiJiang:
ModuleFinder uses ElasticSearch as the backend, so to implement this we will need to do it on the scraper instead of just the website frontend. The scraper will need to load data/venues.json and map of each venue there to their zone. We do need to hard code the borders of the different zones (preferably also in a data file so it can be updated without changing the code), but we only need the boundaries where they touch each other, since the other edges can be larger than the campus (we're not using the boundaries for visualizing the zones, only to categorize each venue). The BTC and Outram campus can just be covered by a large rectangle. 

The more interesting question is where the zone should be in the data. We can add the zone property to the `Venue`, but for ES to be able to filter modules, the `Module` object will need also need a zone property. It might also be useful to add it to `Lesson`, though that is duplicating some information, it might be worth not having to separately look up this information in a separate file.

davidchoo12:
I quickly drew up the zoning separation lines on Google MyMaps: https://www.google.com/maps/d/edit?mid=1YvYaNRFCLXS23eFux7gc4fihhxaab4Gm&usp=sharing
You can check the coordinates of the lines' points by exporting it as KML file.
Is this what you're looking for?

ianfromdover:
Can this issue be closed since the zoning system has now been decomissioned?

PR #2784, opened 2020-07-26
Indexed covid zones in scraper to enable faceted search on modules page 

ZhangYiJiang:
## Context

When we have Covid zoning info about each venue added to the lesson objects, we can use this information in search and for planning. This is 2/3 of the PRs needed for #2748 

## Implementation

Add `covidZone` to each individual lesson, and `covidZones` to `semesterData`. This allows `covidZones` to be indexed in ES, and for the timetable to display zone info 

ES persist was updated to update index mapping and settings instead of just creating it on first run. 

Also extracted an interface for `NusApi` for better mocking 

I don't have ES or scraping config set up locally, can someone else help test if this code works?


scrapers/nus-v2/src/utils/data.test.ts
```diff
@@ -228,10 +229,3 @@ describe(mergeDualCodedModules, () => {
     expect(aliases).toEqual({});
   });
 });
-
```
ZhangYiJiang, line 231: Dead code which somehow never got removed, lol

scrapers/nus-v2/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "outDir": "build",
-    "target": "es2020",
+    "target": "es2018",
```
ZhangYiJiang, line 4: This was a misconfiguration from a previous PR. The `target` is the output's language support level, which should not be `es2020` since Node 12 doesn't support optional chaining. The `lib` should be, because we want `Array#flat` and Node 12 does support that (should probably check if other APIs are also support - we don't use Babel for the scraper, and there might be ES2020 APIs not supported by Node 12 yet)

taneliang (approved):
Thank you!

scrapers/nus-v2/src/services/io/elastic.ts
```diff
@@ -99,6 +59,56 @@ async function createIndex(client: Client): Promise<Client> {
...
+          // An analyzer that produces a level string from a modcode, i.e.
+          // "CNS1010SX" => "1000", "CS2030" => "2000", etc.
+          level_analyzer: {
+            type: 'custom',
+            tokenizer: 'first_digit_tokenizer',
+            filter: ['first_token_limit_filter', 'thousandizer_filter'],
+          },
+        },
+        tokenizer: { first_digit_tokenizer },
+        filter: { first_token_limit_filter, thousandizer_filter },
```
taneliang, line 76: Actually hold on, this errors because they're static settings (putSettings can only update dynamic settings). https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-modules-settings

taneliang (changes requested):

taneliang:
I'll fix this. I'm also gonna upgrade us to Elasticsearch 7 since Searchkit has gained support for it.

taneliang (approved):
Made a few changes:

- Merged master, which now works with ES 7.
- Changed this PR to scrape to a new `modules_v2` ES index as we can't migrate the existing index. More details at [`6c9f598` (#2784)](https://github.com/nusmodifications/nusmods/pull/2784/commits/6c9f598c233baf3309b868679fcb8f9eb426f921).
- Updated website to use the new index.
- Fixed website queries to work with new nested `semesterData`.

@ZhangYiJiang could you take a last look at the changes?

### Test plan

Manually tested scraper:

1. Scraped to `modules` index using scraper on `master`.
1. Checked out this branch and ran `yarn dev all`. A new index was created and populated.

image

Manually checked the fixed exam filter:

| Master | This PR|
|-|-|
|image|image|

Not sure why there's 1 more mod without exam now, but the filter looks correct so I don't think anything broke.

### Deployment plan

We'll want to hold off on merging this until we upgrade to ES 7 so that we don't change too many things at once.

As discussed on Telegram:

1. Upgrade NUSMods to ES 7 with the current code on master.
1. Merge this PR.
1. Deploy and run new scraper.
1. Check Kibana to ensure new data is populated correctly on the production ES server.
1. Deploy new frontend.
1. Delete `modules` index on production ES server.

taneliang:
Also checked that `semesterData.covidZones` exists in the ES index.

image

ZhangYiJiang:
I think it looks good, but again you have more experience with ES than me, and I don't have it set up locally that I can test this

ZhangYiJiang:
Also we should follow up with the last PR to actually add a facet filter to the module page

taneliang:
Great! Let's merge and deploy the scraper now, and we can update the website tonight.

Issue #2040, opened 2019-08-07
Customised Timeslots for DYOM Mods


nlopez98:
### Is your feature request related to a problem? Please describe.

Having enrolled in a DYOM mod this sem, I wonder if it's a possible thing to input them into the timetable.

### Describe the solution you'd like

A template module info sheet with code (if available), lecture slots etc can be input and a slot appears on the timetable

### Describe alternatives you've considered

This may be difficult to implement since DYOM mods are not official, but if anyone has a better idea hopefully this could be pushed forward (when more DYOM mods are presented)

ZhangYiJiang:
Sorry, could you provide more information what DYOM modules are? I've never heard of them before.

nlopez98:
Design-your-own-modules which go through school approval but are not registered as official modules under the major code, but can be pre-allocated on modreg.

http://www.nus.edu.sg/registrar/education-at-nus/undergraduate-education/design-your-own-module.html

sciffany:
I would like to take this feature request

ZhangYiJiang:
@sciffany Sure! Since this is a relatively large change, could you sketch a rough outline about your approach before starting?

agnesnatasya:
Hi, may I ask if I could work on this feature request if it has not been completed yet?

Thank you!

sciffany:
Hello, sorry I was a little caught up and am unable to complete this feature request in the near future.

agnesnatasya:
Okay thank you! I will try to work on this

chrisgzf:
Hello, thanks @agnesnatasya for your work in #2785, and I was wondering if we could refine the scope/use-case of this feature more, as well as work on the UX.

Recently, there has been feature requests for hiding of slots from timetables: see #3334, however I find that hiding entire types of lessons (i.e. hiding all tuts/lec/rec) does not completely solve the problem in #3334, because certain modules (e.g. CS2103T) have multiple lessons of the same type that are taught by different tutors, and there are other issues with this approach which I will expand on in #3334.

In essence, I think that being able to add custom modules is a good compromise between both this issue and #3334:
- DYOM students can add custom modules
- TAs can add slots for their classes that they are teaching
- people on internship can block out their timetable for work (possibly weird use case yes, but they can if they want to)
- tldr: people can add whatever they want to their timetable

[...]

PR #2785, opened 2020-07-26
Custom modules

agnesnatasya:
## Context
#2040  
Feature request to add customised module to the timetable. 

## Implementation
I created a form that will require the user to input necessary information to create class of the customised module. The added class is supposed to appear in the timetable. 
demo feature

This is the responsive design for several screen widths

- Wide screen
Screenshot 2020-07-26 at 3 42 52 PM

- Below md
Screenshot 2020-07-26 at 3 43 03 PM

- Small screen
Screenshot 2020-07-26 at 3 43 15 PM

## Other Information
This is my first PR, I encountered a problem but not really sure how to solve it. After the submit button is clicked, it dispatches the action, and in the log, the difference of the state is correct, but the page is just loading and it just hang in there. In the log, there is no log of waiting or doing for something, so I am not sure what did I do wrong. 

If this problem has been rectified, I will modify the reducer, when it encounters the same module code and title, there are some checks that needs to be done.

Any help is appreciated, thank you!


website/src/views/timetable/CustomModuleTimetableForm.tsx
```diff
@@ -0,0 +1,380 @@
...
+          />
+          <Downshift onChange={this.onSelectLessonType}>
+            {this.renderDropdownLessonType}
+          </Downshift>
+          <input
+            className={classnames(styles.input)}
+            onChange={(e) => this.props.onChangeVenue(e)}
+            placeholder="Venue."
+          />
+          <Downshift onChange={this.onSelectDay}>{this.renderDropdownDay}</Downshift>
```
ZhangYiJiang, line 345: These dropdown menu should just be `<select>` elements - use Downshift only if you want comboboxes where there are too many options to select without searching. See `AvailabilitySearch` for an example of time selection. You might actually be able to reuse that component if you can make it more generic

website/src/views/timetable/CustomModuleTimetableForm.scss
```diff
@@ -0,0 +1,192 @@
+@import "./_variables.scss";
+@import '~styles/utils/modules-entry';
+
+$input-height: 2.75rem;
+$module-list-height: 13.5rem;
+$item-padding-vertical: 0.6rem;
+$btn-margin: 0.5rem;
+
+.formGroup {
```
ZhangYiJiang, line 9: Should use Bootstrap's form and layout system instead of trying to rebuild everything from scratch https://getbootstrap.com/docs/4.5/components/forms/

website/src/views/timetable/CustomModulesForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+      <div>
+        <div
+          className={styles.buttonGroup}
+          role="group"
+          aria-label="Custom module main information"
+        >
+          <input
+            className={classnames(styles.input, styles.titleIcon)}
+            onChange={(e) => this.setState({ moduleCode: e.target.value })}
+            placeholder="Module Code"
```
ZhangYiJiang, line 161: Avoid using the placeholder for label. Each form elements should have a proper label element associated with them (which can be hidden visually if needed)

website/src/views/timetable/CustomModulesForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+  onSelectDay = (item: string) => {
+    this.setState({ day: item });
+  };
+
+  onSelectLessonType = (item: string) => {
+    this.setState({ lessonType: item });
+  };
+
+  isFormValid = () => {
+    const validity = Boolean(
```
ZhangYiJiang, line 136: Try relying on HTML validation instead by adding `required` property to `input` elements

website/src/views/timetable/CustomModulesForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+  endTime: string;
+  weeks: Weeks;
+  venue: string;
+  day: string;
+  lessonType: string;
+  currentTimetableIndex: number;
+  timestamp: number;
+};
+
+class CustomModulesForm extends React.Component<Props, State> {
```
ZhangYiJiang, line 48: If you can, try creating functional components. At least make this a `PureComponent` for better performance

ZhangYiJiang (commented):
Good job on your first PR! 

The reason the page won't stop loading is because the final rendered timetable needs data from two places - the timetable config, which you have implemented, and the actual Module object loaded from the API and cached in module bank, which you have not. If the page finds a config but no module, it assumes the module is not loaded, and displays a spinner. 

To fix this, you need to store the custom module's data in the module bank too. You will also need to persist this. I suggest adding a new `customModules` key to the settings reducer (which is already persisted) and creating a selector to merge that with the module bank. 

Do also run the linter before pushing

agnesnatasya:
Hi @ZhangYiJiang thank you so much for your help and review. I have solved the loading problem by making `moduleBank` reducer process `ADD_CUSTOM_MODULE` action too. I have also implemented all of your review suggestions. 

The UI is responsive, here is the display in larger screen, you can look at the other screen widths in the deployment preview. 
Screenshot 2020-07-29 at 8 54 59 PM

This functionality is also available in the vertical mode.
Screenshot 2020-07-29 at 9 58 39 PM

Several cases that I considered:

- If the user adds a custom module with same code as one of their custom modules, and the title is different, a notification will be displayed.

- If the user adds a custom module with the same code and title as one of their custom modules, the classes willl be added to the Semester data

- All custom module code added will be appended with an apostrophe, to indicate that it is a custom module

Thank you!

-

ZhangYiJiang:
Using the module code field to indicate a module is custom is probably a bad idea. You're using a single field to store two pieces of unrelated information, and that's only going to cause issues in the future. A better way would be to add a `customModule?: boolean` property to the Module type and use that instead. 

The UI is going to take a bit of work. You probably need a way for the user to specify different lesson types. The horizontal mode UI is also a bit too wide - you can probably fit the lesson fields on one line, which would make it easier to specify additional lessons. You'll also need to consider UI for editing existing custom modules and removing them.

agnesnatasya:
Hi @ZhangYiJiang thank you so much for your review again, appreciate it. 
The reason that I add apostrophe, ', in the code of the custom module, is to not confuse the user when they input the same code as a pre-existing module. For example, if someone puts CS1010 in the custom module, the name will be CS1010', so that it is still displayed as a different module from the original CS1010 module from NUSMods, so someone can have both in the timetable. 

Also, it is to prevent custom module with the same code as a pre-existing module to overwrite the pre-existing module in the reducer. For example, in timetables reducer, if the moduleCode is the same, the old one will be overwritten, because it is a key-value pair in the dictionary.
```
    case ADD_MODULE:
    case ADD_CUSTOM_MODULE:
      return {
        ...state,
        [moduleCode]: action.payload.moduleLessonConfig,
      };
```

If I add a `customModule?` field in Module type, it still cannot handle this case because it is key-pair of dictionary. 

May I clarify what do you mean by 'You probably need a way for the user to specify different lesson types'? There has been a 'lesson type' field in the form, or do I understand you wrongly?

Also, for the part 'you can probably fit the lesson fields on one line, which would make it easier to specify additional lessons', do you want the UI for **horizontal mode on large screen or above**, to be like
```
Module Code, Module Title and Module Credit in first line
Lesson Type, Class No, Venue, Day, Start Time, End Time in second line
```
?

Thank you very much!

website/src/views/timetable/CustomModulesForm.tsx
```diff
@@ -0,0 +1,338 @@
...
+          ))}
+        </select>
+      </div>
+    );
+  };
+
+  renderInputClassNo = () => {
+    return (
+      <div className="form-group">
+        <label htmlFor="module-code">Class No.</label>
```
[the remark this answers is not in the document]
agnesnatasya: Okay, I will add '-custom-module' to the end of it to ensure uniqueness. For class number, I think it is still useful for custom modules, there might be several tutorial classes, for example.

ravern:

Issue #2770, opened 2020-07-23
Increase the contrast of navigation links


jeffsieu:
## Problem

The contrast of navigation bar links is poor, leading to readability issues, especially in bad lighting conditions.

### Current state

Poor text contrast of 2.15 on light mode and 2.74 on dark mode (Doesn't fulfill WCAG 2.0 Guidelines)

![image](https://user-images.githubusercontent.com/8487294/88276566-d894f400-cd11-11ea-8626-8b54b71e76a4.png)

![image](https://user-images.githubusercontent.com/8487294/88276123-1d6c5b00-cd11-11ea-9af6-d0750d85fe9c.png)

### Proposed solution

Increase the contrast of navigation bar links, as shown below.

![image](https://user-images.githubusercontent.com/8487294/88276932-712b7400-cd12-11ea-85e4-c5bc02a2852e.png)
![image](https://user-images.githubusercontent.com/8487294/88276345-776d2080-cd11-11ea-82c8-ac6a2ce22283.png)

taneliang:
Fixed by #2786

PR #2786, opened 2020-07-26
Change font color in navigation links to increase contrast

agnesnatasya:
## Context
#2770 

## Implementation
Change the font color of the navigation links to be darker to increase contrast, the color is now the same with the other 'darker' color of the page. The contrast has increased to 5.

This is the new display. It makes the navigation link clearer and more apparent, but at the same time, it makes the page a bit heavier. 
Screenshot 2020-07-26 at 5 02 30 PM


ZhangYiJiang:
@jiholim Thoughts on this? It might be more useful to change the color of `--gray-light` instead - the same color is used in other places and may also not meet contrast requirements

williamhutech:
I agree with @agnesnatasya that the navigation links are too light. How about we increase the contrast ratio from current 2:1 to 3.5:1? -- 5:1 is a bit too strong.

I recommend using color code #848490 for our nav links (screenshot below).

Firefox 2020-10-18 at 18 15 38@2x

taneliang:
Thanks @williamhutech! I think that looks better. @agnesnatasya do you want to change `--gray-light` to that?

Not sure if this will break other components that use `--gray-light`, but we can just check after it's changed.

agnesnatasya:
> Thanks @williamhutech! I think that looks better. @agnesnatasya do you want to change `--gray-light` to that?
> 
> Not sure if this will break other components that use `--gray-light`, but we can just check after it's changed.

Okay sure! I have added a new variable `--gray-mid` for this

website/src/views/layout/Navtabs.scss
```diff
@@ -59,7 +59,7 @@
   flex: 1 0 auto;
   line-height: 1.6;
   text-align: center;
-  color: var(--gray-light);
+  color: var(--gray-mid);
```
taneliang, line 62: Ah, this CSS variable doesn't exist as it needs to be defined in `website/src/styles/utils/css-variables.scss` (line 32). We'll also need a dark mode equivalent (see below). @williamhutech do you want to pick a `gray-mid` for dark mode? I'm actually quite curious how you're picking the colors 😆 We could also change up all these dark mode grays if you want.

https://github.com/nusmodifications/nusmods/blob/089eb585a3751886cb6a2b99a206c6788d31d101/website/src/styles/utils/css-variables.scss#L60-L64

taneliang (commented):
Looks great!

I agree with creating a new `--gray-mid` variable as some components (like form input placeholders) need to remain on `--gray-light`.

All the best for exams! Thanks for doing this, and apologies for taking forever to get this in haha

website/src/views/layout/Navtabs.scss
williamhutech: @taneliang Let's do #7D7D7D for dark mode's gray-mid! It will be dedicated to the navigation links for now. I will probably come back to change some of the dark mode grays later (after we sort out the MPE). 

In terms of how I pick colours, I use our existing color scheme + a contrast checking website (https://webaim.org/resources/contrastchecker/). And then I do Inspect Element on my browser, change the color code to preview how it'll actually look like, and voila, that's about it XD.

Also, I have a suggestion for dark mode, can you change the hover color from #eb1d00 to #aaa (--gray) for the navigation links? I think that way it looks 10x better AHAHA.

taneliang: Thanks @williamhutech! I've added #7d7d7d.

> can you change the hover color

Hahah let's discuss this one on the nav bar redesign PR (#3155). I think the issue with #eb1d00 is that it's not adapted to dark mode at all. I suspect it's computed from our primary brand color #ff5138 somewhere, but I haven't looked into it in detail.

taneliang (approved):
This is a great improvement. Thanks @agnesnatasya and @williamhutech for working on this, and @jeffsieu for suggesting this!

| | Before | After |
|-|-|-|
|Light|image|image|
|Dark|image|image|

PR #2791, opened 2020-07-26
Add  ignore all robots.txt for preview branches only

ZhangYiJiang:
## Context

We're getting reports of confused users ending up on latest.nusmods. This PR forbids crawlers from all CI generated builds using robots.txt, which should cause Google to delist all of those pages 

## Implementation

Added a static-ci folder which is copied only if the build script is running in CI. A bit awkward to create another root level folder, maybe we can reorganize our folder structure later


taneliang (approved):
Thanks! robots.txt exists at https://5f1dce47b5e2497108753a10--nusmods-deploy-preview.netlify.app/robots.txt

PR #2792, opened 2020-07-26
Improve edge case handling for module credits, exam duration and floor numbers 

ZhangYiJiang:
## Context

Module credits can be non-integer, who knew? 

## Implementation

- Use parseFloat for code related to handling module credits
- Improve rendering for examDuration by using nbsp between the duration and the unit to prevent awkward line breaks 
- Use minutes as the unit for exams shorter than 1h
- Handle non-integer floors in the improve venues form better (the browser should prevent non-integer floors already I think, but this handles situations where it doesn't)


taneliang (approved):
Interesting. Doesn't look like there are any 0.5 MC mods this sem anymore, but this is good.

image

Issue #2750, opened 2020-07-17
Request for Adjustable Timeline Feature


Jervoise:
### Is your feature request related to a problem? Please describe.

A clear and concise description of what the problem is. Eg. I'm always frustrated when [...]

agnesnatasya:
I would like to try to work on this feature request. Thank you!

Jervoise:
Thank you! Appreciate it!

Best,
Jervoise

PR #2812, opened 2020-08-10
Timetable start

agnesnatasya:
## Context
#2750 
There is a feature request to make the timetable have an adjustable starting time.

## Implementation
I added two buttons on each side of the start and end time of the timetable, where people can click and adjust their own starting and ending time of their timetable. 

demo feature timetable start

### DIFF
--- a/website/src/utils/timify.test.ts
+++ b/website/src/utils/timify.test.ts
@@ -38,79 +38,110 @@ describe('convertTimeToIndex', () => {
 
 describe('calculateBorderTimings()', () => {
   test('calculate default border timings, with no start and end index preference, correctly', () => {
-    const timings = calculateBorderTimings([
-      createGenericLesson('Anyday', '1100', '1230'),
-      createGenericLesson('Anyday', '1330', '1400'),
-      createGenericLesson('Anyday', '1300', '1500'),
-    ], convertTimeToIndex(DEFAULT_EARLIEST_TIME), convertTimeToIndex(DEFAULT_LATEST_TIME));
+    const timings = calculateBorderTimings(
+      [
+        createGenericLesson('Anyday', '1100', '1230'),
+        createGenericLesson('Anyday', '1330', '1400'),
+        createGenericLesson('Anyday', '1300', '1500'),
+      ],
+      convertTimeToIndex(DEFAULT_EARLIEST_TIME),
+      convertTimeToIndex(DEFAULT_LATEST_TIME),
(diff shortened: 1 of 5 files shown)


ZhangYiJiang:
This is an interesting implementation of the feature request! That said, I think there are a few issues 

- It won't work on mobile - the buttons are extremely small 
- It's not immediately obvious what the buttons would do, not intuitive when you press them 
- This breaks vertical mode 

I would suggest a simpler solution to the problem - a simple select dropdown on the settings page that let students choose when their classes start - Morning (default, current behavior), Afternoon (classes start at 12pm) or Evening (classes start at 5pm). 

This gets out of the way of most students who (I assume) don't need to change when their classes start, is mobile friendly, and is easy to understand

ZhangYiJiang:
Also, this needs to be a persisted preference - there's not much point in changing the start and end time only to find the timetable reset when you come back

agnesnatasya:
Hi @ZhangYiJiang thank you for the feedback. I tried to implement your suggestion but I think it might lead to unexpected behaviour. For example, if the user sets the starting time to 10AM, but the earliest class that the user has is 8AM, then the timetable will get cropped. 

Is it better if I have a settings option, "Trim Timetable", and there is two option, "Yes" or "No", the default choice is "No", but when the user selects "Yes", the timetable will be trimmed, to start on the earliest starting time and end on the latest ending time. 

May I know what do you think about this?

Thank you!

ZhangYiJiang:
In general we have a high bar for adding new settings. In this case I would strongly suggest just showing the lessons if they occur before the selected start time. I doubt users would understand what the trim option mean, and even if they do it's not a particularly meaningful choice - if they have lessons before their expected start time then they definitely want to see them. 

There are of course edge cases, eg. I have a module with lecture at 8am but there are recordings so I don't care about the lecture, but I think those are rare enough to not matter, and there are better ways to support those use cases anyway.

agnesnatasya:
Thank you @ZhangYiJiang for the reply. Okay noted. 

I would implement it such that the user might choose any hour between 6 AM and 9 PM as their timetable start time. May I ask, let's say the user selects start time to be 11AM, but one of their classes is 10AM - 12 PM. Would this class be displayed?

Thank you!

ravern:
It's been a while since this PR has seen activity. I'll close this for now, but definitely feel free to re-open this should you like us to review this / continue working on it.

PR #2823, opened 2020-08-16
Save e-learning venues as E-Learning in scraper

chrisgzf:
Follow up of https://github.com/nusmodifications/nusmods/pull/2736.

This PR changes the scraper to save lesson venues as "E-Learning" when the API returns a `room` of `E-Learn_*` for lessons.

To see current shaped of our scraped data: https://nusmods.com/api/v2/2020-2021/modules/CS1010.json

@ZhangYiJiang does this look right to you? Also would need some help to test this, if possible. 😅

Also, don't merge this yet. Still need to add the frontend changes.

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterTimetable.test.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterTimetable.test.ts
@@ -1613,4 +1613,115 @@ Object {
 }
 `);
   });
+
+  // Lessons with E-Learn_* in venue should have E-Learning as venue
+  test('should save venue of lessons with E-Learning venues correctly', async () => {
+    const task = createTask(
+      CS4238Timetable.map((lesson) => {
+        return { ...lesson, room: 'E-Learn_C' };
+      }),
+    );
+    const timetable = await task.run();
+
+    expect(timetable.CS4238).toMatchInlineSnapshot(`
+Array [
+  Object {
+    "classNo": "1",
+    "day": "Monday",
+    "endTime": "2030",
+    "lessonType": "Lecture",
+    "size": 40,
+    "startTime": "1830",
+    "venue": "E-Learning",
+    "weeks": Array [
+      1,
+      2,
+      3,
+      4,
+      5,
+      6,
+      7,
+      8,
+      9,
+      10,
+      11,
+      12,
+      13,
+    ],
+  },
+  Object {
+    "classNo": "2",
+    "day": "Tuesday",
+    "endTime": "2030",
+    "lessonType": "Lecture",
+    "size": 40,
+    "startTime": "1830",
+    "venue": "E-Learning",
+    "weeks": Array [
+      1,
+      2,
+      3,
+      4,
+      5,
+      6,
+      7,
+      8,
+      9,
+      10,
+      11,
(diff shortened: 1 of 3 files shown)


scrapers/nus-v2/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "outDir": "build",
-    "target": "es2020",
+    "target": "es2018",
+    "lib": ["dom", "es2020"],
```
chrisgzf, line 5: @ZhangYiJiang saw your comment about the `target` here https://github.com/nusmodifications/nusmods/pull/2784#discussion_r460561430, I also added `"dom"` to `lib` because the `URL()` calls were throwing errors in my tests: `Cannot find name 'URL'. Did you mean 'url'?` Not sure if you've encountered this too but adding `"dom"` fixed this for me.

ZhangYiJiang: That's probably not correct - this code will only be run in Node.js, and while `dom` lib will add URL as a global like in Node 10/12, it will also add a bunch of other globals like `document` and `navigator` which are not valid. The correct way to fix this is to use the `@types/node` typing, but it doesn't add globals unless it is imported at least once, so in the end it's probably best to just import `URL` explicitly `import { URL } from 'url';`

scrapers/nus-v2/src/tasks/GetSemesterTimetable.test.ts
```diff
@@ -1611,6 +1611,117 @@ Object {
...
 }
+`);
+  });
+
+  // Lessons with E-Learn_* in venue should have E-Learning as venue
+  test('should save venue of lessons with E-Learning venues correctly', async () => {
+    const task = createTask(
+      CS4238Timetable.map((lesson) => {
+        return { ...lesson, room: 'E-Learn_C' };
+      }),
```
ZhangYiJiang, line 1622:
```suggestion
      CS4238Timetable.map((lesson) => ({ ...lesson, room: 'E-Learn_C' }));
```

scrapers/nus-v2/src/tasks/GetSemesterTimetable.ts
```diff
@@ -111,7 +111,20 @@ export function transformModgrpToClassNo(modgrp: string, activity: string): stri
 }
 
 export function mapTimetableLesson(lesson: TimetableLesson, logger: Logger): TempRawLesson {
-  const { room, start_time, end_time, day, module, modgrp, activity, eventdate, csize } = lesson;
+  const {
+    room: providedRoom,
```
ZhangYiJiang, line 115: This avoids the `|| ''` and `?.` below 

```suggestion
    room: providedRoom = '',
```

taneliang (changes requested):
Oops, one semester has passed. As [zoning restrictions will be lifted on 6 Dec](https://emergency.nus.edu.sg/circulars/wp-content/uploads/2020/10/2020-1031-OSHE-COVID-19-Circular-28-08.14.pdf), I'm not sure how these venue names will change. Let's hold off work on this PR until we have more information.

ravern:

PR #2913, opened 2020-10-17
Add support for year-long modules to module planner

### LINKED ISSUE #1912: Improve support for year long modules in module planner
Similar to #1911, but a bit more difficult because this requires design help. I imagine a possible UI would be something like whole day events in gcal in week mode, where  the events are shown on top of the day's events. Supporting drag and drop would be somewhat difficult though, and if all four semesters are filled, then getting the UI right would be somewhat difficult.

jeffsieu:
## Context
Resolves #1912

### Before adding

image

### After adding

image

Year long modules are represented as modules in semester 0.

### Adding year-long modules

Year-long modules are moved to the year-long module list whenever added from Semester 1 or 2. The year-long semester is hidden unless it is not empty, or being dragged over.

### Dragging modules

- Semester-long modules can only be dragged between regular semesters, exemptions and plan-to-take semesters.
- Year-long modules can only be dragged between year-long semesters, exemptions and plan-to-take semesters.

Upon dragging, their type (year-long or semester-long) is stored as part of their `draggableId`, which is then used to update `PlannerContainer.state.draggedModuleType`. This prompts children `PlannerYear` components to display their year-long semester containers as drag targets.

### Misc
- "+ Add modules" section removed to avoid clutter. Also avoids the case where non year-long modules are added from the year-long "add modules" section
- Adds `type` attribute to `Droppable` targets to disallow dragging of modules to the lists of the wrong module type.
- Some test cases are modified to due to the extra "semester".


website/src/utils/planner.ts
```diff
@@ -3,11 +3,14 @@ import { ModuleCode, PrereqTree, Semester } from 'types/modules';
 import { PlannerModuleInfo } from 'types/planner';
 import config from 'config';
 import { assertNever, notNull } from 'types/utils';
+import { isYearLongModule } from 'utils/modules';
 
 // "Exemption" and "plan to take" modules are special columns used to hold modules
 // outside the normal planner. "Exemption" modules are coded as -1 year so
 // they can always be used to fulfill prereqs, while "plan to take" modules use
 // 3000 so they can never fulfill prereqs
+export const YEAR_LONG_SEMESTER: Semester = 0;
```
ZhangYiJiang, line 12: This is problematic - 0 comes before all the other semesters, so a year long module would fulfill prerequisites for other modules taken in the same year, when the opposite should be true. In general we might want to re-think this hack of using specially coded semester values - it might be better to type this instead 

``` ts 
type Semester = 1 | 2 | 3 | 4;
type PlannerSemester = Semester | 'yearLong' | 'exemption' | 'planToTake';
``` 

but this will have other impact with regards to sorting

website/src/views/planner/PlannerContainer/PlannerContainer.tsx
```diff
@@ -100,26 +109,39 @@ export class PlannerContainerComponent extends React.PureComponent<Props, State>
         moduleCodes.forEach((moduleCode) => {
           this.props.addModule(year, semester, { type: 'module', moduleCode });
           // TODO: Handle error
-          this.props.fetchModule(moduleCode);
+          this.props.fetchModule(moduleCode).then((fetchedModule) => {
+            if (isYearLongModule(fetchedModule)) {
+              this.props.moveYearLongModule(year, semester, moduleCode);
```
ZhangYiJiang, line 114: If I understand the UI correctly, the way to add a year long module is to add it to either semester, then this code will move it to the year long section. 

That's a bit awkward - it means for at least one render the module will be in the wrong section, and there is also no real guarantee that the module will be added to the correct location, eg. if the module failed to load, which would leave the data in an inconsistent state since the module is still in the wrong location in the planner. 

We might want to instead think of adding `yearLong` flag to ModuleCondensed - we need this info for use in module search anyway, so it makes sense. This does require a change to the scraper though.

ZhangYiJiang (changes requested):
This looks great! I do think we might want to change the data structure from the scraper to allow us to do this more naturally. Let me know if you need help with that

website/src/views/planner/PlannerYear.tsx
```diff
@@ -78,7 +81,26 @@ export default class PlannerYear extends React.PureComponent<Props, State> {
         })}
       >
         {this.renderHeader()}
-
+        <div
+          key={yearLongSemester}
+          className={classnames({
+            [styles.hideSemester]:
```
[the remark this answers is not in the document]
jeffsieu: I tried that at first, but that causes the element to disappear right as the module is dropped on it. Then it will throw an error as the drop target is no longer there to receive it.

website/src/utils/modules.ts
```diff
@@ -138,3 +138,7 @@ export function getYearsBetween(minYear: string, maxYear: string): string[] {
 export function isGraduateModule(module: { moduleCode: ModuleCode }): boolean {
   return Boolean(/[A-Z]+(5|6)\d{3}/i.test(module.moduleCode));
 }
+
+export function isYearLongModule(module: Module): boolean {
```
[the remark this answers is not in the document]
jeffsieu: I suppose that could be done, but I was thinking of hiding it under a layer of abstraction.

website/src/utils/planner.ts
```diff
@@ -17,13 +20,19 @@ export const PLAN_TO_TAKE_SEMESTER = -2;
 // We assume iBLOCs takes place in special term 1
 export const IBLOCS_SEMESTER = 3;
 
+export const SEMESTER_LONG = 'SEMESTER';
```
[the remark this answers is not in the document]
jeffsieu: This is used to set the type of PlannerModule as part of its id , so that the drop areas can only accept the currently dragged element if it should accept it (sem 1 shouldn't accept a year-long mod)

jeffsieu:
@ZhangYiJiang How should I proceed with this PR? I really wish to see these changes be incorporated.

scrapers/nus-v2/src/tasks/CollateModules.ts
```diff
@@ -118,10 +118,12 @@ const getModuleCondensed = ({
   moduleCode,
   title,
   semesterData,
+  attributes,
 }: ModuleWithoutTree): ModuleCondensed => ({
   moduleCode,
   title,
   semesters: semesterData.map((semester) => semester.semester),
+  yearLong: Boolean(attributes ? attributes.year : false),
```
ZhangYiJiang, line 126: `Boolean(undefined) === false` 

```suggestion
  yearLong: Boolean(attributes.year),
```

website/src/utils/planner.ts
ZhangYiJiang: Bump

website/src/views/planner/PlannerYear.tsx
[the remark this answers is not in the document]
ZhangYiJiang: Should add a comment then

website/src/views/planner/PlannerModuleSelect.tsx
```diff
@@ -121,7 +121,7 @@ export function PlannerModuleSelectComponent({
                     } else if (inputValue != null) {
                       // Otherwise we use the input value - this allows the user to
                       // enter multiple
-                      onSelect(inputValue);
+                      onSelect(modules.find((module) => module.moduleCode === inputValue) ?? null);
```
ZhangYiJiang, line 124: This would break existing functionality where this field allows multiple module codes to be entered at once here https://github.com/nusmodifications/nusmods/blob/3e844019534912609e38cb546fa3e2b57f7279f6/website/src/views/planner/AddModule.tsx#L133

The logic for validating module code should be on the parent component, not the child.

`onSelect` did have incorrect typing - `ModuleCode` should just be `string`, sorry about that

website/src/views/planner/PlannerSemester.tsx
```diff
@@ -24,6 +30,7 @@ type Props = Readonly<{
 
   showModuleMeta?: boolean;
   className?: string;
+  draggedModuleType: string | null;
```
ZhangYiJiang, line 33: This should be explicitly typed as a union of all valid values. 

``` ts 
type ModuleTypes = 'YEAR_LONG' | 'SEMESTER_LONG'
``` 

It would also make it easier to follow if these are constructed and deconstructed in a type safe manner using functions 

``` ts 
function getDraggableId(moduleCode: ModuleCode, moduleType: ModuleType)

// Alternatively this can also return a tuple
function fromDraggableId(id: string): { moduleCode: ModuleCode, moduleType: ModuleType }
``` 

Also it seems that draggable type can be either `SEMESTER_LONG` or `YEAR_LONG` whereas droppable types are `SEMESTER_LONG`, `YEAR_LONG` and `DEFAULT`, but it's hard to tell what DEFAULT should mean - I think from reading the code it means it accepts all module types, in which case it should probably just be ALL - a comment would be really helpful

ZhangYiJiang (commented):
Sorry for not coming back to review this. This looks almost ready - just a few things 

- The reducers and actions that are now unused should be removed 
- Semester 0 should not be a valid global semester value, instead we can add something specifically for the planner 
- Drag and drop ID and module type should be strongly typed instead of just being strings

website/src/views/planner/PlannerModuleSelect.tsx
jeffsieu: Actually, how does the adding multiple module codes part work? Since `onSelect(inputValue)` seems to call this: https://github.com/nusmodifications/nusmods/blob/486cca0eda81a361485ee6e79ba219d67dd72c9e/website/src/views/planner/AddModule.tsx#L60-L69
which seems to add only 1 module

ZhangYiJiang: It's in this function https://github.com/nusmodifications/nusmods/blob/4ba9c9bb893715c29c98300dff4b4032c650c62c/website/src/views/planner/PlannerContainer/PlannerContainer.tsx#L93. Again the types are wrong and the name is a bit misleading - moduleCode is matched to search for anything that looks like a module code

jeffsieu: @ZhangYiJiang I changed the logic so that the module code extraction is done in `PlannerModuleSelect`, and `PlannerContainer.onAddModule` now only handles a single module at a time. i.e. `AddModuleData`'s type has also been modified. This decision was made because I needed the `yearLong` property of every module. And `PlannerModuleSelect` does have a reference to `modules` from which I can extract ModuleCondensed instances out of.

website/src/types/modules.ts
```diff
@@ -50,7 +51,7 @@ export const WorkingDays: readonly Day[] = [
 
 export const DaysOfWeek: readonly Day[] = [...WorkingDays, 'Sunday'];
 
-export const Semesters: readonly Semester[] = [1, 2, 3, 4];
+export const Semesters: readonly Semester[] = [0, 1, 2, 3, 4];
```
ZhangYiJiang, line 54: Please revert this

website/src/selectors/planner.test.ts
```diff
@@ -45,6 +45,7 @@ describe(getAcadYearModules, () => {
 
   test('should add semesters for empty years', () => {
     const emptyYear = {
+      [0]: [],
```
ZhangYiJiang, line 48: Please revert these changes

website/src/utils/planner.ts
```diff
@@ -134,3 +163,15 @@ export function getTotalMC(
...
 }
+
+/**
+ * Returns if a planner module is year-long
+ */
+export function isYearLong(module: Pick<PlannerModuleInfo, 'moduleInfo' | 'customInfo'>): boolean {
+  return module.moduleInfo ? Boolean(module.moduleInfo.attributes?.year) : false;
+}
+
+export function isSemester(plannerSemester: PlannerModuleSemester | string) {
```
ZhangYiJiang, line 174: This makes the function a type guard which means you won't need to cast https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards 

```suggestion
export function isSemester(plannerSemester: PlannerModuleSemester): plannerSemester is Semester { 
```

website/src/selectors/planner.ts
```diff
@@ -174,10 +173,10 @@ export const getIBLOCs = (state: State): PlannerModuleInfo[] => {
 };
 
 export const getExemptions = (state: State): PlannerModuleInfo[] =>
-  mapNonTimetableModules(state, EXEMPTION_YEAR, EXEMPTION_SEMESTER);
+  mapNonTimetableModules(state, EXEMPTION_YEAR, 'exemption');
 
 export const getPlanToTake = (state: State): PlannerModuleInfo[] =>
-  mapNonTimetableModules(state, PLAN_TO_TAKE_YEAR, PLAN_TO_TAKE_SEMESTER);
+  mapNonTimetableModules(state, PLAN_TO_TAKE_YEAR, 'planToTake');
```
ZhangYiJiang, line 179: On L205 the code is still iterating over `Semesters` which does not include `yearLong`, so year long modules will never correctly count towards fulfilling prereqs.

ZhangYiJiang (commented):
Looking very close! Please revert some of the changes that are outdated, run the linter, and avoid using cast in favor of type guards, and this PR should be good to go

jeffsieu:
@ZhangYiJiang Most issues should be resolved. There's currently still a circular dependency between types/planner.ts and types/reducers.ts.... I'm not sure how to resolve this without shifting the type declarations over from a file to the other.

yangshun:
Guys, merge this plox

ravern:

PR #2919, opened 2020-10-20
Speed up prod-deploy by allowing blue to build from cache

taneliang:
The `prod-deploy.sh` script deploys our dockerized services on our production server. However, the script takes a really long time to run, partly because it builds the containers twice from scratch. This PR changes the script to only build the containers once, drastically improving deployment time.

Already tested in production; this PR just commits the changes on the server to our repository.

Also prunes old Docker data on deploy to prevent the server's hard drive from filling up.


li-kai (approved):
LGTM. Thanks for making this change.

PR #2931, opened 2020-10-27
Fix skipped public holidays in holiday scraper

taneliang:
Seems like the MOM API no longer inserts observance events (the Monday holidays that come after public holidays on Sunday) into its ICS files. Our script originally skips these observance events, preferring to mark the original event as being observed on Monday instead. Because the observance events are no longer there, our script skips an actual holiday instead: in 2020, the existing script skips Polling Day and Deepavali.

Both events are in the original CSV likely because MOM removed the observance events after we ran the scraper.

Breakdown:

- Migrate from Moment.js to date-fns. Once Moment.js has been removed from our Webpack configs, we can remove it from our list of deps.
- Switch to HTTPS URL.
- Deleted unused 2019 holidays.
- Ran both this script and holidays-csv-to-json.js. No change to data/holidays.json.


website/scripts/public-holidays-ics-to-csv.js
```diff
@@ -43,34 +43,37 @@ years
...
           .events()
           .map((event) => ({
-            date: moment(event.getPropertyValue('DTSTART').valueOf()),
+            date: new Date(event.getPropertyValue('DTSTART').valueOf()),
             name: event.getPropertyValue('SUMMARY'),
           }))
           .sort((a, b) =>
             // Order not guaranteed. Have to sort.
-            a.date.isBefore(b.date) ? -1 : 1,
+            isBefore(a.date, b.date) ? -1 : 1,
```
ZhangYiJiang, line 51: /nit If they are both dates it might be simpler to do `a.valueOf() - b.valueOf()`

ZhangYiJiang (commented):
> Specifically, for 2020, the existing script skips Polling Day and Deepavali.

The updated CSV doesn't seem to reflect this?

taneliang:
> The updated CSV doesn't seem to reflect this?

Haha yeah, I think the ICS files changed after we ran our scripts at the start of the year. If you run the script on master now, those holidays will be deleted from our CSV.

website/scripts/public-holidays-ics-to-csv.js
ZhangYiJiang: Actually `a - b` works surprisingly enough, it seems JS coerces Date to number easily enough

taneliang: Nice! I've simplified it.

taneliang:
CI failing because of the Immer upgrade

ZhangYiJiang (approved):

PR #2932, opened 2020-10-27
Update sitemap

taneliang:
Ran `node sitemap.js`.


ZhangYiJiang (approved):
I wonder if we can get a more stable diff if we sorted the venues.

PR #2933, opened 2020-10-27
Upgrade to Webpack 5

taneliang:
Details and breakdown in the commit history.

The biggest change is that `process` has now been removed, along with all other Node.js core modules. I've changed all the env vars to global variables, and introduced a new `__DEV__` variable that replaces `process.env.NODE_ENV`.

I've also replaced our `start.js` script in favor of a minimal `devServer` config. Overlays still work and lint/TS errors still appear in the terminal.

image

Depends on #2931 for the removal of Moment.js in one of our scripts.
Supersedes #2648, #2926.

## Test Plan

- `yarn start`
	|Webpack 4 (`master`)|Webpack 5|
	|-|-|
	|image|image|

- `yarn build`, and `npx serve -s dist`.
	|Webpack 4 (`master`)|Webpack 5|
	|-|-|
	|image|image|

- `docker-compose up --build`

## Known Issues

1. ~[script-ext-html-webpack-plugin](https://github.com/numical/script-ext-html-webpack-plugin) is now deprecated and the maintainer says it doesn't support Webpack 5, but it still seems to work. I'll just leave it in since I couldn't find any maintained alternatives to it.~ Removed in #2934; we can merge that after this PR.

1. ~Chunk naming seems a little broken in dev, and the `3` and `app` chunks seem to be duplicates. I think we can fix this in a future PR though, unless someone knows what's wrong.~ Duplicate `3` and `app` chunks fixed in da9451f.

1. Prod chunks are correct but vendor chunk names use numeric IDs now. I'm not sure how to fix this, but I think it's fine? The only issue this causes is that it's now difficult to identify the original file the chunk came from.

1. ~Hot loading is broken and HMR is likely broken (but I didn't test this).~ Fixed in #2953, which replaces React Hot Loader with Fast Refresh.


li-kai (approved):
LGTM. We can address the hashes in a later PR.

PR #2935, opened 2020-10-28
Upgrade to React 17 and new JSX transform

taneliang:
Breakdown and details in commit messages. A pretty straightforward series of dep bumps and a codemod.

The biggest issue is the use of an unofficial Enzyme adapter. We'll need to remember to switch to the official one once it's released.

After this and #2933 are merged, we can also switch from React Hot Loader to Fast Refresh. Didn't do it in this PR as it requires quite a few changes to our Webpack config.

### Test Plan

- `yarn start`. Hot reload still works.
- `yarn build`. Doesn't seem affected.
	image
- `yarn test`
- `yarn lint:code`


website/package.json
```diff
@@ -46,9 +46,9 @@
     "@types/mousetrap": "1.6.4",
     "@types/no-scroll": "2.1.0",
     "@types/query-string": "5.0.0",
-    "@types/react": "16.9.53",
+    "@types/react": "16.9.54",
```
ZhangYiJiang, line 49: This still on 16.x here and in react-dom below is safe since we don't expect any breaking changes in 17 right?

ZhangYiJiang (commented):
Only concern is switching import style might impact tree shaking. I would use packtracker to confirm that did not regress, but it doesn't seem to be working right now, weird. It's probably fine though since we have a bunch of those in the old code too

website/src/views/components/ExternalLink.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { ReactNode } from 'react';
+import { ReactNode } from 'react';
+import * as React from 'react';
```
ZhangYiJiang, line 2: Should combine this

website/src/views/components/disqus/CommentCount.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useEffect } from 'react';
+import { useEffect } from 'react';
+import * as React from 'react';
```
ZhangYiJiang, line 2: This too

website/src/views/components/filters/DropdownListFilters.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useState, useRef } from 'react';
+import { useState, useRef } from 'react';
```
ZhangYiJiang, line 1: And here

ZhangYiJiang (commented):
The result from the codemod is weirdly inconsistent, but then again our code wasn't consistent before so that might be why. There are two places where we can combine imports (using one or the other import style, there's no need for there to be two imports from React)

website/package.json
taneliang: Yeah, afaik there aren't any stable API differences, so this should be pretty safe.

taneliang:
Hmm yeah it does seem a bit inconsistent, but I think it's accepted to have multiple import statements from React. Here's a sample from the React codebase:

image

If we want to combine them, should we destructure everything (the most problematic one being `React.FC` as I think it'll look weird without the `React.`) or prepend `React.` to everything in files that need `React.`?

ZhangYiJiang (approved):

taneliang:
> Only concern is switching import style might impact tree shaking

As discussed on Telegram, it looks like enabling the JSX transform (0f1c804) caused bundle size to grow by ~1KB, and the new imports (5536cb6) caused the total bundle size to shrink by a few bytes. So this shouldn't be an issue

PR #2937, opened 2020-10-29
Try to fix immer draft escape bug

ZhangYiJiang:
## Context

From the error message it looks like some reference to `draft` escaped the produce function, but it's not clear where that happened. Using the new `current()` function seems to fix the issue, but the bug is not consistent which is kind of maddening. 

I increased the number of moves and games made in the fuzzer, though it doesn't necessarily seem to improve the chance of this repro-ing. 

## Implementation

Using the new `current()` function in `movePieceDown` to fix the places where we try to construct new objects in the original code using parts of draft, which is pretty sus and probably the cause of the crash 

## Other Information

[Screen Shot 2020-10-28 at 11 10 12 PM](https://www.youtube.com/watch?v=BUOri65S1S0)


website/src/views/tetris/TetrisGame.integration.test.tsx
```diff
@@ -6,30 +6,40 @@ import TetrisGame from './TetrisGame';
 
 describe(TetrisGame, () => {
   test('integration test', () => {
-    // Randomly play the game for 250 turns 20 times to check the game
+    // Randomly play the game for 1000 turns 100 times to check the game
```
taneliang, line 9: Should we use the original numbers instead so that the integration tests won't run for so long?

taneliang (approved):
Nice! Looks like we still don't know what the cause is, but since this works let's just go with it 🚢

website/src/views/tetris/TetrisGame.integration.test.tsx
taneliang: Oh, actually integration tests only took 9s. Let's just use this then

ZhangYiJiang: Yeah, the test time doesn't seem to increase linearly with number of games, interestingly enough. Maybe this is the JIT in action

PR #3007, opened 2020-12-01
Remove `withRouter` (1/n)

taneliang:
## Context

As #2922 requires extensive changes to the whole codebase, I've decided to split it into multiple PRs.

This PR works towards React Router 6 compatibility by removing calls to `withRouter`. RR6 is necessary for preloading support, which is necessary in #2922. This is incomplete and will be continued in a future PR, but I think it may be best to get some feedback on the current work first.

## Implementation

1. Frontend tests of hoc-wrapped components will likely need to be changed significantly, because:
	1. With hooks, we can't e.g. pass in `history` and Redux state props. We need to use `MemoryRouter`/`Router` and Redux's `Provider` instead.
	1. Tests frequently assume that data will be passed in as props, which may no longer be the case once we switch to Relay/Apollo. Tests that are written from the end user's perspective will help us prepare for this.
	1. Some tests assume the component will be a class component, and test their methods directly.

1. Added @testing-library/react so that it's easier to test components in context and from a user's perspective.

1. `makeResponsive` is deprecated with a new `useMediaQuery` hook.

1. A new global `__TEST__` variable is declared, as we sometimes need to distinguish between `__DEV__` and `__TEST__`.

`GlobalSearchContainer` tests have been rewritten with this new approach. Would appreciate feedback!


website/src/utils/css.ts
```diff
@@ -34,10 +33,6 @@ export function touchScreenOnly(): QueryObject {
   return { pointer: 'coarse' };
 }
 
-export function queryMatch(query: QueryObject) {
-  return window.matchMedia(json2mq(query));
-}
```
taneliang, line 39: Dead code originally used in module search before we moved to Elasticsearch

website/src/views/AppShell.tsx
```diff
@@ -25,172 +27,160 @@ import ApiError from 'views/errors/ApiError';
...
+  );
 
+  const fetchModuleListAndTimetableModules = useCallback(() => {
     // Retrieve module list
-    const moduleList = this.fetchModuleList();
+    const moduleListPromise = fetchModuleList();
 
     // Fetch the module data of the existing modules in the timetable and ensure all
     // lessons are filled
+    const timetables = store.getState().timetables.lessons;
```
taneliang, line 86: Not using `useSelector` here, because if we bind `timetables` outside this hook:

1. We'll need to pass `timetables` as a dependency to `useCallback`, which will cause the callback to change when the user modifies the timetable. This will cause the effect below to be executed, which is not what we want.
2. If we don't pass `timetables` as a dependency, the callback may capture a old copy of `timetables`, which will cause the incorrect modules to be fetched.

website/src/views/components/KeyboardShortcuts.tsx
```diff
@@ -99,101 +99,83 @@ export class KeyboardShortcutsComponent extends React.PureComponent<Props, State
...
+    }
+
+    bind('z', APPEARANCE, 'Previous Theme', () => {
       dispatch(cycleTheme(-1));
-      this.notifyThemeChange();
+      notifyThemeChange();
     });
 
-    this.bind('c', APPEARANCE, 'Next Theme', () => {
+    bind(['c', '0'], APPEARANCE, 'Next Theme', () => {
```
taneliang, line 130: Hmm... I know I wrote this but I don't know why I did it. Can anyone make sense of this? LOL. Otherwise let's undo this

```suggestion
    bind('c', APPEARANCE, 'Next Theme', () => {
```

website/src/views/components/ScrollToTop.test.tsx
```diff
@@ -55,17 +55,15 @@ describe('ScrollToTopComponent', () => {
...
 
   test('integration test', () => {
-    const wrapper = make({ onComponentDidMount: true, onPathChange: true });
-    const history = getHistory(wrapper);
-
+    const history = make({ onComponentDidMount: true, onPathChange: true });
     expect(window.scrollTo).toHaveBeenCalledTimes(1);
     expect(window.scrollTo).toHaveBeenCalledWith(0, 0);
-    history.push('/foo');
+    act(() => history.push('/foo'));
```
taneliang, line 61: `act` is needed to flush effects

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -19,84 +31,330 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
+const relevantStoreContents = {
+  moduleBank: { moduleList: MODULES },
+  venueBank: { venueList: VENUES },
+};
+
+function make(storeOverrides: Partial<typeof relevantStoreContents> = {}) {
+  const { store } = configureStore(
+    produce(reducers(undefined, initAction()), (draft) => {
+      draft.moduleBank.moduleList = (storeOverrides.moduleBank?.moduleList ??
+        relevantStoreContents.moduleBank.moduleList) as typeof draft.moduleBank.moduleList;
```
ZhangYiJiang, line 43: What's the type error here?

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -19,84 +31,330 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
+    userEvent.type(getByRole('textbox'), '1');
+    expect(queryAllByRole('option')).not.toHaveLength(0);
+  });
+
+  test('shows at most 10 choices when there are many venues and modules', () => {
+    const { getByRole, getAllByRole } = make();
+    userEvent.type(getByRole('textbox'), '1 ');
+    expect(getAllByRole('option').map((elem) => elem.textContent)).toMatchInlineSnapshot(`
+    Array [
+      "ModulesView All ",
```
ZhangYiJiang, line 109: Ooof, this is not ideal for screen readers. We should at the very least add a space here, and maybe using some flexbox shenanigans reverse the order of these in the DOM so that it reads more naturally

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -19,84 +31,330 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
+      "AA1010 TestSem 1",
```
ZhangYiJiang, line 110: Same here - probably should add a space at least

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -19,84 +31,330 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
+      "Venue 60",
+      "Venue 61",
+      "Venue 62",
+      "Venue 63",
+      "Venue 64",
+      "Venue 65",
+      "Venue 66",
+      "Venue 67",
+      "Venue 68",
+      "Venue 69",
```
ZhangYiJiang, line 356: Nice

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -19,84 +31,330 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
+      "CP1010 TestSem 1",
+      "CQ1010 TestSem 1",
+      "CR1010 TestSem 1",
+    ]
+  `);
 
-  let { modules, venues } = instance.getResults('1010')!;
-  expect(modules).toHaveLength(70);
-  expect(venues).toHaveLength(0);
+    userEvent.clear(getByRole('textbox'));
```
ZhangYiJiang, line 281: Feels like this should be split into two tests, though it might also be useful to test state changes like this. Probably as a separate test though

website/src/views/AppShell.tsx
```diff
@@ -25,172 +27,160 @@ import ApiError from 'views/errors/ApiError';
...
-
-export class AppShellComponent extends React.Component<Props, State> {
-  state: State = {};
-
-  componentDidMount() {
-    const { timetables } = this.props;
+/**
+ * Fetch module list on mount.
+ */
+function useFetchModuleListAndTimetableModules(): {
```
ZhangYiJiang, line 42: Do we _need_ to get rid of `connect`? This feels a fair bit more awkward then the old code, though having good typings is nice

ZhangYiJiang (commented):
Man, this is truly a heroic effort. _Props_ to you for doing this :P

website/src/utils/css.ts
```diff
@@ -1,19 +1,18 @@
...
-import json2mq, { QueryObject } from 'json2mq';
-import { entries } from 'lodash';
+import type { QueryObject } from 'json2mq';
 
-export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
-
-const breakpoints: { [breakpoint: string]: number } = {
+// NOTE: Keep in sync with Bootstrap's breakpoints.
+// Breakpoints at time of writing: https://getbootstrap.com/docs/4.5/layout/overview/
+const breakpoints = Object.freeze({
```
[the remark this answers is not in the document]
taneliang: Yeah I figured that the runtime cost is justified since it's only run once and preserves correctness in prod. I'll just change this to `as const`

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang: One of the properties is marked `readonly`. I could look into this more

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang: Hmm, should we fix this in another PR? This is currently the behavior on master

website/src/views/AppShell.tsx
taneliang: What feels awkward? If this hook is weird, maybe we could just inline it into the `AppShell` component. I just thought it was nicer to isolate the fetching logic into its own hook so that we didn't have so many variables lying around. I don't think we need to get rid of `connect`, but I don't think keeping it around helps much?

website/src/views/components/ScrollToTop.tsx
```diff
@@ -1,50 +1,53 @@
...
-      onPathChange,
-      location: { pathname, hash },
-    } = this.props;
+export type Props = {
+  onComponentDidMount?: boolean;
+  onPathChange?: boolean;
+  shouldScrollToHash?: boolean;
+};
 
+const ScrollToTop: React.FC<Props> = ({
```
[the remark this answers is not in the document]
taneliang: Oh yeah! Didn't think of doing this

website/src/views/components/RandomKawaii.tsx
```diff
@@ -1,48 +1,20 @@
...
-    return (
-      <div {...wrapperProps}>
-        <Kawaii size={size} mood={moodProp} color={color} />
-      </div>
-    );
-  }
-}
+const RandomKawaii: FC<Props> = ({ size, color = '#FF715D', mood, ...wrapperProps }) => {
+  const [Kawaii] = useState(() => sample(icons) ?? icons[0]);
+  const [defaultMood] = useState(() => sample(defaultMoods) ?? defaultMoods[0]);
```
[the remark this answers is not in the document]
taneliang: > Using state to only initialize once instead of using refs is rather weird lol.

Hmm, I'm not sure if the additional code overhead is worth it. Also looks like someone else also suggested using `useState` in that issue 😆. IMO `useState` is the best solution here as conceptually what we want is literally state that we don't want to change. I don't think it's exactly a ref either as refs tend to imply some sort of [imperative/mutable variable](https://www.cs.cornell.edu/courses/cs3110/2020fa/textbook/mut/refs.html).

> Also array access like this is not really safe anyway (if sample returns undefined then the array is empty, so [0] would also return undefined), so just use !

Yeah I just wanted to avoid having to eslint-ignore, but sure that works too. What we really want is `nullthrows` but probably it's probably not worth adding it for just one callsite.

website/src/views/components/KeyboardShortcuts.tsx
taneliang: Yeah, seems like I was messing around to test some hypothesis (i.e. was drunk)

website/src/views/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,34 @@
+import { useMemo } from 'react';
+import { Subscription, useSubscription } from 'use-subscription';
+import json2mq, { QueryObject } from 'json2mq';
+
+/**
+ * To be used together with utilities in css.ts.
+ * @returns Whether `mediaQuery` is/are matched.
+ */
+export default function useMediaQuery(mediaQuery: string | QueryObject | QueryObject[]) {
```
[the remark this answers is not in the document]
taneliang: `useState` + `useEffect` will probably work for the vast majority of cases. That's also the approach used by https://usehooks.com/useMedia, but from glancing at their code I think there may be a very minor but unfixable edge case: `getValue` may capture an outdated value if the viewport is resized before effects are flushed. I don't know whether this issue is present in React legacy mode (I don't remember if passive effects are flushed in a separate task/microtask after commit), but I'm pretty sure this will be an issue in concurrent mode as the viewport may be resized between interrupted renders, before any layout/passive effects are flushed.

I don't know how `matchMedia` works though, so I may be wrong.

So TL;DR I think `useSubscription` is the most suitable hook as it's built for this exact use case, at least until `useMutableSource` can be used.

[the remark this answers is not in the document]
taneliang: Ah, looks like `useSubscription` pretty much does `useState` + `useEffect` but also fixes that edge case. Cool!

https://github.com/facebook/react/blob/148ffe3/packages/use-subscription/src/useSubscription.js#L112-L115

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang:
```
Type 'ModuleList' is not assignable to type 'WritableDraft<Readonly<{ moduleCode: string; title: string; semesters: readonly number[]; }>>[]'.
  Type 'Readonly<{ moduleCode: string; title: string; semesters: readonly number[]; }>' is not assignable to type 'WritableDraft<Readonly<{ moduleCode: string; title: string; semesters: readonly number[]; }>>'.
    Types of property 'semesters' are incompatible.
      The type 'readonly number[]' is 'readonly' and cannot be assigned to the mutable type 'number[]'.ts(2322)
```

TL;DR: `semesters` is `readonly` (see snippet below), but Immer types its draft's `semesters` as a mutable array.

https://github.com/nusmodifications/nusmods/blob/089eb585a3751886cb6a2b99a206c6788d31d101/website/src/types/modules.ts#L144-L148

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang: Created issue https://github.com/nusmodifications/nusmods/issues/3010

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang: Created issue https://github.com/nusmodifications/nusmods/issues/3010

website/src/views/components/RandomKawaii.tsx
```diff
@@ -1,48 +1,22 @@
...
-  }
-}
+  return (
+    <div {...wrapperProps}>
+      <Kawaii size={size} color={color} mood={mood || defaultMood} />
+    </div>
+  );
+};
 
 export default RandomKawaii;
```
ZhangYiJiang, line 22: Memo

website/src/views/layout/GlobalSearchContainer.test.tsx
ZhangYiJiang: Ah. I guess it is possible to fix the typing of `MODULES`, but it's a bit awkward (we need to add this util type https://github.com/Microsoft/TypeScript/issues/24509). Cast is probably fine too, since this is just a test

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -17,86 +30,338 @@ const MODULES = _.flatMap(letters, (firstLetter): ModuleCondensed[] =>
...
 
-test('shows no choices when search is too short', () => {
-  const instance = make().instance() as SearchContainerComponent;
-  expect(instance.getResults('1')).toBeNull();
-});
+  test('hides module when screen size is small', () => {
+    let onMediaChangeCallback: () => void;
+    const addEventListener = (_type: string, listener: (...args: unknown[]) => void) => {
+      onMediaChangeCallback = listener;
+    };
```
ZhangYiJiang, line 77: Might be simpler to just mock the entire hook, but this does also add some coverage to the hook itself, so 🤷

ZhangYiJiang (approved):

website/src/views/hooks/useScrollToTopEffect.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import createHistory from 'test-utils/createHistory';
+import { mockDom, mockDomReset } from 'test-utils/mockDom';
+import { scrollToHash } from 'utils/react';
+
+import useScrollToTopEffect from './useScrollToTopEffect';
+
+jest.mock('utils/react');
+const mockedScrollToHash = scrollToHash as jest.MockedFunction<typeof scrollToHash>;
+
+const Tester: FC = () => {
```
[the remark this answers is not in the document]
taneliang: Hmm, I think having an explicit tester component is clearer? I didn't think using that library helps much since this component is only 2 lines long, and we'll still need to provide a `Router` wrapper (similar to the snippet below from hook testing library [docs](https://react-hooks-testing-library.com/usage/advanced-hooks)). The additional indirection seems to provide little value in this case.

```js
  const wrapper = ({ children }) => <CounterStepProvider step={2}>{children}</CounterStepProvider>
  const { result } = renderHook(() => useCounter(), { wrapper })
```

website/src/views/hooks/useScrollToTopEffect.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function scrollToTop() {
+  window.scrollTo(0, 0);
+}
+
+/**
+ * Scrolls to top (or hash, if a hash is present in the URL) after the first
+ * completed render.
+ */
+export default function useScrollToTopEffect() {
```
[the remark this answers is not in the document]
taneliang: I intended to make it more explicit that it's a passive effect (as opposed to just some magic), but yeah I guess the additional noise isn't worth it.

website/src/views/layout/GlobalSearchContainer.test.tsx
taneliang: We probably should add tests for the hook itself and simplify the tests here. Was intending to do it but forgot. I'll file an issue to fix another time

taneliang: Created issue #3011.

taneliang:
Thanks for the quick reviews @ZhangYiJiang 🏃‍♂️

PR #3009, opened 2020-12-02
Remove `withRouter` (2/2)

taneliang:
Continues from #3007.

There's 1 remaining `withRouter` callsite in `VenuesContainer`, but I'll fix that in a separate PR as the venues components looks like they're due for an overhaul.

I've fixed the noob React Testing Library mistakes in this PR, but to prevent this PR from ballooning further, I didn't fix the mistakes from #3007. I'll fix them in a follow up PR and add the Testing Library ESLint plugins.


website/src/test-utils/createHistory.ts
```diff
@@ -11,17 +11,13 @@ type MatchShape = {
   isExact?: boolean;
 };
 
-// This can also be Location, but no test case use that for now so we leave it
```
taneliang, line 14: `createMemoryHistory`'s `initialEntries` can only be a `string[]`, so I think this comment is outdated.

website/src/views/layout/Navtabs.test.tsx
```diff
@@ -1,13 +1,71 @@
...
+        "Today",
+        "Timetable",
+        "Modules",
+        "Venues",
+        "Settings",
+        "Contribute",
+        "Whispers",
+      ]
+    `);
+    expect(container).toMatchSnapshot();
```
taneliang, line 51: Not entirely sure how useful these full DOM snapshots are anymore

taneliang: Removed

li-kai (approved):
LGTM 🍎

PR #3022, opened 2020-12-04
Improve `GlobalSearch` `option` accessibility

### LINKED ISSUE #3010: `GlobalSearch` `option`s have poor readability in screen readers
See test cases in GlobalSearchContainer.test.tsx.

Context:

> Ooof, this is not ideal for screen readers. We should at the very least add a space here, and maybe using some flexbox shenanigans reverse the order of these in the DOM so that it reads more naturally

_Originally posted by @ZhangYiJiang in https://github.com/nusmodifications/nusmods/pull/3007#discussion_r533916966_

bingcheng45:
## Context
 
Fixes #3010 

Screen reader interprets the header of the GlobalSearch component in a manner that is not friendly for non-visual users.
E.g. "Modules View all" header is being interpreted as "ModulesView all" by the screen reader.
## Implementation

Reversed the order of the header elements (so that that the screen reader can read it as "View all Modules") in the GlobalSearch component and use flexbox to reverse the UI (so it will look ordered to visual users).

## Other Information

:D


website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -109,14 +109,14 @@ describe('GlobalSearchContainer', () => {
     userEvent.type(getByRole('textbox'), '1 ');
     expect(getAllByRole('option').map((elem) => elem.textContent)).toMatchInlineSnapshot(`
     Array [
-      "ModulesView All ",
+      "View All Modules ",
       "AA1010 TestSem 1",
```
taneliang, line 113: Nice! Could we fix these as well?

bingcheng45: Hi sorry, but which part needs to be fixed?

taneliang: Ah, I mean that I think it would be better if the options for each mod could also appear as `AA1010 Test Sem 1`

bingcheng45: Ohh, but is `testsem 1` is meant to be together as a single entity and will be replaced by `sem 1` or `sem 2 `? if not then I try to separate it.

taneliang: I don't think they're meant to be a single word -- this is what the UI looks like:

image

website/src/views/layout/GlobalSearch.tsx
```diff
@@ -240,7 +242,9 @@ class GlobalSearch extends Component<Props, State> {
                       [styles.selected]: highlightedIndex === index + 1,
                     })}
                   >
-                    <span>{highlight(`${module.moduleCode} ${module.title}`, tokens)}</span>
+                    {/* Extra space after module.title to aid screen reader to seperate module
+                     title from semester title */}
+                    <span>{highlight(`${module.moduleCode} ${module.title} `, tokens)}</span>
```
taneliang, line 247: Hmm if we do this, won't the tokens come out like `AA1010 Sem1Sem2`?

Could probably verify this by changing the test case modules to contain multiple sems, and seeing what the snapshots look like

taneliang: Maybe there's a way to insert screen reader-only spaces so that it doesn't mess with the DOM. I vaguely recall an API like `<div sr-only=" " />`

bingcheng45: Hi E Liang, I have updated using sr-only for adding a space between the semesters.

taneliang (approved):
I realized screen readers will add those spaces for us, so we don't actually need to add them ourselves. I'll remove them and merge this PR. Thanks so much for doing this! :D

taneliang:
Chrome:

image

Firefox:

image

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -96,17 +96,18 @@ describe('GlobalSearchContainer', () => {
...
     Array [
-      "ModulesView All ",
-      "AA1010 TestSem 1",
-      "AB1010 TestSem 1",
-      "AC1010 TestSem 1",
-      "AD1010 TestSem 1",
-      "AE1010 TestSem 1",
-      "AF1010 TestSem 1",
-      "VenuesView All ",
+      "View All Modules",
```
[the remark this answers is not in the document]
bingcheng45: Yup, View all modules works as intended

PR #3025, opened 2020-12-05
Add tests for `useMediaQuery` hook 

### LINKED ISSUE #3011: Add tests for `useMediaQuery` hook
`useMediaQuery` was added in #3007 to replace `matchBreakpoints`. It's currently tested in `GlobalSearchComponent`'s tests, which isn't ideal. We should:

1. Test the `useMediaQuery` hook itself. This can be based on `GlobalSearchComponent`'s "hides module when screen size is small" test.
1. Simplify `GlobalSearchComponent`'s "hides module when screen size is small" test to only be a sanity check that we've actually run the hook, e.g. by rendering once with `matches: true`.

Original context:

> We probably should add tests for the hook itself and simplify the tests here. Was intending to do it but forgot. I'll file an issue to fix another time

_Originally posted by @taneliang in https://github.com/nusmodifications/nusmods/pull/3007#discussion_r534634177_

teikjun:
## Context
This PR resolves #3011.

## Implementation
The implementation follows the suggested approach in the [original issue](https://github.com/nusmodifications/nusmods/issues/3011#issue-755777774). 
The test for `GlobalSearchComponent` checks if `useMediaQuery` hook is called when `GlobalSearchComponent` is rendered.
The test in `useMediaQuery` checks that the module is hidden when the client's screen size is decreased beyond the breakpoint.

## Other information
I was thinking of adding a test to check that the module is displayed when client's screen size is increased, but I didn't include it because it seems repetitive. Please let me know if you think of anything else I should test :)


website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { act, render } from '@testing-library/react';
+import { breakpointUp } from 'utils/css';
+import { FC } from 'react';
```
taneliang, line 3:
```suggestion
import type { FC } from 'react';
```

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { mockWindowMatchMedia } from 'test-utils/mockDom';
+import useMediaQuery from './useMediaQuery';
+
+const Tester: FC = () => {
+  const matchedBreakpoint = useMediaQuery(breakpointUp('md'));
+
+  if (matchedBreakpoint) {
+    return <div />;
+  }
+  return null;
```
taneliang, line 13: Hmm I think this tester component can be more explicit, which will probably make the test case `expect`s clearer

```suggestion
  return matchedBreakpoint ? <>matched</> : <>no match</>);
```

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+describe(useMediaQuery, () => {
```
taneliang, line 16: We should add the `mockDom` and `mockDomReset` in `beforeEach` and `afterEach` to ensure that the DOM is set up and torn down correctly. These can be copied from the GlobalSearchContainer tests

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  function make() {
+    return render(<Tester />);
+  }
```
taneliang, line 19: Since this is just used once let's inline it hahaha

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -61,29 +65,22 @@ describe('GlobalSearchContainer', () => {
...
-    expect(container).not.toBeEmptyDOMElement();
-
-    mockWindowMatchMedia({ matches: false, addEventListener });
-    // Trigger render when matches changes
-    act(() => onMediaChangeCallback());
-    expect(container).toBeEmptyDOMElement();
+    expect(mockedUseMediaQuery).not.toHaveBeenCalled();
+    mockWindowMatchMedia({ matches: true });
+    make();
+    expect(mockedUseMediaQuery).toHaveBeenCalled();
```
taneliang, line 83: Hmm I don't think we should mock the hook, because:

1. It's an implementation detail of GlobalSearchContainer.
1. We can easily set up the DOM and let the hook do its job.

This should be enough for a sanity check. We test the `matches: true` case in our other test cases anyway.

```suggestion
    mockWindowMatchMedia({ matches: false });
    const { container } = make();
    expect(container).toBeEmptyDOMElement();
```

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  test('should hide module when screen size is small', () => {
```
taneliang, line 21: Test name needs to be updated :)

taneliang (commented):
Thank you! :D

> I was thinking of adding a test to check that the module is displayed when client's screen size is increased, but I didn't include it because it seems repetitive. 

Yep I don't think that's necessary either. This looks good enough!

website/src/views/hooks/useMediaQuery.test.tsx
taneliang: Oh btw, I think it'll be easiest to use inline snapshots with this, e.g.

```
expect(container).toMatchInlineSnapshot()
```

I think the snapshot will be something like `<div>matched</div>`, which will be extremely clear with minimal effort on your part.

teikjun:
Thanks for the review! I've made the requested changes :)

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -1,36 +1,44 @@
...
+  beforeEach(() => {
+    mockDom();
+  });
+
+  afterEach(() => {
+    mockDomReset();
+  });
 
-  test('should hide module when screen size is small', () => {
+  test('should display correct module based on matching', () => {
```
taneliang, line 21: Hmm the "module" stuff was specific to what GlobalSearchContainer was doing.

```suggestion
  test('should return whether media queries were matched', () => {
```

This also means that a more extensive test would expect that `window.matchMedia` was called with the correct media queries passed into `useMediaQuery` ("correct" because `useMediaQuery` transforms the input queries, so they're not simply passed through directly to `window.matchMedia`)

website/src/views/layout/GlobalSearchContainer.test.tsx
```diff
@@ -61,7 +61,7 @@ describe('GlobalSearchContainer', () => {
   beforeEach(() => {
     mockDom();
 
-    // Replace fetchVenueList with a noop action to stop it from firing API requests
+    // Replace fetchVenueList and useMediaQuery with a noop action to stop it from firing API requests
```
taneliang, line 64:
```suggestion
    // Replace fetchVenueList with a noop action to stop it from firing API requests
```

taneliang (commented):
Thanks for making the changes! Looks like I found a test you can add 😆

teikjun:
Added the test! 👍

website/src/types/views.ts
```diff
@@ -1,3 +1,4 @@
+import { QueryObject } from 'json2mq';
```
taneliang, line 1:
```suggestion
import type { QueryObject } from 'json2mq';
```

website/src/views/hooks/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  test('should transform media query correctly', () => {
+    mockWindowMatchMedia({ matches: false });
+    const matchMediaSpy = jest.spyOn(window, 'matchMedia');
+    expect(matchMediaSpy).not.toHaveBeenCalled();
+
+    const jsonMediaQuery = breakpointUp('md');
+    make(jsonMediaQuery);
+    expect(matchMediaSpy).toHaveBeenLastCalledWith(json2mq(jsonMediaQuery));
```
taneliang, line 62: Hmm, I don't particularly like this as it's basically reimplementing the hook. But it's okay, I think this test adds some value (and will be easy to remove it causes problems) so let's keep it

taneliang (approved):
Thank you! :D

Issue #3036, opened 2020-12-08
Unable to export timetable to .ics file


brandoncjh:
### Describe the bug

Unable to export timetable to .ics file using the Download drop-down menu. On clicking the button, nothing happens.

### To Reproduce

1. Go to 'https://nusmods.com/timetable/sem-1'
2. Click on 'Download' button next to 'Share/Sync' button
3. Select option to export as iCalendar (.ics) file

![image](https://user-images.githubusercontent.com/54733974/101461236-8e857580-3975-11eb-9093-3ff3dbcb63c8.png)

### Expected behavior

A download of .ics file should be initiated, just as in the case of pdf.

![image](https://user-images.githubusercontent.com/54733974/101461305-a2c97280-3975-11eb-9bc3-a6b97ec3c7b7.png)

### Desktop:

* OS: Windows 10
* Browser: Chrome 
* Version: 86.0.4240.198 (Official Build) (64-bit)

### Additional Information

Same issue occurs when attempting to export in [Sem 2](https://nusmods.com/timetable/sem-2) as well.

chrisgzf:
Thanks for the report! Triaging now.

PR #3037, opened 2020-12-08
Fix ical export regression after webpack 5 update

chrisgzf:
Closes https://github.com/nusmodifications/nusmods/issues/3036

## Context
https://github.com/nusmodifications/nusmods/pull/2933 seems to have broken the package we used for exporting iCal files. The package we used was frozen at a 2018-ish version from @ZhangYiJiang's fork (PR #779), to fix some macOS iCal bugs. That version of the package made use of some nodejs-specific libraries (`os` and `fs`), which were polyfilled by webpack 4. Webpack 5 no longer provides nodejs polyfills automatically, which broke that library.

## Implementation

EDIT: scrap everything below. the updated package fixes the issue but it includes `moment-timezone` (a huge package) into our bundle. the alternate fix here is to remove all nodejs package calls in our fork of `ical-generator`. which is done at https://github.com/nusmodifications/ical-generator/commit/a2aa62ed82a7c947b3f02172ac2c8ed5a9e32eb8,  and then use that updated package in nusmods

Update `ical-generator` package to latest stable version, and then manually test for regression.

After updating, ical export works and the download is started.

The generated ical file that I tested came up with this timetable:
image

for this timetable:
image

which is correct. Each timetable entry also contains its associated metadata (location, tutorial group number, etc.)

I also exported my sem1 timetable from the last build that had a working iCal export, and exported it again after the new library was added to compare the generated iCal files.

The diff is here: https://gist.github.com/chrisgzf/87524d8f729295827991a8d5a724de8c

tldr: no functional difference, new library's generated files are more concise and smaller

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -142,7 +142,7 @@
     "core-js": "3.8.1",
     "date-fns": "2.16.1",
     "downshift": "6.0.6",
-    "ical-generator": "https://github.com/chrisgzf/ical-generator.git#8fa2fc0e",
+    "ical-generator": "https://github.com/nusmodifications/ical-generator.git#a2aa62ed",
     "immer": "8.0.0",
     "json2mq": "0.2.0",
     "leaflet": "1.7.1",
--- a/website/package.json
+++ b/website/package.json
@@ -142,7 +142,7 @@
     "core-js": "3.8.1",
     "date-fns": "2.16.1",
     "downshift": "6.0.6",
-    "ical-generator": "https://github.com/ZhangYiJiang/ical-generator.git#ed6928fe",
+    "ical-generator": "https://github.com/chrisgzf/ical-generator.git#8fa2fc0e",
     "immer": "8.0.0",
     "json2mq": "0.2.0",
     "leaflet": "1.7.1",
--- a/website/webpack/webpack.config.common.js
+++ b/website/webpack/webpack.config.common.js
@@ -24,13 +24,6 @@ const commonConfig = {
     extensions: ['.js', '.jsx', '.ts', '.tsx', '.json'],
     // We don't use symlinks, so disable for performance
     symlinks: false,
-    // ical-generator imports Node.js core modules, which don't exist in a
-    // browser environment. Tell Webpack to provide empty modules for them so
-    // importing them works.
-    fallback: {
-      fs: false,
-      os: false,
-    },
   },
 
   context: parts.PATHS.src,
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -7845,9 +7845,9 @@ human-signals@^1.1.1:
   resolved "https://registry.yarnpkg.com/human-signals/-/human-signals-1.1.1.tgz#c5b1cd14f50aeae09ab6c59fe63ba3395fe4dfa3"
   integrity sha512-SEQu7vl8KjNL2eoGBLF3+wAjpsNfA9XMlXAYj/3EdaNfAlxKthD1xjEQfGOUhllCGGJVNY34bRr6lPINhNjyZw==
 
-"ical-generator@https://github.com/chrisgzf/ical-generator.git#8fa2fc0e":
+"ical-generator@https://github.com/nusmodifications/ical-generator.git#a2aa62ed":
   version "0.2.10"
-  resolved "https://github.com/chrisgzf/ical-generator.git#8fa2fc0e7495de6bf08c66d922fea24028697c07"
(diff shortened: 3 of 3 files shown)


taneliang:
Hmm, the generated ical diff looks as if it was generated with a reverted https://github.com/ZhangYiJiang/ical-generator/commit/ed6928fecce6789db4d99f856413074de7eb5a1e (the ical-generator patch applied in https://github.com/nusmodifications/nusmods/pull/779). I'm a bit hesitant to accept this as that patch did resolve some magical ical issue that we could not repro (even with a Mac), which may mean that this PR will cause that issue to resurface and we wouldn't be able to repro it.

Maybe you'd like to reapply the patch on `ical-generator`'s current version and used that instead?

website/yarn.lock
```diff
@@ -7845,9 +7845,12 @@ human-signals@^1.1.1:
...
 
-"ical-generator@https://github.com/ZhangYiJiang/ical-generator.git#ed6928fe":
-  version "0.2.10"
-  resolved "https://github.com/ZhangYiJiang/ical-generator.git#ed6928fecce6789db4d99f856413074de7eb5a1e"
+ical-generator@^1.15.2:
+  version "1.15.2"
+  resolved "https://registry.yarnpkg.com/ical-generator/-/ical-generator-1.15.2.tgz#7f4fbe80db8d327b0396fed59219de65beb89832"
+  integrity sha512-+G92MzEouVkQ5/1Qizh8qprvI5a+iMKCGeavUAwWoAuiF19TNAOO9z8qnnrX9UbW7URoE/8PVAo7SWv7V1u4jQ==
+  dependencies:
+    moment-timezone "^0.5.31"
```
taneliang, line 7853: Also why do they require `moment-timezone` now? This is a huge package.

Could we just use the existing ical-generator fork and polyfill the hostname stuff, or further patch @ZhangYiJiang's fork to remove the hostname call?

https://bundlephobia.com/result?p=moment-timezone@0.5.32

image

chrisgzf: Sure sure let me look into why this library needs `moment-timezone`. Anyway, is it possible to enable packtracker for forks as well? So that we can get a better overview on how this might affect the bundle.

taneliang: We do have Packtracker integrated, but it seems to have died suddenly a while back 😆 Feel free to try to fix it! Not sure whether you need our credentials though. I did look at this a bit on Oct 29 but I didn't find a fix.

This is what the logs looked like on that day:
image

taneliang: Oh wait it seems to be working, just not for forks? Interesting. Let me see what I can do.

taneliang: Okay, I just poked around the Packtracker app and tried to do a google search, couldn't find any way to do that :( Any idea how?

chrisgzf: I have no idea how too :( Haven't really used packtracker. I guess a lot of things just naturally don't work with forks sigh

chrisgzf:
I've patched @ZhangYiJiang's fork to not make any calls to nodejs packages in https://github.com/chrisgzf/ical-generator/commit/8fa2fc0e7495de6bf08c66d922fea24028697c07 and tested the iCal export with this package. Works fine, and doesn't include the nasty `moment-timezone` package in the bundle. 🤮

website/package.json
```diff
@@ -142,7 +142,7 @@
     "core-js": "3.8.1",
     "date-fns": "2.16.1",
     "downshift": "6.0.6",
-    "ical-generator": "https://github.com/ZhangYiJiang/ical-generator.git#ed6928fe",
+    "ical-generator": "https://github.com/chrisgzf/ical-generator.git#8fa2fc0e",
```
taneliang, line 145: Thanks! PR to our fork?

website/webpack/webpack.config.common.js
```diff
@@ -24,13 +24,6 @@ const commonConfig = {
     extensions: ['.js', '.jsx', '.ts', '.tsx', '.json'],
     // We don't use symlinks, so disable for performance
     symlinks: false,
-    // ical-generator imports Node.js core modules, which don't exist in a
-    // browser environment. Tell Webpack to provide empty modules for them so
-    // importing them works.
-    fallback: {
-      fs: false,
-      os: false,
-    },
```
taneliang, line 33: Finally we can remove this 👏

website/package.json
chrisgzf: https://github.com/nusmodifications/ical-generator/pull/1 yeap yeap need help merging this

ZhangYiJiang (approved):

devrathc:
can this fix be pulled into prod? reported on #3091 as well.

PR #3040, opened 2020-12-08
Optimize Venues search using React Concurrent Mode APIs

taneliang:
### TL;DR

Uses React Concurrent Mode APIs (added in #3039) to optimize the Venues components reworked in #3038.

### Optimizations in this PR

1. Uses `useDeferredValue` to allow the search box to be immediately updated in a sync render before a second, longer concurrent render is performed.

	|`master`|This PR|
	|-|-|
	|boxm|box|

	This may sound similar to the DIY deferred update using 2 nested `requestAnimationFrame` calls that we currently have on `master`. However, that is inferior to this solution because:
	1. CM allows in-progress renders to be abandoned, allowing multiple keystrokes in quick-ish succession to remain responsive, as previous in-progress renders are simply abandoned. On `master`, the app noticeably freezes up once a legacy render starts as it cannot be interrupted. In the gifs above, notice that on `master`, a bunch of characters appear at once -- this happens because I typed while the app was frozen.
	1. (Independent of CM) #3038 reduces unnecessary renders of expensive components, which reduces the amount of time the browser spends frozen.

	This is what typing a single character looks like in the [React Scheduling Profiler](https://react-devtools-scheduling-profiler.vercel.app/) (that I helped create :D):

	image

	TL;DR of how to read the React data (the stuff above the grey bar):
	- The circles represent state updates in this case.
	- Each horizontal row = a React fiber lane (each of which has a purpose/priority)
	- A blue box in each lane represents a render chunk, and a red box is a commit phase.

	Notice the short sync render on the left (which is in response to the key press), followed by a very long render caused by the deferred value updating. Also notice that `VenueList` takes a long time to render.

	This is what typing a second character during a render looks like:

	image
	
	Notice that a concurrent render got abandoned.

1. Uses `useDeferredValue` to allow the availability filter to open instantly. On `master`, this just freezes as there were no optimizations for this.
	ffr

Note that we do not also defer `searchOptions`, as I couldn't get it to feel right. It seems better to just let the page freeze while the render happens. (Edit: it's now deferred. See GIFs in this comment below: https://github.com/nusmodifications/nusmods/pull/3040#issuecomment-748069917)

### DIFF
--- a/website/src/styles/utils/deferred.scss
+++ b/website/src/styles/utils/deferred.scss
@@ -1,0 +1,11 @@
+// Source: https://reactjs.org/docs/concurrent-mode-patterns.html#delaying-a-pending-indicator
+.deferred {
+  visibility: hidden;
+  animation: 0s linear 0.3s forwards makeVisible;
+
+  @keyframes makeVisible {
+    to {
+      visibility: visible;
+    }
+  }
+}
--- a/website/src/styles/utils/deferred.scss
+++ b/website/src/styles/utils/deferred.scss
@@ -1,7 +1,7 @@
 // Source: https://reactjs.org/docs/concurrent-mode-patterns.html#delaying-a-pending-indicator
 .deferred {
   visibility: hidden;
-  animation: 0s linear 0.3s forwards makeVisible;
+  animation: 0s linear 0.2s forwards makeVisible;
 
   @keyframes makeVisible {
     to {
--- a/website/src/views/components/SearchBox.tsx
+++ b/website/src/views/components/SearchBox.tsx
@@ -19,7 +19,9 @@ export type Props = {
   isLoading?: boolean;
   value: string | null;
   placeholder?: string;
+  /** Called when the search box value changes */
   onChange: (value: string) => void;
+  /** Called when a search should be triggered, potentially debounced by `throttle` milliseconds. */
   onSearch?: () => void;
   onBlur?: () => void;
 };
--- a/website/src/views/components/SearchBox.tsx
+++ b/website/src/views/components/SearchBox.tsx
@@ -15,7 +15,7 @@ import styles from './SearchBox.scss';
 
 export type Props = {
   className?: string;
-  throttle: number;
+  throttle?: number;
   isLoading?: boolean;
   value: string | null;
   placeholder?: string;
--- a/website/src/views/components/SearchBox.tsx
+++ b/website/src/views/components/SearchBox.tsx
@@ -16,18 +16,18 @@ import styles from './SearchBox.scss';
 export type Props = {
   className?: string;
   throttle: number;
-  isLoading: boolean;
+  isLoading?: boolean;
   value: string | null;
   placeholder?: string;
   onChange: (value: string) => void;
-  onSearch: () => void;
+  onSearch?: () => void;
   onBlur?: () => void;
 };
 
 const SearchBox: FC<Props> = ({
   className,
   throttle,
-  isLoading,
+  isLoading = false,
   value,
   placeholder,
   onChange,
@@ -55,7 +55,7 @@ const SearchBox: FC<Props> = ({
       debounce(
         () => {
           isDirty.current = false;
-          onSearch();
+          onSearch?.();
         },
         throttle,
         { leading: false },
--- a/website/src/views/modules/ModuleFinderContainer/ModuleFinderContainer.tsx
+++ b/website/src/views/modules/ModuleFinderContainer/ModuleFinderContainer.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import type { FC } from 'react';
 import {
   Hits,
   HitsStats,
@@ -12,9 +12,10 @@ import {
 } from 'searchkit';
 import classnames from 'classnames';
 
-import { ElasticSearchResult } from 'types/vendor/elastic-search';
-import { ModuleInformation } from 'types/modules';
+import type { ElasticSearchResult } from 'types/vendor/elastic-search';
+import type { ModuleInformation } from 'types/modules';
 
+import LoadingOverlay from 'views/components/LoadingOverlay';
 import ModuleFinderSidebar from 'views/modules/ModuleFinderSidebar';
 import ModuleSearchBox from 'views/modules/ModuleSearchBox';
 import ModuleFinderNoHits from 'views/errors/ModuleFinderNoHits';
@@ -38,7 +39,7 @@ const searchkit = new SearchkitManager(esHostUrl, {
 
 const pageHead = <Title>Modules</Title>;
 
-const ModuleInformationListComponent: React.FC<HitsListProps> = ({ hits }) => (
+const ModuleInformationListComponent: FC<HitsListProps> = ({ hits }) => (
   <ul className={styles.modulesList}>
     {hits.map((hit) => {
       const result = hit as ElasticSearchResult<ModuleInformation>;
@@ -55,7 +56,7 @@ const ModuleInformationListComponent: React.FC<HitsListProps> = ({ hits }) => (
   </ul>
 );
 
-const ModuleFinderContainer: React.FC = () => (
+const ModuleFinderContainer: FC = () => (
   <div className={classnames(styles.modulesPageContainer, 'page-container')}>
     {pageHead}
     <SearchkitProvider searchkit={searchkit}>
@@ -73,7 +74,7 @@ const ModuleFinderContainer: React.FC = () => (
             />
 
             <LoadingComponent>
-              <div className={styles.loadingOverlay} />
+              <LoadingOverlay />
             </LoadingComponent>
 
             <InitialLoader component={LoadingComponent} />
--- a/website/src/views/venues/VenuesContainer.scss
+++ b/website/src/views/venues/VenuesContainer.scss
@@ -113,17 +113,6 @@ $venue-list-width: 16rem;
   border-radius: 0 0 $border-radius $border-radius;
 }
 
-.availabilitySpinner {
-  display: inline-block;
-  width: 1.4rem;
-  height: 1.4rem;
-  margin: 0 0.3rem -0.1rem 0;
-  border-width: 0.2rem;
-  // Use parent button's text color so that the spinner is still visible when
-  // the button is hovered over.
-  border-left-color: initial;
-}
-
 .noVenueSelected {
   $icon-size: 6rem;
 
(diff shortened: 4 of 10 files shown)


website/src/views/venues/VenuesContainer.tsx
```diff
@@ -153,23 +167,25 @@ export const VenuesContainerComponent: FC<Props> = ({ venues }) => {
...
             'btn btn-block btn-svg',
-            styles.availabilityToggle,
             isAvailabilityEnabled ? 'btn-primary' : 'btn-outline-primary',
           )}
           onClick={onFindFreeRoomsClicked}
           type="button"
         >
-          <Clock className="svg" /> Find free rooms
+          {isAvailabilityEnabled !== deferredIsAvailabilityEnabled ? (
+            <LoadingSpinner className={styles.availabilitySpinner} white={isAvailabilityEnabled} />
```
taneliang, line 184: This causes the spinner to flash on faster browsers. No idea how to fix it, but I think storing `isAvailabilityEnabled !== deferredIsAvailabilityEnabled` in a `useState` and using our `defer` may be able to fix it at the cause of more renders?

website/src/views/venues/VenuesContainer.tsx
```diff
@@ -153,23 +167,25 @@ export const VenuesContainerComponent: FC<Props> = ({ venues }) => {
...
           value={searchQuery}
           placeholder="e.g. LT27"
           onChange={setSearchQuery}
-          onSearch={noop}
         />
 
         <button
           className={classnames(
             'btn btn-block btn-svg',
-            styles.availabilityToggle,
```
taneliang, line 166: This style didn't exist

website/src/views/venues/VenuesContainer.scss
```diff
@@ -113,6 +113,17 @@ $venue-list-width: 16rem;
...
+.availabilitySpinner {
+  display: inline-block;
+  width: 1.4rem;
+  height: 1.4rem;
+  margin: 0 0.3rem -0.1rem 0;
+  border-width: 0.2rem;
+  // Use parent button's text color so that the spinner is still visible when
+  // the button is hovered over.
+  border-left-color: initial;
+}
```
taneliang, line 125: This is copied and customized from `LoadingSpinner.scss`. Not particularly clean, but I couldn't find any other way using CSS specificity to get CSS to use these styles over the `LoadingSpinner` ones, and I didn't have much time to work on this. Help appreciated!

website/src/views/venues/VenuesContainer.tsx
chrisgzf: How do I repro this flashing? I tried to profile it but I couldn't see the flashing

image

taneliang: Oh, just click the Find free rooms button. The clock icon in the button will disappear for a moment as the spinner flashes

website/src/views/venues/VenuesContainer.scss
ZhangYiJiang: Won't nesting them under `.venueSearch` work?

website/src/views/venues/VenuesContainer.tsx
ZhangYiJiang: I'm not sure if putting the spinner here is the best idea - the thing that's "loading" is the venue list below, and it's quite easily missed (though the animation helps). Won't putting a spinner there instead work? 

Also thinking about https://reactjs.org/docs/concurrent-mode-patterns.html#delaying-a-pending-indicator, but I don't know how to make that work with useDeferredValue

taneliang:
Rebased on parent branch, and fixed conflict with [`d5b270e` (#3038)](https://github.com/nusmodifications/nusmods/pull/3038/commits/d5b270ed29d5b7ddeccc55388bca5d0915aba348)

website/src/views/venues/VenuesContainer.tsx
taneliang: Ooh, deferring the loading indicator in CSS is brilliant. I'll switch us to `useTransition` then, especially since the current condition for determining whether to show the indicator or not is a bit of a hack.

website/src/views/venues/VenuesContainer.scss
taneliang: Oh yes. Derped because I was rushing. Anyway this style has been removed, so I'm resolving this convo

website/src/views/venues/VenuesContainer.tsx
taneliang: I lied, `useTransition` isn't a good fit for our use case. It's back to `useDeferredValue`, but with a deferred spinner :D

taneliang:
Main changes:

- Spinner is now overlaid on `VenueList`
- Spinner and overlay are now deferred by 0.2s. The experience is far better.
- `searchOptions` is now deferred -- selecting stuff in the Find free rooms is now super smooth 😍 

Updated gifs:

|No throttling|6x CPU throttling|
|-|-|
|defno|def6|

taneliang:
Rebased on parent branch, no diff change

website/src/styles/utils/deferred.scss
```diff
@@ -0,0 +1,11 @@
+// Source: https://reactjs.org/docs/concurrent-mode-patterns.html#delaying-a-pending-indicator
+.deferred {
+  visibility: hidden;
+  animation: 0s linear 0.2s forwards makeVisible;
```
ZhangYiJiang, line 4: We could in theory make this configurable from the component side using CSS variables or even inline styles, but probably YAGNI

website/src/views/components/SearchBox.tsx
```diff
@@ -16,18 +16,18 @@ import styles from './SearchBox.scss';
 export type Props = {
   className?: string;
   throttle: number;
-  isLoading: boolean;
+  isLoading?: boolean;
   value: string | null;
   placeholder?: string;
   onChange: (value: string) => void;
-  onSearch: () => void;
+  onSearch?: () => void;
```
ZhangYiJiang, line 23: Might be useful adding a small comment on the difference between this and onChange

website/src/views/modules/ModuleFinderContainer/ModuleFinderContainer.tsx
```diff
@@ -73,7 +74,7 @@ const ModuleFinderContainer: React.FC = () => (
             />
 
             <LoadingComponent>
-              <div className={styles.loadingOverlay} />
```
ZhangYiJiang, line 76: Should remove the unused duplicate class here from `ModuleFinderContainer.scss`

website/src/views/venues/VenuesContainer.tsx
```diff
@@ -46,18 +54,16 @@ export const VenuesContainerComponent: FC<Props> = ({ venues }) => {
...
 
-  // Search state
-  const [
-    /** Value of the controlled search box; updated real-time */
-    searchQuery,
-    setSearchQuery,
-  ] = useState<string>(() => qs.parse(location.search).q || '');
-  /** Actual string to search with; deferred update */
-  const deferredSearchQuery = searchQuery; // TODO: Redundant now. Use React.useDeferredValue after we adopt concurrent mode
+  const [searchQuery, setSearchQuery] = useState<string>(() => qs.parse(location.search).q || '');
```
ZhangYiJiang, line 57: /nit 

```suggestion
  const [searchQuery, setSearchQuery] = useState(() => qs.parse(location.search).q || '');
```

website/src/views/venues/VenuesContainer.tsx
```diff
@@ -46,18 +54,16 @@ export const VenuesContainerComponent: FC<Props> = ({ venues }) => {
...
+  const handleSearchChange = useCallback((newSearchQuery: string) => {
+    setSearchQuery(newSearchQuery);
+  }, []);
```
ZhangYiJiang, line 60: Isn't `handleSearchChange` exactly the same as `setSearchQuery`?

website/src/views/venues/VenuesContainer.tsx
```diff
@@ -213,11 +233,15 @@ export const VenuesContainerComponent: FC<Props> = ({ venues }) => {
       <div className={styles.venuesList}>
         {renderSearch()}
 
-        {size(matchedVenues) === 0 ? (
-          renderNoResult()
-        ) : (
+        <div className="position-relative">
```
ZhangYiJiang, line 236: I think without specifying height this will not take up the whole height of the venues list right?

ZhangYiJiang (commented):
I'm going to need a bit more time reading up the concurrent mode docs to understand what useDeferredValue is doing here

taneliang:
@ZhangYiJiang I think reading `useDeferredValue`'s implementation may be easiest. https://github.com/facebook/react/blob/99554dc36fa9f5ef29b75dd59836dcc720d831b9/packages/react-reconciler/src/ReactFiberHooks.old.js#L1498-L1538

The `mount*` functions are called the first time the hook is called, and the `update*` functions are called on subsequent calls. The `*State` and `*Effect` functions correspond to `useState` and `useEffect`.

TL;DR it looks like it's just storing the previous state in a `useState`, then doing a state update in a passive effect. But interestingly it also sets a `transition` flag on the fiber (which is now just `1` or `0`, but in the past it was the suspense config object that you see in the CM docs). I'm not sure yet what `transition` accomplishes, but `useTransition` does that as well.

website/src/views/venues/VenuesContainer.tsx
taneliang: Oh yes, I think I forgot to change this back after my brief experiment with `useTransition`

website/src/styles/utils/deferred.scss
taneliang: Yeah, probably not a good idea to overengineer this right now hahaa

website/src/views/modules/ModuleFinderContainer/ModuleFinderContainer.tsx
taneliang: Yep, it was already removed

website/src/views/venues/VenuesContainer.tsx
taneliang: That doesn't work because `qs.parse` returns `any`. We could do the below type assertion instead, but the current code seems good enough

```suggestion
  const [searchQuery, setSearchQuery] = useState(
    () => (qs.parse(location.search) as Params).q || '',
  );
```

website/src/views/venues/VenuesContainer.tsx
taneliang: I think it does since it wraps the venue list component

image

ravern:

PR #3041, opened 2020-12-09
Replace `ScrollToTop` with `useScrollToTop`

### LINKED ISSUE #3034: Replace `ScrollToTop` with `useScrollToTop`
`useScrollToTop` was added in #3007 to replace the `ScrollToTop` component. It's currently still used in 2 class components (`ModulePageContent` and `SettingsContainer`), which need to be converted to functional components before we can delete `ScrollToTop`.

This issue exists so that we remember to remove `ScrollToTop` the codebase.

### Steps

1. Convert `ModulePageContent` and/or `SettingsContainer` to function components (both should be easy to convert).
1. Replace `<ScrollToTop />` with a `useScrollToTop()` call.
1. Delete `ScrollToTop` from the codebase.

teikjun:
## Context
This PR resolves #3034. It also improves a test in `useMediaQuery.test.tsx` (see [comment](https://github.com/nusmodifications/nusmods/pull/3025#discussion_r537043351)).

## Implementation
The implementation follows the suggested approach in the original issue. In addition, the tests for `ModulePageContent` have been rewritten using RTL because `useScrollToTop` requires the router's `location`.


website/src/views/settings/SettingsContainer.tsx
```diff
@@ -59,37 +59,27 @@ type Props = {
...
       }
 
-      this.setState({ allowTracking: !tracker.isUserOptedOut() });
+      setAllowTracking(!tracker.isUserOptedOut());
     });
   };
 
-  renderModRegNotificationRounds() {
-    const { modRegNotification } = this.props;
+  const renderModRegNotificationRounds = () => {
```
ZhangYiJiang, line 81: This and `renderNightModeOption` should be inlined

website/src/views/settings/SettingsContainer.tsx
```diff
@@ -128,192 +118,179 @@ class SettingsContainer extends Component<Props, State> {
...
       </div>
-    );
-  }
+      <hr />
+    </div>
+  );
 
-  render() {
-    const { currentThemeId, modRegNotification } = this.props;
+  const { currentThemeId, modRegNotification } = props;
```
ZhangYiJiang, line 144: All props should be spread on line 62 - there's no need I think to keep `props` unspread in a functional component

website/src/views/modules/ModulePageContent.test.tsx
```diff
@@ -1,21 +1,45 @@
-import ScrollSpy from 'react-scrollspy';
-import { shallow } from 'enzyme';
-
 import { Module } from 'types/modules';
 /** @var {Module} */
 import { CS1010S } from '__mocks__/modules';
 
+import { initAction } from 'test-utils/redux';
+import { Provider } from 'react-redux';
```
ZhangYiJiang, line 6: /nit sort imports so that external imports come first, then ours separated by a line break

website/src/views/modules/ModulePageContent.test.tsx
```diff
@@ -1,21 +1,45 @@
...
   function make(module: Module = CS1010S) {
-    return shallow(<ModulePageContent module={module} />);
+    const initialState = reducers(undefined, initAction());
+    const { store } = configureStore(initialState);
+    return renderWithRouterMatch(
+      <Provider store={store}>
+        <ModulePageContent module={module} />
+      </Provider>,
+      {
+        location: '/archive/CS1010S/2017-2018',
```
ZhangYiJiang, line 24: Kind of weird to be testing with an archive URL. Shouldn't matter here though

website/src/views/settings/SettingsContainer.tsx
```diff
@@ -59,37 +59,27 @@ type Props = {
...
-        allowTracking: !tracker.isUserOptedOut(),
-      }),
-    );
-  }
+  useEffect(() => {
+    withTracker((tracker: Tracker) => setAllowTracking(!tracker.isUserOptedOut()));
+  }, []);
 
-  onToggleTracking = (allowTracking: boolean) => {
+  const onToggleTracking = (isTrackingAllowed: boolean) => {
```
taneliang, line 69: Nit: wrap in `useCallback`

website/src/views/settings/SettingsContainer.tsx
```diff
@@ -59,37 +59,27 @@ type Props = {
...
-  componentDidMount() {
-    withTracker((tracker: Tracker) =>
-      this.setState({
-        allowTracking: !tracker.isUserOptedOut(),
-      }),
-    );
-  }
+  useEffect(() => {
+    withTracker((tracker: Tracker) => setAllowTracking(!tracker.isUserOptedOut()));
+  }, []);
```
taneliang, line 67: Won't this cause the wrong value of `allowTracking` to be displayed for a while if the analytics script takes some time to load? Maybe we should initialize this component with `allowTracking = null` instead of `true`, and handle the `null` case separately.

This is also an issue with the original implementation, so we can investigate this further in a follow up issue/PR. I'll spin this off into its own issue when this PR is about to be merged.

taneliang (commented):
Thank you! :D

website/src/views/modules/ModulePageContent.tsx
```diff
@@ -50,274 +46,279 @@ export const SIDE_MENU_LABELS = {
...
-
-  render() {
-    const { module, archiveYear } = this.props;
-    const { moduleCode, title } = module;
-
-    const pageTitle = `${moduleCode} ${title}`;
-    const semesters = getSemestersOffered(module);
-    const isArchive = !!archiveYear;
-    const offered = isOffered(module);
+const ModulePageContent: React.FC<Props> = (props) => {
```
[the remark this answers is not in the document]
taneliang: Oh, this comment is a duplicate of https://github.com/nusmodifications/nusmods/pull/3041#discussion_r541818710.

website/src/views/settings/SettingsContainer.tsx
teikjun: I've spread all props except for the ones that shadow the import variables (`selectTheme`, `selectMode`, etc.)

website/src/views/modules/ModulePageContent.test.tsx
```diff
@@ -1,21 +1,45 @@
...
-    const component = make();
-    expect(component.find('[id]').map((ele) => ele.prop('id'))).toEqual(
-      component.find(ScrollSpy).prop('items'),
-    );
+    make();
+    const orderedSideMenuItems = screen
+      .getAllByRole('link')
+      .map((elem) => elem.textContent?.toLowerCase());
+    const orderedDocumentItems = screen.getAllByTestId('side-menu-items').map((elem) => elem.id);
+    expect(orderedSideMenuItems).toEqual(orderedDocumentItems);
```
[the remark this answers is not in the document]
teikjun: I agree. Sadly, I don't know of any way to write this test without using test IDs, so this is the best approach that I could find.

The other queries in RTL won't work because the relevant elements are mixed (one of the item appears as a `div` element while the rest are `section` elements). 

I also tried looking into custom queries. Ideally, I want to query for all the elements containing `id` attribute (so I can do the same thing as the original Enzyme test).  Strangely, I don't see any way to write the equivalent query in RTL. 

#### Closest attempt
Based on these sources:
- https://testing-library.com/docs/dom-testing-library/api-helpers/#custom-queries
- https://stackoverflow.com/a/53003981/11537371
```js
// Custom query, similar to getAllByTestId except that it looks for id attribute instead of data-testid
const getAllById = queryAllByAttribute.bind(null, 'id');
```

This is not exactly what we are looking for. This query requires a particular value as argument, and returns all elements whose`id` has that particular value (so it will only return one element). But, we want all elements containing `id` attribute regardless of its value. Querying by `id` attribute also goes against RTL's philosophy, so I think this isn't a good approach.

website/src/views/settings/SettingsContainer.tsx
teikjun: I'm not sure about how to handle the null case, so I left this issue unresolved for now

teikjun:
Thanks for the reviews! I've made the requested changes 🙂 

Some previous commits got mixed into this PR while I was rebasing, so I did a force-push to clean up the commits.

Sorry for the delayed response, I got caught up with ICT as well.

website/src/views/modules/ModulePageContent.test.tsx
teikjun: > Alternatively, renderWithRouterMatch could also be changed to make the path/location stuff optional (maybe just have a default path and location)

I followed this one!

website/src/views/modules/ModulePageContent.test.tsx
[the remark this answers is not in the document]
ZhangYiJiang: RTL returns a DOM element, right? So we can just do `const documentIds = querySelectorAll('h3[id]').map(ele => ele.id)`, and forget about test IDs. The feature this test is checking is "if I click on an item on the sidebar then I should navigate to that location in the document", but this is impossible to test directly since JSDOM doesn't simulate scrolling, so we're really testing "does each item on the sidebar have a matching ID in the document in the same order". We are directly testing implementation, but IMO this is fine rather than try to bend over backwards trying to force RTL to do something it cannot do.

ZhangYiJiang (commented):
Looks close - I can fix the settings page issue in another PR

teikjun:
Thanks! I've made the requested changes! :)

website/src/views/modules/ModulePageContent.test.tsx
```diff
@@ -1,21 +1,45 @@
...
-      component.find(ScrollSpy).prop('items'),
-    );
+    const { view } = make();
+    const { container } = view;
+    const sideMenuItems = screen
+      .getAllByRole('link')
+      .map((elem) => elem.textContent?.toLowerCase());
+    const documentIds = Array.from(container.querySelectorAll('[id]'))
+      .map((ele) => ele.id)
+      .filter((ele) => sideMenuItems.includes(ele));
```
ZhangYiJiang, line 42: /nit Would be a touch more efficient to convert `sideMenuItems` to a set before this (set.has is O(1), Array.includes is O(n))

ZhangYiJiang (approved):

website/src/views/modules/ModulePageContent.test.tsx
teikjun: Done! 😄

Issue #3032, opened 2020-12-07
Replace `makeBreakpoint` with `useMediaQuery`


taneliang:
`useMediaQuery` was added in #3007 to replace the `makeBreakpoint` higher order component. This issue tracks the removal of the deprecated `makeBreakpoint` from the codebase.

At the time of writing, there are 5 remaining components that use `makeBreakpoint`, 2 of which are function components. The remaining 3 class components will need to be replaced with function components first (similar to work done in #3007 and #3009).

This can be done over multiple PRs depending on your comfort level.

BransonNg:
Hi! Can i take a work on this :)

taneliang:
@BransonNg sure, thank you! Just leave out `VenuesContainer` for now as I'm reworking it.

Let me know if you have any questions!

BransonNg:
Hi @taneliang, was going through TodayContainer.test.tsx, and am having trouble testing the useEffect cycle as I'm unable to mount or render TodayContainer due to some Link components being rendered outside of a Router component.

Any suggestions as to how to test this?

Also related to the same test suite, any suggestions for checking the weather state? (i'm unaware of a way to access a component's useState to assert any tests)

taneliang:
Hey @BransonNg!

Sorry for the delay in replies, I'm having In-Camp Training these 2 weeks.

Would `renderWithRouterMatch` fix the Link stuff? https://github.com/nusmodifications/nusmods/blob/master/website/src/test-utils/renderWithRouterMatch.tsx Alternatively we could also manually wrap the `TodayContainer` in a `MemoryRouter` like we do with some other tests, since we don't actually need a route match.

Regarding the weather test, off the top of my head (I haven't looked at the test or the code yet) maybe we could try one of these approaches:

1. Use https://mswjs.io/ to mock the weather service. This approach is recommended by RTL's author.
1. Mock the Axios/fetch request for weather data (assuming we're using one of them). This would be a lot easier but there's a good chance that we'll want to dump Axios soon in favour of SWR or React Query, so I'd slightly prefer the MSW approach if it's not too complicated.

We can then just assert on the UI contents with a `find*` matcher.

BransonNg:
oh okay, i will look into that. Thanks @taneliang!

Also, would it be possible for me to add this dependency? 
https://github.com/immerjs/use-immer
Would like to use it in the FC in VenuesContainer to manage the state

taneliang:
[...]

PR #3042, opened 2020-12-10
Refactor to use FC and useMediaQuery [SideMenuComponent]

BransonNg:
## Context
#3032 
Replacing HOC MakeResponsive with hook useMediaQuery

## Implementation
Change of class component to functional component so that we can use the hook.


BransonNg:
Once this passes, I'll do a separate PR for all the other components using MakeResponsive at one go

website/src/views/components/SideMenu.tsx
```diff
@@ -14,66 +14,55 @@ type Props = {
...
-    disableScrolling(this.isSideMenuShown());
-  }
+  const isSideMenuShown = props.isOpen && !matchBreakpoint;
 
-  componentWillUnmount() {
-    // Force unset noscroll when unmounting so the user gets scrolling back if
-    // they navigate out of the parent component without closing the menu
-    disableScrolling(false);
-  }
+  const { isOpen, toggleMenu, children } = props;
```
ZhangYiJiang, line 35: There shouldn't be a need for this - just put these on line 29

website/src/views/components/SideMenu.tsx
```diff
@@ -14,66 +14,55 @@ type Props = {
...
 
-  isSideMenuShown() {
-    return this.props.isOpen && !this.props.matchBreakpoint;
-  }
+  useEffect(() => {
+    disableScrolling(isSideMenuShown);
```
ZhangYiJiang, line 38: This should have a cleanup function, otherwise scrolling will never be re-enabled if this component unmounts when the menu is open

ZhangYiJiang (changes requested):

website/src/views/components/SideMenu.tsx
BransonNg: Sure, latest commit addresses both issues! Thank u for the review :)

ZhangYiJiang (approved):

PR #3070, opened 2020-12-19
Remove eslint-loader and stylelint-webpack-loader

taneliang:
Since eslint-loader has been deprecated, I figured we should just take this opportunity to remove it and the corresponding stylelint plugin.

Because of eslint-loader or the replacement eslint-webpack-loader (I tried), `yarn start` took a really long time (on my computer: ~93s with eslint-loader, ~23s without).

Besides, eslint-loader's overlay appeared below the fast refresh overlay.

Since we have CI I think we should just use that to catch lint errors instead.


ZhangYiJiang:
Is it still possible to start these in watch mode? I personally don't use the loaders either but I think CRA has something similar included because it does make for better devex with new developers .Experienced developers tend to have their own preferences for how they want to deal with these, but new devs don't want to bother with setting them up_

ZhangYiJiang (approved):

taneliang:
> Is it still possible to start these in watch mode?

Hmm unfortunately it doesn't seem like it's possible without adding extra packages.

> new devs don't want to bother with setting them up

Agree. Personally I think it's fine if new devs don't set them up, since we can just catch them with CI. A new-dev friend also once told me years ago that the linter error overlays we used were really annoying because they kept getting in the way, so I'm not sure if the loader is doing more harm than good. Maybe if we disabled the overlays it would have been an even better devex.

Anyway I suspect CRA has them enabled for new devs who manage their own projects and thus don't have CI or other workflows/checks, so it seems to me that the value of these loaders is much clearer for CRA than it is for us.

PR #3071, opened 2020-12-19
Bump query-string and enable transpilation of it

taneliang:
query-string was previously frozen to v5 because it broke on Safari 9.
See https://github.com/nusmodifications/nusmods/pull/1053

We should just transpile it ourselves to support our browser matrix.


ZhangYiJiang (approved):

ZhangYiJiang:
Ah it looks like the typings changed. I wonder if this represents a change in the API or if the types just got better. Let me poke at it

taneliang:
Derp, didn't notice the CI failure. I can come back to this in a few days if it's too much trouble for you

PR #3092, opened 2020-12-28
Set up /website Vercel deployment

taneliang:
## Goal

Setting up Vercel now, which may be able to replace Netlify (though we're on the old free plan that has unlimited team members but only a meagre 2 build hours) and even our Docker website setup.

The commit history is rather messed up -- it'll be easier to review this just by looking at the diff. ~Also ignore the changes to the CircleCI config; that has been spun out to #3094 to reduce the mess in this PR.~ #3094 has been merged

## Breakdown

- [x] Fix Vercel build failure caused by Packtracker plugin.
- [x] Overhaul scripts/build.js to parallelize website and timetable-only builds. There's now a ~15% reduction in build time when built locally.
- [x] Nest timetable-only website within the main website's `dist` folder. This will allow one Vercel project to serve both projects. (Edit: done: https://nusmods-website-git-eliang-vercel.nusmodifications.vercel.app/timetable-only)
- [x] Configure caching headers.
	- [x] Move long-lived assets into an `assets` folder.
	- [x] Configure Vercel to provide long-lived caching headers for everything in `assets`.

	image

This should only be merged until after we can confirm that the export server can consume the new timetable-only bundle hosted by Vercel. (**EDIT:** it can! 🎉 see #3095)

## What's purposely *not* done

- We'll no longer proxy `/api` to `api.nusmods.com`. Visiting /api in the browser still works as React Router has a redirect configured, ~but it won't be redirected at the proxy level as imo it's unnecessary.~ /website will use api.nusmods.com directly. **EDIT:** we'll redirect /api to api.nusmods.com since it's convenient.

### DIFF
--- a/website/scripts/build.js
+++ b/website/scripts/build.js
@@ -21,20 +21,21 @@ function runWebpack(config) {
 
 /**
  * Print out errors.
- * @param {function} log
+ * @param {function} log console.log-compatible function used to print errors
  * @param {string} summary
- * @param {Error[]} errors
+ * @param {Error|Error[]} errorOrErrors
  */
-function printErrors(log, summary, errors) {
+function printErrors(log, summary, errorOrErrors) {
   log(chalk.red(summary));
   log();
+  const errors = _.castArray(errorOrErrors);
   errors.forEach((err, i) => {
     log(err);
     if (i !== errors.length - 1) {
       log();
     }
   });
-  console.log();
+  console.log(); // Use native console.log to force an unprefixed line after the error block
 }
 
 /**
@@ -99,7 +100,7 @@ async function buildProd(previousDistFileSizes) {
     log('File sizes after gzip:');
     printFileSizesAfterBuild(mainStats, previousDistFileSizes, parts.PATHS.build);
   } catch (err) {
-    printErrors(log, 'Failed to compile prod.', _.castArray(err));
+    printErrors(log, 'Failed to compile prod.', err);
     throw err;
   }
 }
@@ -117,7 +118,7 @@ async function buildTimetableOnly() {
 
     log(chalk.green('Compiled timetable-only successfully.'));
   } catch (err) {
-    printErrors(log, 'Failed to compile timetable-only.', _.castArray(err));
+    printErrors(log, 'Failed to compile timetable-only.', err);
     throw err;
   }
 }
--- a/website/scripts/build.js
+++ b/website/scripts/build.js
@@ -79,10 +79,6 @@ async function buildProd(previousDistFileSizes) {
   try {
     log(chalk.cyan('Creating build...'));
 
-    // Remove dist folders
-    await fs.remove(parts.PATHS.build);
-    log(`${parts.PATHS.build} has been removed`);
-
     // Build the browser warning bundle first so we can pass it to the main bundle
     const browserWarningStats = await runWebpack(browserWarning);
     handleErrors((...args) => console.log('prod (browser-warning):', ...args), browserWarningStats);
@@ -102,8 +98,6 @@ async function buildProd(previousDistFileSizes) {
     log(chalk.green('Compiled successfully.'));
     log('File sizes after gzip:');
     printFileSizesAfterBuild(mainStats, previousDistFileSizes, parts.PATHS.build);
-
-    log(`The ${chalk.cyan(parts.PATHS.build)} folder is ready to be deployed.`);
   } catch (err) {
     printErrors(log, 'Failed to compile prod.', _.castArray(err));
     throw err;
@@ -117,14 +111,11 @@ async function buildTimetableOnly() {
   const log = (...args) => console.log('timetable-only:', ...args);
   try {
     log(chalk.cyan('Creating build...'));
-    await fs.remove(parts.PATHS.buildTimetable);
-    log(`${parts.PATHS.buildTimetable} has been removed`);
 
     const timetableOnlyStats = await runWebpack(timetableOnly);
     handleErrors(log, timetableOnlyStats);
 
     log(chalk.green('Compiled timetable-only successfully.'));
-    log(`The ${chalk.cyan(parts.PATHS.buildTimetable)} folder is ready to be deployed.`);
   } catch (err) {
     printErrors(log, 'Failed to compile timetable-only.', _.castArray(err));
     throw err;
@@ -166,6 +157,10 @@ async function main() {
   // This lets us display how much they changed later.
   const previousDistFileSizes = await measureFileSizesBeforeBuild(parts.PATHS.build);
 
+  // Clear previous build
+  await fs.remove(parts.PATHS.build);
+  console.log(`${parts.PATHS.build} has been removed`);
+
   try {
     await buildFn(previousDistFileSizes);
   } catch (e) {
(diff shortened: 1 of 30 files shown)


ZhangYiJiang:
I wonder if we can replace this with the GitHub action or CircleCI orb - https://github.com/packtracker/report

taneliang:
Great idea! Let me try configuring a CircleCI Orb

taneliang:
So I think we can't use either the orb or the GitHub Action:

1. Both the orb and the action build the app directly from the Webpack config provided to it, independently of our build script. Apart from just being wasteful and requiring more CI time, our prod build script also requires the browser warning bundle to be built first.
1. The orb is erroring because it [tries](https://github.com/packtracker/report/blob/251139ec80c330f695587bd839614c1402cc7a22/orb.yml#L7) to look for a package.json *before* `cd`ing into our `project_root` 🤦 

I'll add the Webpack plugin back to our Webpack config after lunch.

taneliang:
We currently redirect /news to https://blog.nusmods.com in our Caddyfile; I'll add that to our vercel.json tomorrow, but this PR is ready for review.

website/src/config/app-config.json
```diff
@@ -2,7 +2,7 @@
   "brandName": "NUSMods",
   "defaultDescription": "NUSMods is a timetable builder and knowledge platform, providing students with a better way to plan their school timetable and useful module-related information that are community-driven.",
   "academicYear": "2020/2021",
-  "apiBaseUrl": "https://nusmods.com/api",
+  "apiBaseUrl": "https://api.nusmods.com",
```
ZhangYiJiang, line 5: Should add this as a DNS prefetch into either our website or our headers

website/vercel.json
```diff
@@ -0,0 +1,13 @@
+{
+  "headers": [
+    {
+      "source": "/assets/(.*)",
+      "headers": [
+        {
+          "key": "Cache-Control",
+          "value": "public, max-age=31536000, immutable"
```
ZhangYiJiang, line 8: I think 7 days (604800) is the most I would cache this for. We've definitely had bugs in this file before, and it would be awkward for this file to be borked for an entire year. Also I just took another look at the file and it's somehow 164kb (!?) will investigate

website/scripts/build.js
```diff
@@ -72,45 +93,68 @@ async function build(previousFileSizes) {
...
-  // Sync filename with `scripts/promote-staging.sh`.
-  fs.writeFileSync(path.join(parts.PATHS.build, 'commit-hash.txt'), `${commitHash.slice(0, 7)}\n`);
+async function buildAll(previousDistFileSizes) {
+  const results = await Promise.allSettled([
+    buildProd(previousDistFileSizes),
+    buildTimetableOnly(),
+  ]);
+  const errors = results.filter((r) => r.status === 'rejected').map((r) => r.reason);
+  if (errors.length > 0) {
+    throw errors;
```
ZhangYiJiang, line 132: A bit awkward since you're throwing an array not an Error, but not a huge issue

website/scripts/build.js
```diff
@@ -72,45 +93,68 @@ async function build(previousFileSizes) {
...
+  // This lets us display how much they changed later.
+  const previousDistFileSizes = await measureFileSizesBeforeBuild(parts.PATHS.build);
+
+  // Clear previous build
+  await fs.remove(parts.PATHS.build);
+  console.log(`${parts.PATHS.build} has been removed`);
+
+  try {
+    await buildAll(previousDistFileSizes);
+  } catch (e) {
```
ZhangYiJiang, line 149: The version of Node we're using should support this syntax 

```suggestion
  } catch {
```

website/scripts/build.js
```diff
@@ -18,20 +19,32 @@ function runWebpack(config) {
...
+function printErrors(log, summary, errors) {
+  log(chalk.red(summary));
+  log();
+  errors.forEach((err, i) => {
+    log(err);
+    if (i !== errors.length - 1) {
+      log();
+    }
   });
+  console.log();
```
ZhangYiJiang, line 37: This is to force an unprefixed line after the error block right? Might want to add a comment, it looks unintentional at first

ZhangYiJiang (approved):

website/scripts/build.js
taneliang: Yeah, I'm a bit desensitized to this now after seeing React throw \*promises\* 😆 

Figured it should be fine since we do want to keep track of multiple errors, and creating an `Errors extends Error` class seemed a bit overkill since these thrown errors are only used by this file.

website/vercel.json
taneliang: Actually hold on, all the files in the /assets/ folder have content hashes in their filenames. I think we potentially may have a problem if we accidentally produce unhashed files, but I think that'll be unlikely?

My intention with this 1 year cache is to help with our "Error loading chunk xxx" errors, by basically letting the CDN serve them even after they're outdated. Maybe 1 year is too long, but given that students tend to keep NUSMods open for a really long time I think 1 week may be a bit too short. Maybe a month?

website/src/config/app-config.json
taneliang: Never knew this existed! But because we already preload the module list endpoint and thus presumably already perform this DNS resolution very early, will this have any significant effect? I'm not too familiar with prefetching in general, but if the improvement is negligible I'm concerned that the cost of sending down that additional DNS prefetching line will negate any gains. Plus there's some (minimal) eng cost to adding more fields to our HTML Webpack Plugin config.

website/vercel.json
taneliang: I've changed it to a month. Let's discuss this again if there's a better choice :)

taneliang:
/news and /api now redirect to blog.nusmods.com and api.nusmods.com respectively, with a 307 status code. Merging now to unblock #3095, but we can continue discussing the unresolved comments

website/vercel.json
ZhangYiJiang: I'm not sure why you'd think the browser cache would help with the error? We've always kept the older versions of our assets on our current server and it never seem to help. But yeah the URL must contain hash, in which case a 1 year expiration date actually make sense and we can use `immutable` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Revalidation_and_reloading

website/src/config/app-config.json
ZhangYiJiang: I was thinking Vercel would have a config for that in the headers field

taneliang: Yeah we can add a header to our Vercel config, but I'm still not sure if it's worth it. It'll still be one extra line sent down to the client, plus now we'll have `api.nusmods.com` in multiple places in our codebase. Very minimal issues of course, but if the gains are negligible these may cost more

website/vercel.json
taneliang: I'm thinking the proxy cache/CDN (i.e. Vercel) can help with this. I'm not sure if our previous Caddy/Traefik set up was ever configured to continue serving cached versions of these files after they were deleted by the build script, but I doubt they were. I know the pre-Docker setup had all the old files sitting around, but I don't recall whether we had these errors then.

And yep this uses `immutable`. I'll bump the max-age back to 1 year in another PR

PR #3093, opened 2020-12-28
Rounder Add to Semester button

dlqs:
## Context

(This is a trivial visual change, feel free to ignore)

I was getting a little annoyed by half-round half sharp corners... so I made them totally round when there's only one semester.

Before
image
After
image
More than one sem: unchanged
image

## Implementation

The "Add to default Semester" button had sharp corners because it wasn't the last child in the button group. It wasn't the last child because of the dropdown buttons ("Add to other Semester") which appear when there's more than 1 semester. So I simply moved that logic into the `{!!otherSemesters.length ... }` block too.

### DIFF
--- a/website/src/views/components/module-info/AddModuleDropdown.tsx
+++ b/website/src/views/components/module-info/AddModuleDropdown.tsx
@@ -120,34 +120,39 @@ export class AddModuleDropdownComponent extends PureComponent<Props, State> {
               </button>
 
               {!!otherSemesters.length && (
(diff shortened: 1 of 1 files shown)


ZhangYiJiang (approved):
This looks great! We should also try to remove that double border on the dropdown button but that can go into another PR

taneliang:
Thanks @dlqs! This has been bugging me for years but we never got around to fixing it 😆

PR #3095, opened 2020-12-29
Turn export service into a Vercel serverless function

taneliang:
## Context

Stacked on #3092, part of work to dedockerize NUSMods to streamline our deployment processes.

## Observations

- Cold startup time can take a while (~6s?), but I think it's well worth it.

- No docs, but I think we should just update all the docs at once go once we've removed Docker.

- Not sure if local dev environment is working. I couldn't get `vercel dev` to work locally, but I think that may just be because I'm using WSL. We could just fix this in a future PR.

- The /export project has quite a bit of code that's not used, specifically the PM2 deployment and ability to load timetable data from the filesystem. With the move to serverless, all the router stuff will also not be needed anymore. I've left all the old code untouched in case this serverless stuff doesn't work out.

- If we do decide to keep this serverless setup for good, we'll want to consider inlining the /export project into /website. This will allow us to have a working export button in every deploy 😍 

## Biggest risks

- Not sure if we'll hit any pricing limits, but I guess we'll find out once real traffic comes in with the semester starting.

## Test URLs

- [Image export](https://nusmods-export-git-eliang-vercel-export.nusmodifications.vercel.app/api/export/image?data=%7B%22semester%22%3A2%2C%22timetable%22%3A%7B%22CS1010%22%3A%7B%22Sectional%20Teaching%22%3A%221%22%2C%22Tutorial%22%3A%221%22%7D%7D%2C%22colors%22%3A%7B%22CS1010%22%3A6%7D%2C%22hidden%22%3A%5B%5D%2C%22theme%22%3A%7B%22id%22%3A%22eighties%22%2C%22timetableOrientation%22%3A%22HORIZONTAL%22%2C%22showTitle%22%3Atrue%2C%22_persist%22%3A%7B%22version%22%3A-1%2C%22rehydrated%22%3Atrue%7D%7D%2C%22settings%22%3A%7B%22mode%22%3A%22LIGHT%22%7D%7D&pixelRatio=1)
- [PDF export](https://nusmods-export-git-eliang-vercel-export.nusmodifications.vercel.app/api/export/pdf?data=%7B%22semester%22%3A2%2C%22timetable%22%3A%7B%22CS1010%22%3A%7B%22Sectional%20Teaching%22%3A%221%22%2C%22Tutorial%22%3A%221%22%7D%7D%2C%22colors%22%3A%7B%22CS1010%22%3A6%7D%2C%22hidden%22%3A%5B%5D%2C%22theme%22%3A%7B%22id%22%3A%22eighties%22%2C%22timetableOrientation%22%3A%22VERTICAL%22%2C%22showTitle%22%3Atrue%2C%22_persist%22%3A%7B%22version%22%3A-1%2C%22rehydrated%22%3Atrue%7D%7D%2C%22settings%22%3A%7B%22mode%22%3A%22LIGHT%22%7D%7D
)
- 422 page: https://nusmods-export-git-eliang-vercel-export.nusmodifications.vercel.app/api/export/pdf
- 404 page: just visit any random path

## TODO after this PR

- Set up export.nusmods.com DNS records
- Update /website to use https://export.nusmods.com/api/export as the export base URL
- Point nusmods.com to Vercel 🎉 
- Docs


export/api/export/image.ts
```diff
@@ -0,0 +1,87 @@
...
+async function main(request: NowRequest, response: NowResponse) {
+  throwIfAcademicYearNotSet();
+  setUpSentry();
+
+  // Validate input before starting the browser (which is expensive)
+  let data = undefined;
+  try {
+    const dataCandidate = JSON.parse(request.query.data as never);
+    validateExportData(dataCandidate);
+    data = dataCandidate;
```
ZhangYiJiang, line 40: What is that `as never` cast? Isn't this the same as 

``` ts 
data = JSON.parse(request.query.data);
validateExportData(data);
```

export/api/export/image.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    if (e instanceof HttpError) {
+      if (e.code === 422) {
+        response.status(422).send(render422());
+        return;
+      }
+    }
+    response.status(500).send(render500(eventId));
+  }
+};
```
ZhangYiJiang, line 29: This should definitely be lifted to a higher order function, and we can probably deduplicate most of the two `main` functions - right now they are very similar

export/yarn.lock
```diff
@@ -286,6 +291,15 @@
   dependencies:
     "@types/node" "*"
 
+"@vercel/node@1.8.5":
```
ZhangYiJiang, line 294: Man, vercel node bundling it's own TypeScript is super awkward, especially when it diverge from our own version. It seems they forked ts-node for some reason https://github.com/vercel/vercel/blob/master/packages/now-node/src/typescript.ts

export/api/export/image.ts
```diff
@@ -0,0 +1,87 @@
...
+  try {
+    page = await render.open(url);
+  } catch (e) {
+    if (e.message.includes('ERR_CONNECTION_REFUSED')) {
+      throw new HttpError(
+        500,
+        `Could not open the page located at process.env.PAGE (${url}). Try opening it in your browser?`,
+        e,
+      );
+    }
```
ZhangYiJiang, line 73: Is this just for local development? If so a local env check + logger log might be better

ZhangYiJiang (commented):
Looks good as a direct port, but you might want to restructure the code a bit more since the code looks like it has a lot of copy-pasted chunks from the old Koa middleware code

export/api/export/image.ts
taneliang: Not necessarily, if the function was wrongly configured in prod or if nusmods.com goes down for some reason this will be shown too. But it sounds like we can improve this error message so that it'll make more sense in prod

export/yarn.lock
taneliang: Hmm this could be problematic if they use their version of TS in prod

export/api/export/image.ts
taneliang: Hmm okay.

I originally decided against that because I thought we'll want the handler file itself to perform all the response sending and error throwing, so that it's easier to see what's sent back to the client and also to trace the thrown error and where they're caught. So almost all the duplicated code just throws errors and creates responses.

But I drafted a hof and I think it'll still be easy to trace, so let's do that:

```ts
function makeHandler<T>(
  validateData: (request: NowRequest) => T,
  performExport: (page: Page, data: T, response: NowResponse) => void,
): NowApiHandler {
  // ...
}
```

taneliang:
@ZhangYiJiang Yeah, there's quite a lot of copy pasted code in this PR 😆 

I think apart from the handler code duplication discussed above, the other big copy paste is the render.ts -> render-serverless.ts file. I decided to fork it so that our handlers don't import `puppeteer`, and also so that it'll be easy to delete if/when we decide to drop the old Koa implementation.

I think part of this quick and dirty approach is also just to deploy it to prod and see how well it functions (pun not intended, but ha), before we really clean it up for real.

ZhangYiJiang (approved):
Fair - let's go for the MVP then and see how well this works. We should also think of some way of testing this in prod just to see if anything comes up, maybe route a small amount of traffic to this

taneliang:
I think we don’t need progressive rollout for this one. Since the new Vercel-hosted /website will be deployed by a DNS change, we can just change the DNS record back if export (or website) fails. Should be pretty painless. Also, yolo :D

export/api/export/image.ts
taneliang: Nope:

image

Figured we should cast to `never` instead of the ideal `string` type since we aren't actually sure.

export/api/export/image.ts
taneliang: I'm going to leave this message as is because I expect this to only be encountered if the function isn't set up properly. nusmods.com shouldn't ever have downtime unless Vercel goes down, and if that ever happens the serverless functions will probably be dead too.

export/yarn.lock
taneliang: Okay, thankfully they don't.

From the [build log](https://vercel.com/nusmodifications/nusmods-export/r3ub9qji4) (internal-only):

```
...
22:02:42.582  	Using TypeScript 4.1.2 (local user-provided)
...
```

taneliang:
Rebased on master and added [`071d895` (#3095)](https://github.com/nusmodifications/nusmods/pull/3095/commits/071d895bc2d7832f8913c3d664b5f6af6bb94be0).

Issue #3098, opened 2020-12-30
Migrate /website and /export from Docker to Vercel


taneliang:
## Context

About 1.5 years ago, we moved both projects to Docker (#1826). Our goal was to make it easier for new contributors to get started by simplifying their getting started steps to a simple `docker-compose up`.

My secondary vision was also to be able to start all of NUSMods' services with a single command, and even potentially have deploy previews that previewed the entire suite of services instead of just /website. This would have been useful as we intended to write a new GraphQL API server.

## Problems with Docker

While the `docker-compose up` experience was pretty sweet, it did not quite live up to its promises.

1. As of now, only /website and /export were moved to Docker. It turns out that porting the scraper, Matomo, and other services that live on our 1.nusmods.com droplet was just not easy or not worth the effort. Also, although our Elasticsearch server was [originally in our Docker setup](https://github.com/nusmodifications/nusmods/pull/1874), we [moved them to Elastic Cloud](https://github.com/nusmodifications/nusmods/pull/1937).
2. Because only 2 services were started, there wasn't much point using 
3. Deployment was much slower and messier. The script would build our services from scratch, a process which takes about 15 minutes, before doing a blue-green deployment. For a simple SPA and service, this seemed a little extreme.
4. Using Docker to serve the /website SPA was inefficient. We didn't really need an instance-level Caddy reverse proxy to serve them behind a machine-level Traefik reverse proxy to serve a simple static site, but we had them so that it would be easy to start multiple instances of NUSMods (because of our vision).

As a result, I think we should drop our unified Docker set up. We should definitely maintain Docker setups for each individual project, including project-specific docker-compose.yml configs.

## Solution

Moving forward, I think both Dockerized services can be moved to Vercel. Here's why:

### /website

/website is a simple SPA, and Vercel/Netlify have both been serving SPAs well for years. Although we have Netlify already set up, they've introduced build minute limits recently, forcing us to [work around them](https://github.com/nusmodifications/deploy-preview-bot). Vercel doesn't have these limits as they currently charge by team seats. Though, we're on a grandfathered Free plan with limits, so we may hit those limits. We'll see.

### /export

The biggest problem with /export is Puppeteer, as it's big and resource-heavy. Thankfully, with third party packages, it looks like it's still possible to run them in a serverless environment. The choice now is between the following serverless function providers:

- AWS Lambda/GCP: the OG serverless function providers, but they're a real pain to work with.
- Netlify Functions: seems good, with 125k free invocations per month.
[...]

PR #3099, opened 2020-12-30
Point /website timetable export button to Vercel serverless function

taneliang:
#3098.

1. The export buttons on /website now always work, because they use the export service running in prod :D
1. This allows the export server to operate independently of a local instance of /website.

URLs in the exported timetables point to a very long Vercel URL, but that's because the `PAGE` env var in prod isn't pointing to `nusmods.com` yet.

This is the last step before we can point nusmods.com to Vercel. We'll probably want to do a thorough test now :)

### DIFF
--- a/website/src/apis/export.ts
+++ b/website/src/apis/export.ts
@@ -9,7 +9,7 @@ export type ExportOptions = {
   pixelRatio?: number;
 };
 
-const baseUrl = '/export';
+const baseUrl = 'https://export.nusmods.com/api/export';
 
 function serializeState(
   semester: Semester,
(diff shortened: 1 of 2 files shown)


ZhangYiJiang (approved):

website/src/apis/export.ts
```diff
@@ -9,7 +9,7 @@ export type ExportOptions = {
   pixelRatio?: number;
 };
 
-const baseUrl = '/export';
+const baseUrl = 'https://export.nusmods.com/api/export';
```
taneliang, line 12: Actually this should probably be in app-config.json, but ah well. Let’s move that another time

PR #3103, opened 2020-12-30
Remove covid zoning from LocationMap

### LINKED ISSUE #3102: Stop displaying covid zoning on the venues map
### Describe the bug

NUS covid zoning is now lifted, so we should probably stop displaying it. We can still keep the data for now in case they somehow bring it back, but removing it from the UI is pretty easy

ZhangYiJiang:
## Context

Fixes #3102 - Remove covid zoning from the frontend since it is lifted 

## Implementation

Only removing the frontend parts for now - hopefully we won't have to go back but let's leave the backend stuff for now in case we do


taneliang (approved):
So glad

PR #3105, opened 2020-12-31
Truncate long exam duration string

dlqs:
## Context

Long repeating decimals in exam duration e.g. https://nusmods.com/modules/LC1025/singapore-law-in-context
image

Some exams are 70 or 100 minutes, which produce repeating decimals when divided by 60. Fortunately there are only 4 of these:
```
curl -X GET "https://api.nusmods.com/v2/2020-2021/moduleInfo.json" | jq '[ .[] | .semesterData | .[] | .examDuration | select(. != null and (. == 70 or . == 100)) ]' | jq length

4
```

After truncating (5 chars is rather arbitary, we can always change it):
image

## Other Information
happy new year LOL

### DIFF
--- a/website/src/utils/modules.test.ts
+++ b/website/src/utils/modules.test.ts
@@ -168,6 +168,7 @@ describe(renderExamDuration, () => {
     [45, '45 mins'],
     [60, '1 hr'],
     [90, '1.5 hrs'],
+    [70, '1.166 hrs'],
     [120, '2 hrs'],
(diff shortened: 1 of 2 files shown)


taneliang (commented):
Nice catch! Why not display them as minutes if they're not divisible by 30 instead?

dlqs:
oh yeah that sounds better actually

ZhangYiJiang (approved):

PR #3124, opened 2021-01-02
Remove unused deps from /website

taneliang:
Breakdown and details in commits.


ZhangYiJiang (approved):
lol this doesn't actually remove any dependencies since they are all transitively required by other packages. Still good for housekeeping purposes though

taneliang:
Yeah 😆 biggest surprise is that Express is required

PR #3125, opened 2021-01-02
Bump scraper deps

taneliang:
Details and breakdown in commit messages.

## Test plan

```sh
cd nusmods/scrapers/nus-v2

git checkout a9f0d # Commit checked out in prod
yarn
yarn dev all

cp -r data data-old

git fetch
git checkout eliang/bump-scraper-deps
yarn
yarn dev all

diff data data-old
```

To be run on 2.; scraping is failing now because the NUS API is throwing random errors 🤷


scrapers/nus-v2/src/services/logger/SentryStream.ts
```diff
@@ -71,20 +71,22 @@ export default function getSentryStream(config: StreamConfig = {}): Stream {
...
+            .mapValues((v) => String(v))
+            .value(),
+        );
+
+        scope.setExtras(
+          chain(record as { [key: string]: unknown })
+            .pick(extraProps)
+            .mapValues((v) => v ?? null)
+            .value(),
+        );
```
ZhangYiJiang, line 89: The typings for `_.chain` is pretty gnarly. I think it's better to just use the plain versions of these, especially since it's just two calls 

```suggestion
        scope.setTags(
          mapValues(
            pick(record as { [key: string]: unknown }, tagProps),
            (v) => String(v),
          ),
        );

        scope.setExtras(
          mapValues(
            pick(record as { [key: string]: unknown }, extraProps),
            (v) => v ?? null,
          ),
        );
```

ZhangYiJiang (approved):

taneliang:
So I guess this isn't very testable because the output JSON files don't have stable, sorted content, so diff outputs a ton of stuff. Let's just merge this?

PR #3131, opened 2021-01-02
Fix no scroll behavior on iOS

taneliang:
This PR fixes the no-scrolling behavior on iOS by replacing the no-scroll package, `.no-scroll`, and `disableScrolling` with the body-scroll-lock package.

## Implementation

- body-scroll-lock is about 0.5kB larger than no-scroll, but I think it's worth it.
- The enable/disableBodyScroll function calls are not extracted into a function/custom hook because they are subtly different.

## Other Information

Left = this PR, right = prod

https://user-images.githubusercontent.com/12784593/103456291-9f849500-4d2f-11eb-99c1-e82b9a3126d3.mp4


website/src/views/components/SideMenu.scss
```diff
@@ -60,7 +60,7 @@ $offset-top: $navbar-height + 1rem;
     $closed-menu-translation: calc(100% + 120px);
 
     width: 100%;
-    padding: 1rem 0 3rem;
+    padding: 3rem 0 3rem;
```
taneliang, line 63: This results in quite a big, ugly gap above the "Refine by" header, but if we don't do that the "Refine by" header will go below the search box on iOS (if you open the filters without scrolling). This issue did not exist previously because the old code kicked the navbar and search box to the top of the screen, which was far more jarring.

taneliang:
Merging to unblock nav bar overhaul

PR #3132, opened 2021-01-02
Fix tsconfig path lookup for IntelliJ

ZhangYiJiang:
IntelliJ's ESLint plugin suffers from a weird bug - it tries to resolve relative paths from the file that it is reading, even if the path is inherited from a config from another file. 

Screen Shot 2021-01-02 at 3 18 40 PM

I don't know if this is the correct behavior, and ESLint is just better at resolving upwards for tsconfig.json, but either way using __dirname changes the path to an absolute path which fixes the error

### DIFF
--- a/website/.eslintrc.js
+++ b/website/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
 
   parserOptions: {
-    project: './tsconfig.json',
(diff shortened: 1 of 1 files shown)


taneliang:
Wow thanks! This also fixed VS Code's ESLint error if the root folder wasn't /website. We should do this for the scraper too; I can do that later once we're done with the existing PRs -- I don't want to cause too many Renovate rebases

image

PR #3136, opened 2021-01-03
Minor build, lint, and CI fixes

taneliang:
Breakdown in commits. Redoes #3132; @ZhangYiJiang could you check that IntelliJ's behavior isn't broken?


ZhangYiJiang (approved):
Can confirm that the new config also works in IntelliJ

PR #3137, opened 2021-01-03
Introduce `NUSMODS_ENV` and fix deploy preview robots.txt and Sentry env

taneliang:
Introduces a new global `NUSMODS_ENV` to replace the existing `__DEV__` and `__TEST__`, because we have 3 different production-like environments. There are thus a total of 5 defined envs. This variable makes it easy to generate Sentry envs and to copy the correct static files.

Fixes 3 issues:

1. Netlify deploy previews have "development" Sentry envs:
	* Before: Sentry performance traces [have development env](https://sentry.io/organizations/nusmods/performance/summary/?project=213986&query=event.type%3Atransaction+transaction.duration%3A%3C15m+url%3A%22https%3A%2F%2F5ff0bc17fd243e3644913b7d--nusmods-deploy-preview.netlify.app%2Ftimetable%2Fsem-2%22&statsPeriod=24h&transaction=%2Ftimetable&unselectedSeries=p100%28%29) (internal only)
	* After: they [have preview env](https://sentry.io/organizations/nusmods/performance/v3:9608cda1d58c4b6b8f0374e27b0fd95b/?environment=preview&project=213986&query=event.type%3Atransaction+transaction.duration%3A%3C15m&statsPeriod=24h&transaction=%2Fcontributors&unselectedSeries=p100%28%29) (internal only)
1. Vercel and Netlify deploy previews don't have the correct robots.txt, causing latest.nusmods.com and other deploy previews to appear on google:
	* Netlify before (i.e. past PRs): https://5fcf94be0d264ddee36ab0f8--nusmods-deploy-preview.netlify.app/robots.txt
	* Netlify after (i.e. this PR): https://5ff1551df090071d9df141a9--nusmods-deploy-preview.netlify.app/robots.txt
	* Vercel preview before: https://nusmods-website-git-eliang-bump-scraper-deps.nusmodifications.vercel.app/robots.txt
	* Vercel preview after: https://nusmods-website-git-eliang-build-fixes.nusmodifications.vercel.app/robots.txt
	* Vercel production before (i.e. current nusmods.com): https://nusmods-website-1fxw49pja.vercel.app/robots.txt
	* Vercel production after: we'll only know after we deploy so... yolo
1. Reduces number of perf samples sent to Sentry in prod (but we'll still send all traces for previews and staging), as we've burst our quota. Internal-only link: https://sentry.io/organizations/nusmods/performance/?project=213986&statsPeriod=30d
	image


website/src/bootstrapping/sentry.ts
```diff
@@ -6,11 +6,18 @@ import { isBrowserSupported } from './browser';
 
 // Decide Sentry environment based on some basic heuristics.
 function sentryEnv(): string | undefined {
+  // Vercel deployments
```
ZhangYiJiang, line 9: Maaaybe `switch(VERCEL_ENV)`?

website/webpack/webpack.config.prod.js
```diff
@@ -80,19 +81,19 @@ const productionConfig = ({ browserWarningPath }) =>
...
-        (IS_CI || IS_NETLIFY) &&
-          new CopyWebpackPlugin({
-            patterns: [{ from: 'static-ci', context: parts.PATHS.root }],
-          }),
+        // Copy files from static folder over to dist
+        new CopyWebpackPlugin({
+          patterns: [
+            (IS_CI || IS_NETLIFY) &&
+              !IS_VERCEL_PROD && { from: 'static-ci', context: parts.PATHS.root },
+            { from: 'static', context: parts.PATHS.root },
```
ZhangYiJiang, line 94: We do need static-ci to copy over static. The docs don't say anything about the order in `patterns` though, I think it might be clearer to have two copies of the plugin?

ZhangYiJiang: https://webpack.js.org/plugins/copy-webpack-plugin/#flatten-copy suggests the ordering here is non-deterministic, so I think this is subjected to race conditions 

Also I suspect plugins may not run in order either - the first answer here https://stackoverflow.com/questions/41470771/webpack-does-the-order-of-plugins-matter suggest it does, but the other answer links to a GitHub issue where Webpack's maintainer says it's not a task runner and so you shouldn't rely on that 🤷‍♂️

ZhangYiJiang: One way to fix this is to ignore `robots.txt` from `static` (or more generally everything in `static-ci`) when `static-ci` is included. Awkward, but it should fix the issue regardless whether the patterns or plugins run in order

taneliang: > We do need static-ci to copy over static.

Oh, we do? I thought that apart from robots.txt, there are no filename clashes, so the order wouldn't matter

> ordering here is non-deterministic, and the source code uses Promise.all to run copies in parallel... Also I suspect plugins may not run in order either

Oh wow nice find. I didn't know that.

Maybe we could move the robots.txt files into another directory, then conditionally copy one of them in. I'm thinking instead of having static and static-ci folders we can have static/base, static/ci-only, static/robots

website/src/bootstrapping/sentry.ts
taneliang: I didn't really want to use a switch because it'll prevent us from mixing different conditions, e.g. maybe we'd want to add a domain host check after the vercel env === production line

website/webpack/webpack.config.prod.js
ZhangYiJiang: robots.txt is the file I'm worried about and the one that matters here, though we might have more in the future. It shouldn't be a bit too hard to write a tiny helper function that reads the content of each folder using `fs` and diffs them to construct a config that copies files over each other correctly

taneliang: Wouldn't it be clearer if we restructured the folders and wrote explicit conditions in the Webpack config?

I feel like a helper function may be an unnecessary amount of indirection for this. Plus it probably won't work if we want to have more complex conditions as diffing only works for 2 targets

ZhangYiJiang: Yeah I think the folders approach is fine. We hopefully shouldn't have too many files that differ for each deployment target

taneliang: > Main worry about the separate folder approach is that it's easy to accidentally add duplicate files since there's no checks

Yeah I agree. Is there an fs-linter that we can use to prevent this 😆

website/jest.common.config.js
```diff
@@ -15,15 +15,11 @@ module.exports = {
   },
   // Mimic the globals we set with Webpack's DefinePlugin
   globals: {
-    // Default to development
-    __DEV__: process.env.NODE_ENV === 'development',
-    __TEST__: process.env.NODE_ENV === 'test',
+    NUSMODS_ENV: 'test',
```
taneliang, line 18: This forces a test env in Jest tests, regardless of whether `NODE_ENV === "test"`.

website/src/entry/main.tsx
```diff
@@ -38,6 +37,6 @@ if (
   registerServiceWorker(store);
 }
 
-if (!__DEV__ && !__TEST__) {
+if (NUSMODS_ENV === 'production') {
```
taneliang, line 40: This logic is different -- this now only enables analytics on nusmods.com, i.e. we'll no longer have analytics in deploy previews and staging.

website/webpack/webpack.config.dev.js
```diff
@@ -74,9 +60,7 @@ const developmentConfig = merge([
       }),
       // Copy files from static folder over (in-memory)
       new CopyWebpackPlugin({
-        patterns: [
-          { from: 'static', context: parts.PATHS.root, globOptions: { ignore: ['short_url.php'] } },
```
taneliang, line 78: short_url.php wasn't in the static folder

website/webpack/webpack.config.prod.js
```diff
@@ -76,23 +66,27 @@ const productionConfig = ({ browserWarningPath }) =>
             };
           },
         }),
-        !IS_CI &&
-          new CompressionPlugin({
-            test: /\.(js|css|html|json|svg|xml|txt)$/,
```
taneliang, line 81: Vercel/Netlify/Nginx gzips for us, so this isn't necessary

taneliang (commented):
Merging since there aren't any further comments

PR #3139, opened 2021-01-06
Decode HTML entities in module description

### LINKED ISSUE #3100: &amp; in module descriptions
### Describe the bug
Some module descriptions render `&` as `&amp;` e.g. https://nusmods.com/modules/PC1222/fundamentals-of-physics-ii

image

### To Reproduce
I used `jq` to count modules with `&amp;` in the description:
```
curl -X GET "https://api.nusmods.com/v2/2020-2021/moduleInfo.json" | jq '[ .[] | select(.description|test("&amp;")) ]' | jq length 

18
```
Remove `| jq length` to see the actual modules

### Expected behavior

& appears as &

Manjaro, Firefox 84

### Additional context
I took a look at `/scrapers` and I don't think this is caused by any code there. Perhaps the NUS API is returning these `&amp;`s? If it's possible, I'm interested in fixing this. Perhaps this could be as simple as an additional `.replace("&amp;", "&")`?

dlqs:
## Context

Fixes #3100 

## Implementation

Adds a new package [he](https://github.com/mathiasbynens/he) and use it to decode. I added it to existing cleanModuleInfo.

More inline


scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
```diff
@@ -151,6 +151,28 @@ describe(cleanModuleInfo, () => {
...
+    ).toEqual({
+      acadYear: '2020/2021',
+      description: 'These concepts pertain to the structure of "ultimate reality"...',
+      title: 'Metaphysics',
+      department: 'Philosophy',
+      faculty: 'Arts and Social Science',
+      moduleCredit: '4',
+      moduleCode: 'PH2213',
+    });
+  });
```
dlqs, line 175: Let me know if this is too verbose - since we're really only testing the `.description` field anyways, I could squash this test into the one above

scrapers/nus-v2/src/utils/data.ts
```diff
@@ -46,6 +47,10 @@ export function titleize(string: string) {
   return capitalized;
 }
 
+export function decodeHTMLEntities(string: string) {
+  return decode(string);
```
dlqs, line 51: I'm using he here in non-strict mode. There is [strict mode](https://github.com/mathiasbynens/he#strict-1), which will decode any HTML text content even if it contains entities that cause parse errors. 

e.g. `&amp` turns into `&` in non-strict mode, but will raise an error in strict mode.

I believe it's alright to leave it in non-strict. 

(LMK if I should do strict mode, we probably want sentry logging and all, and I should move decodeHTMLEntities to https://github.com/nusmodifications/nusmods/blob/master/scrapers/nus-v2/src/tasks/GetSemesterModules.ts#L78 instead.)

scrapers/nus-v2/src/tasks/GetSemesterData.ts
```diff
@@ -93,6 +94,10 @@ export function cleanModuleInfo(module: SemesterModule) {
     cleanedModule.title = titleize(cleanedModule.title);
   }
 
+  if (typeof cleanedModule.description !== 'undefined') {
```
ZhangYiJiang, line 97: A bit shorter and more idiomatic way to check for null-like values 

```suggestion
  if (cleanedModule.description != null) {
```

scrapers/nus-v2/package.json
```diff
@@ -20,6 +20,7 @@
     "@types/bunyan": "1.8.6",
     "@types/deep-diff": "1.0.0",
     "@types/fs-extra": "9.0.6",
+    "@types/he": "^1.1.1",
```
ZhangYiJiang, line 23: Needed for Renovate bot 

```suggestion
    "@types/he": "1.1.1",
```

scrapers/nus-v2/package.json
```diff
@@ -53,6 +54,7 @@
     "date-fns": "2.16.1",
     "deep-diff": "1.0.2",
     "fs-extra": "9.0.1",
+    "he": "^1.2.0",
```
ZhangYiJiang, line 57: Ditto above. Also run `yarn` to fix the lock file after these changes 

```suggestion
    "he": "1.2.0",
```

ZhangYiJiang (commented):
Looks good, just some minor changes needed

scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
taneliang: I'm in favor of keeping it as a separate test so that it's not accidentally deleted.

Maybe we could make it less verbose by doing this?

```ts
    expect(
      cleanModuleInfo({
        acadYear: '2020/2021',
        description: 'These concepts pertain to the structure of &quot;ultimate reality&quot;...',
        title: 'Metaphysics',
        department: 'Philosophy',
        faculty: 'Arts and Social Science',
        moduleCredit: '4',
        moduleCode: 'PH2213',
      }),
    ).toHaveProperty('description', 'These concepts pertain to the structure of "ultimate reality"...');
```

scrapers/nus-v2/src/utils/data.ts
taneliang: Non-strict sounds good to me. We probably don't want the (admittedly small) burden of dealing with additional errors. Hopefully this HTML is generated by some spec-compliant rich text editor and not manually written by people hahaha

scrapers/nus-v2/src/tasks/GetSemesterData.ts
taneliang: Actually why not just

```suggestion
  if (cleanedModule.description) {
```

since we don't need to care about empty strings as well

ZhangYiJiang: Technically true, but I think `!= null` makes it clear this is a null check. But yeah it is functionally the same

scrapers/nus-v2/package.json
dlqs: fixed in 0bee168

scrapers/nus-v2/package.json
dlqs: fixed in 0bee168

scrapers/nus-v2/src/tasks/GetSemesterData.ts
dlqs: going with `!= null` in c0e6be4

scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
dlqs: changed in dd1380c

ZhangYiJiang (approved):

Issue #3035, opened 2020-12-08
Implement Module Preference Exercise form for NUS


taneliang:
Umbrella issue that tracks the addition of an MPE feature to our planner (https://nusmods.com/planner). Hard deadline in March, but we'd like to complete this by the start of AY20/21 Sem 2 (11 Jan 2021).

cc @chuabingquan @williamhutech

#### Must have

- [ ] UI/UX design (both planner (if necessary) and MPE form, and associated user flows) (@williamhutech)
- [ ] Integrate NUS Exchange
- [ ] Implement MPE form
- [ ] Implement MPE form submission
- [ ] Add logic to only show MPE form during MPE periods
- [ ] Bring Planner out of beta

#### Should have

- [ ] Implement importing of previous submission

#### Could have

- [ ] Improve existing Planner if necessary

### Technical unknowns

To coordinate with NUSIT.

- [ ] NUS Exchange integration details
- [ ] VFS API
- [ ] MPE JSON file specification

alissayarmantho:
I'll be doing 'Add an empty-ish MpeContainer component on a new nusmods.com/mpe route.'

alissayarmantho:
I'll be doing 'Add MPE page introduction content.'

alissayarmantho:
I'm going to add the MPE page UI

PR #3159, opened 2021-01-15
Setup empty MpeContainer

alissayarmantho:
## Context
 #3035
 
Add an empty-ish MpeContainer component on a new nusmods.com/mpe route.

## Implementation
 
Added an MpeContainer in `/website/src/views`. The MpeContainer is accessible on the browser via `/mpe`.

## Other Information
Is it possible to create a new branch for MPE work and transfer this pull request over to that branch?


chrisgzf:
Hmm @taneliang why is CI failing? I ran tests on her branch and it doesn't seem like any tests are failing o.o

website/src/views/routes/Routes.tsx
```diff
@@ -40,6 +41,9 @@ const Routes: React.FC = () => (
     <Route path="/contributors" component={ContributorsContainer} />
     <Route path="/apps" component={AppsContainer} />
 
+    {/* NUS' requested features routes */}
+    <Route path="/mpe" component={MpeContainer} />
+
```
taneliang, line 46: Nit: Probably not necessary to separate NUS features from the others. Could we move this up to above /tetris?

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -0,0 +1,13 @@
+import styles from './MpeContainer.scss';
```
taneliang, line 1: ESLint is erroring on this  line, as the `styles` variable is unused, and Stylelint is erroring because the .scss file is empty.

If you haven't done it, it may be a good idea to add stylelint and eslint plugins to your editor so that you'll automatically get warnings and errors while you're coding :)

taneliang:
> Is it possible to create a new branch for MPE work and transfer this pull request over to that branch?

Sure, if you want to. I think the higher level goals are to have frequent checkins so that we don't go too far down the wrong direction implementation-wise, to have small diffs so that it's easier and quicker to review, and also to keep your code updated with master so that you don't end up with giant merge conflicts at the end.

I would thus typically prefer working directly on our master branch, but I know it can be a bit slow sometimes. What do you think? One option I can think of is for you to work on your own fork instead of on our main repo. You can PR to yourself and cc me in your PRs haha. Also happy to hear any other suggestions

We could also work on a feature branch on this main repo but I don't really see the benefit of doing that if you can't push directly to it.

taneliang:
> I ran tests on her branch and it doesn't seem like any tests are failing o.o

@chrisgzf looks like both linters are erroring haha. You should be able to see it if you run `yarn ci` in /website

website/src/views/mpe/MpeContainer.tsx
alissayarmantho: Ok I've commmented it out

website/src/views/routes/Routes.tsx
alissayarmantho: Ok sure

website/src/views/mpe/MpeContainer.tsx
alissayarmantho: > > Is it possible to create a new branch for MPE work and transfer this pull request over to that branch?
> 
> Sure, if you want to. I think the higher level goals are to have frequent checkins so that we don't go too far down the wrong direction implementation-wise, to have small diffs so that it's easier and quicker to review, and also to keep your code updated with master so that you don't end up with giant merge conflicts at the end.
> 
> I would thus typically prefer working directly on our master branch, but I know it can be a bit slow sometimes. What do you think? One option I can think of is for you to work on your own fork instead of on our main repo. You can PR to yourself and cc me in your PRs haha. Also happy to hear any other suggestions
> 
> We could also work on a feature branch on this main repo but I don't really see the benefit of doing that if you can't push directly to it.

I guess we can just do it this way, I'll just PR to nusmods master directly if that's okay for you. Thanks!

alissayarmantho: I've solved the build error by removing the empty .scss and the import comment.

taneliang (approved):
Let's go! 🚀

PR #3165, opened 2021-01-16
Fix keyboard shortcuts snackbar text showing stale values

chrisgzf:
In #3007 there was a regression when we functionalized `KeyboardShortcuts`. Whenever we press <kbd>x</kbd> to toggle light/dark mode or <kbd>z</kbd>/<kbd>c</kbd> to cycle themes, the snackbar that pops out will always show the old value. This can be repro-ed in prod.

This is due to [our useSelector hook](https://github.com/nusmodifications/nusmods/blob/f019ae1445d24496e1113d56faadf8b79944fa00/website/src/views/components/KeyboardShortcuts.tsx#L36) always capturing the old stale value when displaying the notification snackbar, due to the rendering lifecycle of functional components.

To fix this, we directly access the redux store when we want to display the notification snackbars, to get the most current values.


website/src/views/components/KeyboardShortcuts.tsx
```diff
@@ -33,8 +33,7 @@ const KeyboardShortcuts: React.FC = () => {
   const [helpShown, setHelpShown] = useState(false);
   const closeModal = useCallback(() => setHelpShown(false), []);
 
-  const mode = useSelector(({ settings }: State) => settings.mode);
-  const themeId = useSelector(({ theme }: State) => theme.id);
+  const store = useStore();
```
taneliang, line 36: Is it possible to use this instead so that we don't have to keep casting the store type?

```suggestion
  const store = useStore<State>();
```

taneliang (approved):
Thanks @chrisgzf for fixing a bug I caused 😆

website/src/views/components/KeyboardShortcuts.tsx
chrisgzf: Woops, forgot that this exists. Fixed in https://github.com/nusmodifications/nusmods/pull/3165/commits/c282a2d31fe03787f1ee2bf9544d1c2453a29673.

taneliang: Thank you!

PR #3167, opened 2021-01-16
feat: add support for system color scheme

chrisgzf:
# Context

Very WIP, just the preliminary implementation first. PRing for sanity check.

Based on https://github.com/nusmodifications/nusmods/issues/1289.

## Implementation

Roughly:
 - Uses matchMedia to check for `prefers-color-scheme` media query
 - Adds a third default option to the mode selector
 - When it is set to default, AppShell checks for OS dark mode preference, and applies dark mode if there is dark mode preference
 - Whenever the user is in Default mode and presses <kbd>x</kbd> to toggle their mode, they have made a conscious choice to switch away from the Default, and it will be set to Light/Dark accordingly

Todo
- [x] Make it work with export service
- [x] Write tests

Note:
The keyboard toggling and setting of dark mode feels a bit laggy in dev, but isn't sluggish once exported. There is also incorrect snackbar notification text, which will be fixed in #3165.

Picture:
image

GIF:
ezgif-7-77d0500e1ca8


website/src/types/settings.ts
```diff
@@ -5,6 +5,8 @@ export type Theme = {
   readonly name: string;
 };
 
-export type Mode = 'LIGHT' | 'DARK';
+export type Mode = 'DEFAULT' | 'LIGHT' | 'DARK';
+
+export const DEFAULT_MODE: Mode = 'DEFAULT';
```
taneliang, line 10: Hmm, is there a better name for this variable? "Default mode" doesn't mean much out of context hahaha. I can't think of anything though

taneliang (commented):
This looks like it won't automatically update when the browser changes mode. Plus, having a reducer read data from outside feels like a violation of FP principles as the reducer is no longer pure.

Maybe we should implement this a different way. The Redux store should go back to being a store of the user's setting. We can do the theme switching in a new `useIsDarkMode` hook similar to [this `useBrowserTheme` hook](https://github.com/facebook/react/blob/95feb0e701a5ae20996e8cc6c4acd0f504d5985a/packages/react-devtools-scheduling-profiler/src/hooks.js#L24-L62), which can read from the user's setting and listen to changes to the browser's dark mode. This can be easily done using our existing `useMediaQuery` hook.

`AppShell` can then just do `const isDarkMode = useIsDarkMode()`.

website/src/types/settings.ts
chrisgzf: What about `DEFAULT_DISPLAY_MODE`? Or `OS_DEFAULT_DISPLAY_MODE`? Was actually thinking of refactoring all the Mode names to DisplayMode, i.e. `LIGHT_DISPLAY_MODE` or `DARK_DISPLAY_MODE` etc, so that it sounds less generic.

chrisgzf:
@taneliang Thanks for the review and sanity check! Agree on the reducer reading data from the outside being pretty iffy! I kinda know how to move on from here with your suggestions. I will slowly work on this over the next week.

website/src/types/settings.ts
taneliang: `DisplayMode` also sounds a bit vague though. I've looked up a bunch of potential alternatives:

- `DarkColorPreference`, inspired by [Electron's `nativeTheme.shouldUseDarkColors`](https://www.electronjs.org/docs/api/native-theme#nativethemeshouldusedarkcolors-readonly)
- `ColorScheme`, from the [CSS media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)
- `Appearance`, from macOS's System Preferences pane
- `UserInterfaceStyle`, from [Apple's APIs](https://developer.apple.com/documentation/uikit/uiuserinterfacestyle)
- `Theme`, from [Microsoft's APIs](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.application.requestedtheme?view=winrt-19041)

Unfortunately most of them also sound vague. Some also sound similar to our existing `Theme`, which I guess is fine as I blame `Theme` for being even more vague -- we should probably rename or remove it instead.

I think `DarkColorPreference` is the clearest, as this var and type will only be used to store the user's preference. The `useIsDarkMode` hook can return a boolean based on the preference + the CSS media query.

taneliang:
Btw we'll also need to display the segmented control better on `xs` 😆 

image

ravern:
I think if we just change the label to "Auto" (which is still telling) it looks fine.

image

This is on iPhone SE.

website/src/types/settings.ts
ravern: See global PR comment for what I've done to resolve this.

ravern:
I've rebased this onto the latest `master` and made some changes to how the "auto" light/dark mode works, taking into account @li-kai's comments too.

`Mode` is now split into two different concepts: `ColorScheme` and `ColorSchemePreference`.

- `ColorSchemePreference` is the user-selected preference in the settings. It is either "auto", "light" or "dark".
- `ColorScheme` is the final decided color scheme. It is either "light" or "dark".

The `useColorScheme` hook transforms the `ColorSchemePreference` stored in the Redux store into a `ColorScheme`, which is used to determine if we'll render light or dark mode palette. Should the user preference be set to "auto", I used a `useMediaQuery` hook to determine the system-preferred color scheme and return "light" or "dark" accordingly.

There was an issue where the behaviour of a `toggleMode`/`toggleColorScheme` action is now unclear — how do we toggle an "auto" preference if we don't have access to the external state i.e. the media query? To resolve this, I removed the action entirely, and changed the only dispatcher of this action — the 'x' keybinding handler — to use a combination of `useColorScheme` and `setColorScheme` instead. I also removed the corresponding tests.

Finally, regarding export, I refactored the export component to be a functional component instead, giving me access to `useColorScheme`, which I then used to determine the color scheme of the rendered output.

website/src/types/settings.ts
```diff
@@ -1,10 +1,29 @@
...
   readonly id: ThemeId;
   readonly name: string;
 };
 
-export type Mode = 'LIGHT' | 'DARK';
-export const LIGHT_MODE: Mode = 'LIGHT';
-export const DARK_MODE: Mode = 'DARK';
+/**
+ * Color schemes simply define whether the website is in light or dark mode. This
+ * is not to be confused with themese, which define the color palette of the website.
```
kokrui, line 14: nit: themese -> themes

website/src/types/settings.ts
```diff
@@ -1,10 +1,29 @@
...
+/**
+ * Themes are the "color palette" of the website. They define the set of colors
+ * being used across the website.
+ */
 export type Theme = {
   readonly id: ThemeId;
   readonly name: string;
 };
 
-export type Mode = 'LIGHT' | 'DARK';
```
kokrui, line 8: wow this is a much-needed refactoring LOL good stuff

website/src/views/settings/SettingsContainer.tsx
```diff
@@ -114,7 +114,7 @@ const SettingsContainer: React.FC<Props> = ({
               </p>
```
kokrui, line 114: it might be good to add the fact that "Auto" uses your system preferences in this protip here^

kokrui (approved):
LGTM and works on my machine! a few nits but overall ready to merge. good stuff!

PR #3177, opened 2021-01-17
website: css: Add fallbacks for props using env()

### LINKED ISSUE #3146: Venues page split pane layout broken on Edge 18
### Describe the bug

See screenshots

### To Reproduce

Open https://nusmods.com/venues

### Expected behavior

See screenshots

### Screenshots

| Edge 18 | Expected |
|-|-|
|image|image|

### Desktop:

- OS: Windows
- Browser: Edge 18

zwliew:
Pre-Chromium Microsoft Edge (<= v18) does not have support for
the env() function. Add fallbacks for the properties that use it.

Fixes #3146

## Implementation

Added fallbacks to all CSS properties which use the `env()` function.
I believe that the `env(safe-inset-area-*)` calls resolve to 0 on any device using Windows and Microsoft Edge. Hence, the fallbacks should be functionally equivalent to the originals.

### DIFF
--- a/website/src/styles/layout/site.scss
+++ b/website/src/styles/layout/site.scss
@@ -32,6 +32,7 @@ a {
 
 .main-container {
   flex: 1 auto;
+  padding: 0; // For Microsoft Edge 18
   padding: env(safe-area-inset-top) env(safe-area-inset-right) env(safe-area-inset-bottom)
     env(safe-area-inset-left);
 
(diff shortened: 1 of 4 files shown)


zwliew:
This is what I've come up with. I'm curious as to whether there are better ways of resolving this issue?

taneliang (approved):
Ah, so that's why! Thanks for investigating and solving this :D I didn't notice that the footer was also broken. This seems like the best solution to me too, and the comment should make it easy to delete the fallbacks when we eventually drop support for Edge 18.

Since you already have Edge 18 set up, do you want to also take a look at https://github.com/nusmodifications/nusmods/issues/3147?

PR #3178, opened 2021-01-17
webpack/babel: Properly transpile spread operator in react-leaflet

### LINKED ISSUE #3147: Venues page React Leaflet crash on Edge 18
Sentry Issue: [V3-395](https://sentry.io/organizations/nusmods/issues/2076142373/?referrer=github_integration) (internal only; [external link](https://sentry.io/share/issue/660b34ad456c4041b398e68806d21856/))

```
Error: Expected identifier, string or number
  at None (../node_modules/@react-leaflet/core/esm/pane.js:5:17)
```

### Steps to repro

1. In Edge 18, open browser console.
2. Navigate to https://nusmods.com/venues
3. See error.

image

### Additional info

It's possible that React Leaflet isn't being transpiled correctly. We use the same spread operator in our code and it hasn't caused any issues. The Sentry link shows the offending line of code.

Also, anyone fixing this may also want to fix #3146 at the same time :)

zwliew:
react-leaflet currently does not work on Microsoft Edge 18 (and possibly
Mobile Safari 11.2). A fix was attempted in commit 8e7d591af825e53b1cb199a9b59458264c4fc1a3.

However, the fix was incomplete for a few reasons:
1. The stated path did not reference the correct directory for
   react-leaflet.
2. There are 2 react-leaflet directories (react-leaflet and
   @react-leaflet) which use the spread operator.

Properly fix the issue by including the full path to both directories.

Note:
An alternate fix is to use regex again (e.g. /../node_modules\/react-leaflet/).
However, I used absolute paths according to the best practices stated in [1].

[1]: https://webpack.js.org/configuration/#:~:text=//%20-%20Use%20arrays%20of%20absolute%20paths%20in%20include%20and%20exclude%20to%20match%20the%20full%20path

Fixes #3147

### DIFF
--- a/website/webpack/webpack.config.common.js
+++ b/website/webpack/webpack.config.common.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 const parts = require('./webpack.parts');
 
 const commonConfig = {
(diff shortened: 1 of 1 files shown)


taneliang (approved):
Ah, so that's why! I didn't even know we had a @react-leaflet/core package. Thanks also for linking to those best practices! Should've looked at the docs when I wrote this 😆 

Thanks for the contributions, looking forward to more :D

PR #3155, opened 2021-01-14
Redesign nav bar, GlobalSearch, and dropdowns

taneliang:
## Summary

This PR redesigns our nav bar, global search, and all dropdowns throughout the app.

The design is inspired by FB5 (the recent facebook.com redesign) and @jiholim's NUSMods V4 mockups from 2018.

Because a huge number of things have been changed, it'll probably be easier to review this PR without looking at the existing code or behavior.

Fixes #3060 and resolves #1646 (no sidebar => no problem)

Thanks to @eugenood, @williamhutech, @chrisgzf, and @chuabingquan for the valuable input!

#### Why

We need space to put a lot more buttons in the future, including user account/info, and possibly an MPE button and planner button. The tab bar is full on mobile, and the top bar is full on desktop, leaving no good place to put auth info. This PR unifies both the mobile and desktop nav experience, and provides space on the right for logged-in user info + a dropdown menu for extra pages e.g. Contribute, Settings, MPE, etc.

#### Design Inspiration

FB5:

image

@jiholim's mockup:

image

GitHub:

image

#### What Changed

- Nav tabs on `sm`-width and above are moved to the nav bar.
- A nav dropdown has been added.
- AY week string has been moved into the dropdown, and an academic calendar link has been added.
- NUS Business link added.
- Global search is now always expanded, for simpler code and greater visibility.
- Centered page content, and tweak Today and Venues page layouts.

| |Before|After|
|-|-|-|
|320px (`xs`)|nusmods com_modules_CS1010E_programming-methodology|nusmods-website-git-eliang-new-nav nusmodifications vercel app_modules_CS1010E_programming-methodology|
|1024px (`sm`)|nusmods com_modules_CS1010E_programming-methodology (1)|nusmods-website-git-eliang-new-nav nusmodifications vercel app_modules_CS1010E_programming-methodology (2)|
|1440px (`xl`)|nusmods com_modules_CS1010E_programming-methodology (2)|nusmods-website-git-eliang-new-nav nusmodifications vercel app_modules_CS1010E_programming-methodology (3)|

Nav dropdown on xs

nusmods-website-git-eliang-new-nav nusmodifications vercel app_modules_CS1010E_programming-methodology (1)

#### Why

The existing dropdown did not look particularly good, and would have been weird as a nav dropdown. As I had designed a new one for the nav dropdown based on FB5's design (but it's also pretty common elsewhere), I just moved that into our global styles for consistency throughout the whole app.

[...]

PR #3197, opened 2021-01-26
Replace node-sass with Dart Sass

taneliang:
LibSass has been deprecated. Their maintainers [recommend](https://sass-lang.com/blog/libsass-is-deprecated) switching to Dart Sass instead.

### Test plan

1. `git checkout master`
1. Remove `contenthash` from `filename` and `chunkFilename` in `MiniCssExtractPlugin` config object in webpack.parts.js.
4. `yarn && yarn build`
3. `mv dist dist-nodesass`
3. `git checkout eliang/rip-node-sass`
4. `yarn && yarn build`
5. `diff dist dist-nodesass`: no changes in generated css files.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -17,7 +17,7 @@
     "test:integration": "jest --config jest.integration.config.js",
     "test:watch": "jest --watch",
     "e2e": "nightwatch -e default,safariMobile,safariDesktop,edge",
-    "ci": "cross-env NODE_ENV=production yarn lint && yarn typecheck && yarn test && yarn test:integration && npm rebuild node-sass && yarn build"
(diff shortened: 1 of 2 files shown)


taneliang:
Merging without approval to unblock #3155

ZhangYiJiang:
Is there a diff in the CSS output? I think that's the easiest way to check for regressions

taneliang:
I did try to diff the generated CSS, but it was minified so it wasn't useful 😆 Anyway I figured it was unlikely that anything would break so I didn't pursue it further. This is what I did:

```sh
# Remove `contenthash` from `filename` and `chunkFilename` in `MiniCssExtractPlugin` config object in webpack.parts.js.

git checkout master
yarn && yarn build

mv dist dist-nodesass

git checkout eliang/rip-node-sass
yarn && yarn build

diff dist dist-nodesass
```

PR #3206, opened 2021-01-27
Enable CORS on /api redirect

taneliang:
Fixes v2.nusmods.com, old deploy previews, and other old clients by changing the nusmods.com/api 307 temporary redirect to a 301 and adding CORS headers. I had to use 301 instead of 307 to fix CORS errors.

Header config copied from [Vercel docs](https://vercel.com/knowledge/how-to-enable-cors#enabling-cors-using-vercel.json).

### Test

- [test-cors.org with nusmods.com/api](https://www.test-cors.org/#?client_method=GET&client_credentials=false&server_url=https%3A%2F%2Fnusmods.com%2Fapi%2F&server_enable=true&server_status=200&server_credentials=false&server_tabs=remote): fails to load due to missing CORS headers
- [test-cors.org with deploy preview /api](https://www.test-cors.org/#?client_method=GET&client_credentials=false&server_url=https%3A%2F%2Fnusmods-website-git-eliang-api-redirect-cors.nusmodifications.vercel.app%2Fapi&server_enable=true&server_status=200&server_credentials=false&server_tabs=remote): loads

### DIFF
--- a/website/vercel.json
+++ b/website/vercel.json
@@ -31,8 +31,17 @@
       ]
     }
   ],
+  "rewrites": [
+    {
+      "source": "/api(/.*)?",
+      "destination": "https://api.nusmods.com$1"
+    }
+  ],
   "redirects": [
-    { "source": "/api(/.*)?", "destination": "https://api.nusmods.com$1", "permanent": false },
-    { "source": "/news(/.*)?", "destination": "https://blog.nusmods.com$1", "permanent": false }
+    {
+      "source": "/news(/.*)?",
+      "destination": "https://blog.nusmods.com$1",
+      "permanent": false
+    }
   ]
 }
--- a/website/vercel.json
+++ b/website/vercel.json
@@ -8,6 +8,27 @@
           "value": "public, max-age=2629800, immutable"
         }
       ]
+    },
+    {
+      "source": "/api(/.*)?",
+      "headers": [
+        {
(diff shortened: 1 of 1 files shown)


taneliang:
Merging this to do an unrelated test of Cloudflare edge caching on latest.nusmods.com (which I've just enabled, and which sits in front of Vercel's CDN). I'm trying to fix the issue where Vercel's CDN doesn't cache our static assets (at /assets) even though we've [set](https://github.com/nusmodifications/nusmods/blob/d473f78552eaf550e93d2300326cfb6523a38aea/website/vercel.json#L8) a correct Cache-Control header, which seems to be because Vercel [purges](https://vercel.com/docs/edge-network/caching#cache-invalidation) its cache on every deploy.

I couldn't find a way to disable this behavior, so putting Cloudflare in front of Vercel should allow Cloudflare to cache our assets long-term, even if the resource has been purged from Vercel. This should not affect the caching behavior for other resources because:

- /index.html: no changes since Vercel sets `s-maxage=0`, disabling all shared caches (Vercel and Cloudflare alike).
- All other non-/assets resources: no changes since Vercel sets `public, max-age=0, must-revalidate`, so no stale data will be served.

I'll also keep Vercel CDN enabled as it'll be easier to configure these caching headers in Vercel's config and we have a limited number of Cloudflare Page Rules.

Issue #294, opened 2016-08-17
NUSMods R


yangshun:
NUSMods v2.0 was launched in July 2014. 2 years have passed and with the emergence of ES6, React, Redux, Webpack and Babel, it's probably nice to incorporate some of these new tools into NUSMods and also fix some of the long-standing issues in the current NUSMods.

**Problems with the current NUSMods:**
- Timetable page (along with some other pages) is not mobile friendly. It's hard to change the timetable on mobile.
- Rendering uses a mix of jQuery and Backbone/Marionette Views. Ideally we shouldn't use jQuery at all and [views should be a function of the state](https://facebook.github.io/react/docs/thinking-in-react.html).
- Doesn't support lessons with non-half hour interval timings (there were lessons that end at 2359).
- Current approach of theming is not very scalable. A lot of overrides have to be done just for the dark mode.
- CSS is in a mess.
- No JS / CSS styles enforced
- No tests 😂 

**What the newer version can potentially include:**
- ES6, React, Redux, Webpack and Babel combo
- UI Library:
  - React:
    - Declarative UI makes the code easier to follow
    - Utilize virtual DOM diff-ing for efficient DOM updates
    - React Hot Loader makes development easier
- Flux State Management:
  - [Redux](redux.js.org):
    - The de facto flux choice out there
    - Single state tree - Which is very suitable because changes are made to the timetable from a few places (search bar, module page, timetable page)
  - Undo / redo functionality is easier to implement
- Build System
  - [Webpack 2.0](https://webpack.github.io/docs/roadmap.html)
    - Hot module reload
    - Code splitting (Many people do not need to access the other pages like Venues / News / Apps)
    - Tree shaking
  - [Brocoli](http://broccolijs.com/) and [Rollup](http://rollupjs.org/) are not mature enough yet at the moment.
- Babel
  - Cuz es6 is kool
- Mobile friendly editable timetable using CSS Flexbox
  - Proof of concept: http://v3.nusmods.com
- [Progressive Web App](https://developers.google.com/web/progressive-web-apps/) Features
  - Might be able to replace the need for a native web app
  - [Offline capabilities with service workers](https://developers.google.com/web/fundamentals/getting-started/your-first-offline-web-app)
- CSS
  - Use Bootstrap 4 or other CSS framework if suitable.
  - Adopt a CSS architecture ([OOCSS](https://github.com/stubbornella/oocss) / [BEM](http://getbem.com/) / [SMACSS](https://smacss.com/))
[...]

PR #3212, opened 2021-01-28
Fix modsn.us URL shortening script

### LINKED ISSUE #2057: Timetable share URL is no longer shortened
### Describe the bug

image

The URL should've been shortened to something like `https://modsn.us/<somehash>`.

### Cause

We have a [short_url.php](https://github.com/nusmodifications/nusmods/blob/master/infra/instance/short_url.php) script which mediates requests to our modsn.us URL shortener service. This was previously supported by Nginx. However, as #1937 changes the reverse proxy to Traefik, we need to add back PHP support for this to work again.

### Possible actions

1. Rewrite the script to not use PHP.
1. Add fastcgi support to our Traefik reverse proxy.

taneliang:
## Context

Implements a fix for #2057 behind a new `enableShortUrl` feature flag.

**NOTE:** This PR doesn't actually fix the issue for end users as the flag is disabled. See the TODO below for more info.

This PR reimplements the 6-year-old short_url.php in a Vercel serverless function using TypeScript. short_url.php had been lying unused since around the time we moved to a Dockerized setup (in #1937). The Dockerized setup was abandoned recently (#3098), and gives us the opportunity to deploy this in a serverless manner.

Also lays groundwork for adding more serverless functions to /website, required by #3035.

**TODO**: because the serverless function imposes a huge delay on URL shortening (it takes 2-3s), we'll want to display the unshortened URL immediately while the shortener loads. We can alternatively make it shorten on demand (i.e. a shorten button) or optionally (i.e. have a toggle). (**EDIT:** as this requires some UI work and iteration, we'll punt this to a future PR to unblock the urgent-ish #3035)

## Implementation

- TS used instead of PHP for consistency with the rest of NUSMods V3 (#294). short_url.php dates back to V1/2 times.
- Using TS also allows us to integrate Sentry.
- Using a serverless function so that we can version and deploy the script with the rest of /website.
- One downside is that we're now storing a secret in Vercel, so external PRs won't have automatic /website deployments on Vercel until one of us goes in to approve it.

In the future, we could investigate implementing our own URL shortener using Cloudflare Workers, or folding URL shortening into the new GraphQL server (#1726). YOURLS seems overcomplicated for what's essentially just a persisted dictionary.


ZhangYiJiang:
Curious about this living under `/website`. It seems logically these serverless functions should live under `/serverless` or something (and we can merge in https://github.com/nusmodifications/serverless-functions/). Does this enable better Vercel deployment? 

Also I can't believe I've not noticed the shorturl endpoint is literally just a proxy to `modn.us`

taneliang:
Unfortunately no. By default, Vercel expects serverless functions to be placed under an /api directory at the root of the project, and the Vercel nusmods-website project is rooted at /website. I think we could try to configure that, but iirc support for that was a little sketchy, and I'm pretty sure it won't accept functions located outside the project root.

Alternatively, we could create a new nusmods-serverless Vercel project rooted at a new /serverless directory. We'll just need to host it on a different domain, e.g. maybe serverless.nusmods.com or similar. I'm not sure if there's any benefit to doing that though.

> Also I can't believe I've not noticed the shorturl endpoint is literally just a proxy to `modn.us`

Yeah ikr, I realized that when working on Dockerizing it.

It also seems pretty dumb because that script/endpoint is publicly accessible, so it basically allows anyone to shorten a URL with modsn.us 🤷 I'm actually surprised nobody has abused it yet.

taneliang:
@nusmodifications/nusmods-developers This is ready for review. TL;DR I've fixed but disabled URL shortening as shortening is a bit too slow. But we need to merge this asap to unblock the MPE work (#3035)

website/package.json
```diff
@@ -125,6 +126,7 @@
     "@material/fab": "0.40.1",
     "@material/snackbar": "0.40.1",
     "@sentry/browser": "5.30.0",
+    "@sentry/node": "6.0.2",
```
ZhangYiJiang, line 129: This uses a different major version from our website lol

ZhangYiJiang (approved):
I think we should try to get the Sentry packages on the same version just to reduce the amount of duplication in node_modules, but otherwise this looks good. Also a nice test run for serverless functions.

website/src/featureFlags.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * See:
+ * - https://github.com/nusmodifications/nusmods/issues/2057
+ * - https://github.com/nusmodifications/nusmods/pull/3212
+ */
+export const enableShortUrl = false;
```
ZhangYiJiang, line 8: Some way to override this at runtime would be nice, maybe reuse the debug flags code by reading them from query params

taneliang: I think we should actually move all the `debug.ts` stuff into dynamic feature flags that check query params on module init, which will prevent pages that change the params (e.g. Venues, Module search) from flipping the debug flags. This can also replace `useGlobalDebugValue`, with the downside of losing the ability to flip flags after our JS loads.

I'll move this into another issue.

taneliang: https://github.com/nusmodifications/nusmods/issues/3215

website/package.json
taneliang: Yes... yes it does 😆 I didn't want the overhead of upgrading packages at this time but looks like v6 doesn't actually change any APIs. I'll bump all @sentry/* before merging.

PR #3217, opened 2021-01-29
Implement MPE and NUS Exchange serverless functions

taneliang:
## Context

#3035. Largely implemented by @chuabingquan, I'm just deploying it 😆 

## TODO

- [x] Ensure that we can actually make the service provider config public (it has no secrets, so... we should be able to)
- [ ] Manually test, or write automated tests using https://mswjs.io/ to mock NUS Exchange and VFS.

## Process

Similar to a React PR that I made, the goal of this import is to preserve history from https://github.com/chuabingquan/mpe-serverless (a private repo). Verified that `git log --follow` worked as expected on imported files.

#### Prepare mpe-serverless clone for merge

1. `git checkout -b importable`
1. Deleted repo configs
2. Inlined config.ts
3. Moved files into final locations
4. Commit

#### Import into NUSMods

```sh
cd ../nusmods

git remote add mpe-merge ../mpe-serverless
git fetch --all

git checkout -b mpe-functions
git merge --allow-unrelated-histories mpe-merge/importable
git remote rm mpe-merge
```

PR #3237, opened 2021-02-09
Mpe functions

chuabingquan:
## Context

#3217 

## Implementation

Instead of duplicating switch statements to handle method routing across various serverless APIs, this pull request introduces a new `Router` abstraction to encourage code reusability.


taneliang:
Merging without review on request

PR #3238, opened 2021-02-12
Mpeform (working draft)

alissayarmantho:
## Context

#3035 

Add the MPE page UI (draft)

## Implementation

Implementing the UI design and integrating the front-end to the back-end

image
image


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
ZhangYiJiang, line 1: I don't use VS Code. @taneliang are these settings useful to have?

website/src/types/mpe.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface UnrestrictedElective {
+  type: '04';
+}
+
+export type ModuleType = EssentialMajor | EssentialSecondMajor | Elective | UnrestrictedElective;
+
+export type MpePreference = {
+  moduleTitle?: ModuleTitle;
+  moduleCode: ModuleCode;
+  moduleType: ModuleType;
```
ZhangYiJiang, line 24: I think a better way to model this data is to represent type as a union of constant strings 
```suggestion
  moduleType: 'essential' | 'essentialSecondMajor' | 'elective' | 'unrestrictedElective';
```

You would only represent the types as objects themselves if you want to store additional info on type, but we don't need to in our case. You can then store data associated with each ModuleType in a separate Record 

``` ts 
interface ModuleTypeInfo {
  label: string;
  code: string;
}

const MODULE_TYPES: Record<MpePreference['moduleType']> = {
  essential: {
    label: 'Essential Major',
    code: '04',
  },
  // ...
};
```

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,105 @@
...
+  updatedConfig.headers.Authorization = getToken();
+  return updatedConfig;
+});
+
+mpe.interceptors.response.use(
+  (resp) => resp,
+  (err) => {
+    if (err?.response?.status === 401) {
+      removeToken();
+      return Promise.reject(ERR_SESSION_EXPIRED);
```
ZhangYiJiang, line 39: This technically works but it's a bit weird. Usually you'd do custom errors like this in JS 

``` ts 
class MpeSessionExpiredError extends Error {}

// To throw 
throw new MpeSessionExpiredError();

// To handle
catch (e) {
  if (e instanceof MpeSessionExpiredError) {
    // Do something special for MpeSessionExpiredError
  }
}
``` 

The main benefit here is that you can attach additional info to the thrown error object

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (token !== null) {
+      setToken(token);
+      params.delete(TOKEN_URL_QUERY);
+      history.replace({
+        search: params.toString(),
+      });
+    }
+  });
+  return getToken() !== null;
+};
```
ZhangYiJiang, line 58: I'm a little confused by this hook. If the intention is to parse out and store the token when the component is first rendered, I don't think this will work because on first render `useLayoutEffect` will not be called, so `getToken()` will return null.

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+      })
+      .catch((err) => {
+        // this is a temporary fix
+        // eslint-disable-next-line no-console
+        console.log(err);
+      })
+      .finally(() => {
+        setIsInitialLoad(false);
+      });
+  }, [props]);
```
ZhangYiJiang, line 36: `props` is an object, so this will always re-render. You probably want `props.getPreferences` instead

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,98 @@
...
+      throw err;
+    }
+  };
+
+  return (
+    <div className={styles.pageContainer}>
+      <div className={classnames(styles.innerContainer, 'col-md-8')}>
+        <header className={styles.header}>
+          <h2>Module Preference Exercise</h2>
+          <h4>AY2021/2022 - Semester 2</h4>
```
ZhangYiJiang, line 58: /nit Usually headings go in order, ie. you'd put h3 here because the previous heading is a h2. Although I think the previous heading should actually be a h1 since it is the title of the page

website/src/views/mpe/form/ModuleFormBeforeSignIn.tsx
```diff
@@ -0,0 +1,26 @@
...
+const ModuleFormBeforeSignIn: React.FC<Props> = (prop) => (
+  <div className={classnames(styles.container, 'col-md-6')}>
+    <div className={styles.image}>
+      <img src={mpePlaceholder} alt="Timetable" />
+    </div>
+    <h4>Start Module Preference Exercise</h4>
+    <p>Select your modules and we will automatically save your changes</p>
+    <button
+      type="button"
+      className={classnames('btn btn-outline-primary btn-svg')}
```
ZhangYiJiang, line 18:
```suggestion
      className='btn btn-outline-primary btn-svg'
```

website/src/views/mpe/form/ModuleMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <button
+            className={classnames('btn close')}
+            type="button"
+            onClick={() => toggleMenu()}
+            data-toggle="dropdown"
+            aria-haspopup="true"
+            aria-expanded={isOpen}
+          >
+            Module Type
+            <ChevronDown color="#6f6f6f" />
```
ZhangYiJiang, line 61: In general we try to set color using CSS instead

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  async function updateModuleType(moduleCode: ModuleCode, moduleType: ModuleType) {
+    if (!preferences.find((p) => p.moduleCode === moduleCode)) return;
+    setIsUpdating(true);
+    const previousPreferences = [...preferences];
+    const updatedPreferences = preferences.reduce<MpePreference[]>(
+      (acc, p) => [...acc, p.moduleCode === moduleCode ? { ...p, moduleType } : p],
+      [],
+    );
```
ZhangYiJiang, line 107:
```suggestion
    const updatedPreferences = preferences.map(
      (p) => p.moduleCode === moduleCode ? { ...p, moduleType } : p,
    );
```

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+    setPreferences(updatedPreferences);
+    try {
+      await props.updatePreferences(updatedPreferences);
+    } catch (err) {
+      setPreferences(previousPreferences);
```
ZhangYiJiang, line 112: Optimistic UI is a pretty neat idea, but I'm a bit worried here - if the user has made some additional change while `props.updatePreferences` is running, then this reset could discard those additional changes. I think it's better to handle errors by informing the user that their changes have not been saved due to some error, and provide a button for them to manually retry saving

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,105 @@
...
+import { useLayoutEffect } from 'react';
+import getLocalStorage from 'storage/localStorage';
+import { Location, History } from 'history';
+import NUSModsApi from './nusmods';
+import type { MpePreference } from '../types/mpe';
+import type { Module, ModuleCode } from '../types/modules';
+
+const SSO_PATH = '/auth/sso';
+const MPE_PATH = '/mpe/submissions';
+const TOKEN_STORAGE_KEY = 'nus-auth-token';
```
ZhangYiJiang, line 11: Import this from `keys.ts` instead of redefining the same constant here

ZhangYiJiang (commented):
I think this looks like a good start! Let me know if you need help with this

website/src/views/mpe/MpeContainer.tsx
alissayarmantho: ok

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,98 @@
...
-  placeholder: true; // Remove this when new props are added.
-};
+const MpeContainer: React.FC = () => {
+  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);
+  const isLoggedInOnLoad = useProcessLogin(useLocation(), useHistory());
+  const [isSessionExpired, setIsSessionExpired] = useState<boolean>(false);
+
+  useLayoutEffect(() => {
+    setIsLoggedIn(isLoggedInOnLoad);
+  }, [isLoggedInOnLoad]);
```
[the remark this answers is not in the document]
taneliang: @alissayarmantho btw updating state in effects should be avoided as you'll cause React to do another render. Doing a state update in `useLayoutEffect` is worse as it forces React to do its nested render during the commit phase.

In concurrent mode (which we aren't using yet, but we maaay be using it soon (#2914)), this nested render is de-opted and is executed in sync mode. In practice, this means that a concurrent mode app may see bigger stutters because the commit phase cannot be interrupted (unlike the render phase) and is now made longer because of the sync render happening within it.

website/src/views/mpe/form/ModuleForm.tsx
chuabingquan: Yep that makes sense! I've discussed with @taneliang about this and we decided that since we are auto-saving the user's changes, we would commit the latest request right after an ongoing request completes, discarding (not processing) the requests in-between the ongoing request and the latest request. Said changes are implemented in commits 6cf4530 and b467bc3.

website/src/apis/mpe.ts
chuabingquan: Yes, the intent is to parse out and store the token when the component is first rendered. We've changed `useLayoutEffect` to `useEffect` and it's working as expected. Would you require any further changes for this?

chuabingquan: Resolved by removing `useLayoutEffect`, `useProcessLogin` (now `getLoginState`) solely returns the current login state of the MPE page.

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,110 @@
...
+const removeToken = (): void => storage.removeItem(NUS_AUTH_TOKEN);
+
+const mpe = axios.create({
+  baseURL: '/api/nus',
+});
+
+mpe.interceptors.request.use((config) => {
+  const updatedConfig = { ...config };
+  updatedConfig.headers.Authorization = getToken();
+  return updatedConfig;
```
ZhangYiJiang, line 39: This does not immutably update `config` - `config.headers` is mutated. I'm not sure if that is important, but I'd suggest either this if immutability is not important 

```suggestion
  config.headers.Authorization = getToken();
```

or this if it is 

```suggestion
mpe.interceptors.request.use(produce(draft => {
  draft.headers.Authorization = getToken();
}));
```

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,110 @@
...
+export const fetchModuleDetails = (moduleCode: ModuleCode): Promise<Module> =>
+  axios.get<Module>(NUSModsApi.moduleDetailsUrl(moduleCode)).then((resp) => resp.data);
+
+export const getSSOLink = (): Promise<string> =>
+  new Promise((resolve, reject) => {
+    mpe
+      .get(SSO_PATH)
+      .then((resp) => resolve(resp.data))
+      .catch((err) => reject(err));
+  });
```
ZhangYiJiang, line 75: The Promise wrapper is unnecessary - the code below is equivalent 

```suggestion
    mpe
      .get(SSO_PATH)
      .then((resp) => resp.data);
```

website/src/apis/mpe.ts
```diff
@@ -0,0 +1,110 @@
...
+    })
+    .then((modules) => {
+      const moduleTypes = responsePreferences.map((p) => p.moduleType);
+      return modules.map<MpePreference>((m, index) => ({
+        moduleTitle: m.title,
+        moduleCode: m.moduleCode,
+        moduleType: moduleTypes[index],
+        moduleCredits: parseInt(m.moduleCredit, 10),
+      }));
+    })
```
ZhangYiJiang, line 93: /nit The moduleTypes map seems a bit unnecessary 

```suggestion
    .then((modules) => modules.map<MpePreference>((m, index) => ({
      moduleTitle: m.title,
      moduleCode: m.moduleCode,
      moduleType: responsePreferences[index].moduleType,
      moduleCredits: parseInt(m.moduleCredit, 10),
    })))
```

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
+  const getPreferences = (): Promise<MpePreference[]> =>
+    new Promise((resolve, reject) => {
+      getMpePreferences()
+        .then((preferences) => resolve(preferences))
+        .catch((err) => {
+          if (err instanceof MpeSessionExpiredError) {
+            setIsSessionExpired(true);
+          }
+          return reject(err);
+        });
```
ZhangYiJiang, line 41: As above, the `new Promise` is unnecessary 

```suggestion
    getMpePreferences()
      .catch((err) => {
        if (err instanceof MpeSessionExpiredError) {
          setIsSessionExpired(true);
        }
        throw err;
      });
```

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
-      <p>
-        Do take note that this is only a planning exercise;{' '}
-        <strong>it does not enroll you in the modules.</strong>
-      </p>
+  return (
+    <div className={styles.pageContainer}>
+      <div className={classnames(styles.innerContainer, 'col-md-8')}>
+        <header className={styles.header}>
+          <h1>Module Planning Exercise</h1>
+          <h4>AY2021/2022 - Semester 2</h4>
```
ZhangYiJiang, line 61: Should make a note to make this configurable

website/src/views/mpe/form/MpeFormContainer.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+  isLoggedIn: boolean;
+  isLoggingIn: boolean;
+  onLogin: () => Promise<void>;
+  getPreferences: () => Promise<MpePreference[]>;
+  updatePreferences: (preferences: MpePreference[]) => Promise<string>;
+};
+
+const MpeForm: React.FC<Props> = (props) => (
```
ZhangYiJiang, line 13: This component seems unnecessary. The content should just be inlined into `MpeContainer`

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
+    return getSSOLink()
+      .then((ssoLink) => {
+        window.location.href = ssoLink;
+      })
+      .finally(() => {
+        setIsGettingSSOLink(false);
+      });
+  };
+
+  const getPreferences = (): Promise<MpePreference[]> =>
```
ZhangYiJiang, line 32: Wrap this in `useCallback` - otherwise it will be called whenever this component re-renders since it is used in an useEffect below

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+            nextOperation = null;
+            processLastRequest(toExecute, onEnd);
+          }
+        });
+    }
+  };
+  return processLastRequest;
+};
+
+const processLastRequest = initProcessLastRequest();
```
ZhangYiJiang, line 49: This seems like an extremely roundabout way to get global state. Use `useState` or `useRef` instead. We can talk about this in chat

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        // eslint-disable-next-line no-console
+        console.log(err);
+      })
+      .finally(() => {
+        setIsInitialLoad(false);
+      });
+  }, [getPreferences]);
+
+  const onDragEnd = async (result: DropResult): Promise<void> =>
+    new Promise((resolve) => {
```
ZhangYiJiang, line 80: These probably won't work as expected because `resolve` is not used in a callback or in a promise chain. Again, we should talk about this in chat

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        />
+      </div>
+      <p className={styles.Status}>{isUpdating ? 'Saving...' : 'All changes are saved'} </p>
+      <Modal
+        isOpen={hitMaxModsLimit}
+        onRequestClose={() => setHitMaxModsLimit(false)}
+        shouldCloseOnOverlayClick={false}
+        animate
+      >
+        You are unable to add more than 7 modules in this exercise.
```
ZhangYiJiang, line 223: This is poor user experience. The interface should never allow the user to trigger this in the first place. Either disable the add module search box when the user has reached the limit, or we allow them to keep changing, but show an error where the saving status is and only submit when they go back down to 7 modules

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        return;
+      }
+      setIsUpdating(true);
+      fetchModuleDetails(moduleCode).then((moduleInfo) => {
+        const updatedPreferences: MpePreference[] = [
+          ...preferences,
+          {
+            moduleTitle: moduleInfo.title,
+            moduleCode,
+            moduleType: (Object.keys(MODULE_TYPES) as Array<MpePreference['moduleType']>)[0],
```
ZhangYiJiang, line 117: I think it's better to just hard code in `'01'`. The type checker will make sure the string value here is valid and it avoids this very ugly cast caused by `Object.keys` throwing away key typing

website/src/views/mpe/form/ModuleTypeMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+      value={props.type}
+      className={classnames('btn close', styles.menu)}
+      onChange={(e) => {
+        const modType = e.target.value as MpePreference['moduleType'];
+        props.updateModuleType(props.moduleCode, modType);
+      }}
+    >
+      {Object.entries(MODULE_TYPES).map(([moduleTypeCode, { label }]) => (
+        <option
+          key={props.moduleCode + moduleTypeCode}
```
ZhangYiJiang, line 27: `key`, unlike `id`, only needs to be unique between all children. Since each item has a unique `moduleTypeCode`, that is sufficient 

```suggestion
          key={moduleTypeCode}
```

website/package.json
```diff
@@ -139,12 +139,13 @@
     "core-js": "3.8.2",
     "date-fns": "2.16.1",
     "downshift": "6.0.15",
+    "history": "^5.0.0",
```
ZhangYiJiang, line 142:
```suggestion
```

react-router-dom uses history@^4.9.0. Importing this means that you'll import a different incompatible version. I think you can just `import type` without error, but if there is an error then just `eslint-ignore` it

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
+        <h4 className={styles.subtitle}>Overview</h4>
+
+        <p>
+          The Module Planning Exercise (MPE) is a project initiated by NUS to better understand
+          students’ demand for specific modules (as decided by the Module Host Departments) and
+          facilitate the Departments in their resource and timetable planning.
+        </p>
+        <p>
+          All current undergraduate and graduate students can{' '}
+          <strong>select up to 7 modules per semester.</strong> Do note that there are no validation
```
ZhangYiJiang, line 72: The number 7 should be extracted as a constant

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
+          updatePreferences={updatePreferences}
+        />
+        <Modal
+          isOpen={isSessionExpired}
+          onRequestClose={() => setIsSessionExpired(false)}
+          shouldCloseOnOverlayClick={false}
+          animate
+        >
+          Your session has expired. Please sign in again!
+          <br /> <br />
```
ZhangYiJiang, line 96:
```suggestion
          <p>Your session has expired. Please sign in again!</p>
```

website/src/views/mpe/MpeContainer.tsx
```diff
@@ -1,29 +1,110 @@
...
+          <button
+            type="button"
+            className={classnames('btn btn-outline-primary btn-svg', styles.ErrorButton)}
+            onClick={() => setIsSessionExpired(false)}
```
ZhangYiJiang, line 100: This should call `onLogin`, or at least provide an option to. It might also be better implemented as an `<div className="alert alert-warning">` instead of a modal (which would block the MPE instructions if the student wants to refer to it)

website/src/views/mpe/form/ModuleCard.scss
```diff
@@ -0,0 +1,94 @@
+.side {
+  align-self: center;
+  padding: 1rem 1.5rem 1rem 1.5rem;
+  font-weight: bold;
+  font-size: 18px;
+  text-decoration-line: #c0c0c0;
+}
+.modulecode {
```
ZhangYiJiang, line 8: /nit Naming convention for class names is camelcase

website/src/views/mpe/form/ModuleCard.scss
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.mctext {
+  padding: 0.1rem 0.5rem 0.1rem 0.5rem;
+  margin-bottom: 0;
+  border-radius: 0.35rem;
+  font-weight: bold;
+  font-size: 15px;
+  color: white;
+  background-color: #c0c0c0;
```
ZhangYiJiang, line 25: Gray text needs one of these - https://github.com/nusmodifications/nusmods/blob/master/website/src/styles/utils/css-variables.scss#L29. The grays need to change depending on whether it is dark or light mode

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+              )}
+            </Droppable>
+          </DragDropContext>
+        )}
+      </div>
+      <div className={styles.SelectContainer}>
+        <ModulesSelectContainer
+          moduleList={[]}
+          preferences={preferences}
+          semester={2021}
```
ZhangYiJiang, line 211: ?

website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+                  {provided.placeholder}
+                </div>
+              )}
+            </Droppable>
+          </DragDropContext>
+        )}
+      </div>
+      <div className={styles.SelectContainer}>
+        <ModulesSelectContainer
+          moduleList={[]}
```
ZhangYiJiang, line 209: ?

website/src/views/mpe/form/ModulesSelectContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const RESULTS_LIMIT = 500;
+
+function isModuleinPreference(preferences: MpePreference[], moduleCode: ModuleCode) {
+  return preferences.reduce<boolean>((acc, curr) => acc || curr.moduleCode === moduleCode, false);
+}
+function makeModuleList(state: StoreState, props: Props): ModuleSelectListItem[] {
+  return state.moduleBank.moduleList.map((module) => ({
+    ...module,
+    isAdded: isModuleinPreference(props.preferences, module.moduleCode),
```
ZhangYiJiang, line 32: Avoid using linear search for membership test, since linear search is O(n) while set membership test is just O(1) 

```suggestion
function makeModuleList(state: StoreState, props: Props): ModuleSelectListItem[] {
  const selectedMpeModules = new Set(props.preferences.map(p => p.moduleCode));
  return state.moduleBank.moduleList.map((module) => ({
    ...module,
    isAdded: selectedMpeModules.has(module.moduleCode),
```

website/src/views/mpe/form/ModulesSelectContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+type Props = OwnProps & {
+  preferences: MpePreference[];
+  addModule: (moduleCode: ModuleCode) => void;
+  removeModule: (moduleCode: ModuleCode) => Promise<void>;
+};
+
+const RESULTS_LIMIT = 500;
+
+function isModuleinPreference(preferences: MpePreference[], moduleCode: ModuleCode) {
+  return preferences.reduce<boolean>((acc, curr) => acc || curr.moduleCode === moduleCode, false);
```
ZhangYiJiang, line 27: Even if linear search is needed you should still only need `.some` which would short circuit when at least one match is found instead of scanning the whole array

website/src/views/mpe/form/ModulesSelectContainer.tsx
li-kai: Since preferences' length is 7, changing this to a set to complicate code is perhaps unnecessary. Will review in the future.

taneliang (approved):
Thanks for all the work @alissayarmantho @chuabingquan @KOOTSTHEHOOTS @ZhangYiJiang @li-kai @chrisgzf! 🎉

chuabingquan:
> Thanks for all the work @alissayarmantho @chuabingquan @KOOTSTHEHOOTS @ZhangYiJiang @li-kai @chrisgzf! 🎉 

Thanks William for helping to layout the foundations for the design! 🍾

PR #3241, opened 2021-02-16
Add support for MPE attributes

taneliang:
## Context

#3035

Following this email sent from NUSIT to @chuabingquan:

image

This PR adds support for the MPE course attribute to /scraper and /website.

cc @alissayarmantho

## Implementation

Though the MPE attribute has multiple possible values, I've flattened them into multiple separate boolean attributes in our API. Definitely not an ideal solution, but I think our Elasticsearch index expects the attributes to be a [list of strings](https://github.com/nusmodifications/nusmods/blob/4f8cf0cef3fd296021bd0d00f21f40481bce1a71/scrapers/nus-v2/src/services/io/elastic.ts#L144).

## Other Information

Not tested because it seems like no module in the NUS API has the MPE course attribute yet, so the new code isn't executed. @chuabingquan @alissayarmantho do yall know when NUS plans to add these attributes?

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterData.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterData.ts
@@ -57,6 +57,12 @@ const attributeMap: { [attribute: string]: keyof NUSModuleAttributes } = {
   // MPE handled separately as it maps to multiple attributes
 };
 
+const mpeValueMap: { [attribute: string]: (keyof NUSModuleAttributes)[] } = {
+  S1: ['mpes1'],
+  S2: ['mpes2'],
+  'S1&S2': ['mpes1', 'mpes2'],
+};
+
 export function mapAttributes(
   attributes: ModuleAttributeEntry[],
   logger: Logger,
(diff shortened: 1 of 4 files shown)


scrapers/nus-v2/src/tasks/GetSemesterData.ts
```diff
@@ -54,6 +54,7 @@ const attributeMap: { [attribute: string]: keyof NUSModuleAttributes } = {
   LABB: 'lab',
   ISM: 'ism',
   HFYP: 'fyp',
+  // MPE handled separately as it maps to multiple attributes
```
ZhangYiJiang, line 57: We could add a MPE map to make this a bit more declarative, what do you think? 

``` ts 
const mpeValueMap: { [attribute: string]: (keyof NUSModuleAttributes)[] } = {
  S1: ['mpes1'],
  S2: ['mpes2'],
  'S1&S2': ['mpes1', 'mpes2'],
};

// ... 

if (entry.CourseAttribute === 'MPE') {
  const mappedAttributes = mpeValueMap[entry.CourseAttributeValue];
  if (!mappedAttributes) {
    logger.warn(
      { value: entry.CourseAttributeValue, key: entry.CourseAttribute },
      'Non-standard course attribute value',
    );
  } else {
    for (const attr of mappedAttributes) {
      nusAttributes[attr] = true;
    }
  }
  continue;
}
```

chuabingquan:
@taneliang NUS have updated their modules with the new attributes. We can now test this PR with the for the following values:

**Available for Sem 1**
- AH3204
- AH3205
- BAA6002
- BBP6781
- BDC6112
- CS6285
- DBA4761

**Available for Sem 2**
- AS2237
- BDC6305
- BDC6307
- CS6240
- DBA3712

**Available for Sem 1 & 2**
- AH2101
- BMA5011
- BMA5104
- EBA5006
- EC1101E
- EC1301

scrapers/nus-v2/src/tasks/GetSemesterData.ts
taneliang: Beautiful

taneliang:
Tested locally, seems to work.

image

taneliang:
Merging without approval due to deadlines

PR #3245, opened 2021-02-23
Expose MPE attributes to mpeModules endpoint

chrisgzf:
Follow up of #3241.

MPE frontend requires filtering of all modules by whether they are participating in MPE for that semester. The way the frontend currently fetches module data means that we only know if the MPE attributes are present only after we view the module's detail page.

<s>
This PR exposes the MPE attributes to the `moduleList` endpoint, so that the frontend can fetch the MPE attributes of every module from the API.

Note: This is not ideal. This couples our `moduleList` endpoint to MPE, which is a NUS-specific thing. Might need further clean up in the future to extract the MPE data to:
1. A separate endpoint on the scraper
</s>

This PR creates a new endpoint  `mpeModules` which exposes only the modules that are participating in MPE, along with the semesters they are participating in.

### DIFF
--- a/scrapers/nus-v2/src/tasks/CollateModules.ts
+++ b/scrapers/nus-v2/src/tasks/CollateModules.ts
@@ -118,10 +118,15 @@ const getModuleCondensed = ({
   moduleCode,
   title,
   semesterData,
+  attributes,
 }: ModuleWithoutTree): ModuleCondensed => ({
   moduleCode,
   title,
   semesters: semesterData.map((semester) => semester.semester),
+  // Flag that marks if module is included in Semester 1's Module Preference Exercise
+  mpes1: attributes?.mpes1 ?? false,
+  // Flag that marks if module is included in Semester 2's Module Preference Exercise
+  mpes2: attributes?.mpes2 ?? false,
 });
 
 // Avoid using _.pick here because it is not type safe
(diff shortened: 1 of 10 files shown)


website/src/types/modules.ts
```diff
@@ -149,6 +149,10 @@ export type ModuleCondensed = Readonly<{
   moduleCode: ModuleCode;
   title: ModuleTitle;
   semesters: readonly number[];
+  // Flag that marks if module is included in Semester 1's Module Preference Exercise
+  mpes1?: boolean;
+  // Flag that marks if module is included in Semester 2's Module Preference Exercise
+  mpes2?: boolean;
```
chrisgzf, line 155: Doesn't have to be optional, since our scraper guarantees a boolean. I made it optional so that we don't have to add MPE information to our tests, where it will be coupled with the MPE implementation, for now.

scrapers/nus-v2/src/tasks/CollateModules.ts
```diff
@@ -118,10 +118,15 @@ const getModuleCondensed = ({
...
   semesterData,
+  attributes,
 }: ModuleWithoutTree): ModuleCondensed => ({
   moduleCode,
   title,
   semesters: semesterData.map((semester) => semester.semester),
+  // Flag that marks if module is included in Semester 1's Module Preference Exercise
+  mpes1: attributes?.mpes1 ?? false,
+  // Flag that marks if module is included in Semester 2's Module Preference Exercise
+  mpes2: attributes?.mpes2 ?? false,
```
ZhangYiJiang, line 129: Do not add these if they are false. Including these properties when they are unnecessary adds a lot of additional bytes

chrisgzf: Ah right, great point.

taneliang (commented):
Tested locally. Looks great! I'll merge and deploy

taneliang (approved):

PR #3246, opened 2021-02-25
Add title and module credit to MPE endpoint

ZhangYiJiang:
## Context

Since the only thing we need for the frontend are the module title and the number of MCs we can push all these data into the list to simplify the frontend. We are loading the module title twice - might be a good idea to look it up via moduleList on the frontend instead? Might not be worth it if it makes the frontend more complicated though


taneliang (approved):
Thank you! This is a great idea

PR #3250, opened 2021-02-27
Fix bug where module card sometimes renders weird on iOS

chuabingquan:
## Context

1. On some iOS devices, the module card component's styling is broken.
2. Intended MCs input field will always have one character, leading to frustrating experience when changing values.

## Implementation

Broken module card styling on iOS
Broken module card styling on iOS

### DIFF
--- a/website/src/views/mpe/form/ModuleForm.tsx
+++ b/website/src/views/mpe/form/ModuleForm.tsx
@@ -115,12 +115,8 @@ const ModuleForm: React.FC<Props> = ({
     });
   };
 
-  // TODO: Remove leading/padded zero for the intended MCs to take field.
   const updateIntendedMCs = (moduleCredits: number) => {
-    if (Number.isNaN(moduleCredits)) {
-      return;
-    }
-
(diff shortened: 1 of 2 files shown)


website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -164,7 +160,7 @@ const ModuleForm: React.FC<Props> = ({
             min="0"
             inputMode="numeric"
             className="form-control"
-            value={intendedMCs}
+            defaultValue={intendedMCs}
```
taneliang, line 163: This solution isn’t great as it turns this controlled component into an uncontrolled one. This means that there are now 2 sources of truth: the DOM, and the component state (or I guess wherever `updateIntendedMCs` saves to), and there’s only a 1 way data flow from the DOM to the component, so they can get out of sync.

It’ll be better if we validated and transformed the MC value on blur instead

taneliang (approved):
Strange that only you see this, but let's get this in and hope it doesn't cause any issues :)

PR #3251, opened 2021-02-27
Change Input MC Text field Description to be less confusing

alissayarmantho:
## Context
 #3238 

Many users think the MC textbox is linked to the number of modules you can select in the exercise.

## Implementation

Change the description of the Textfield for intended MC taken in ModuleForm.tsx


website/src/views/mpe/form/ModuleForm.tsx
```diff
@@ -156,7 +156,8 @@ const ModuleForm: React.FC<Props> = ({
     <div className={styles.formContainer}>
       <label className={classnames('row', styles.mcTextField)}>
         <div className="col-sm-8">
-          Please indicate how many MCs you are planning to pursue this semester:
+          Intended number of MCs (Note that this does not affect how many modules you can select in
+          this exercise) :
```
taneliang, line 160:
```suggestion
          Intended number of MCs (does not affect how many modules you can select):
```

taneliang (approved):

taneliang:
Yeah, unfortunately this field is extremely confusing.

PR #3252, opened 2021-02-28
Improve intended MCs input field empty state

chuabingquan:
## Context

In the existing intended MCs input field, there will always be at least one digit left to ensure that autosave would update with at least a numerical value. However, this constraint has caused some frustrating experience, e.g. if the current value in the input is `1`, someone who wants to type in `20` MCs would have to add `20` after `1` first (resulting in `120`), then navigate backwards to remove `1`.

## Implementation

A new state is introduced to maintain the intended MCs input field state separately. When user backspaces all the way in the field, the field will reflect the changes but not trigger the autosave. If a user enters a valid numerical value however, the main `intendedMCs` state will be updated, along with the API call that follows. If a user leaves the input field empty and clicks elsewhere, the input field will automatically repopulate their latest intended MCs count from the `intendedMCs` state.

## Other Information

NIL

### DIFF
--- a/website/src/views/mpe/form/ModuleCard.scss
+++ b/website/src/views/mpe/form/ModuleCard.scss
@@ -79,7 +79,6 @@
     flex-direction: column;
     align-self: center;
     padding-left: 0;
-    align-self: center;
   }
 
   .moduleType {
--- a/website/src/views/mpe/form/ModuleCard.scss
+++ b/website/src/views/mpe/form/ModuleCard.scss
@@ -78,6 +78,7 @@
   .container {
     flex-direction: column;
     padding-left: 0;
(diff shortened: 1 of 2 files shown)


website/src/views/mpe/form/ModuleCard.scss
```diff
@@ -79,6 +79,7 @@
     flex-direction: column;
     align-self: center;
     padding-left: 0;
+    align-self: center;
```
taneliang, line 82: Gotta revert this

```suggestion
```

taneliang (approved):
Too tired to review properly, but seems good

taneliang:
Manually poked, seems to work.

PR #3261, opened 2021-03-03
MPE: Add FAQ and fix crashing module details fetch

taneliang:
## Context

#3035. Details in email and our private NUSMods Telegram group (sorry everyone else! We're trying to ship this asap. Would be happy to explain the context if you comment)


taneliang:
Verified to work:

Original crash repro steps:

1. Add MA2002
2. Reload
3. Observe crash as MA2002.json doesn't exist in our modules

Issue #3202, opened 2021-01-26
Prerequisite tree right-to-left progression is confusing


taneliang:
### Is your feature request related to a problem? Please describe.

Some module pages ([example](https://nusmods.com/modules/YCC2137/scientific-inquiry-2%23prerequisites)) have prerequisite trees, where a given modules' dependents and prerequisites are displayed on the module's left and right respectively. However, this is counter-intuitive to people who expect a left to right progression.

![30FD5527-AFF8-46D4-98BD-D78081C8796E](https://user-images.githubusercontent.com/12784593/105841738-1439af00-6010-11eb-8278-460ad16c7ca6.jpeg)

### Describe the solution you'd like

We should probably flip the tree to show dependents on the right.

### Describe alternatives you've considered

@li-kai any idea why we did it this way?

li-kai:
I think the original thinking is people want to know the dependencies rather than the dependents. You're right, it's a little confusing to display it this way. Top to bottom is more intuitive but it's hard to do in css.

sciffany:
I'd like to take on this issue please!

sciffany:
> I'd like to take on this issue please!

This is harder than I thought

alvynben:
hey @schalkneethling do you mind if I take this up? I think I've created a working version with reverse ordering.

alvynben:
I've created a pull request for this issue above.

chrisgzf:
Hey @alvynben, thanks for the contribution! However since @sciffany already has a working implementation in #3263, we'll review her version first. (I think there was a miscomm somewhere above.) If you'd like to, you can review her PR too! I'm going to close your PR first to avoid confusion (we can reopen it again if the need arises). Sorry about that!

zehata:
I have taken this issue and opened a PR, but I would favour closing this issue and creating a new one for the same but with an option in the settings that allow users to choose their preference

PR #3263, opened 2021-03-05
Reverse Prerequisite Tree

sciffany:
## Context
Solves #3202 

## Implementation
Use CSS to reverse formatting of Prerequisite Tree

Result

Before
image

After
image

## Other Information
I renamed some variables from `prereq` to `depedent`, as they are mostly used to refer to the dependent modules (e.g. CS3247 above)


chrisgzf:
for ref: alternate implementation in https://github.com/nusmodifications/nusmods/pull/3291 but we'll review this PR first @taneliang

website/src/views/modules/ModuleTree.scss
```diff
@@ -4,6 +4,7 @@ $connector-size: 0.75rem;
 
 .container {
   display: flex;
+  flex-direction: row-reverse;
```
li-kai, line 7: For accessibility reasons, we should not use row-reverse here. 
I know it's a lot more work to flip the dom order, but I think it's worth it to prevent confusion for our fellow students with visual disabilities.

li-kai (commented):
Thanks for working on this! I really like the renaming, it definitely makes it easier to understand!

zwliew:
Thanks a lot for this -- it served as a basis for a more complete implementation in #3842! Since that has already been merged, I am closing this.

PR #3268, opened 2021-03-14
Disable MPE

taneliang:
## Context

#3035

To be deployed to prod midnight 15 March SGT


taneliang:
Verified MPE is closed and the announcement banner is no longer shown. Merging so that @li-kai can deploy at midnight (thanks! :D)

image


image

PR #3282, opened 2021-05-30
Add semesters 3 and 4 to semesters with exam available

taneliang:
As requested by Ai Lin from RO (see internal email to mods@)


taneliang:
Tested and works. Merging

image

PR #3295, opened 2021-06-27
Fix query parameters being removed on module search retry

### LINKED ISSUE #3288: "Click here to try again" button deletes the 'q' query parameter
### Describe the bug

When there is a network/server issue and we can't load modules, the "Click here to try again" button deletes our query, although all filters are retained.

### To Reproduce

1. Search for a module at a busy time or somehow prevent connection to the server, so that the "Oh no... We can't load the module information" page appears (for example, we might go to `https://nusmods.com/modules?q=geh1042&no-exam=true`)
2. Click on the "Click here to try again" button.
3. Observe that the 'q' parameter of the query is deleted (so we are sent to `https://nusmods.com/modules?no-exam=true`).  This also happens if there were no additional selection parameters after the 'q' parameter.

### Expected behavior

Don't clear the 'q' parameter.

### Screenshots

https://user-images.githubusercontent.com/6948096/122679513-bf484080-d21d-11eb-8966-b6303ad75ea2.mp4

- OS: Windows 10
- Browser: Chrome
- Version: 91.0.4472.106

chrisgzf:
## Context

Closes #3288

searchkit provided a default resetSearchFn for our ES modules search to reset the search. However, we used the resetSearchFn to retry the search instead, which is incorrect. This causes the query params to be cleared when the user attempts to retry the search.

Did this with @alissayarmantho - we were debugging together so the commits are tagged with her too!

## Implementation

We fix this by using the actual searchkit function to retry the search.

### DIFF
--- a/website/src/views/errors/ModuleFinderApiError.test.tsx
+++ b/website/src/views/errors/ModuleFinderApiError.test.tsx
@@ -1,0 +1,53 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { SearchkitManager } from 'searchkit';
+
+import renderWithRouterMatch from 'test-utils/renderWithRouterMatch';
+import ModuleFinderApiError from './ModuleFinderApiError';
+
+jest.mock('views/components/RandomKawaii');
+
+const mockedUrl = 'https://someEsUrl.com';
+const CANONICAL = '/modules?q=GER1000&sem[0]=1&sem[1]=2&sem[2]=3&sem[3]=4';
+
+const searchkit = new SearchkitManager(mockedUrl, {
+  searchUrlPath: '_search?track_total_hits=true',
+});
+
+describe(ModuleFinderApiError, () => {
+  test('should render', async () => {
+    const { container } = render(<ModuleFinderApiError searchkit={searchkit} />);
+    expect(container).toMatchSnapshot();
+  });
+
+  test('should render with url', async () => {
+    const { view } = renderWithRouterMatch(<ModuleFinderApiError searchkit={searchkit} />, {
+      path: '/modules',
+      location: CANONICAL,
+    });
+    const { container } = view;
+    expect(container).toMatchSnapshot();
+  });
+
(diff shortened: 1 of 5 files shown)


website/src/views/errors/ModuleFinderApiError.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+        location: CANONICAL,
+      },
+    );
+    expect(history.location.pathname).toBe('/modules');
+    expect(history.location.search).toBe('?q=GER1000&sem[0]=1&sem[1]=2&sem[2]=3&sem[3]=4');
+
+    const { getByRole } = view;
+    searchkit.history = {
+      push: jest.fn(),
+    };
```
ZhangYiJiang, line 46: If you're replacing this with a mock function, would the `expect(history.location)` below ever fail? 

Also patching the object directly worries me a little since it kinda looks like you're updating a global shared object. I think the Jest runner makes it safe but this might be a bit better (spyOn also executes the original function which I think is what we want here) 

``` ts 
const push = jest.spyOn(searchkit.history, 'push');
``` 
const

ZhangYiJiang (commented):
Changes look correct, the test looks a bit sketchy though. 

I can't believe I had to dig into the SearchKit source code to confirm that the `errorComponent` prop also takes in React elements - the documentation doesn't mention it at all >_<

website/src/views/errors/ModuleFinderApiError.test.tsx
chrisgzf: Update on this: when I was writing this test, there were some struggles trying to get searchkit to use the right `history` object. It's also pretty weird that I can't really make an assertion on `history.push`. When I do `expect(searchkit.history.push).toHaveBeenCalledWith` searchkit always calls `history.push` with `/?`, which is pretty strange. Let me take a closer look at this over the coming weekend.

chrisgzf:
Merging for now, with a new message to direct people to use the alternate module search, as ES seems to be having issues again.

Issue #3294, opened 2021-06-27
Design Discussion for integrating NUS Timetable Optimizer


frizensami:
This is a fairly long issue - my apologies! I hope it will provide a good base for discussion. 

---------------------------

### Is your feature request related to a problem? Please describe.

Students have many possibilities each semester when choosing lessons and modules for their timetable. This is especially difficult and time-consuming in earlier academic years. 
* Introductory modules (e.g., for CS, CS1101S, CS1231S) may have tens of choices for their lessons, some of which may conflict with other modules. 
* Senior students (with fewer lesson choices) may benefit from a way to choose a subset of _modules_ instead. 
* Generally, students also may prefer to start their lessons no earlier than a certain time, have more free days, allow some time for lunch, etc. 

Therefore, a feature that automatically optimizes a student's NUSMods timetable subject to their preferences might be very useful. 

---------------------------

#### Description 
[NUS Timetable Optimizer](https://nus-optimizer.com/) ([GitHub link](https://github.com/frizensami/nus-timetable-optimizer)) is a standalone MVP implementation of this proposed feature. It takes as input a list of modules and a set of user-supplied constraints (e.g., free day / lunch hours / lesson start-end requirements), and outputs a timetable that meets the constraints. If optimization constraints are specified (e.g., compact the timetable as much as possible), the output timetable attempts to maximize/minimize the stated constraints. The list of possible new features are at [the project's Issues page](https://github.com/frizensami/nus-timetable-optimizer/issues), with more to be added. 

**This issue aims to discuss the possibility of merging its functionality into NUSMods, and if so, to discuss the design considerations around its implementation.** 

#### Some results from MVP implementation
* 3,233 unique users since June 1st
* No timetable-generation-related errors reported  
* Mean of 2 - 4 minutes spent optimizing on the site 

#### Prior work 
#390 describes a previous attempt to solve this issue, which inspired the current approach. Notable **similarities** are:
* We use a SMT solver/optimizer to compute matching timetables
* We treat each half-hour slot as an integer-valued variable 
* Some similarities in encoding lessons as integer values

However, we also build upon #390. This is a short description of some of the **improvements**:
* We do not require **any** server-side computation. The constraints are converted to the SMTLIB2 format on the client
* The SMT solver is a WebAssembly-compiled version of the Microsoft Z3 SMT solver, instead of BoolectorJS. This solves timetables at near-native speeds (~1 second after removing outliers)
[...]

PR #3296, opened 2021-06-28
[WIP] NUSMods Optimizer Prototyping

frizensami:
## Context
This WIP PR accompanies the discussions on timetable optimization in #3294. 

This PR is for rapid prototyping based on the discussions in the issue + auto deployment to Vercel so that changes can be viewed easily.

## Implementation

**Note: won't try to fulfil `npm run ci` warnings/errors mostly for now, will implement functionality first**

### Discussion points
Some discussion points in no particular order:
- [ ] Main entry point to view optimizer (on the main timetable? separate tab? Former seems quite natural.)
  - [ ] Should we modify the main timetable and then give users an option to undo to their initial timetable? 
- [ ] Where to display constraints? (Just below modules? Popup/Modal? Not immediately in favour of the popup/modal idea since users perhaps can't change constraints + rerun optimizer quickly) 
- [ ] How to guide users on how to use constraints? 
- [ ] When should we download the ~ 4 MB solver file? During initial NUSMods page load? Or only when the user starts to use the optimizer?
- [ ] In the future, we might have _competing_ optimization constraints, e.g., (a) maximize my free days but (b) give me lunch hours. This might require users to rank or weight their optimization preferences. 

### Implementation Plan
After further discussion I can update this PR with an implementation plan. Minimally, this needs:
- [ ] Constraints UI / UX
- [X] Integrating the Z3 worker, wasm file, and wrappers
- [x] Classes and utils to translate from an NUSMods timetable --> SMTLIB2 --> solver results --> back to a timetable
- [x] (In progress) Non-UI/UX cleanup
- [x] Properly publish the `smtlib-ext` library so we can stop the github reference
- [ ] A LOT of tests
- [x] Vendor types for Z3 and smtlib?
Of the above, the **Constraints UI** will need the most work. The latter two already have been somewhat implemented in [the standalone NUS Timetable Optimizer](https://github.com/frizensami/nusmods_smt_web/).

Translation utils:
- [ ] (In progress) NUSMods timetable --> Optimizer's GenericTimetable
- [ ] Z3 results --> list of changes to make to the existing timetable
- [ ] More to be updated

### Known bugs / limitations
- [ ] Expects only normal semesters (1, 2)
- [ ] Assumes normal weeks array (`NumericWeeks: number[]`) and not `WeekRange` (strange week ranges)
 
### Things to test
- Lesson types with spaces need to be processed before going into the solver as variable names. Solver varnames cannot have spaaces.

### Current changes 

These are just to indicate the idea, please no punterino on the UI :D

1. Implements a "Show/Hide Optimizer" button on the main timetable UI
image
2. Showing the optimizer (for now) displays a set of constraints below the timetable 
image

## Other Information
This is more or less the first time I'm using React, and I'm not much of a front-end designer (clearly, haha), so I'll try to keep up! :slightly_smiling_face:


website/package.json
```diff
@@ -168,7 +168,8 @@
     "reselect": "4.0.0",
     "samlify": "^2.7.6",
     "searchkit": "2.4.1-alpha.5",
-    "use-subscription": "1.5.1"
+    "use-subscription": "1.5.1",
+    "worker-loader": "^3.0.8"
```
li-kai, line 172: This should be in devDependencies

website/src/types/optimizer.ts
```diff
@@ -0,0 +1,144 @@
...
+// Mapping between lesson types -> classNo -> lessons.
+//  We have to take one classNo of each lessonType, so this indicates all the slots to be filled
+//    per classNo per lessonType
+export type LessonsByGroupsByClassNo = {
+    [lessonType: string]: { [classNo: string]: readonly RawLesson[] };
+}
+
+
+// User-selected constraints to pass to optimizer
+export interface GlobalConstraints {
```
li-kai, line 54: for booleans, we try to prefix them with `isX` or `hasX` (e.g. `isWorkloadActive`)

website/src/types/optimizer.ts
```diff
@@ -0,0 +1,144 @@
...
+  lunchHalfHours: number;
+  lunchBreakActive: boolean;
+  // Ask optimizer to compact timetable to leave as few gaps between lessons as possible
+  preferCompactTimetable: boolean;
+}
+
+/**
+ * Defs for communicating between Optimizer <-> WebWorker <-> WASM wrapper
+ * */
+export enum Z3MessageKind {
```
li-kai, line 82: Is there something stopping us from using numbers instead? Iirc, when we pass things to wasm, it is more efficient.

frizensami: The Z3 message is maybe a misnomer, it's a message for the Z3 Worker. I'll rename it.

frizensami: Little easier to debug with string enums, so if it's not a performance concern (very few messages are sent generally), I think the string enum might be ok.

ravern:

PR #3322, opened 2021-07-07
Fix module aliases false-positives in scraper

### LINKED ISSUE #2995: Scraper: Some module `aliases` contain unrelated modules
### Describe the bug

CS1101S (and others) should not contain the other unrelated mods in `aliases`.

* https://nusmods.com/modules/CS1101S
* https://nusmods.com/api/v2/2020-2021/modules/CS1101S.json

Reported by Prof. Martin Henz over email.

### Screenshots

image

### Additional context

I suspect it's due to the scraper being too eager when deciding if modules are dual-coded, probably because a large number of classes are in "E-Learn_*" venues this AY.

chrisgzf:
## Context

Closes #2995 

This PR adds an avoid list for merging of dual-coded modules.

Yeah, I know this looks kind of hacky and isn't really a rule-based solution. Let me explain.

Currently, how the merging of dual coded module works is as follows:
1. Find modules with some classes with similar timings and **venues**
2. Check if these modules have the same name, if yes, merge them as dual coded modules

This breaks down slightly in today's E-Learning world because a lot of modules have the same venue (E-Learn_A, E-Learn_B, or E-Learn_C), and causes a lot of false-positives.

One way to avoid this is by excluding E-Learning modules from this merging algorithm, but I tried that and it gave me a lot of false-negatives (a lot of E-Learning modules also legitimately require merging). This is the implementation I tried:

Screenshot 2021-07-07 at 9 36 57 PM

and the diff of the aliases.json: https://hastebin.com/goruralafi.shell (these are the aliases that got removed)

A lot of legitimate aliases ended up being removed.

## Current Implementation

But from the diff above, it helped me to sift through to see which are the mods that shouldn't be merged, and hence it helped me to arrive to this.

Note: To figure out if a mod should or shouldn't be aliased with another mod, I used 2019-2020 data as the "ground truth".

After this PR, these false-positive aliases will be removed: https://hastebin.com/satunekase.shell

Tested locally.

### DIFF
--- a/scrapers/nus-v2/README.md
+++ b/scrapers/nus-v2/README.md
@@ -49,7 +49,7 @@ We use ElasticSearch for our module search page. For local development it is not
 Run these through `yarn scrape` in production or `yarn dev` in development piped through `yarn bunyan` for formatting - eg. `yarn dev test | yarn bunyan`. You can also run `yarn dev help` to see a list of all commands.
 
 - `test` - run some simple API requests to check you have set everything up correctly
-- `department` - download department and faculty codes
+- `departments` - download department and faculty codes
 - `semester [sem]` - download module and timetable data for the given semester
 - `venue [sem]` - collate venue data into the shape needed by the frontend
 - `combine` - combine the module data for all four semesters together
--- a/scrapers/nus-v2/src/index.ts
+++ b/scrapers/nus-v2/src/index.ts
@@ -32,7 +32,8 @@ function run(fn: (...args: any[]) => Promise<any>) {
 
 const parameters: Record<string, yargs.Options> = {
   sem: {
-    choices: Semesters.map(String),
+    type: 'number',
+    choices: Semesters,
   },
   year: {
     type: 'string',
@@ -65,7 +66,7 @@ yargs
     handler: run(({ year }) => new GetFacultyDepartment(year).run()),
   })
   .command({
-    command: 'semester [year] <sem>',
+    command: 'semester <sem> [year]',
     describe: 'download all data for the given semester',
     builder: {
       sem: parameters.sem,
--- a/scrapers/nus-v2/src/utils/data.ts
+++ b/scrapers/nus-v2/src/utils/data.ts
@@ -201,6 +201,12 @@ export const activityLessonType: Record<string, LessonType> = {
   ...unrecognizedLessonTypes,
 };
 
+export const modulesToAvoidMerging = new Set<string>([
+  'Programming Methodology',
+  'Data Structures and Algorithms',
+  'Graduate Research Seminar',
+]);
+
 export function isModuleOffered(module: {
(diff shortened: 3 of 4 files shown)


scrapers/nus-v2/README.md
```diff
@@ -49,7 +49,7 @@ We use ElasticSearch for our module search page. For local development it is not
 Run these through `yarn scrape` in production or `yarn dev` in development piped through `yarn bunyan` for formatting - eg. `yarn dev test | yarn bunyan`. You can also run `yarn dev help` to see a list of all commands.
 
 - `test` - run some simple API requests to check you have set everything up correctly
-- `department` - download department and faculty codes
+- `departments` - download department and faculty codes
```
chrisgzf, line 52: Changed this to match the main command in yargs

scrapers/nus-v2/src/index.ts
```diff
@@ -32,7 +32,8 @@ function run(fn: (...args: any[]) => Promise<any>) {
 
 const parameters: Record<string, yargs.Options> = {
   sem: {
-    choices: Semesters.map(String),
+    type: 'number',
+    choices: Semesters,
```
chrisgzf, line 36: The original one somehow didn't work even if I ran `yarn dev semester 1 | yarn bunyan` or `yarn dev semester "1" | yarn bunyan` or even `yarn dev semester \"1\" | yarn bunyan`, somehow yargs keeps parsing it as a number and then it keeps saying it expects "1" or "2" or "3" or "4". So I just changed it to accept a number instead.

scrapers/nus-v2/src/index.ts
```diff
@@ -65,7 +66,7 @@ yargs
     handler: run(({ year }) => new GetFacultyDepartment(year).run()),
   })
   .command({
-    command: 'semester [year] <sem>',
+    command: 'semester <sem> [year]',
```
chrisgzf, line 69: From my cursory search on yargs syntax, seems like optional args should be put at the back.

scrapers/nus-v2/src/utils/data.ts
```diff
@@ -201,6 +201,12 @@ export const activityLessonType: Record<string, LessonType> = {
   ...unrecognizedLessonTypes,
 };
 
+export const modulesToAvoidMerging = new Set<string>([
```
ZhangYiJiang, line 204: There should be a comment here explaining how this list was derived, otherwise this would just look arbitrary

ZhangYiJiang (approved):

scrapers/nus-v2/src/utils/data.ts
chrisgzf: Got it! Fixed in https://github.com/nusmodifications/nusmods/pull/3322/commits/e863970e3b7ba1529c78ef40a15dcbd42a66ef61

PR #3327, opened 2021-07-22
Add regex to remove gce prerequisites

### LINKED ISSUE #2686: The relationship between CS4240 and CS3240 is not shown in the prerequisite tree
### Describe the bug

CS4240's Prerequisite Tree does not show any prerequisites despite "CS3240 and (MA1301 or A-level / H2 Mathematics)" being listed as prerequisites. In CS3240's page, CS4240 is also not listed as a module needing CS3240. 

### To Reproduce

1. Go to [CS4240](https://nusmods.com/modules/CS4240/interaction-design-for-virtual-and-augmented-reality)
2. Scroll down to Prerequisite Tree
image

Also, 
1. Go to [CS3240](https://nusmods.com/modules/CS3240/interaction-design)
2. CS4240 is not shown. 
image

### Expected behavior
image

CS3240 is listed as a prerequisite both in the descriptions above and on Luminus, so it should appear in the prerequisite tree for CS4240. Similarly, CS4240 should appear in the prerequisite tree for CS3240.

andlimey:
## Context

Fixes #2686 

Removes GCE prerequisites that the parser cannot recognize.

## Implementation

The solution had to specifically identify keywords pertaining to A/O level prerequisites so that it will not affect `removeModuleTitles()`. See the unit tests for more information.

## Other Information

ramda.pipe() from `@types/ramda` seem to only support up to 10 arguments. Hence, `removeModuleTitles()` had to be composed with `removeGCEPrerequisites()`.

### DIFF
--- a/scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
+++ b/scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
@@ -3,23 +3,51 @@ import { normalize } from './normalizeString';
 /* eslint-disable max-len */
 
 describe(normalize, () => {
-  it('removes GCE prerequisites 1', () => {
-    const testString =
-      'GCE ‘A’ Level or H2 Mathematics or H2 Further Mathematics or MA1301 or MA1301FC or MA1301X';
-    const expected = 'MA1301 or MA1301FC or MA1301X';
-    expect(normalize(testString)).toBe(expected);
+  describe.each([
+    { testString: 'MA1301 or O-level', expected: 'MA1301' },
+    { testString: 'CS3240 and (A-level or MA1301)', expected: 'CS3240 and (MA1301)' },
+  ])('removes A/O level prerequisites', ({ testString, expected }) => {
+    test(`return value should be ${expected}`, () => {
+      expect(normalize(testString)).toBe(expected);
+    });
   });
 
-  it('removes GCE prerequisites 2', () => {
-    const testString = 'CS3240 and (MA1301 or A-level / H2 Mathematics)';
-    const expected = 'CS3240 and (MA1301)';
-    expect(normalize(testString)).toBe(expected);
+  describe.each([
+    { testString: 'MA1301 or GCE ‘O’ Level or MA1301FC', expected: 'MA1301 or MA1301FC' },
+    { testString: 'GCE ‘A’ Level Mathematics or MA1301 or MA1301X', expected: 'MA1301 or MA1301X' },
+  ])('removes GCE ‘A’/‘O’ Level prerequisites', ({ testString, expected }) => {
+    test(`return value should be ${expected}`, () => {
+      expect(normalize(testString)).toBe(expected);
+    });
   });
 
-  it('removes GCE prerequisites 3', () => {
-    const testString = 'MA1301 or GCE ‘A’ Level or MA1301FC';
-    const expected = 'MA1301 or MA1301FC';
-    expect(normalize(testString)).toBe(expected);
+  describe.each([
+    { testString: 'MA1301 or H1 Mathematics', expected: 'MA1301' },
+    { testString: 'H2 Further Mathematics or MA1301', expected: 'MA1301' },
+    {
+      testString: 'CS3240 and (MA1301 or H3 Further Mathematics)',
+      expected: 'CS3240 and (MA1301)',
+    },
+  ])('removes H1/H2/H3 subject prerequisites', ({ testString, expected }) => {
+    test(`return value should be ${expected}`, () => {
+      expect(normalize(testString)).toBe(expected);
+    });
+  });
+
+  describe.each([
+    {
+      testString:
+        'GCE ‘A’ Level or H2 Mathematics or H2 Further Mathematics or MA1301 or MA1301FC or MA1301X',
+      expected: 'MA1301 or MA1301FC or MA1301X',
+    },
+    {
+      testString: 'CS3240 and (MA1301 or A-level / H2 Mathematics)',
+      expected: 'CS3240 and (MA1301)',
+    },
+  ])('remove all GCE prerequisites', ({ testString, expected }) => {
+    test(`return value should be ${expected}`, () => {
+      expect(normalize(testString)).toBe(expected);
+    });
   });
 
   it('converts commas to delimiter or', () => {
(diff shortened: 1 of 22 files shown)


scrapers/nus-v2/src/services/requisite-tree/normalizeString.ts
```diff
@@ -106,6 +106,14 @@ function removeModuleTitles(string: string): string {
   return result;
 }
 
+const gceRegex = /((GCE\D+)|(H[1-3]\D+)|([A|O][\W*][level]+))\b[\s]*(?:and|or)?/g;
```
ZhangYiJiang, line 109: I think this regex might be a bit buggy https://regex101.com/r/nUk5pS/1 - 

``` regex
(
  // Why do we want to capture non-digits after these two groups here here?
  (GCE\D+)|
  (H[1-3]\D+)|
  // [\W*] looks like a typo - this would only capture one character that's either non-word or * literally 
  // [A|O] also captures | literally which is likely unintended 
  (
    [A|O][\W*]
    // [level]+ would capture any combination of characters as long as they consist of the letters e, l or v
    [level]+
  )
)
\b
// The [] is not necessary since this is a class of one character
[\s]*
// This should already capture trailing and/or right? 
(?:and|or)?
``` 

in general we probably want to avoid classes like `\D` and if we can be more specific (since not digit can be, well, almost anything)

scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
```diff
@@ -3,6 +3,25 @@ import { normalize } from './normalizeString';
 /* eslint-disable max-len */
 
 describe(normalize, () => {
+  it('removes GCE prerequisites 1', () => {
```
ZhangYiJiang, line 6: In general we try to name the test cases as specifically as possible. In this case you probably want to test the ability to remove H1-H3, GCE 'A' and 'O' separately first, and trailing and/or, then a few more cases using real data. If we have multiple cases for the same scenario we can use the `describe.each` function to define a table of test cases https://jestjs.io/docs/api#describeeachtablename-fn-timeout

scrapers/nus-v2/src/services/requisite-tree/normalizeString.ts
andlimey: >   Why do we want to capture non-digits after these two groups here?  
(GCE\D+) | (H[1-3]\D+)

I wanted to capture phrases like `H2 Mathematics` and `GCE 'A' Level`.  I'll replace `\D` with a more specific regex.

> [\W*] looks like a typo - this would only capture one character that's either non-word or * literally 
[A|O] also captures | literally which is likely unintended 

`[\W*]` was meant to capture the dash in `A-level`. Will rectify it to make it more specific as well.

> [level]+ would capture any combination of characters as long as they consist of the letters e, l or v
The [] is not necessary since this is a class of one character. [\s]*

Noted on these. Will rectify accordingly!

>This should already capture trailing and/or right? (?:and|or)?

I'm not sure what you are trying to say here.

scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
andlimey: Will rectify accordingly. Thanks for the suggestion!

scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
```diff
@@ -3,6 +3,53 @@ import { normalize } from './normalizeString';
 /* eslint-disable max-len */
 
 describe(normalize, () => {
+  describe.each([
+    { testString: 'MA1301 or O-level', expected: 'MA1301' },
```
ZhangYiJiang, line 7: Sorry, I think I linked to the wrong function - I think each of these should be a single test block, not describe block - https://jestjs.io/docs/api#testeachtablename-fn-timeout. We can still use describe to organize the scenarios. So this might look something like 

``` ts 
describe('Remove A and O level requirements', () => {
  test.each([
    ['MA1301 or O-level', 'MA1301'],
    // ...
  ])('%s should be normalized to %s', (input: string, expected: string) => {
    expect(normalize(input)).toBe(expected);
  });
});
```

scrapers/nus-v2/src/services/requisite-tree/normalizeString.ts
```diff
@@ -106,6 +106,14 @@ function removeModuleTitles(string: string): string {
   return result;
 }
 
+const gceRegex = /((GCE\s‘[AO]’\sLevel\s*([A-Z]\w+\s*)?)|(H[1-3]\s([A-Z]\w+\s*)+)|([AO]-level\s*))((and|or)\s)?/g;
```
ZhangYiJiang, line 109: More suggestions - 

`(GCE\s‘[AO]’` - The quote regex would only match "smart quotes" - we should probably use `[‘'"]?` to match all quotes, and `?` to so we can match even if quotes were not used. All `\s` should also be changed to `\s+` to match multiple whitespaces 

`([A-Z]\w+` - this matches one capitalized word, but this might be unnecessarily strict - is there a reason we expect the word matched here to always be capitalized? Otherwise I think we might want to just add `i` flag and just use `\w+` to match words 

I think we might also be able to combine the "GCE 'A' Levels" and the "A-levels" cases by making the GCE part optional, so we only need to match two cases, and in that case it might be more readable to split this into two regexes testing each case individually.

ZhangYiJiang (commented):
This is looking quite close - thanks for doing this!

scrapers/nus-v2/src/services/requisite-tree/normalizeString.ts
ZhangYiJiang: There's a separate `trailingAndOrRegex` regex which seems to be doing the same thing as the and|or block at the end of the regex. Do we still need `trailingAndOrRegex`?

scrapers/nus-v2/src/services/requisite-tree/normalizeString.test.ts
andlimey: I see. Will rectify accordingly!

chrisgzf:
I tested this, turns out that the scraper task that eventually normalizes the string does a check at first to see if the prereq string contains GCE prereqs - and skips the parsing entirely if it does.

I removed the check, and added some tests in https://github.com/nusmodifications/nusmods/pull/3327/commits/c9279c674f78721ba11f30116175b272d27f5644.

chrisgzf (approved):
Tested with prod data, works well! Thanks for fixing this issue @andlimey!

PR #3332, opened 2021-07-26
Modify module code regex

### LINKED ISSUE #3328: Planner doesn't work with GESS modules
### Bug Description

GESS-prefixed modules cannot be placed into the planner.

### To Reproduce

1. Go to 'Module Planner'
2. Click on 'Add Modules'
3. Type in any GESS-prefixed module (i.e. GESS1029)
4. Module doesn't appear.

### Expected behavior

GESS-prefixed module appears.

### Desktop:

- OS: Windows
- Chrome Browser
- Version 91.0.4472.124

andlimey:
## Context

Fixes #3328 

The issue was due to `MODULE_CODE_REGEX`. 

## Implementation

Modifying the regex will fix the bug, as seen below.

image


chrisgzf:
Thank for looking into this @andlimey! I've just tested on the deploy preview and the fix is working, merging now.

PR #3337, opened 2021-08-08
Expand 'Coll.' abbrieviation in scraper

chrisgzf:
image

Currently CHS's name is written as "Coll. of Humanities and Sciences", because` The "College" is shorten due to length restriction.` (from RO staff over email).

This PR expands that abbrieviation.

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetFacultyDepartment.ts
+++ b/scrapers/nus-v2/src/tasks/GetFacultyDepartment.ts
@@ -22,10 +22,12 @@ const abbreviationMap = {
   Grad: 'Graduate',
   DO: "Dean's Office",
   Stud: 'Studies',
+  'Coll.': 'College',
 };
 
 const abbreviationRegex = map(abbreviationMap, (expanded, abbr): [RegExp, string] => [
-  new RegExp(`\\b${escapeRegExp(abbr)}\\b`, 'gi'),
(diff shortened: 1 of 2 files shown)


scrapers/nus-v2/src/tasks/GetFacultyDepartment.ts
```diff
@@ -22,10 +22,12 @@ const abbreviationMap = {
   Grad: 'Graduate',
   DO: "Dean's Office",
   Stud: 'Studies',
+  'Coll.': 'College',
 };
 
 const abbreviationRegex = map(abbreviationMap, (expanded, abbr): [RegExp, string] => [
-  new RegExp(`\\b${escapeRegExp(abbr)}\\b`, 'gi'),
+  // (?!\S) instead of \b because \b (word boundary) doesn't account for the . in "Coll."
+  new RegExp(`\\b${escapeRegExp(abbr)}(?!\\S)`, 'gi'),
```
ZhangYiJiang, line 30: Interesting, you're using the lookahead because it won't match the whitespace itself, right? I was wondering why this isn't just `(\b|\s+)`

ZhangYiJiang (approved):
👍

scrapers/nus-v2/src/tasks/GetFacultyDepartment.ts
chrisgzf: Yup! That's right! We don't want to replace the whitespace itself, so we don't match it.

PR #3341, opened 2021-08-11
Increase ModReg Notification's z-index to be over navtabs'

chrisjwelly:
## Context

The currently-deployed NUSMods (https://github.com/nusmodifications/nusmods/commit/ebb5cf52b6f50b9a4179e1ede05b965668d40dd5) has an issue whereby the ModReg notification banner's z-index is incompatible with the side navigation tabs. It appears to be "sandwich"ed by the navigation tabs text and the timetable cells.
image

## Implementation
This PR resolves it by increasing the z-index of the ModReg Notification banner, specifically I have chosen a value that is slightly over the `$navtabs-z-index` variable. 

image

Attached is a video of the fixed version:
https://user-images.githubusercontent.com/39089453/129068200-7b75ba67-f69e-4e4b-b296-5660ae81c0c6.mp4

A slight point of discussion is that in the old version, the Navbar(?) z-index seems to be higher than the notifications:
OLD version:
image

This PR:
image

I think in terms of UX, it might be unlikely that users would minimise the screen to an extent that the ModReg Notification banner reaches the top Navbar. Furthermore, I think either one might actually work. However, I am open to comments from the developers if the ModReg Notification Banner is intended to be covered by the Navbar.

## Other Information

I'm not very proficient in frontend, but I was wondering how I could see the z-index of this component from the devtools?

I tried opening the devtools in Firefox and could not find the z-index CSS:
image

I resolved this by finding the correct React component in the source code, and studying how this component is styled by its class, noticing that there is a `z-index` field for the `container`:
https://github.com/nusmodifications/nusmods/blob/ebb5cf52b6f50b9a4179e1ede05b965668d40dd5/website/src/views/components/notfications/ModRegNotification.scss#L5

Any help would be appreciated! Thank you very much.

### DIFF
--- a/website/src/styles/constants.scss
+++ b/website/src/styles/constants.scss
@@ -184,7 +184,7 @@ $venue-map-z-index: 760;
 $venue-detail-z-index: 750;
 $side-menu-z-index: 700;
 $side-menu-overlay-z-index: 690;
-$modreg-notification-z-index: 500;
+$modreg-notification-z-index: 910;
 $timetable-warning-z-index: 200;
 $timetable-scolled-day-z-index: 120;
 $timetable-selected-cell-z-index: 110;


chrisjwelly:
I ran `yarn run ci` on the current `master`(ebb5cf5) locally, and it seems to fail as well:
image

I'm not quite sure what was failing and what I should do to resolve this

chrisgzf (approved):
Thanks for the fix!

> I think in terms of UX, it might be unlikely that users would minimise the screen to an extent that the ModReg Notification banner reaches the top Navbar.

Yeap, so we can safely ignore this non-issue.

>  ran yarn run ci on the current master(ebb5cf5) locally, and it seems to fail (and consequently, in my branch as well):

I cloned your branch and ran CI on it, was working fine. Probably a node version mismatch or yarn version issue. See what you get from `node -v`: if you get node v10 and lower, then there will likely be some issues. We officially support `lts/erbium` which is node 12 LTS, so it would be a good idea for you to use a tool like [fnm](https://github.com/Schniz/fnm) (much faster than the widely-used `nvm`) to manage your node versions.

Thanks again for the contribution!

chrisgzf:
Tested on deploy preview, working. Merging.

chrisjwelly:
Thanks for the merge!

> I cloned your branch and ran CI on it, was working fine. Probably a node version mismatch or yarn version issue. See what you get from `node -v`: if you get node v10 and lower, then there will likely be some issues. We officially support `lts/erbium` which is node 12 LTS, so it would be a good idea for you to use a tool like [fnm](https://github.com/Schniz/fnm) (much faster than the widely-used `nvm`) to manage your node versions.

Seems like I am running one of the latest versions (v16.6.1). This might be the cause. Will keep this in mind if I contribute to the repository again in the future!

PR #3342, opened 2021-08-12
Prevent overscroll-behavior in scrollable timetable

EmilyOng:
## Context

https://user-images.githubusercontent.com/21200681/129133919-57191023-72bd-42c7-b11c-12f644f2ad89.mp4

In the TimeTable view, when there are multiple classes to be chosen for a certain module, exceeding the scroll boundary of the TimeTable causes to:
    1) navigate to the previous page (for scrolling in x-direction)
    2) triggering pull-to-refresh behavior on mobile phones (for scrolling in y-direction).

## Implementation

Set `overscroll-behavior` to be `none` so there is no scroll chaining to neighbouring scrolling areas. (ref: https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior).

## Other Information

Can be tested with modules such as: `GEQ1000`.

### DIFF
--- a/website/src/views/timetable/TimetableContent.scss
+++ b/website/src/views/timetable/TimetableContent.scss
@@ -9,6 +9,7 @@
 .timetableWrapper {
   composes: scrollable from global;
   margin-bottom: 1rem;
+  overscroll-behavior: none;
 
   @include media-breakpoint-down(sm) {
     // Maximize the space taken by the timetable by overflowing into the grid gutters


website/src/views/timetable/TimetableContent.scss
```diff
@@ -9,6 +9,7 @@
 .timetableWrapper {
   composes: scrollable from global;
   margin-bottom: 1rem;
+  overscroll-behavior-x: none;
```
ZhangYiJiang, line 12: Do we need to apply this to the y axis in horizontal mode? I see in your video that you're using vertical mode

chrisgzf:
Thanks for your contribution @EmilyOng! I've tested the fix on the deploy preview and merged the fix!

PR #3347, opened 2021-08-17
Fix vertical overscroll behavior

chrisgzf:
#3342 broke vertical scrolling on mobile, because scrolling in the timetable now doesn't scroll the entire page.

We will change this to only disable horizontal overscroll - the main purpose here is to not trigger the browser's default back/forward mechanism when you scroll to the left or right, inside the timetable.


chrisgzf:
Merging this without review because it's breaking experience for mobile users.

PR #3348, opened 2021-08-18
Fix Cropping Out of Dropdown menus 

### LINKED ISSUE #2747: Module Planner - Dropdown for module card is cropped out
### Describe the bug

screenshot_13-07-20_23:44:08
On the planner, when you click on the card for a module that is on the far left (right next to the scroll bar), the dropdown gets cropped out.

### To Reproduce

1. Go to https://nusmods.com/planner
2. Shrink your browser window so only 1 semester can be seen
3. Click on the arrow

### Expected behavior

Dropdown should appear on the left side for this case so that it isn't cropped out.

* OS: Ubuntu 18.04
* Browser [e.g. Chrome, Safari] Firefox
* Version [e.g. 22] 78.0.1

seanlowjk:
## Context

This PR fixes #2747 which involves the overflowing of module dropdown menus 
for modules located close to the right end of the screen. 

## Implementation

It makes use of `getBoundingClientRect` in order to retrieve the current right boundary for the module 
selected in the module planner. And as a result, it will allow to adjust the position of the dropdown accordingly 
with the help of `sass` 

This is done thanks to the work of `useLayoutEffect`

https://user-images.githubusercontent.com/42912708/129859866-b41184e8-fb74-40e4-a317-a95a1faf3367.mp4

## Other Information

N/A


seanlowjk:
Not sure if this is the best way to go about fixing the cropping out, as a result I will not add the tests yet!

Would need some advice thank you! orz @chrisgzf @ZhangYiJiang @taneliang

website/src/views/planner/ModuleMenu.tsx
```diff
@@ -37,15 +68,20 @@ const ModuleMenu = memo((props: Props) => {
...
             aria-expanded={isOpen}
           >
             <ChevronDown />
           </button>
           <div
-            className={classnames(styles.menu, 'dropdown-menu', { show: isOpen })}
+            className={classnames(isVisible ? styles.menuRight : styles.menuLeft, 'dropdown-menu', {
+              show: isOpen,
+            })}
             {...getMenuProps()}
```
ZhangYiJiang, line 85: Pass the HTMLDivElement menuRef into getMenuProps instead as ref. Downshift will pass it through. Getting the Downshift instance is not guaranteed to work since it's not part of the public API and is just a lot more work

website/src/views/planner/PlannerModule.scss
```diff
@@ -40,7 +40,15 @@
   }
 }
 
-.menu {
+.menuLeft {
+  right: 0;
+  left: auto;
+  width: 12rem;
```
ZhangYiJiang, line 46: This is a common menu style. Leave this as a base style and use `&.menuLeft` and `&.menuRight` for the positions

website/src/views/planner/ModuleMenu.tsx
```diff
@@ -17,13 +17,44 @@ type MenuItem = {
 };
 
 const ModuleMenu = memo((props: Props) => {
+  const [isVisible, setIsVisible] = useState<boolean>(true);
```
ZhangYiJiang, line 20: visible is a bit awkward since the menu isn't visible when it is closed either 

```suggestion
  const [isMenuOverflowing, setMenuOverflowing] = useState(true);
```

website/src/views/planner/ModuleMenu.tsx
```diff
@@ -37,15 +68,20 @@ const ModuleMenu = memo((props: Props) => {
           <button
             className={classnames('btn close')}
             type="button"
-            onClick={() => toggleMenu()}
+            onClick={() => {
+              toggleMenu();
+              toggleVisibility();
```
ZhangYiJiang, line 73: I'm not sure if this will work. When I pause execution here and check the position of the menu div, I get this - 

``` 
$0.getBoundingClientRect()
DOMRect { x: 0, y: 0, width: 0, height: 0, top: 0, right: 0, bottom: 0, left: 0 }
``` 

this is because the element is not yet visible at this point and so it does not have a bounding rect, so you actually need to wait until the menu is actually rendered (probably inside a useLayoutEffect) instead of checking here

ZhangYiJiang (dismissed):
I think an easier approach to measuring the element in the DOM is instead to always make the cards in the last semester open on the left. Since we know only the last semester's card menu can be clipped, since the menu is narrower than the width of a semester, we can just always open that semester's menu on the left.

website/src/views/planner/ModuleMenu.tsx
seanlowjk: Alright! :+1:

website/src/views/planner/PlannerModule.scss
seanlowjk: Okay, understood! :smile:

website/src/views/planner/ModuleMenu.tsx
seanlowjk: :+1:

website/src/views/planner/ModuleMenu.tsx
seanlowjk: Alright, will use the `useLayoutEffect` hook and will try it out!

seanlowjk:
@ZhangYiJiang thank you for the review, I have made the changes necessary and thanks to your suggestion, the logic seems much better now and managed to fix it!

I have also added simple tests to test the behaviour of `ModuleMenu` 

Thank you! orz

seanlowjk:
Apologies, will require a bump with regards to this PR, thank you!

website/src/views/planner/ModuleMenu.tsx
```diff
@@ -17,11 +17,26 @@ type MenuItem = {
 };
 
 const ModuleMenu = memo((props: Props) => {
+  const [isMenuOverflowing, setMenuOverflowing] = useState<boolean>(false);
+  const [hasRendered, setHasRendered] = useState<boolean>(false);
```
li-kai, line 21:
```suggestion
  const [updateState, forceUpdate] = useReducer((x) => x + 1, 0);
```

This is a slightly different way of forcing renders. The current `toggleRender` function could lead to stale state as `hasRendered` may not be consistent with the actual state during re-renders.
Alternatively, you could change `toggleRender` to use a function instead:
```js
const toggleRender = () => setHasRendered((hasRendered) => !hasRendered);
```

li-kai (commented):
@seanlowjk Thanks so much for submitting this patch! Most of our contributors are working full time so we are a little shorthanded on reviewers. This PR is 99% there and I'm just nitpicking so I will merge this in a week regardless 👍.

website/src/views/planner/ModuleMenu.tsx
seanlowjk: Hi @li-kai , thank you for the comment. 

I have fixed the above to rectify. I have chosen the alternative method which is to change `toggleRender` to use a function instead, thank you!

seanlowjk:
I will do some checking on my side later on, and let you know once it is ready, thank you! @li-kai

seanlowjk:
Hi @li-kai the PR should be ready for a re-review, thank you!

li-kai (approved):
LGTM! 🎉

li-kai:
@chrisgzf could you look into the vercel CI failing check? I don't think this PR breaks anything, so it's safe to merge, but just in case.

li-kai:
I don't think this should break master, so I'll go ahead with the merge. Once again, thanks for the PR, @seanlowjk!

PR #3355, opened 2021-09-29
Make vertically displayed day names centred

SuibianP:
## Context

On narrow screens, the vertical day names are left-justified. This makes it aesthetically displeasing in a variable-width font, especially for wide letters such as 'W'.
before

## Implementation

Make them centred to look better.
after

## Other Information

I guess this might be too trivial to spam an issue for it, so I did not open one :rofl: 

I am new to the entire thing, and my apologies in advance if the fix is erroneous in some way.

### DIFF
--- a/website/src/views/timetable/TimetableDay.scss
+++ b/website/src/views/timetable/TimetableDay.scss
@@ -57,13 +57,13 @@
 
     .dayNameText {
       // On small screens, display the day names vertically
+      display: flex;
+      justify-content: center;
       width: 0.6rem;
       font-size: $font-size-s;
       line-height: 1.1;
-      word-break: break-all;
-      display: flex;
-      justify-content: center;
       text-align: center;
+      word-break: break-all;
     }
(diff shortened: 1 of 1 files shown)


chrisgzf (changes requested):
You have very sharp eyes! Thanks for the fix!

CI is failing though, because of

```
$ stylelint "src/**/*.scss" --syntax scss

src/views/timetable/TimetableDay.scss
 64:7  ✖  Expected "display" to come before              order/properties-order
          "word-break"
```

Do you mind fixing the lint errors? You can run CI locally to check if anything will fail with `yarn ci`.

SuibianP:
Sorry, dunno how I could have missed that glaring step when reading CONTRIBUTING.md :flushed:

chrisgzf (approved):
LGTM! Thanks for fixing this! And no worries about the CI thingy - it takes a while to get used to all the steps.

PR #3357, opened 2021-10-04
Update MPE UI strings for term 2120

chrisgzf:
This updates the UI strings for the upcoming MPE in AY21/22 S1.

Changes are live at https://mpe-staging.nusmods.com/mpe (note: this uses test data and not production data)

### DIFF
--- a/website/src/views/mpe/constants.ts
+++ b/website/src/views/mpe/constants.ts
@@ -1,3 +1,3 @@
 export const MAX_MODULES = 7;
-export const MPE_SEMESTER = 1;
+export const MPE_SEMESTER: 1 | 2 = 2;
 export const MPE_AY = '2021/2022';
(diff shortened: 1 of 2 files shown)


website/src/views/mpe/constants.ts
```diff
@@ -1,3 +1,3 @@
 export const MAX_MODULES = 7;
-export const MPE_SEMESTER = 1;
+export const MPE_SEMESTER: 1 | 2 = 2;
```
chrisgzf, line 2: This change is needed because L48 in ModuleForm.tsx has:
```js
const semesterProperty = MPE_SEMESTER === 1 ? 'inS1MPE' : 'inS2MPE';
```
and will give a lint error saying "This condition will always return 'false' since the types have no overlap."

If I were to add a `number` type to `MPE_SEMESTER`, then TS complains that trivially inferred type declarations are not needed.

PR #3363, opened 2021-10-20
Fix wrong API doc of `semesterData` key in `Module`

SuibianP:
## Context

In `#/components/schemas/Module`, `semesterData` key is documented as one single object of type `#/components/schemas/SemesterData`, while it should be an array of objects of such type, as per the actual 200 response of endpoint `/{acadYear}/modules/{moduleCode}.json` (including modules that are only available in one semester).

```shell
$ curl -sL "https://api.nusmods.com/v2/2019-2020/modules/CG1111.json" | jq '.semesterData | type, length'
"array"
1
```

## Implementation

Change `semesterData` key in `#/components/schemas/Module` to be an array of `#/components/schemas/SemesterData`s to match the actual API response.

This should have no impact on anything else since the only change is in the OpenAPI documentation.

### DIFF
--- a/scrapers/nus-v2/swagger.yaml
+++ b/scrapers/nus-v2/swagger.yaml
@@ -635,7 +635,9 @@ components:
           type: 'string'
           example: 'CS2100'
         semesterData:
-          $ref: '#/components/schemas/SemesterData'
+          type: 'array'
+          items:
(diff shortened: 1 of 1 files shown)


SuibianP:
@ZhangYiJiang Sorry but just in case this is forgotten to be merged, bump 🆙

PR #3367, opened 2021-12-03
Shorten timetable share URL

### LINKED ISSUE #2057: Timetable share URL is no longer shortened

winphong:
## Context

This PR fixes #2057 where timetable share url is not shortened

#### Display show original URL button when URL is shortened
_NOTE: Used a separate url shortener service to mock url shortening behaviour_

## Other Information

👋🏻  my first time contributing to a project, implemented the solution based on my personal experience on user experience, open to suggestion if there's a better UX. Tried my best but not too familiar with UI test so do let me know if there's something that I can improve on. Thanks 🙏🏻

### DIFF
--- a/website/src/featureFlags.ts
+++ b/website/src/featureFlags.ts
@@ -1,12 +1,5 @@
 /* eslint-disable import/prefer-default-export */
 
-/**
- * See:
- * - https://github.com/nusmodifications/nusmods/issues/2057
- * - https://github.com/nusmodifications/nusmods/pull/3212
- */
-export const enableShortUrl = false;
-
 /** Enable Course Planning Exercise */
 export const enableCPEx = false;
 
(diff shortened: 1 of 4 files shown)


li-kai (approved):
Thanks for the great work, and apologies for the delay. The code and tests look good to me. @chrisgzf I'll leave the merging to you? We should also open a task to bring our url shorterner back up.

PR #3374, opened 2021-12-22
Add Toggling of Prerequisites Checks for Planner

### LINKED ISSUE #3369: Allow removal of yellow warnings in Planner

The UI gets very cluttered when I add a module to the Planner but haven't filled up the prereq mods from previous AYs. There's this yellow warning box that pops up below the module title/detail. It'd be nice to make it collapsible.

A cross ("X") at the top of the yellow warning box would be nice. Once clicked, it means I acknowledge that there will be some issue/problem (eg: have to take prereq mods). It can either remain yellow but collapsed (only mod title/MCs seen) **OR** go back to being green. That way, I can see all my mods at a glance so I can share it with friends or plan the future AYs better. 

image

There's no way to remove it unless I put in all the prereq mods in the previous AYs on the Planner. 

Thank you for your consideration!!!

seanlowjk:
## Context

Fixes #3369 

## Implementation

Under `PlannerState`, adds a boolean flag `prereqsCheck` to allow the 
toggling of prerequisites check. This ensures that the the numerous warnings 
do not overwhelm the user when planning his or her users. 

## Other Information

https://user-images.githubusercontent.com/42912708/147061651-2d1924f1-29fc-44b9-8cfc-f06075258ff0.mp4

Attached above is the implementation video mock-up done.


website/src/views/planner/PlannerSettings.tsx
```diff
@@ -121,6 +128,19 @@ export const PlannerSettingsComponent: React.FC<Props> = (props) => {
...
         />
       </section>
+
+      <section className={styles.toggleSection}>
+        <div>
+          <h2 className={styles.label}>Toggle Prerequisites Check</h2>
+
+          <p>
+            By toggling this off, it becomes your responsibility to check that prerequisites of the
+            modules you are taking are enforced.
```
chrisgzf, line 138: Hmm.. sounds very scary. Perhaps something like:

"Prerequisite checking for some modules might be inaccurate, giving planner warnings. Toggling this off removes these checks. Please ensure that you manually check that the prerequisites for the modules you would like to take are sufficiently met."
might sound slightly better?

website/src/selectors/planner.ts
```diff
@@ -220,7 +224,7 @@ export function getAcadYearModules(state: State): PlannerModulesWithInfo {
 
       const conflictChecks = [
         noInfoConflict(moduleBank.moduleCodes, planner.custom),
-        prereqConflict(moduleBank.modules, modulesTaken),
+        prereqConflict(moduleBank.modules, modulesTaken, planner.prereqsCheck),
```
chrisgzf, line 227: This is a small nit but I find it slightly weird that the actual checking if prereqs should be checked (damn, what a mouthful) is done inside `prereqConflict`.

I'd prefer to leave `prereqConflict` as-is, where it does its job of checking for the prereqs conflict without having to access the planner's state.

Instead, we decide if we want to actually run the `prereqConflict` function in this selector itself, so that the state is contained within this selector.

Something like:
```
if (planner.prereqsCheck) {
  conflictChecks.push(prereqConflict(...));
}
```

website/src/types/reducers.ts
```diff
@@ -161,6 +161,7 @@ export type PlannerState = Readonly<{
   minYear: string;
   maxYear: string;
   iblocs: boolean;
+  prereqsCheck: boolean; // To enforce check of prerequisites.
```
chrisgzf, line 164:
```suggestion
  prereqsCheck: boolean; // To turn checking of prerequisites on/off
```

chrisgzf (changes requested):
Very good stuff! Thanks for your work! Some small changes else LGTM!

website/src/views/planner/PlannerSettings.tsx
seanlowjk: Ah alright!

website/src/views/planner/PlannerSettings.tsx
```diff
@@ -111,7 +118,7 @@ export const PlannerSettingsComponent: React.FC<Props> = (props) => {
 
           <p>
             <ExternalLink href="http://www.nus.edu.sg/ibloc/iBLOC.html">iBLOCs</ExternalLink> is a
-            program that allow full-time NSmen to read some modules before matriculating.
+            program that allows full-time NSmen to read some modules before matriculating.
```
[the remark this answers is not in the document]
seanlowjk: Welcome! 💯

website/src/selectors/planner.ts
seanlowjk: ah yes, this is a much better way of handling it instead of modifying the method signature!

Thank you for your suggestion!

website/src/types/reducers.ts
seanlowjk: 👍

seanlowjk:
Have made the following modifications, thank you! 🙏 @chrisgzf

website/src/reducers/planner.ts
```diff
@@ -23,6 +24,7 @@ const defaultPlannerState: PlannerState = {
   minYear: config.academicYear,
   maxYear: config.academicYear,
   iblocs: false,
+  prereqsCheck: true,
```
ZhangYiJiang, line 27: This defaults true for new users, but for users who have already used NUSMods this will be `undefined` which is not allowed under types, and when evaluated as a boolean becomes false. You'll want to set up a migration like this one https://github.com/nusmodifications/nusmods/blob/33df1bcca8898b7997dd2f7e728734edf3caa7ad/website/src/reducers/planner.ts#L161 to set up the initial state correctly for existing users

website/src/views/planner/PlannerSettings.tsx
```diff
@@ -121,6 +128,20 @@ export const PlannerSettingsComponent: React.FC<Props> = (props) => {
           onChange={(checked) => props.setIBLOCs(checked)}
         />
       </section>
+
+      <section className={styles.toggleSection}>
```
ZhangYiJiang, line 132: It might be worth considering allowing this to be toggled off per module instead of globally. This would mean the data is stored as an array of ignored module + year + semester instead to identify which module the user has ignored, and the user can ignore them via either menu item on the card or on the warning itself.

ZhangYiJiang (changes requested):
Thanks for doing this! The migration is blocking, the UX improvement is good to have.

website/src/reducers/planner.ts
ZhangYiJiang: Alternatively, define the state as `ignorePrereqCheck?: boolean` instead - this will satisfy the typing without having to write a migration. The values need to be reversed of course

seanlowjk: Ah thanks for the catch! Will fix this in due time!

website/src/views/planner/PlannerSettings.tsx
seanlowjk: Ah yes, that would definitely be more concise. 
Think it would better to do so, but adjustments will definitely be needed to be made :P

Do you think we can still keep the global toggle or is it best to just have the feature for toggles per module?

seanlowjk: Think we can leave this for a future PR which I can work on, as we would need to flesh out the details and implementation details! 

Will this be fine? >< @chrisgzf @ZhangYiJiang

chrisgzf (changes requested):
Last change from me: Can we rename all `setPrerequisitesCheck` and `SET_PREREQUISITES_CHECK` to `setIgnorePrerequisitesCheck` and `SET_IGNORE_PREREQUISITES_CHECK` instead? So that it's not ambiguous what the boolean values mean.

seanlowjk:
Changes fixed, thanks for pointing it out! @chrisgzf

website/src/views/planner/PlannerSettings.tsx
chrisgzf: Yup! Definitely okay to merge this first to alleviate the problem for now. I feel like a better long term fix might be to make the prerequisite checking more robust. We can do so by writing an algo for the prereq-checking detection that assumes that preclusions of a module (e.g. CS1010 precludes CS1101S) also fulfil the same prereqs fulfilled by the module. I will create an issue for this soon.

website/src/views/planner/PlannerSettings.tsx
chrisgzf: Updated the wordings in https://github.com/nusmodifications/nusmods/pull/3374/commits/482cadf31e194f38c316937de47bd51c13096dec

chrisgzf (approved):
I've updated the wordings for the UI change in https://github.com/nusmodifications/nusmods/pull/3374/commits/482cadf31e194f38c316937de47bd51c13096dec to reflect the boolean negation. Everything looks good now. Thanks for your hard work!

PR #3375, opened 2021-12-29
Update ModReg dates for 21/22 S2

chrisgzf:
## Context

Information from:

- http://www.nus.edu.sg/ModReg/docs/UGCPE_ModRegSchedule.pdf
- http://www.nus.edu.sg/ModReg/docs/GD_ModRegSchedule.pdf
- Correct as at 29 Dec 2021 11pm.


chrisgzf:
Merging for now because ModReg is ongoing and this is a minor data change.

PR #3376, opened 2021-12-29
Upgrade Babel, Jest, Webpack, webpack-dev-server

chrisgzf:
## Context

Current master (https://github.com/nusmodifications/nusmods/commit/67ca41a9854e666e28b0641d64850879e00ae8dc) has lots of warnings from Babel and Webpack:
image

This PR attempts to fix these warnings while upgrading some of our deps.

### Babel (7.14 -> 7.16)

1. Read minor version changelogs from https://babeljs.io/blog/2021/07/26/7.15.0 and https://babeljs.io/blog/2021/10/29/7.16.0 and check for major API changes
2. Go through babel config files, check for plugins being included in `@babel/present-env`, clean up config
3. Use babel config's preferred `assumptions` (see [this](https://babeljs.io/docs/en/assumptions)) to specify some of the plugin options that we previously set to `loose`.
4. `yarn ci`

### Jest (26.6 -> 27.4)

Upgrading `@babel/core` caused Jest to fail because it relied on an older version of `@babel/core` which doesn't recognise the newer Babel APIs, specifically the `assumptions` option in the config, so I decided to update Jest too.

1. `yarn upgrade --latest --exact --pattern jest`
2. Read [changelogs](https://jestjs.io/blog/2021/05/25/jest-27) which state that the default testing environment has changed in Jest 27.
3. Change the default test environment to `jsdom`, so that DOM APIs are exposed (they are required by our tests, for obvious reasons).
4. `yarn ci`

### Webpack (5.15 -> 5.65)

There was previously a webpack warning
```
(node:51130) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
	Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
	Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
```
caused by `html-webpack-plugin` not being compatible with webpack 5, which was rather cutting-edge when we transited to webpack 5.

I upgraded webpack along with all webpack associated plugins.

1. `yarn upgrade --latest --exact --pattern webpack`
2. `yarn ci`
3. Manually test build output

### Webpack Dev Server (3.11 -> 4.7)

Updating webpack with the command above updated dev server as well and `yarn start` had issues, due to API changes.

[...]

PR #3378, opened 2022-01-03
Fix timetable-only build HtmlWebpackPlugin JS injection

chrisgzf:
## Context

Timetable export wasn't working because https://nusmods.com/timetable-only wasn't rendering the timetable, so puppeteer couldn't screenshot it. This also caused the serverless function to timeout.

This regression is caused by #3376.

On investigation, the output HTML files generated by webpack before and after #3376 is different.

**Before:** Copy paste `view-source:https://nusmods-website-9k4pwqc12-nusmodifications.vercel.app/timetable-only`
 - Bundle JS is loaded in `<body>`, and hence would be blocking

**After:** Copy paste `view-source:https://nusmods-website-9dyvagl9e-nusmodifications.vercel.app/timetable-only`
 - Bundle JS is loaded in `<head>`, and hence would be deferred

This gives timetable export trouble due to https://github.com/nusmodifications/nusmods/blob/bf368cb3979943ef846ccff93a808edd29216174/website/src/entry/export/main.tsx#L53
which expects the bundle to have been loaded and rendered.

## Implementation

On looking at the webpack configs for timetable-only export, `html-webpack-plugin` seems to have caused the above change.

Looking at https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md#500-2021-02-03, the default script injection behaviour has changed: `Entry javascript resources are now beeing loaded deferred in the <head> tag to improve the page load performance by default - You can set the scriptLoading option to 'blocking' to keep the previous behaviour`

Cross referencing https://github.com/jantimon/html-webpack-plugin#options, we get the fix.

### DIFF
--- a/website/webpack/webpack.config.timetable-only.js
+++ b/website/webpack/webpack.config.timetable-only.js
@@ -28,6 +28,7 @@ const timetableOnlyConfig = merge([
     },
     plugins: [
       new HtmlWebpackPlugin({
+        scriptLoading: 'blocking',
         template: path.join(parts.PATHS.src, source('index.html')),
         inject: true,
       }),


chrisgzf:
Merging without review is timetable export is currently broken.

Issue #3101, opened 2020-12-30
Department of Physics email address is incorrect 


ZhangYiJiang:
### Describe the bug

We've received a bounced email from the email address phybox3@nus.edu.sg

> This message was created automatically by mail delivery software.
> A message that you sent could not be delivered to one or more of its recipients. This is a permanent error.

> phybox3@nus.edu.sg INVALID_ADDRESS

This email address is still listed on http://www.nus.edu.sg/ModReg/docs/UGFac_Contacts.pdf 

We should remove this address from the contacts JSON, 

https://github.com/nusmodifications/nusmods/blob/ed21a41d247631af4af4ac2d08c3c52e36bdf8c1/website/src/data/facultyEmail.json#L384

and contact NUS to check where did the Physics department go

chrisgzf:
note: with the CHS changes we might need to investigate to see if this is more than just a "remove/update" physics dept email address

RussellDash332:
It seems like physngwl@nus.edu.sg is the current contact email
Source: https://www.physics.nus.edu.sg/about-us/contact-us/

Darylgolden:
This issue should be closed.

Darylgolden:
This issue should be closed.

PR #3382, opened 2022-01-22
Update Physics Faculty email

dingyuchen:
## Context

#3101 

## Implementation

Email updated manually

## Other Information

New email is now also reflected in http://www.nus.edu.sg/ModReg/docs/UGFac_Contacts.pdf
so information is correct as of 17 Dec 2021


chrisgzf (approved):
Thanks for the update!

PR #3387, opened 2022-04-30
Improve copy for reading/examination week

li-kai:
## Context
We want to avoid giving everyone a fake scare about non-existent exams with our current wording "Time to study for your exams!"

Screen Shot 2022-04-29 at 22 58 49

I ran a poll and it seems we have overall support for "It's x week!" structure.

## Implementation
Screen Shot 2022-04-29 at 22 56 03

## Other Information
📚


taneliang (approved):
Thanks for the quick PR!

Issue #3402, opened 2022-07-23
bug: able to add negative modular credits to custom module


tohhengxing:
### Describe the bug

You can add negative MC when creating a custom module in module planner

### To Reproduce

1. go to module planner
2. add a custom mod
3. click add data
4. input a negative number

### Expected behavior

should not allow negative number to be added

### Screenshots

https://user-images.githubusercontent.com/102740235/180600693-18652abe-4674-446a-a24a-6ab29eb05820.mp4

wr1159:
I believe this issue has been resolved and can be closed

kokrui:
Thanks for the catch @wr1159!

PR #3403, opened 2022-07-23
fix: negative MC being valid when creating custom mod in planner

tohhengxing:
## Context
create custom module form can save negative modular credits.

resolves issue #3402 

## Implementation
prevents create custom module form from saving if there is a negative number for modular credit

## Screenshots
<details>
<summary>before</summary>

https://user-images.githubusercontent.com/102740235/180600847-1490f612-530f-423e-98d7-53aa4ebcc990.mp4
</details>

<details>
<summary>after</summary>

https://user-images.githubusercontent.com/102740235/180600820-21034cea-d749-4491-a07f-450037b81077.mp4
</details>

### DIFF
--- a/website/src/views/planner/CustomModuleForm.tsx
+++ b/website/src/views/planner/CustomModuleForm.tsx
@@ -36,8 +36,8 @@ export const CustomModuleFormComponent: React.FC<Props> = (props) => {
     const inputTitleCurrent = inputTitle.current;
     const title = inputTitleCurrent && inputTitleCurrent.value;
 
-    // Module credit is required
-    if (moduleCredit == null) return;
+    // Module credit is required, module credit cannot be negative
+    if (moduleCredit == null || parseInt(moduleCredit, 10) < 0) return;
 
     props.addCustomModule(props.moduleCode, {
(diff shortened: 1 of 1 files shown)


website/src/views/planner/CustomModuleForm.tsx
```diff
@@ -36,8 +36,8 @@ export const CustomModuleFormComponent: React.FC<Props> = (props) => {
     const inputTitleCurrent = inputTitle.current;
     const title = inputTitleCurrent && inputTitleCurrent.value;
 
-    // Module credit is required
-    if (moduleCredit == null) return;
+    // Module credit is required, module credit cannot be negative
```
ZhangYiJiang, line 39: You can also add `min` attribute to the input below, that should prevent the user from inputting invalid values. Right now they can input invalid value and try to save, but nothing will happen which is likely to be confusing

tohhengxing: hi @ZhangYiJiang thank you for reviewing my code! i have added the changes you have requested

chrisgzf (approved):
Thanks Heng Xing!

PR #3406, opened 2022-08-05
Change type when downloading as iCal to text/calendar

### LINKED ISSUE #3405: Downloading calendar on iOS results in .txt file instead of .ics file
### Describe the bug

Downloading timetable as calendar file on iOS Safari and Chrome saves it as `.ics.txt` instead of `.ics`, which does not allow it to be recognised as a calendar file. 

### To Reproduce

1. On iOS, download timetable as Calendar FIle

### Expected behavior

File to be downloaded without `.txt` extension, which allows events to be added to calendar after download. 

- Device: iPhone 13
- OS: iOS 15.6
- Browser: Safari
- Version: 15

### Additional context

Changing MIME Type in `downloadAsIcal` in `website/src/actions/export.ts` from `text/plain` to `text/calendar` seems to fix this.

w-yuchen:
## Context
Resolves #3405. 

## Implementation
Changes content type from `text/plain` to `text/calendar`, since according to [this](https://apple.stackexchange.com/questions/227113/safari-automatically-appends-txt-extension-to-ino-files/239319), appending `.txt` is intentional.

### DIFF
--- a/website/src/actions/export.ts
+++ b/website/src/actions/export.ts
@@ -42,7 +42,7 @@ export function downloadAsIcal(semester: Semester) {
           events,
         });
 
-        const blob = new Blob([cal.toString()], { type: 'text/plain' });
+        const blob = new Blob([cal.toString()], { type: 'text/calendar' });
         downloadUrl(blob, 'nusmods_calendar.ics');
       })
       .catch(captureException);


chrisgzf (approved):
Thanks for the fix!

Issue #3404, opened 2022-08-04
NUSMODs for TAs


FeanTMX:
### Is your feature request related to a problem? Please describe.

A clear and concise description of what the problem is. Eg. I'm always frustrated when [...]
Hi, as an undergraduate TA, I would like to suggest that nusmods can have a functionality to add tutorials/ labs only, since lectures are generally irrelevant to TAs. If possible, I would like to see a function which introduces modularity, so that if I can add one or more tutorial slots that I am tutoring, or only include labs and not tutorials if I am tutoring for labs (vice versa).  I will really be very very thankful if there is such a functionality.

HollaG:
Hacky workaround for now:
1) Export the timetable as a link
2) Remove the specific lecture slot from the link
3) Re-import and save the timetable

for e.g.:
https://nusmods.com/timetable/sem-1/share?CS1101S=TUT:09F,REC:11B,LEC:1
becomes
https://nusmods.com/timetable/sem-1/share?CS1101S=TUT:09F,REC:11B

FredericChow00:
Can I ask how do you remove the specific lecture slot from the link?

PR #3420, opened 2023-01-01
Change ModuleLessonConfig value to array

zhoukerrr:
## Context

Part of #3404 

## Implementation

Changed the value of `ModuleLessonConfig` type to `classNo[]`.
This change does not affect anything for its current functionality.
New redux actions will be added to add/remove classes to the array.
This will allow TAs to customise the module.

## Other Information

The current idea is to add an extra button next to the remove and hide button of the module and allow student to add/remove classes as they click.

### DIFF
--- a/website/src/reducers/timetables.test.ts
+++ b/website/src/reducers/timetables.test.ts
@@ -125,41 +125,41 @@ describe('lesson reducer', () => {
           lessons: {
             [1]: {
               CS1010S: {
-                Lecture: '1',
-                Recitation: '2',
+                Lecture: ['1'],
+                Recitation: ['2'],
               },
               CS3216: {
-                Lecture: '1',
+                Lecture: ['1'],
               },
             },
             [2]: {
               CS3217: {
-                Lecture: '1',
+                Lecture: ['1'],
               },
             },
           },
         },
         setLessonConfig(1, 'CS1010S', {
-          Lecture: '2',
-          Recitation: '3',
-          Tutorial: '4',
+          Lecture: ['2'],
+          Recitation: ['3'],
+          Tutorial: ['4'],
         }),
       ),
     ).toMatchObject({
       lessons: {
         [1]: {
           CS1010S: {
-            Lecture: '2',
-            Recitation: '3',
-            Tutorial: '4',
+            Lecture: ['2'],
+            Recitation: ['3'],
+            Tutorial: ['4'],
           },
           CS3216: {
-            Lecture: '1',
+            Lecture: ['1'],
           },
         },
         [2]: {
           CS3217: {
-            Lecture: '1',
+            Lecture: ['1'],
           },
         },
       },
@@ -172,7 +172,7 @@ describe('stateReconciler', () => {
     '2015/2016': {
       [1]: {
         GET1006: {
-          Lecture: '1',
+          Lecture: ['1'],
         },
       },
     },
@@ -181,13 +181,13 @@ describe('stateReconciler', () => {
   const oldLessons = {
     [1]: {
       CS1010S: {
-        Lecture: '1',
-        Recitation: '2',
+        Lecture: ['1'],
+        Recitation: ['2'],
       },
     },
     [2]: {
       CS3217: {
-        Lecture: '1',
+        Lecture: ['1'],
       },
     },
   };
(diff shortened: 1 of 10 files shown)


li-kai (commented):
Hi! Thanks for submitting the PR, the code looks good to me.

Enabling custom removal/addition of lessons is such a powerful utility that we want to make sure users wield it right. That means we have to make it obvious which lessons were added/removed, and also the ability to undo them. The consequence of a someone missing lessons/lectures for a semester is pretty big.

As a team we just couldn't come up with a good solution. Ultimately NUSMods is more of a planner than a calendar, so we've been encouraging users customize their schedule via calendar apps instead.

Despite that, it does seem that demands for such a feature never back down. If you're keen on building it, could you propose how your TA-mode would work? Once the details are ironed out, you'll have the go-ahead.

zhoukerrr:
@li-kai This is how I plan to do this for now.

UI side:
* A new button with the setting icon will be added next to the delete/hide button for each of the modules at the bottom
* To avoid accidental modification, a pop up window will display a warning 
* If the user chooses to proceed, the timetable will enter the state where all possible slots for that module are now selectable/removable
* At this point, buttons next to the module below will turn into a **DONE** button, this will end the customisation mode
* A special symbol/word (e.g. **TA**) could be placed on **ALL** of the slots for modules to emphasis that the module has been customised

Code side:
* Implement the UI mentioned above
* A new redux state that stores the modules that are customised will be added.
* Upon entering the customisation state, classes will be added/removed from the array done in this PR

Let me know your thoughts on this approach. Feel free to share any ideas/comments that you have. Thank you!

li-kai:
On the UI, it might be better to collapse all 3 into a dropdown so we have:
    - Hide All
    - Delete All
    - Edit Individually <- this triggers modal

But the rest of the details sound good to me. @chrisgzf do you have bandwidth to review the PRs?

zhoukerrr:
Hi @li-kai @chrisgzf, may I know if this PR can be merged first is that I can push out my next PR? I am almost done with the changes but I do not want to send in a giant PR that is difficult to review. Thank you!

chrisgzf (changes requested):
Hi @zhoukerrr, changes look okay to me but one more thing. Could you:
- increment the redux-persist schema version in `timetables.ts:25`
- write a migration in the same `persistConfig` to migrate the users' existing data to the new data shape?

This prevents this change from breaking the user's existing timetable data, which will be very messy.

zhoukerrr:
@li-kai @chrisgzf Hi! May I know if this PR can be merged?

website/src/reducers/timetables.test.ts
```diff
@@ -239,3 +239,56 @@ describe('stateReconciler', () => {
...
+          Lecture: '1',
+        },
+      },
+    },
+    colors: {},
+    hidden: {},
+    academicYear: '2022/2023',
+    archive: {},
+    _persist: {
+      version: 2,
```
chrisgzf, line 263: should this be `1`?

website/src/reducers/timetables.test.ts
```diff
@@ -239,3 +239,56 @@ describe('stateReconciler', () => {
...
+      rehydrated: false,
+    },
+  };
+  test('should migrate from V1 to V2', () => {
+    expect(migrateV1toV2(reduxDataV1)).toEqual(reduxDataV2);
+  });
```
chrisgzf, line 293: Thanks for writing tests for the migration!

website/src/reducers/timetables.test.ts
chrisgzf: I think the version incrementing is done by the framework and not your function,  so if thats the case I think we can just leave a comment here so that it isn't confusing to others (or just remove the `_persist` key all together (though I guess that might give you type errors).

chrisgzf (approved):
Some last queries else LGTM

website/src/reducers/timetables.test.ts
zhoukerrr: @chrisgzf yes. Removing `_persist` here will give type error. This is just for the purpose of testing. I had to keep the value identical here to pass the test because in the unit test here the framework doesn't do the magic for us. I will make a comment next to it to avoid any future confusion.

Issue #3422, opened 2023-01-11
Sidebar clips through colour selection


RyanCheungJF:
### Describe the bug

When choosing a colour for the module, there are times where the sidebar on the left overlays and causes users to redirect when choosing a colour. (Screenshot might provide more details)

### To Reproduce

1. Add module, try on the leftmost module
2. Scroll till one of the sidebar components overlap
3. Open colour option

### Expected behavior

User should be able to choose the colour instead of clicking the sidebar

### Screenshots

![image](https://user-images.githubusercontent.com/63341400/211819762-7897a036-58e2-4a56-864f-a54366ad13a1.png)

Can't see my mouse pointer, but im actually hovering over the red colour box with the cross, but 'Contribute' is being selected here instead of red when i click on the colour.

### Desktop (please complete the following information):

- OS: Windows
- Browser Firefox
- Version 105.0.3

### Smartphone (please complete the following information):

NA

### Additional context

My laptop's screen size might be small at 1366 x 768 leading to this issue. On my desktop it is completely fine! :-)

Not a huge bug to solve but just wanted to raise it! Thank you for all your hard work!

PR #3424, opened 2023-01-18
fix: color selection dropdown overlaps the sidebar

Kaby-Lake:
## Context

The following change is meant to address and resolve issue #3422 where colour selection dropdown overlaps the sidebar.

Before:
image

After: 
image

## Implementation
image
the z-index for dropdown should be some number greater than the navtabs.


li-kai (approved):
Lgtm. Thanks for the fix!

PR #3434, opened 2023-03-19
Add support for Timetable for TAs

### LINKED ISSUE #3404: NUSMODs for TAs

zhoukerrr:
## Context

Closes #3404 
## Implementation

UI side:
* A new button with the setting icon will be added next to the delete/hide button for each of the modules at the bottom
* Once the edit button is pressed, the timetable will enter the state where all possible slots for that module are now selectable/removable
* At this point, buttons next to the module below will turn into a **DONE** button, this will end the customisation mode
* A special symbol (*) is placed on **ALL** of the slots for modules to emphasis that the module has been customised

Code side:
* Implement the UI mentioned above
* A new redux state that stores the modules that are customised will be added.
* New redux actions needed for this state
## Other Information

It will only available in beta

### DIFF
--- a/website/src/actions/__snapshots__/timetables.test.ts.snap
+++ b/website/src/actions/__snapshots__/timetables.test.ts.snap
@@ -10,6 +10,7 @@ Object {
 exports[`changeLesson should return updated information to change lesson 1`] = `
 Object {
   "payload": Object {
+    "activeLesson": "1",
     "classNo": "1",
     "lessonType": "Recitation",
     "moduleCode": "CS1010S",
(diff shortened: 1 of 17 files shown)


chrisgzf (approved):
Thank you for your contribution!

Issue #3444, opened 2023-06-25
Hidden courses on Timetable builder are exported and visible when imported


HollaG:
### Is your feature request related to a problem? Please describe.

Hidden courses on NUSMods Timetables are shared even when they're hidden. When the shared link is opened, the hidden courses become visible, which may confuse people when they share their timetables with others -- the timetable will look fuller than it actually is

### Describe the solution you'd like
Hidden mods should not be visible when imported

### Describe alternatives you've considered

There are two solutions I thought of:
1) don't include hidden mods in the share URL generation
2) include the hidden mods in the share URL generation, but add a new query parameter which will be an array of the course codes that are hidden

### Additional context

Reproducing - hide a course on timetable builder, export it and import in Incognito - the hidden course is now visible

I'm able to implement the fix for this, depending on which solution is better, I'm not sure at the moment!

zwliew:
I think the 2nd solution would be a great addition :D

HollaG:
Alright, I'll attempt to do it now

kokrui:
Fixed in #3445

PR #3445, opened 2023-06-25
Hide hidden courses in shared link

HollaG:
## Context

This PR is meant to resolve issue #3444, to allow for the exporting and importing of hidden modules in NUSMods Timetables.

*Because the codebase still uses Module instead of Course, I will be using Module for this PR.

## Implementation

Implementing this turned out to be a bit bigger than expected.
There are 3 main parts:
1) Generating the shared link if there are hidden modules

- An example URL would be https://nusmods.com/timetable/sem-1/share?CS2040S=LAB:02,LEC:1&CS2103T=LEC:G22&ES2660=SEC:G21&hidden=CS2040S,ES2660

2) Parsing the imported URL link when loading the imported Timetable
3) Ensuring the hidden modules are copied over when the user decides to replace the Timetable.

1: Generating shared link
I put the task of adding on the &hidden=... in src/views/routes/path.ts. It just does a simple array join, since all module codes are alphanumeric.

2: Parsing imported URL
Since the imported timetable is already being parsed in TimetableContainer.tsx, I parse the imported hidden modules and dispatch an action that sets the value of `timetables.hidden[-1]` to the array of hidden modules.

3: Copying over modules
I modified setTimetable in actions/timetable.ts and in reducers/timetable.ts to set hidden[semester] to hidden[-1], and to delete hidden[-1].

## Other Information

It's been a while since I worked with Redux, so I apologise if I chose the wrong place to put functions etc.


kokrui:
Hi @HollaG, one of the new maintainers here -- thanks for this PR! I'll take this on and try to get it merged during recess week!

I think the overall implementation idea makes sense, and from a cursory look the PR looks pretty well-written (thanks for writing tests :D), so for any nits I find and for merge conflicts I'll just push directly to your branch and then approve + merge to master once it's done

kokrui (approved):
LGTM

PR #3447, opened 2023-06-29
feat: new NUSMods prerequisite tree parser using ANTLR

Fidget-Spinner:
## Context
NUS API changed their prerequisite tree format. They now have a custom DSL of sorts. This PR fixes the prerequisite tree for NUSMods.

## Implementation
I created a EBNF grammar file for the new DSL and used `antlr4ts` to generate a lexer plus parser. Then I implemented the visitor pattern to walk the AST and convert binary operations ('and', 'or') and `N_OF` operations to the prereq tree.

Tests pass on sample specifications given by NUS. I do not have API access to NUS so I cannot test on a wider range of courses.

## Other Information
This adds a new node to the prereq tree called `nOf` which will need to be supported.

I prefer the ANTLR parser generator approach as it's a lot more scalable should NUS change their DSL again. The current method using our own `chevrotain` DSL requires a significant amount of manual labour. Also it means we don't need to normalize stuff as the parser can handle everything (the entire normalize is gone).


chrisgzf (approved):
Thank you so much for working on this! And for squashing all the painful bugs with undocumented stuff! Really appreciate it!

PR #3450, opened 2023-07-03
Upgrade css loaders

li-kai:
## Context

Update css and postcss to latest versions so it doesn't break on node 18

## Implementation

1. Check for breaking changes
1. `yarn install && yarn test`

## Other Information

💅


taneliang (approved):
Thanks! It broke for me too lol

PR #3451, opened 2023-07-03
Upgrade circle ci node version

li-kai:
## Context

Use lts/galium which follows our .nvmrc setting

## Implementation

Change node version in circle ci config

## Other Information

⚫


taneliang (approved):
Thanks!

PR #3452, opened 2023-07-03
Bump prettier from 2.2.1 -> 2.8.8

taneliang:
Prettier 2.2.1 doesn't support TS's new `satisfies` keyword. Before we bump TS, we should bump Prettier first.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -3,6 +3,5 @@ module.exports = {
   semi: true,
   singleQuote: true,
   trailingComma: 'all',
-  jsxBracketSameLine: false,
   bracketSpacing: true,
 };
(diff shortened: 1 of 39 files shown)


.prettierrc.js
```diff
@@ -3,6 +3,5 @@ module.exports = {
   semi: true,
   singleQuote: true,
   trailingComma: 'all',
-  jsxBracketSameLine: false,
```
taneliang, line 6: This option is deprecated, but it defaulted to `false` anyway, so this can safely be deleted. https://prettier.io/docs/en/options.html#deprecated-jsx-brackets

li-kai (approved):

Issue #3459, opened 2023-07-09
Show total number of active units in Timetable


HollaG:
### Is your feature request related to a problem? Please describe.
When planning timetables, it might be helpful to see the total number of active units, instead of just the total number.

![image](https://github.com/nusmodifications/nusmods/assets/31369072/68bb8b9f-db8a-4ba5-8387-17eab84342ed)

For example, only 12 Units are active here.

### Describe the solution you'd like

Add an extra line above 'Total Units: 28 Units' to say 'Active Units: 12 Units'

### Describe alternatives you've considered

Possible to add as a bracket e.g. 'Total Units: 28 Units (12 active)' ?

### Additional context

Saw feature request from this [Reddit thread](https://www.reddit.com/r/nus/comments/14q5o1r/comment/jqlpxpf/?utm_source=share&utm_medium=web2x&context=3)

zwliew:
Resolved in #3460

PR #3460, opened 2023-07-09
New UI Feature: Show the # of active units in Timetable

HollaG:
## Context

Implements #3459 

In the recent NUSMods Ideas thread on Reddit, I saw someone suggest to implement a display for `Active Units` alongside the `Total Units` when viewing a timetable.
image

## Implementation

1. Passed `TimetableContent.tsx`'s `hiddenInTimetable` prop (array of module codes) to `ModulesTableFooter.tsx` as a new prop
2. Calculated the active unit count by filtering the list of modules (previously passed as prop), removing hidden modules
3. Ran the existing `renderMCs()` function to get the correct text output

### Video showing changes
https://github.com/nusmodifications/nusmods/assets/31369072/a26b1132-a9dc-49f0-a87e-ea58875ba046

### Mobile view
image

## Other Information

[Reddit thread](https://www.reddit.com/r/nus/comments/14q5o1r/comment/jqlpxpf/?utm_source=share&utm_medium=web2x&context=3)

### DIFF
--- a/website/src/views/timetable/ModulesTableFooter.tsx
+++ b/website/src/views/timetable/ModulesTableFooter.tsx
@@ -54,7 +54,7 @@ const ModulesTableFooter: React.FC<Props> = (props) => {
       </div>
       <div className="col">
         <div>
-          Active Units: <strong>{renderMCs(hiddenMCs)}</strong>
+          Active Units: <strong>{renderMCs(shownMCs)}</strong>
         </div>
         <div>
           Total Units: <strong>{renderMCs(totalMCs)}</strong>
--- a/website/src/views/timetable/ModulesTableFooter.tsx
+++ b/website/src/views/timetable/ModulesTableFooter.tsx
@@ -37,9 +37,10 @@ type Props = {
 
 const ModulesTableFooter: React.FC<Props> = (props) => {
   const totalMCs = sumBy(props.modules, (module) => parseFloat(module.moduleCredit));
-  const shownMCs = props.modules
-    .filter((module) => !props.hiddenInTimetable.includes(module.moduleCode))
-    .reduce((acc, module) => acc + parseFloat(module.moduleCredit), 0);
+  const shownMCs = sumBy(
+    props.modules.filter((module) => !props.hiddenInTimetable.includes(module.moduleCode)),
+    (module) => parseFloat(module.moduleCredit),
+  );
 
   return (
     <div className={classnames(styles.footer, 'row align-items-center')}>
--- a/website/src/views/timetable/ModulesTableFooter.tsx
+++ b/website/src/views/timetable/ModulesTableFooter.tsx
@@ -53,9 +53,11 @@ const ModulesTableFooter: React.FC<Props> = (props) => {
         <hr />
       </div>
       <div className="col">
-        <div>
-          Active Units: <strong>{renderMCs(shownMCs)}</strong>
-        </div>
+        {shownMCs !== totalMCs && (
+          <div>
+            Active Units: <strong>{renderMCs(shownMCs)}</strong>
+          </div>
+        )}
         <div>
           Total Units: <strong>{renderMCs(totalMCs)}</strong>
         </div>
--- a/website/src/views/timetable/ModulesTableFooter.tsx
+++ b/website/src/views/timetable/ModulesTableFooter.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import classnames from 'classnames';
-import { filter, map, sumBy } from 'lodash';
+import { map, sumBy } from 'lodash';
 import { connect } from 'react-redux';
 
 import { ModuleTableOrder } from 'types/reducers';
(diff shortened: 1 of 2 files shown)


website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -30,13 +30,17 @@ type Props = {
...
+  hiddenInTimetable: string[];
 
   setModuleTableOrder: (moduleTableOrder: ModuleTableOrder) => void;
 };
 
 const ModulesTableFooter: React.FC<Props> = (props) => {
   const totalMCs = sumBy(props.modules, (module) => parseFloat(module.moduleCredit));
-
+  const hiddenMCs = sumBy(
+    filter(props.modules, (module) => !props.hiddenInTimetable.includes(module.moduleCode)),
```
zwliew, line 41: Would prefer not to introduce yet another dependency on lodash. Can we use `props.modules.filter` instead? Also, would prefer to use `.reduce()` instead of `sumBy()`, like so:

```typescript
  const hiddenMCs = props.modules.filter((module) => ...).reduce((acc, module) => ...);
``` 

IMO, Lodash should really not be used much anymore nowadays, given how many of its features have been present in browsers for years.

website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -30,13 +30,17 @@ type Props = {
...
   moduleTableOrder: ModuleTableOrder;
   modules: Module[];
+  hiddenInTimetable: string[];
 
   setModuleTableOrder: (moduleTableOrder: ModuleTableOrder) => void;
 };
 
 const ModulesTableFooter: React.FC<Props> = (props) => {
   const totalMCs = sumBy(props.modules, (module) => parseFloat(module.moduleCredit));
-
```
zwliew, line 39: nit: Could you re-add the newline after the variables declarations? Nice to have it to visually separate the declarations from the JSX

website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -49,8 +53,14 @@ const ModulesTableFooter: React.FC<Props> = (props) => {
...
       <div className="col">
-        Total Units: <strong>{renderMCs(totalMCs)}</strong>
+        <div>
+          Active Units: <strong>{renderMCs(hiddenMCs)}</strong>
+        </div>
+        <div>
+          Total Units: <strong>{renderMCs(totalMCs)}</strong>
+        </div>
       </div>
+
```
zwliew, line 63: nit: redundant newline

zwliew (commented):
Looks good! Have a few minor comments

website/src/views/timetable/ModulesTableFooter.tsx
HollaG: Noted! I thought it would be best to follow what was used in line 39 haha. Would you like me to change line 39 as well or is that better served in another PR?

HollaG:
@zwliew I've applied your requests! Didn't change `totalMCs` though, if you'd like me to change that too, I can

website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -49,7 +53,12 @@ const ModulesTableFooter: React.FC<Props> = (props) => {
         <hr />
       </div>
       <div className="col">
-        Total Units: <strong>{renderMCs(totalMCs)}</strong>
+        <div>
+          Active Units: <strong>{renderMCs(hiddenMCs)}</strong>
+        </div>
```
ZhangYiJiang, line 58: To avoid confusing users who don't hide modules, I would consider hiding this when there are no hidden modules. It also provides no useful information in that case

website/src/views/timetable/ModulesTableFooter.tsx
ZhangYiJiang: I think we should actually not use reduce. Reduce can be used for almost any array operation, but it is less clear precisely because it can represent any operation, so you have to read the inner logic to determine what is actually happening. sumBy is more readable since it's clear this is just adding numbers up. I would only avoid lodash when the available native function does the same thing.

website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -30,12 +30,16 @@ type Props = {
...
   moduleTableOrder: ModuleTableOrder;
   modules: Module[];
+  hiddenInTimetable: string[];
 
   setModuleTableOrder: (moduleTableOrder: ModuleTableOrder) => void;
 };
 
 const ModulesTableFooter: React.FC<Props> = (props) => {
   const totalMCs = sumBy(props.modules, (module) => parseFloat(module.moduleCredit));
+  const hiddenMCs = props.modules
```
ZhangYiJiang, line 40: This is MCs of modules _not_ hidden 

```suggestion
  const shownMCs = props.modules
```

website/src/views/timetable/ModulesTableFooter.tsx
zwliew: Fair enough. Personally, this use case is simple enough that it would take more time to look up what `sumBy` does than to read `.reduce(x => x + y, 0)`. I don't have strong opinions on this though so `sumBy` is fine too.

website/src/views/timetable/ModulesTableFooter.tsx
HollaG: I didn't consider that, will do!

website/src/views/timetable/ModulesTableFooter.tsx
HollaG: Hi @ZhangYiJiang , may I check if I still keep Array.filter() so that we don't introduce another import from Lodash? That's how I've done it in the latest commit. Thanks & sorry for all the trouble!

HollaG:
New behaviour with hiding when no hidden modules:

https://github.com/nusmodifications/nusmods/assets/31369072/c8158e4c-6591-478f-aaac-a27058b5928e

zwliew (approved):

zwliew:
@ZhangYiJiang do you think there's anything else to work on for this PR? Otherwise, I think it's ready to be merged.

zwliew:
Thanks again for this!

PR #3467, opened 2023-07-27
Include colors and theme when sharing timetable

### LINKED ISSUE #3423: Colours are not preserved when sharing timetables
When sharing timetables to different devices via QR code, the colours assigned to each module are not preserved across devices.

leslieyip02:
## Context

Resolves #3423 by allowing module colors and theme to be shared via URL.

## Implementation

Module `colorMapping` and `themeId` are passed into the query parameters.

e.g.

```
modules = {
   CS2104: { Lecture: '1', Tutorial: '2' },
   CS2107: { Lecture: '1', Tutorial: '8' },
}

colors = {
   CS2104: 3,
   CS2107: 4,
}

themeId = 'monokai'

=> CS2104=LEC:1,TUT:2,COL:3&CS2107=LEC:1,TUT:8,COL:4&THEME=monokai
```

https://github.com/nusmodifications/nusmods/assets/90888680/5ef8e2a5-0cae-48dd-b695-87ddd258e3d5


website/src/utils/timetables.ts
```diff
@@ -434,18 +443,68 @@ export function formatNumericWeeks(weeks: NumericWeeks): string | null {
...
+      if (colors && moduleCode in colors) {
+        moduleParams[COLOR_PARAM] = colors[moduleCode];
+      }
+      return [moduleCode, moduleParams];
+    }),
+  );
+  const timetableParams = {
+    ...mapValues(timetableModuleParams, serializeModuleConfig),
+    ...((themeId && { [THEME_PARAM]: themeId }) || EMPTY_OBJECT),
+  };
```
ZhangYiJiang, line 468: This is a bit easier to read, but also use `?:` ternary to represent ternary logic instead of `&& ||`

```suggestion
  const timetableParams = mapValues(timetableModuleParams, serializeModuleConfig);
  if (themeId != null) {
    timetableParams[THEME_PARAM] = themeId };
  }
```

website/src/utils/timetables.ts
```diff
@@ -434,18 +443,68 @@ export function formatNumericWeeks(weeks: NumericWeeks): string | null {
...
   // We are using query string safe characters, so this encoding is unnecessary
-  return qs.stringify(mapValues(timetable, serializeModuleConfig), { encode: false });
+  return qs.stringify(timetableParams, { encode: false });
 }
 
+// Extracts module configs from query string
 export function deserializeTimetable(serialized: string): SemTimetableConfig {
```
ZhangYiJiang, line 474: It's best to rewrite this so that it parses from query string, then delegate to objects to extract individual parts instead of having three functions each re-parse the query string.

website/src/views/timetable/ShareTimetable.tsx
```diff
@@ -34,8 +38,13 @@ type State = {
   shortUrl: string | null;
 };
 
-function shareUrl(semester: Semester, timetable: SemTimetableConfig): string {
-  return absolutePath(timetableShare(semester, timetable));
+function shareUrl(
+  semester: Semester,
+  timetable: SemTimetableConfig,
+  colors?: ColorMapping | null,
+  themeId?: ThemeId | null,
```
ZhangYiJiang, line 45: Optional parameters should not be `| null` 

```suggestion
  colors?: ColorMapping,
  themeId?: ThemeId,
```

website/src/views/timetable/TimetableContainer.tsx
```diff
@@ -156,7 +161,18 @@ export const TimetableContainerComponent: FC = () => {
...
+    if (semester && importedColors) {
+      Object.entries(importedColors).forEach((colorMapping: [string, ColorIndex]) => {
+        const [moduleCode, colorIndex] = colorMapping;
+        dispatch(selectModuleColor(semester, moduleCode, colorIndex));
+      });
+    }
+    const importedThemeId = deserializeTimetableThemeId(location.search);
+    if (semester && importedThemeId) {
+      dispatch(selectTheme(importedThemeId));
+    }
```
ZhangYiJiang, line 174: Note that imported timetable is temporary - the user has to confirm before the existing timetable is overwritten. This makes the imported module colour and theme take effect immediately when opening the URL, which is inconsistent with the timetable behavior

website/src/utils/timetables.ts
```diff
@@ -434,18 +443,68 @@ export function formatNumericWeeks(weeks: NumericWeeks): string | null {
...
+        if (!serializedModule) return;
+        serializedModule.split(QUERY_PARAM_SEP).forEach((param) => {
+          const [paramName, paramValue] = param.split(QUERY_PARAM_VALUE_SEP);
+          if (paramName === QUERY_PARAM_ABBREV[COLOR_PARAM]) {
+            colorMapping[moduleCode] = parseInt(paramValue, 10);
+          }
+        });
+      });
+    });
+  if (isEqual(colorMapping, EMPTY_OBJECT)) return null;
```
ZhangYiJiang, line 499:
```suggestion
  if (isEmpty(colorMapping)) return null;
```

website/src/utils/timetables.test.ts
```diff
@@ -402,6 +404,68 @@ test('timetable serialization/deserialization', () => {
   });
 });
 
+test('timetable serialization/deserialization with colors', () => {
+  const configs: SemTimetableConfig[] = [
```
ZhangYiJiang, line 408: Use `test.each` https://jestjs.io/docs/api#each instead for testing multiple test cases

website/src/utils/timetables.ts
```diff
@@ -434,18 +443,68 @@ export function formatNumericWeeks(weeks: NumericWeeks): string | null {
...
+  const colorMapping: ColorMapping = {};
+  Object.keys(params)
+    .filter((param) => param !== THEME_PARAM)
+    .forEach((moduleCode) => {
+      castArray(params[moduleCode]).forEach((serializedModule) => {
+        if (!serializedModule) return;
+        serializedModule.split(QUERY_PARAM_SEP).forEach((param) => {
+          const [paramName, paramValue] = param.split(QUERY_PARAM_VALUE_SEP);
+          if (paramName === QUERY_PARAM_ABBREV[COLOR_PARAM]) {
+            colorMapping[moduleCode] = parseInt(paramValue, 10);
```
ZhangYiJiang, line 494: Since colours are bound to a sub-param inside each module code, this is better extracted as part of deserializeTimetable instead

website/src/utils/timetables.test.ts
```diff
@@ -402,6 +404,68 @@ test('timetable serialization/deserialization', () => {
...
+    CS1010S: 0,
+    GER1000: 1,
+    CS2104: 2,
+  };
+
+  configs.forEach((config) => {
+    // if mod config is not present, no color mapping is serialized
+    const colors = Object.fromEntries(
+      Object.entries(colorMapping).filter(([moduleCode]) => moduleCode in config),
+    );
```
ZhangYiJiang, line 435: _.filter can operate on objects

ZhangYiJiang (changes requested):
Sorry the previous review submitted early. The serialization and deserialization looks good, but the changes cannot be applied immediately when the import URL is opened if we want to match existing import behavior

website/src/views/timetable/TimetableContainer.tsx
leslieyip02: https://github.com/nusmodifications/nusmods/assets/90888680/f108db6a-cddf-44e5-ba51-d8cffc156b74

Is this better?

leslieyip02:
I wanted to implement undoing colors and theme imports, but it seems that the "undo" button in the notifications isn't working. I have opened a new issue for that: #3475

leslieyip02:
> I wanted to implement undoing colors and theme imports, but it seems that the "UNDO" button in the notifications isn't working. I have opened a new issue for that: #3475.

The issue has been resolved. Imported colors can be undone, but the theme does not become undone. I'm not sure why the theme is not stored in the history. I'm guessing it has something to do with this:

https://github.com/nusmodifications/nusmods/blob/8ebbf2fd7d3751a04069ddfd9b357cc6c4e1e180/website/src/views/timetable/TimetableContainer.tsx#L50

Issue #3466, opened 2023-07-27
Open mods and venues in new tab in global search


jloh02:
### Is your feature request related to a problem? Please describe.

As a user, I am able to click on mods (for e.g. in timetable) to navigate to the module's page. I can also open it in a new tab. However, when using global search, opening modules/venues in new tab is not possible.

### Describe the solution you'd like

- Switch from using React history router with downshift's `onChange` listener to react-dom `Link`
- This allows pushing of new routes without reloading the entire page
- Remove the callback prop from `GlobalSearchContainer` to `GlobalSearch` as `modulePage` and `venuePage` can be referred to directly in `Link`'s `to` property

### Describe alternatives you've considered

- Switch to `<a>` tag (Not ideal as reloads entire dom)

### Additional context
NIL

zwliew:
Resolved in #3468

PR #3468, opened 2023-07-28
Change global search dropdown to Link

jloh02:
## Context
Currently, global search results cannot be opened in a new tab. This is due to the implementation using downshift's onChange rather than `<a>` tags or `<Link>`
#3466 

## Implementation
By switching from using the `onChange` listener from downshift to a <Link> tag, we can allow users to open search results in new tabs like a normal `<a>` link. Using `<Link>` mimics the `router.push` behavior where other UI components are not reloaded when the link is clicked normally.

Current context menu only shows back
Screenshot (67)

Proposed changes allows opening in new tab
Screenshot (66)

## Other Information
- First contribution and first time using React component
- Not sure if using the <Link> tag affected the CSS styles in any other way. Looks visually similar on my end


website/src/views/layout/GlobalSearch.tsx
```diff
@@ -292,7 +277,6 @@ class GlobalSearch extends Component<Props, State> {
       <Downshift
         isOpen={isOpen}
         onOuterClick={this.onOuterClick}
-        onChange={this.onChange}
```
ZhangYiJiang, line 295: This breaks keyboard navigation. One way to fix this is to keep the original logic and add `onClick={(evt) => evt.preventDefault()}` to the anchors so they still use the same logic when clicked (we also don't need them to be `Link` in that case but that's not a big deal)

jloh02: Keyboard navigation fixed in c6d65db9430b0fd79d9c85eef455505e1f799020

zwliew (changes requested):

website/src/views/layout/GlobalSearch.scss
```diff
@@ -1,4 +1,4 @@
-@import '~styles/utils/modules-entry';
+@import "~styles/utils/modules-entry";
```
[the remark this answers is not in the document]
jloh02: e7b177a827afcd601c3456498d166e1b9e2c0270

website/src/views/layout/GlobalSearch.tsx
```diff
@@ -35,6 +26,25 @@ type State = {
...
+    case SEARCH_RESULT: {
+      // TODO: Move this into a proper function
+      const path = item.result === VENUE_RESULT ? '/venues' : '/courses';
+      return `${path}?q=${encodeURIComponent(item.term)}`;
+    }
+
+    default:
+      throw Error(`Unknown item type: ${item}`);
+  }
+};
```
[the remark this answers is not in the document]
jloh02: Not sure why but a few weeks back the linter gave me warnings requiring me to put a `default`. Anyways resolved in 2acbd15ba237914a47876e948518b806c3921b7d

zwliew (approved):

website/src/views/layout/GlobalSearch.tsx
```diff
@@ -35,6 +26,27 @@ type State = {
...
+
+    case SEARCH_RESULT: {
+      // TODO: Move this into a proper function
+      const path = item.result === VENUE_RESULT ? '/venues' : '/courses';
+      searchUrl = `${path}?q=${encodeURIComponent(item.term)}`;
+      break;
+    }
+  }
+  return searchUrl;
+};
```
zwliew, line 48: It's cases like this when I think the `consistent-return` rule should be disabled, but eh -- this is fine. :)

zwliew:
Thank you!

PR #3499, opened 2023-08-05
fix(export): watch the correct output directory for index.js

### LINKED ISSUE #3455: /exports build structure does not match historical output
### Describe the bug

[/export](https://github.com/nusmodifications/nusmods/tree/master/export) needs some fixing:
the commands and docs all point to build/index.js. But when you run `yarn build`, you get `build/api` and `build/src`. I think previously `/src` files used to be on the `build/` root, but it got moved into `/src` at some point.

Unfortunately I'm not familiar with any of the code, nor the vercel infra, so I can only help up to here.

cc: @chrisgzf 

### To Reproduce

1. Go to '/export`
2. Run `yarn build`
3. Look at `/build` structure

### Expected behavior

Either:
1. Commands and docs need to point to `build/src/index.js`, or
2. `/api` needs to be nested under `/src`, with the rootDir set to `./src` in `tsconfig.json`.

### Additional context

This problem is reproducable at https://github.com/nusmodifications/nusmods/commit/73ec10b0126f3a94cd4c094f7a8bce6c266405f0, before any of the node/ts upgrades.

zwliew:
Closes #3455.

This doesn't affect production since we use Serverless Functions on Vercel, but it affects local development.


chrisgzf (approved):
cool didnt actually try to dev this locally before so didnt realise this was a problem. LGTM

PR #3519, opened 2023-08-08
ModuleTree: improve the UI of the prereq tree

### LINKED ISSUE #3469: Prettify unwieldy phrasing in prereq tree

The current prereq tree looks unwieldy in cases where courses are prefixed with `minimally` and `course starting with`.

Screenshot 2023-07-28 at 5 56 40 PM

Screenshot 2023-07-28 at 5 56 20 PM

We could modify the CSS of the prereq tree item to make `minimally` and `course starting with` appear as follows:

Screenshot 2023-07-28 at 6 01 10 PM

I've discussed two alternatives with @chrisgzf, namely:
1. Appending "X"s to the course code of prereq tree items that begins with `course starting with` (i.e. `MA11X`, `MA11XX`)
2. Same as the first, but with "*"s instead of "X"s (i.e. `MA11*`, `MA11**`)

In general, both solutions aren't optimal as they introduce ambiguity. For the first option, there exist course codes which ends with an "X". As for the second option, it's unclear if "*" would be correctly interpreted to mean what we intend. This leaves us with the proposed solution, which seeks to avoid these ambiguities by retaining the prefixes for the clarity they provide, while making them aesthetically more compact for ease of readability.

zwliew:
## Context

Resolves #3469.

## Implementation

Define a new `prefix` element that we style and render separately from the usual course name.

**Important change:** nodes on the same level are no longer shrunk to the same width. This ended up looking pretty good in my opinion.

### DIFF
--- a/website/src/utils/react.tsx
+++ b/website/src/utils/react.tsx
@@ -43,6 +43,15 @@ export function replaceWithNode(
   );
 }
 
+export function replaceAllWithNode(
+  str: string,
+  regex: RegExp,
+  replacement: (match: string, index: number) => React.ReactNode,
+): React.ReactElement {
+  const matches = str.match(regex);
+  return <>{matches?.length ? replacement(matches[0], 0) : str}</>;
+}
+
 export function highlight(str: string, search: string | string[], Tag = 'mark'): React.ReactNode {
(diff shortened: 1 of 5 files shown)


zwliew:
image
image
image

Issue #3522, opened 2023-08-08
Module terminology still present archived modules banner


wongzw:
### Describe the bug

Module terminology still observed in banner for archive modules. It should be referred as a course instead of a module for consistency with the rest of the website. 

### To Reproduce

1. Go to any module that is not being offered this AY (e.g. https://nusmods.com/archive/MKT4813/2020-2021/design-thinking-business-innovations)

### Expected behavior

The banner should read "You are looking at archived information of this **course**" instead of "You are looking at archived information of this module"

### Screenshots

![image](https://github.com/nusmodifications/nusmods/assets/53685088/60db5eda-5364-4143-86b5-1be23afebf93)

zwliew:
Thanks for reporting this!

zwliew:
Related issue: 
![Screenshot 2023-08-09 at 12 37 28 AM](https://github.com/nusmodifications/nusmods/assets/4538946/2cf596f3-0bb9-44ee-80c2-5191cda1a7c6)

zwliew:
Resolved

PR #3542, opened 2023-08-11
Update module terminology to course

jloh02:
## Context
As per #3522, there are areas on the website which label courses as modules and this does not follow the new conventions.

## Other Information
Linked issue only specifies one page but upon searching, there are a number of files that also contain this inconsistency in naming and have updated this. Did not touch the MPE related files despite numerous mentions of "modules".

### DIFF
--- a/website/src/views/errors/ModuleNotFoundPage.tsx
+++ b/website/src/views/errors/ModuleNotFoundPage.tsx
@@ -29,7 +29,7 @@ type Props = OwnProps & {
 export class ModuleNotFoundPageComponent extends PureComponent<Props> {
   override componentDidMount() {
     Sentry.withScope(() => {
-      Sentry.captureMessage('404 - Course Not Found');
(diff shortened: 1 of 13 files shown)


website/src/views/errors/ModuleNotFoundPage.tsx
```diff
@@ -29,7 +29,7 @@ type Props = OwnProps & {
 export class ModuleNotFoundPageComponent extends PureComponent<Props> {
   override componentDidMount() {
     Sentry.withScope(() => {
-      Sentry.captureMessage('404 - Module Not Found');
+      Sentry.captureMessage('404 - Course Not Found');
```
zwliew, line 32: I'd rather not change this because it affects analytics.

zwliew (changes requested):
Thanks for taking this up! I have just one minor comment.

zwliew (approved):

Issue #3449, opened 2023-07-02
Possible issue in exam clash checking function?


weiseng18:
### Describe the bug

CM2112 has a 120 minute exam according to the module list data from [here](https://api.nusmods.com/v2/2023-2024/modules/CM2112.json). However, the timetable does not seem to detect an exam clash.

![image](https://github.com/nusmodifications/nusmods/assets/20338724/95c7d3e1-0b94-4e13-b3d0-c4519c2456c3)

From my brief understanding, CM2112 is a core module for Chemistry majors, and HSI2001 can be used to fulfil Scientific Inquiry 2, so hopefully this is a reasonable combination.

### To Reproduce

Add CM2112 and HSI2001 to the timetable in Semester 2.

### Expected behavior

Exam clash warning should appear

### Additional context

From briefly looking through the code, it seems that this code is responsible for detecting exam clashes.

```js
// Find all exam clashes between modules in semester
// Returns object associating exam dates with the modules clashing on those dates
export function findExamClashes(modules: Module[], semester: Semester): ExamClashes {
  const groupedModules = groupBy(modules, (module) =>
    get(getModuleSemesterData(module, semester), 'examDate'),
  );
  delete groupedModules.undefined; // Remove modules without exams
  return omitBy(groupedModules, (mods) => mods.length === 1); // Remove non-clashing mods
}
```

From what I understand, this function considers a clash to be two exams starting at the same time. This is usually a reasonable assumption assuming that exams start at a fixed time (9am/1pm/5pm or 9am/2.30pm), but it appears that there may be counterexamples, so it is important to check the exam duration as well.

Another test case: SH5110 (5pm) and FE5222 (7pm) in Sem 1 but SH5110 has a 2.5 hour exam.

weiseng18:
Implementation idea:

Note that `examDate` is a string like this: `2023-09-23T08:00:00.000Z` and `examDuration` is a number like this: `120`

1. Convert `examDate` to GMT+8 (to make it easier to detect clash)
2. Perform `groupBy` on a key containing `examDate` and `examDuration`, e.g. `` `${examDate}_${examDuration}` ``
3. For each key like `${date}T${time}Z_${duration}`, extract `date`, `time`, `duration`.
4. For each date
    - For every pair of time and duration, form the time interval
    - We can check for interval overlaps using `.getTime()` and inequalities
[...]

PR #3560, opened 2023-08-12
Add additional check for findExamClashes() for non-identical start times

EssWhyy:
## Context

Fix issue #3449

## Implementation

Per the discussion in the issue, I added in an extra check in findExamClashes() in timetables.ts. The steps for this check are as follows:

1. Group modules in timetable by exam date only 
2. For each array of modules[] assigned to an exam date, we do a brute force O(n^2) check on each unique pair of modules 
3. If two modules have the same exam start date and time, then we ignore since its already accounted for in the existing clash check
4. If two modules with different exam start time has clash based on time intervals, then we add it into the existing clashes dictionary via the dateTime string key 

The result can be seen here for both test cases:

case1
case2

## Other Information

The current exam clash warning text can only show one date & time, as such for two exams with different start times, I used the earlier start time for the warning text as a workaround for now.

The function could probably be less verbose and more efficient. But on testing my end, there are no major slowdowns on the performance when adding or deleting a typical number of modules to the timetable.


kokrui (approved):
Thanks @EssWhyy for getting this started! It's an important change and it's ready to land now with some algo changes + tests

PR #3611, opened 2023-08-18
Fix `run-p` script glob on `yarn` 2+

SuibianP:
## Context

Yarn 2+ expands glob patterns using its own shell and fails the `lint` script.

```console
$ yarn --version
3.6.2
$ yarn run ci
No matches found: "lint:**"
```

No issue created due to its trivial nature.

## Implementation

Quote the pattern to prevent premature expansion. Escaped double quotes are used for compatibility with `cmd.exe`.

## Other Information

Related-Bug: mysticatea/npm-run-all#200 yarnpkg/berry#1814

Does the project officially support/plan to migrate to yarn 2+? If so, would it be better if such configuration is added to the test matrix?


zwliew (approved):

zwliew:
Thank you! I haven't looked into yarn 2 yet but I'm open to the idea of upgrading to it. Someone just has to do it :P

PR #3622, opened 2023-10-01
Add standalone CPEx scraper

chrisgzf:
## Context

This PR adds the standalone MPE scraper I've started using since MPE from the last semester.

### Why is this needed?

Prior to 2310 MPE, the MPE scraper was coupled with the NUSMods scraper's acad year.

Meaning, in AY2022/2023 S1, when the NUSMods scraper is set to AY2022/2023 S1, we can scrape MPE data for AY2022/2023 S2, which works fine.

However, in AY2022/2023 S2, when the NUSMods scraper is set to AY2022/2023 S2, we cannot scrape MPE data for AY2023/2024 S1, **because it is a different acad year**. We could just run a separate instance of the NUSMods scraper for AY2023/2024 yes, but because the module data for AY2023/2024 will not be ready by then, the scraper (where we built in a lot of validation) will error out.

The solution, is hence, to build a very simple, standalone MPE scraper, that is not coupled with the NUSMods scraper.

Note: this is definitely not the cleanest, but it's simple, it works, and it does its job well.


ravern (approved):
Tested and LGTM!

ravern:
This doesn't

1. Remove the old MPE scraper, which should still produce `mpeModules.json` in the semester-level directory
2. Modify the frontend to use the new `cpexModules.json` file

scrapers/cpex-scraper/src/index.ts
```diff
@@ -1,11 +1,13 @@
 import axios from 'axios';
-import env from '../env.json';
 import fs from 'fs';
 import path from 'path';
 
+import env from '../env.json';
+
 // Configure this!
 const term = '2310';
```
chrisgzf, line 8: Need to set this to `2320`

scrapers/cpex-scraper/src/index.ts
```diff
@@ -124,70 +127,73 @@ const scraper = async () => {
...
-      if (collatedMpeModulesMap.has(moduleCode)) {
+      if (collatedCPExModulesMap.has(moduleCode)) {
         continue;
       }
 
       const moduleCredit = module.ModularCredit;
-      const mpeAttribute = module.ModuleAttributes.find(
-        (attribute) => attribute.CourseAttribute === 'MPE',
+      const cpexAttribute = module.ModuleAttributes.find(
+        (attribute) => attribute.CourseAttribute === 'MPE', // this still isn't changed to CPEx
```
chrisgzf, line 136: wtf turns out querying CS2100 for 2320 gives us this for ModuleAttributes
```
            "ModuleAttributes": [
                {
                    "CourseAttributeValue": "YES",
                    "CourseAttribute": "MOEF"
                },
                {
                    "CourseAttributeValue": "S1&S2",
                    "CourseAttribute": "MPE"
                },
                {
                    "CourseAttributeValue": "YES",
                    "CourseAttribute": "PRQN"
                },
                {
                    "CourseAttributeValue": "NO",
                    "CourseAttribute": "SSGF"
                }
            ],
```

why cant they just be consistent? now im wondering if this caused regressions in the NUSMods scraper because we didnt know they changed to `CourseAttributes` on the API

chrisgzf (commented):
Some changes

Issue #3623, opened 2023-10-01
CPE 2320 - Update frontend strings and enable MPE on staging


chrisgzf:
1. For everything that is visible to the user (i.e. any strings on the frontend):
- `MPE` -> `CPEx`
- `Module Planning Exercise` -> `Course Planning Exercise`

2. Check all strings and URLs are accurate

3. Set the `enableMpe` flag to true in `website/src/featureFlags.ts`

4. Deploy to the MPE staging site (go find out the URL in Vercel), the staging site already has all the environment variables set correctly.

5. Test in MPE staging site

PR #3626, opened 2023-10-03
Change MPE to CPE and Open MPE (#3623)

alissayarmantho:
## Context

#3623 
Open MPE and Change MPE strings to CPEx


alissayarmantho:
No longer needed. Will merge cpex-staging instead

Issue #3625, opened 2023-10-01
CPE Tech Debt - Remove generation of MPE modules from NUSMods scraper


chrisgzf:
After #3622, we now have a dedicated MPE scraper.

NUSMods scraper no longer needs to read and generate the `mpeModules.json` file. We should remove the parts of NUSMods scraper that generates that file.

PR #3635, opened 2023-10-14
Merge CPEx Staging (#3623))

alissayarmantho:
## Context

Merging for CPEx (#3623)

### DIFF
--- a/scrapers/nus-v2/src/tasks/CollateModules.ts
+++ b/scrapers/nus-v2/src/tasks/CollateModules.ts
@@ -135,8 +135,8 @@ const getModuleMPEParticipation = ({
   title,
   moduleCode,
   moduleCredit,
-  inS1MPE: attributes?.mpes1,
-  inS2MPE: attributes?.mpes2,
+  inS1CPEx: attributes?.mpes1,
+  inS2CPEx: attributes?.mpes2,
 });
 
 // Avoid using _.pick here because it is not type safe
--- a/scrapers/nus-v2/src/types/mpe.ts
+++ b/scrapers/nus-v2/src/types/mpe.ts
@@ -5,6 +5,6 @@ export type MPEModule = Readonly<{
   title: ModuleTitle;
   moduleCode: ModuleCode;
   moduleCredit: string;
-  inS1MPE?: boolean;
-  inS2MPE?: boolean;
+  inS1CPEx?: boolean;
+  inS2CPEx?: boolean;
 }>;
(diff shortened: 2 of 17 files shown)


scrapers/nus-v2/src/tasks/CollateModules.ts
```diff
@@ -135,8 +135,8 @@ const getModuleMPEParticipation = ({
   title,
   moduleCode,
   moduleCredit,
-  inS1MPE: attributes?.mpes1,
-  inS2MPE: attributes?.mpes2,
+  inS1CPEx: attributes?.mpes1,
+  inS2CPEx: attributes?.mpes2,
```
chrisgzf, line 139: just a quick note that this is actually not necessarily cos this is part of the old scraper that is alr unused refer to #3625

scrapers/nus-v2/src/types/mpe.ts
```diff
@@ -5,6 +5,6 @@ export type MPEModule = Readonly<{
   title: ModuleTitle;
   moduleCode: ModuleCode;
   moduleCredit: string;
-  inS1MPE?: boolean;
-  inS2MPE?: boolean;
+  inS1CPEx?: boolean;
+  inS2CPEx?: boolean;
```
chrisgzf, line 9: same as above

website/src/apis/nusmods.js
```diff
@@ -82,7 +82,7 @@ class NUSModsApi {
    * @returns {string}
    */
   static mpeModuleListUrl() {
-    return `${NUSModsApi.baseUrl('')}mpeModules.json`;
+    return `${NUSModsApi.baseUrl('')}cpexModules.json`;
```
chrisgzf, line 85: i find it quite funny that we renamed it to `cpexModules` but nobody thought to rename it to `cpexCourses` 😆

website/src/views/routes/Routes.tsx
```diff
@@ -36,7 +36,7 @@ const Routes: React.FC = () => (
     <Route path="/venues/:venue?" component={VenuesContainer} />
     <Route path="/today" component={TodayContainer} />
     <Route path="/planner" component={PlannerContainer} />
-    <Route path="/mpe" component={MpeContainer} />
+    <Route path="/cpex" component={MpeContainer} />
```
chrisgzf, line 39: also adding another note here that changing routes is typically an operation that involves a little more thought. if you see above, when we changed from /modules to /courses, i actually left the /modules route to redirect to /courses, for search engine purposes.

i.e. after changing the routes, google doesn't scrape nusmods immediately after, so the results will still point to the old url, so the old url should still redirect to the new url.

so it would be preferable to leave the old route in.

but in this case its not a big problem because NUS will be sending the students the actual link, so we'll merge it in for now.

chrisgzf (approved):
LGTM, some non-blocking comments, otherwise good work!

Issue #3639, opened 2023-11-09
Autoscroll Course Planner to current year on load


jacobkwan:
Small QoL improvement as I always find myself having to scroll to the right to see my Year 3 and 4 modules when the page loads

PR #3640, opened 2023-11-09
Autoscroll Course Planner to current year card

jacobkwan:
## Context

#3639: Autoscroll Course Planner to current year on first render

## Implementation

1. Add `currentYearCardRef` state variable 
2. Only pass ref to the current PlannerYear card
3. Define `componentDidMount` method in `PlannerYear`
    - Uses `currentYearCardRef` to get parent container and calculate offsets for child and parent
    - scrolls the desired amount (with `smooth` effect)

## Before
https://github.com/nusmodifications/nusmods/assets/77665182/73a54f87-4da6-43ed-9bac-267ef656a392

## After
https://github.com/nusmodifications/nusmods/assets/77665182/2fa24e4a-0337-4eea-8d34-39bcde10d12f

### DIFF
--- a/website/src/views/planner/PlannerYear.tsx
+++ b/website/src/views/planner/PlannerYear.tsx
@@ -1,4 +1,4 @@
-import { PureComponent } from 'react';
+import { PureComponent, Ref, createRef } from 'react';
 import classnames from 'classnames';
 import { flatMap, size, sortBy, toPairs, values } from 'lodash';
 
@@ -24,12 +24,14 @@ type Props = Readonly<{
 
 type State = {
   readonly showSpecialSem: boolean;
+  readonly currentYearCardRef: Ref<HTMLDivElement>;
 };
 
 export default class PlannerYear extends PureComponent<Props, State> {
   override state = {
     // Always display Special Terms I and II if either one has modules
     showSpecialSem: this.hasSpecialTermModules(),
+    currentYearCardRef: createRef<HTMLDivElement>(),
   };
 
   hasSpecialTermModules() {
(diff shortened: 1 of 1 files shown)


website/src/views/planner/PlannerYear.tsx
```diff
@@ -24,14 +24,31 @@ type Props = Readonly<{
 
 type State = {
   readonly showSpecialSem: boolean;
+  readonly currentYearCardRef: Ref<HTMLDivElement>;
```
ZhangYiJiang, line 27: ref should not be stored in state, see https://react.dev/reference/react/createRef

website/src/views/planner/PlannerYear.tsx
```diff
@@ -24,14 +24,31 @@ type Props = Readonly<{
...
+      return;
+    }
+    const currentYearCard = this.state.currentYearCardRef.current;
+    const parentContainer = currentYearCard?.parentElement;
+    if (!currentYearCard || !parentContainer) {
+      return;
+    }
+    parentContainer.scrollTo({
+      left: currentYearCard.offsetLeft - parentContainer.offsetLeft,
+      behavior: 'smooth',
```
ZhangYiJiang, line 48: For scrolling on component mount you probably _don't_ want smooth, it's better for the component to appear already scrolled to the correct location.

website/src/views/planner/PlannerYear.tsx
```diff
@@ -24,14 +24,31 @@ type Props = Readonly<{
...
+  override componentDidMount() {
+    if (this.props.year !== config.academicYear) {
+      return;
+    }
+    const currentYearCard = this.state.currentYearCardRef.current;
+    const parentContainer = currentYearCard?.parentElement;
+    if (!currentYearCard || !parentContainer) {
+      return;
+    }
+    parentContainer.scrollTo({
```
ZhangYiJiang, line 46: Could consider using `scrollIntoView` instead, that would avoid having to do math to figure out exact offset

ZhangYiJiang (changes requested):
I think this is a good idea, although ref is used incorrectly here

zwliew:
It's been a long time but thanks for your contribution!

jacobkwan:
> It's been a long time but thanks for your contribution!

ty chads

PR #3644, opened 2023-12-17
Add serverless fn NUS auth user endpoint

chrisgzf:
Adds a new serverless fn endpoint that provides user data given a valid SAML token.


ravern (approved):
LGTM!

Issue #3473, opened 2023-07-31
Grey out courses that are no longer offered (pre-req tree, etc.)


yimqiy:
### Is your feature request related to a problem? Please describe.

A clear and concise description of what the problem is. Eg. I'm always frustrated when [...]

EssWhyy:
Hey I’ve made a working implementation of the greyed out nodes in PR #3647, feel free to leave feedback

PR #3647, opened 2023-12-25
Grey Out Deprecated Courses In PreReq Tree

EssWhyy:
## Context

Implements #3473 

## Implementation

Connects ModuleTree.tsx to the Redux Store, and uses the getModuleCondensed() function to check if the module code in each node points to a valid module in the current module bank. If the module has been deprecated, we remove the hoverable color-${layer} classname tag from the module node. The node will be greyed out and no longer appear hoverable as shown below

UI Example:

deprecated courses


website/src/views/modules/ModuleTree.tsx
```diff
@@ -95,11 +108,29 @@ const Tree: React.FC<TreeDisplay> = (props) => {
...
+        [styles.prereqNode]: isPrereq,
+      })}
+    >
+      {prefix && <span className={styles.prefix}>{prefix}</span>}
+      <LinkModuleCodes className={styles.link}>{name}</LinkModuleCodes>
+    </div>
+
+    )
+
+  }
```
ravern, line 133: Instead of repeating the rendering code below, is it possible to leverage conditional classes via `classnames` to achieve the same behaviour? For e.g.

```typescript
return (
    <div
      className={classnames(styles.node, styles.moduleNode, {
        [`hoverable color-${layer}`]: !!moduleActive,
        [styles.prereqNode]: isPrereq,
      })}
    >
      {prefix && <span className={styles.prefix}>{prefix}</span>}
      <LinkModuleCodes className={styles.link}>{name}</LinkModuleCodes>
    </div>
  );
```

ravern (changes requested):
Thanks for the PR!

I have one nitpick on the conditional rendering code, but overall LGTM!

website/src/views/modules/ModuleTree.tsx
EssWhyy: Ah ok, I was wondering if there was a way to avoid repeating the code, this seems to work fine, thanks!

kokrui (approved):
LGTM (and ravern!), merging now

yimqiy:
Ooh just saw this :) thanks for implementing!

Just a note - pre-reqs that are not individual courses to begin with are also greyed out, like the one below:
image

kokrui:
@yimqiy thanks for the comment! I do think it makes sense to have different behaviour (i.e. don't just grey them out) for these pre-req tree nodes -- let's track this as an issue

PR #3654, opened 2024-02-04
Pin binary version in PM2 config file

kokrui:
## Context

We've had a few scraper outages caused by the scraper being restarted and PM2 using old versions of node incompatible with some of our code.

This pins the PM2 binary to an updated version, which ensures a sufficiently-updated node version is used, which ensures that the scraper can run correctly independent of other env and path changes we might make in our production environment.

## Implementation

We previously used a larger set of env flags to deal with outages, but I've tested that specifying the PM2 binary as I've done so here is sufficient. Let's also print out the node version when starting up our scraper

### DIFF
--- a/scrapers/nus-v2/ecosystem.config.js
+++ b/scrapers/nus-v2/ecosystem.config.js
@@ -16,6 +16,7 @@ module.exports = {
 
       env: {
         NODE_ENV: 'production',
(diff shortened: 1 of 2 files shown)


ravern (approved):
LGTM

Issue #3658, opened 2024-03-06
Information should displayed properly on UI for E learning modules


JCSnap:
### Describe the bug

When the module is an E-Learning mod, it does not occupy the whole cell. As a result, the information is unable to be displayed properly:
<img width="420" alt="image" src="https://github.com/nusmodifications/nusmods/assets/111076731/0e72361c-003d-4dc9-aac4-51b4c3111e69">

### To Reproduce
Add an E-Learning module (BPM1702, BPM1701) to the timetable

### Expected behavior

The UI should not hinder the display of information (Suggestion: use a min length of 1 hour cell even if there is no predefined start and end time as in the case of e learning)

### Desktop (please complete the following information):
- Browser Chrome

PR #3659, opened 2024-03-06
Fix cell size for e learning by setting minimum size

JCSnap:
## Context
#3658 

This is a very short PR.

## Implementation
Currently, E-learning module information often use placeholder time since they do not have a fixed time (eg. 0900 - 0901). Since the size of the cell is calculated based on the lesson time, the placeholder time causes the UI to be a bit weird as the calculated background width is short, and the module information is hindered by the UI.
image  

This implementation ensures that the size of the background is at least half the cell:
image

My initial implementation was to make the cell occupy at least the whole cell but I thought maybe there could be lessons that are just 30 minutes, which is why I changed to such that the minimum is half a cell.

### DIFF
--- a/website/src/views/timetable/TimetableRow.tsx
+++ b/website/src/views/timetable/TimetableRow.tsx
@@ -40,7 +40,7 @@ const TimetableRow: React.FC<Props> = (props) => {
         const startIndex = convertTimeToIndex(lesson.startTime);
         const endIndex = convertTimeToIndex(lesson.endTime);
 
-        const size = endIndex - startIndex;
+        const size = Math.max(endIndex - startIndex, 2);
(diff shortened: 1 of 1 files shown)


kokrui (approved):
Thanks for this change, good catch! TIL that there are <30minute lesson slots.

LGTM! Codecov is screaming due to an indirect change on `Timetable.tsx` that's not relevant to this PR (seems like we don't actually have tests for that component yet 😬) so I'm happy to ignore Codecov and get this merged!

kokrui:
I also don't love how some courses show up when the border timing range is large, especially in dark mode

image

but that's for future further discussion where we can build on this work

Issue #3661, opened 2024-03-09
Show Notifications for CPEx opening / closing


kokrui:
Right now, we show notifications for ModReg Rounds opening / closing that pop up on the NUSMods homepage

- The component is here: https://github.com/nusmodifications/nusmods/blob/master/website/src/views/components/notfications/ModRegNotification.tsx
- And the way we config the timings for when the notifications should show is here: https://github.com/nusmodifications/nusmods/blob/master/website/src/data/modreg-schedule.json

It would be good to also show notifications for CPEx when it opens / closes

PR #3665, opened 2024-03-11
Add CPEx notifications

kokrui:
Fixes https://github.com/nusmodifications/nusmods/issues/3661

This PR:
- adds CPEx as a type of modreg round
- includes CPEx dates in modreg-schedule json
- adds a NotificationLink component which decides whether clicking on the notification should navigate user to the EduRec CourseReg link or the CPEx page
- shows minutes in notification text, since we now have 11.59pm times (which would otherwise display 11pm)

image


ravern (approved):
LGTM!

Issue #3648, opened 2024-01-14
Issue with lecture schedule for ST2334


awlarpi:
### Describe the bug

The lecture schedule for ST2334 is displayed as "Weeks 1, 1, 2, 2, 3, 3, 4, 4, 5, 5 ...".

### To Reproduce

1. Add ST2334 to Semester 2 timetable.
2. View the timetable under the timetable tab.

### Expected behavior

Expected the schedule to be shown as "Weeks 1-13".

### Screenshots

<img width="1275" alt="image" src="https://github.com/nusmodifications/nusmods/assets/64833276/93225933-0c3c-4204-85dd-dab65c2b11e2">

### Additional context

The issue occurs on desktop as well as on mobile devices.

jloh02:
Expected behavior specifically for Weeks 1-13 should have no weeks shown

kokrui:
Thanks for the fix @jloh02 !

PR #3677, opened 2024-03-20
Add unique week check in formatNumericWeeks

jloh02:
## Context
#3648 

## Implementation
`TimetableCell` 
`formatNumericWeeks` does a check to ensure listed weeks are unique. Also update condition for change in week <= 1 instead

## Other Information
Instead of simply ensuring unique, should we also sort the weeks and maintain the <= 1 condition?


kokrui (approved):
LGTM! I'll create a new issue to add warnings in our scraper (so we notice these sorts of incorrectly-entered data if we want to let the departments know)

PR #3678, opened 2024-03-20
Update exam clash detection in the Module search page

### LINKED ISSUE #3663: Improve module search page's exam clash detection
In https://github.com/nusmodifications/nusmods/pull/3560, we improved our exam clash detection algorithm. Previously, we only checked for exam start times, but it's possible that an exam's end time overlaps with another exam's start time.

However, the "No Exam Clash (Sem X)" filters in our Module Search page still use the previous detection method. We should change it to the better exam clash detection algorithm.

https://github.com/nusmodifications/nusmods/blob/d99218bad72e5bed4245669a95d42aa27e1cf14b/website/src/views/modules/ModuleFinderSidebar.tsx#L53

This will require some elasticsearch-fu. Relevant docs: https://www.elastic.co/guide/en/elasticsearch/reference/8.12/query-dsl-range-query.html

uyencfi:
## Context
 Fix #3663

## Implementation

Given `timetable = [m1, m2, m3]` which is the list of modules currently added in the timetable. 
To detect whether a module `X` clashes with `timetable`:
* current behaviour is to compare `X.start == m.start` (exact match) for each m in timetable
* updated behaviour:
  * assert that `(X.start < m.end) && (X.end > m.start)` for each m in timetable. This is the condition for 2 intervals to overlap, see https://stackoverflow.com/questions/325933/determine-whether-two-date-ranges-overlap

❗ However, the module struct only stores examStartTime (ISO timestamp, **assuming UTC(?)**) and examDuration (in minutes). Thus to get the endTime:
* `m.end` can be easily obtained since `m` is a known value from the timetable.
* `X.end` is more difficult since we need to compute it on-the-fly inside the Elasticsearch query.

The new Elasticsearch query I have looks like
```javascript
query: {
  bool: { must_not: [  // this document must not match *any* condition in the list, i.e. not overlap any exam
    
    bool: { must: {               // condition 1: true if this exam overlaps m1
      range: { document.start < m1.start + m1.duration },
      script: { document.start + document.duration > m1.start }   // <-- can't use 'range' b/c ES doesn't allow accessing 2 fields at the same time
    }},

    bool: { must: { ... } },     // condition 2: true if this exam overlaps m2
    bool: { must: { ... } },     // condition 3: true if this exam overlaps m3
  ]},
},
```

## TBD

- **Is `examStartTime` really stored in Elasticsearch in UTC or UTC+08?**
- (Unit?) Testing and how to verify correctness

### DIFF
--- a/website/src/views/modules/ModuleFinderSidebar.tsx
+++ b/website/src/views/modules/ModuleFinderSidebar.tsx
@@ -10,7 +10,13 @@ import {
 import { Filter } from 'react-feather';
 import { State as StoreState } from 'types/state';
 
-import { attributeDescription, NUSModuleAttributes, Semester, Semesters } from 'types/modules';
+import {
+  attributeDescription,
+  Module,
+  NUSModuleAttributes,
+  Semester,
+  Semesters,
+} from 'types/modules';
 import { RefinementItem } from 'types/views';
 
 import SideMenu, { OPEN_MENU_LABEL } from 'views/components/SideMenu';
@@ -26,6 +32,11 @@ import { notNull } from 'types/utils';
 import config from 'config';
 import styles from './ModuleFinderSidebar.scss';
 import ChecklistFilter, { FilterItem } from '../components/filters/ChecklistFilter';
+
+type ExamTiming = {
+  start: string;
+  duration: number;
+};
 
 const RESET_FILTER_OPTIONS = { filter: true };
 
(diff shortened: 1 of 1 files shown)


kokrui:
Thanks for the PR! We'll test out scraper changes over the next few days and get back to you ☺️

kokrui:
Hey, it looks like the exam for CM2112 got changed to be 2.30pm instead of 1pm, which makes it match HSI2001, so it's no longer that good of a candidate for a test case. That said, let me take a look and test some things with your query and get back to you -- it looks pretty good right now, but for one, I think the query is inverted! (i.e. "No Exam Clash" gives us courses which _do_ have an exam clash)

ravern:

PR #3680, opened 2024-03-23
Set autoFocus for activeLesson

### LINKED ISSUE #3484: ESC Key not cancelling slot modifying on some specific slots
### Describe the bug

In some (so far, only found 1) courses, when modifying a lesson slot (after clicking on the slot in the timetable), pressing the ESC key does not cancel it.

Seems to only occur when clicking on very specific slots.

### To Reproduce

1. Import this timetable in incognito 
https://nusmods.com/timetable/sem-1/share?CS2103T=LEC:G15

2. Click on the 1hr slot between 1100 - 1200 (important, as clicking the 2hour slots works somehow...?)
3. Try to press ESC to cancel
4. Timetable fails to quit modifying mode

### Expected behavior

Timetable should quit modifying mode

### Screenshots

image
The buggy slot

Bugged behaviour on Windows observed
- OS: Windows 11
- Browser: Chrome, Firefox
- Version 115.0.5790.111 (Chrome), 116.0 (Firefox)

This bug does NOT occur on:
OS: MacOS 13.3.1
Browser: Safari
Version: 16.4 (18615.1.26.110.1)

Not applicable on mobile

### Additional context

This bug also occurs for slots G11, G18, G22 and only when starting modification of slot from the 1hr timeslot.

This is such an obscure bug and I have no idea why it happens, the `onKeyUp` handler in TimetableContent.tsx is not even firing.
However, it DOES fire if you press the 'd' key (which opens the Download menu)

jloh02:
## Context
Fix #3484 

## Implementation
Set `autoFocus` on `activeLesson`. Although this leads to multiple elements with `autoFocus`, the first one in the DOM will be chosen which would still be an `activeLesson` button. This allows the `onKeyUp` event to fire in `TimetableContent.tsx`

### DIFF
--- a/website/src/views/timetable/TimetableCell.tsx
+++ b/website/src/views/timetable/TimetableCell.tsx
@@ -128,6 +128,7 @@ const TimetableCell: React.FC<Props> = (props) => {
       onTouchStart={() => onHover(getHoverLesson(lesson))}
       onMouseLeave={() => onHover(null)}
       onTouchEnd={() => onHover(null)}
+      autoFocus={lesson.isActive}
       {...conditionalProps}
     >
       <div className={styles.cellContainer}>
--- a/website/src/views/timetable/TimetableContent.tsx
+++ b/website/src/views/timetable/TimetableContent.tsx
@@ -356,7 +356,7 @@ class TimetableContent extends React.Component<Props, State> {
           verticalMode: isVerticalOrientation,
         })}
         onClick={this.cancelModifyLesson}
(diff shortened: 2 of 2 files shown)


li-kai (approved):
The root cause of this bug is due to how the timetable is rendered.

If we have to render a new row above the clicked cell, what react actually does underneath the hood is asking can we re use this node? 
If yes -> change the node contents (focus is preserved)
If no, replaced with a different dom node (focus is lost) 

This solution solves it but I wanted to make sure the underlying cause. Thanks for the submission. LGTM.

Issue #3681, opened 2024-03-23
Exclude Hidden Modules from being exported into iCal


maverick-g00se:
### Describe the bug

Hello! Modules that are hidden on the timetable appear in the iCal when exported.

### To Reproduce

1. Add a couple of sample modules
2. Hide at least one module
3. download/export as iCal
4. Open the file with TextEdit and you can see an error where hidden modules are exported into the iCal

### Expected behavior

Technically what the users export is usually based on what they can see on the timetable, hence, modules that are not visible on the timetable at the time of export should not appear in the exported iCal. 

If this does get accepted as a bug, I have already looked through the code and I would like to give it a whirl :)

### Screenshots
Module hidden:
<img width="1283" alt="Screenshot 2024-03-23 at 12 12 02 PM" src="https://github.com/nusmodifications/nusmods/assets/164596138/b5094e77-055c-4ebb-bc45-634be08c0bb8">

Exported iCal info:
<img width="550" alt="Screenshot 2024-03-23 at 12 12 42 PM" src="https://github.com/nusmodifications/nusmods/assets/164596138/6e7ab11d-e7a7-427a-a98e-ed00863a6209">

kokrui:
I think it's reasonable behaviour for hidden courses to be excluded from ics exports. Go for it! @maverick-g00se

maverick-g00se:
Hi @kokrui, I've submitted PR #3682 to address the current issue. Whenever you have some free time, could you please help to review it? I would greatly appreciate any feedback you could provide. Thank you!

kokrui:
Yep, I'll do it by the end of this week's office hours on Wednesday :)

PR #3682, opened 2024-03-23
Exclude Hidden Modules from being exported into iCal

maverick-g00se:
## Context

#3681 

## Implementation

Added an extra parameter `hiddenModules: string[]` to `iCalForTimetable` function in ical.ts to pass in hidden modules and filter accordingly.


kokrui (approved):
LGTM, thanks for adding this (and writing + updating tests!)

PR #3696, opened 2024-03-27
Release Planner 

### LINKED ISSUE #3689: Move planner out of "beta"
Our planner (https://nusmods.com/planner) has been "in beta" for around 5 years. While it doesn't have all the features it was planned to have at the start, it is useful enough (and used enough) (according to our analytics)

The big feature that was holding us back was graduation requirement validation. The best implementation looks like at https://nusplanner.com/ https://github.com/nus-planner/ right now, where people can contribute requirements in a DSL (or eventually somehow generate it from what-if reports).

There's some discussion here (https://github.com/nusmodifications/nusmods/issues/359) on creating such a grad requirement validator that may or may not be integrated in NUSMods. We're happy to support that work but for now let's set our planner free!

wr1159:
## Context

Resolves #3689 

## Implementation

* Display planner in Navbar on all devices (Including mobile devices which didn't show the planner in the past, was it intended or was it a bug?)
* Remove `currentTests` in `BetaToggle`
* Remove beta feedback for Planner page

| Description | Before | After |
|-------|--------|--------|
| No more beta toggle| image| image |
|Planner tab in mobile navbar| image | image |
| No more leave feedback button| image | image| 

## Other Information

Since there's no more test, the beta test toggle in settings doesn't appear by default anymore. Is this intended?

Noticed that the plan to take, exemptions and trash section overflows for mobile. Should we fix it first before releasing this to public? We can make another issue for it, because it's UI related I think we need some discussion for it.


kokrui (commented):
Thanks for the PR (including the screenshots and all)!

I think this PR looks good to me and I approve of it in-principle, but I agree with you that we should fix the mobile UI first before merging this. Thanks for pointing that out!

kokrui:
I'll approve this pending merge conflicts and stuff, but merging will be on hold for now due to some internal requirements with NUS. We expect to be able to merge in around a month or so 🙏

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -24,6 +24,8 @@
     h1 {
       margin: 0.1rem 0;
       font-size: $font-size-xlg;
+      white-space: nowrap;
+      margin: auto;
```
zwliew, line 28: Can we remove these style changes? They make the layout look worse on small form factors.

zwliew (changes requested):

zwliew (approved):

zwliew:
LGTM, thanks!

PR #3700, opened 2024-03-31
Add well-formed timetable validation check

### LINKED ISSUE #3673: Add well-formed timetable validation check
Right now, NUSMods just checks on load for whether a timetable object exists. This could potentially cause issues since we never explicitly check that `data` is of type `PageData`. Notice here that `data` is implicitly converted from `any` as returned in `JSON.parse` to `PageData`, received as a parameter to `validatePageData`.

https://github.com/nusmodifications/nusmods/blob/47238fa8203fb0b6432b74a9d737e7fc4314b4ae/export/src/data.ts#L49-L50

https://github.com/nusmodifications/nusmods/blob/089993f56557c729572284b762384424f7806d53/export/src/data.ts#L66-L71

We could implement a basic check that the data it contains is well-formed according to `PageData`'s type definition.

leslieyip02:
## Context
Resolves #3673 

## Implementation

Added logic to check if a given timetable is well-formed:
- Add lesson clash check
- Add exam clash check

## Other Information
One thing I wasn't sure about is this part in `export.ts` where the URL is hard coded:

https://github.com/nusmodifications/nusmods/blob/7334b074fc4fee253592418b28528784129c0f3d/website/src/apis/export.ts#L12

For local testing, I changed this to a local host manually, but I think this should be put into an environment variable?

Also, let me know if I should add any additional checks. I'm new to this so let me know if there's anything I missed!

### DIFF
--- a/website/src/data/team.json
+++ b/website/src/data/team.json
@@ -45,7 +45,6 @@
   {
     "name": "Li Kai",
     "description": "Happy to help. Software Engineer @ Quora, Ex-Front End Engineer Intern @ Facebook",
-    "active": true,
     "link": "https://www.github.com/li-kai",
     "gravatar": "3d2bcece0409c7578b2e9d87f0c1b6d7",
     "gravatarAlt": "Li Kai",
@@ -76,7 +75,6 @@
   {
     "name": "Zhang Yi Jiang",
     "description": "Fan of pretty purple pony princesses. Software Engineer @ Ironclad",
-    "active": true,
     "link": "https://meebleforp.com",
     "gravatar": "263859aefe62ee6cc1aa993e8e2293e0",
     "gravatarAlt": "Zhang Yi Jiang",
@@ -86,7 +84,6 @@
   {
     "name": "E-Liang Tan",
     "description": "Makers gotta make. Software Engineer @ Ironclad, Ex-Software Engineer Intern @ Facebook",
-    "active": true,
     "link": "https://www.eliangtan.com/",
     "gravatar": "3064783701c7bb1b4451c79318e1b536",
     "gravatarAlt": "E-Liang Tan",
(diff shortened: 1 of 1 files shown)


kokrui:
Hey @leslieyip02, thanks for the PR -- it's cool to see that you managed to identify the relevant code to look for timetable / exam clashes and stuff. However, I think I provided too little detail in my original issue about what "well-formed" entails -- I meant that we should write a basic check that the implicit conversion of `data` to `PageData` worked (and something like the fact that `timetable` is indeed a map from `moduleCode` to `ModuleLessonConfig`). 

I'm not super sure about whether we should block the exporting of a timetable of there are lesson/exam clashes right now, especially since there isn't an option to customise timetable slots yet. We could discuss this in a new issue though (maybe a warning when exporting might be useful? idk)

ravern (changes requested):
Thank you for your effort and contribution!

I think the changes you've made here are great, and do provide a much more effective check here instead of just `_.isObject`.

Unfortunately, I believe the original intention of #3673 is to improve the checks at a much lower-level i.e. that `data` is actually of type `PageData` as passed in the argument to `validatePageData(data: PageData)`. In the existing code, this is never validated:

https://github.com/nusmodifications/nusmods/blob/47238fa8203fb0b6432b74a9d737e7fc4314b4ae/export/src/data.ts#L49-L50

`data: any` is passed to `validatePageData` which is then implicitly converted to `data: PageData`.

Thus, the intended fix was more akin to checking the structure of the `data` object itself rather than if it obeyed our business logic i.e. no clashes / conflicts.

---

The reason I'm not going to merge this PR is because the code you've added [duplicates functionality](https://github.com/nusmodifications/nusmods/blob/master/website/src/utils/timetables.ts) in `website` for the conflict detection. Should we want conflict detection in `export`, I'd rather we implement a generalised code-sharing mechanism (i.e. shared NPM package) rather than copying complicated code.

---

That being said, thanks again for your contribution! We apologise for not being super clear in the original issue — I've updated #3673 to more accurately reflect what we're looking for. If you want to work on the (corrected) version of this, feel free to continue working on this PR, or close and open a new PR.

Let me know here too if you disagree or have any questions or clarifications.

PR #3701, opened 2024-03-31
Fix overflow in planner module list

### LINKED ISSUE #3698: Planner's "Plan To Take" / "Exemptions" / "Trash" section overflows on mobile
### Describe the bug

On mobile, the bottom section of the planner which contains the "plan to take", "exemption", and "trash" cards overflows on mobile

This is blocking https://github.com/nusmodifications/nusmods/pull/3696

The "Beta - Send Feedback" button also doesn't really fit on mobile, but that doesn't need to be fixed in this PR (and it doesn't block #3696)

jloh02:
## Context
Fix #3698 with additional fixes to ensure header remains in size

## Implementation
- CSS fixes for overflows including scrollable and `no-wrap`
- Set minimum width for trash button
- `useMediaQuery` to allow settings button to reduce to an icon button
- Minor fix to capitalize the word "Course"/"Courses" to match the capitalization of "Unit"

### DIFF
--- a/website/src/views/planner/PlannerButton.tsx
+++ b/website/src/views/planner/PlannerButton.tsx
@@ -1,0 +1,18 @@
+import { Settings } from 'react-feather';
+import { breakpointDown } from 'utils/css';
+import useMediaQuery from 'views/hooks/useMediaQuery';
+
+export type Props = Readonly<{
+  onClick: () => void;
+}>;
+
+export default function PlannerButton(props: Props) {
+  const narrowViewport = useMediaQuery(breakpointDown('sm'));
+
+  return (
+    <button className="btn btn-svg btn-outline-primary" type="button" onClick={props.onClick}>
+      <Settings className="svg" />
+      {narrowViewport ? '' : 'Settings'}
+    </button>
+  );
+}
--- a/website/src/views/planner/PlannerContainer/PlannerContainer.scss
+++ b/website/src/views/planner/PlannerContainer/PlannerContainer.scss
@@ -1,5 +1,5 @@
-@import '~styles/utils/modules-entry';
-@import '../variables';
+@import "~styles/utils/modules-entry";
+@import "../variables";
 
 .pageContainer {
   composes: page-container from global;
@@ -60,6 +60,8 @@
 
 .moduleLists {
   display: flex;
+  width: 100%;
+  overflow-x: auto;
 }
 
 .modListHeaders {
@@ -70,12 +72,13 @@
 
 .trash {
   opacity: 0.6;
+  min-width: $semester-width;
   width: $semester-width;
   border: 1px dashed;
   border-radius: 0.4rem;
   text-align: center;
-  color: theme-color('danger');
-  background: rgba(theme-color('danger'), 0.1);
+  color: theme-color("danger");
+  background: rgba(theme-color("danger"), 0.1);
 
   &.isDraggingOver {
     opacity: 1;
@@ -83,8 +86,8 @@
 
   @include night-mode {
     opacity: 0.75;
-    color: lighten(theme-color('danger'), 20);
-    background: rgba(lighten(theme-color('danger'), 20), 0.2);
+    color: lighten(theme-color("danger"), 20);
+    background: rgba(lighten(theme-color("danger"), 20), 0.2);
   }
 }
 
--- a/website/src/views/planner/PlannerContainer/PlannerContainer.scss
+++ b/website/src/views/planner/PlannerContainer/PlannerContainer.scss
@@ -14,12 +14,19 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
+  gap: 1rem;
+  width: 100%;
 
-  h1 {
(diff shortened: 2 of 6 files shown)


website/src/views/planner/PlannerButton.tsx
```diff
@@ -0,0 +1,18 @@
+import { Settings } from 'react-feather';
+import { breakpointDown } from 'utils/css';
+import useMediaQuery from 'views/hooks/useMediaQuery';
+
+export type Props = Readonly<{
+  onClick: () => void;
+}>;
+
+export default function PlannerButton(props: Props) {
```
kokrui, line 9: I think I'd prefer if this file / component was called something like `PlannerSettingsButton`, since there are a few buttons on this page

kokrui: also, even though this is a simple component (and we don't use the `children`) prop, I believe in this repo we lean towards writing React components as functional component with destructured props, so let's do that here

i.e.

```js
const PlannerButton: React.FC<Props> = ({
  onClick
}) => {

// (then props.onClick -> onClick)
```

and then `export default PlannerButton;` at the bottom

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -71,20 +89,21 @@
 .trash {
   opacity: 0.6;
   width: $semester-width;
+  min-width: $semester-width;
   border: 1px dashed;
   border-radius: 0.4rem;
   text-align: center;
-  color: theme-color('danger');
-  background: rgba(theme-color('danger'), 0.1);
+  color: theme-color("danger");
```
kokrui, line 96: I have no current opinion on this, and I'm sure we'd be happy to consider/accept a different PR where we migrate all colours names from single-quotes to double-quotes, but for now it looks like the repo uses single-quotes for colour names like this and I'd prefer consistency

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -1,5 +1,5 @@
-@import '~styles/utils/modules-entry';
-@import '../variables';
+@import "~styles/utils/modules-entry";
```
kokrui, line 1: similar to below, right now it seems the repo uses single-quotes for import paths, so I'd prefer that here

kokrui (changes requested):
Thanks for the PR! It's really well-written and there are a lot of thoughtful decisions being made here, great work!

I'm happy with the feature and functionality, just a few small style and consistency nits before we get this merged

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -33,11 +38,21 @@
   .headerRight {
     display: flex;
     align-items: center;
+    width: 100%;
```
ZhangYiJiang, line 41: I'm confused by this - it seems this just makes the right side items take up a lot of space while squashing the heading and "Send feedback" button 


https://github.com/nusmodifications/nusmods/assets/445650/8adef033-8bc4-4ac5-9765-2f1d261a8678

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -14,12 +14,17 @@
...
   justify-content: space-between;
   align-items: center;
+  width: 100%;
+  gap: 1rem;
 
-  h1 {
+  .headerLeft {
     display: flex;
     align-items: center;
     margin: 0;
```
ZhangYiJiang, line 23: This should be moved back to under h1 - the div does not intrinsically have margin so this won't do anything here, and having margin: 0 on the h1 overrides the default so the h1 is correctly positioned vertically 


https://github.com/nusmodifications/nusmods/assets/445650/ff05249f-992c-4c54-9b34-2b44678d34b8

website/src/views/planner/PlannerSettingsButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+  onClick: () => void;
+}>;
+
+const PlannerSettingsButton: React.FC<Props> = ({ onClick }) => {
+  const narrowViewport = useMediaQuery(breakpointDown('sm'));
+
+  return (
+    <button className="btn btn-svg btn-outline-primary" type="button" onClick={onClick}>
+      <Settings className="svg" />
+      {narrowViewport ? '' : 'Settings'}
```
ZhangYiJiang, line 16: This should wrap the text in `sr-only` utility class instead of removing it so that this text is still available to screen readers https://getbootstrap.com/docs/4.6/utilities/screen-readers/ (there's a number of places in the app where we use `display: none` to do this which is unfortunately has the same issue). 

If we just want to change the layout on mobile, prefer CSS media queries instead

website/src/views/planner/PlannerContainer/PlannerContainer.scss
jloh02: Rectified in #3704

PR #3702, opened 2024-04-01
Add close button in planner settings modal

### LINKED ISSUE #3699: Hard to close Planner Settings modal on mobile

On mobile, on the Planner page, the settings modal may not be easy to close on some screens, as it requires the user to tap on the space outside of the modal.

I think a good solution would be to add a close button at the top-right like the "Beta - Send Feedback" modal does:

I think it's fine if the close button appears on both mobile and desktop too

IMO this is NOT a blocker for https://github.com/nusmodifications/nusmods/pull/3696

jloh02:
## Context
Fix #3699 

## Implementation
- Use existing `CloseButton` component
- Changes the container to flexbox so that 0 height item still maintains the `margin-bottom`

### DIFF
--- a/website/src/views/planner/PlannerSettings.scss
+++ b/website/src/views/planner/PlannerSettings.scss
@@ -8,10 +8,6 @@
   section {
     margin-bottom: 1rem;
 
-    &:first-child {
-      margin-bottom: 0;
-    }
-
     &:last-child {
       margin-bottom: 0;
     }
--- a/website/src/views/planner/PlannerSettings.scss
+++ b/website/src/views/planner/PlannerSettings.scss
@@ -2,8 +2,15 @@
 @import 'variables';
 
 .settings {
+  display: flex;
+  flex-direction: column;
+
   section {
-    margin-bottom: 2rem;
+    margin-bottom: 1rem;
+
+    &:first-child {
+      margin-bottom: 0;
+    }
 
     &:last-child {
       margin-bottom: 0;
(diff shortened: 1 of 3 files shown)


website/src/views/planner/PlannerSettings.scss
```diff
@@ -2,8 +2,15 @@
...
+  display: flex;
+  flex-direction: column;
+
   section {
-    margin-bottom: 2rem;
+    margin-bottom: 1rem;
+
+    &:first-child {
+      margin-bottom: 0;
+    }
```
kokrui, line 14: Looks like this CSS selector matches on nothing, since there doesn't exist any `section` element that is the first child of its parent. With the addition of the close button diff, the first section element is now the second child of the settings div. We can remove this dead code!

kokrui (changes requested):
Thanks for the PR again! Just a tiny bit of dead code to shave off before I LGTM and merge

website/src/views/planner/PlannerSettings.scss
jloh02: Oops sorry about that. Forgot to remove it after moving the CloseButton out of a section

kokrui (approved):
LGTM!

ZhangYiJiang:
The approach is not wrong, but this means that the close button always takes up a row on its own. The close button already has styles so it is on the right side of the modal (using floats). You can also adjust the positioning after that using negative margin top and right. 


https://github.com/nusmodifications/nusmods/assets/445650/7f502ed4-c27c-4940-b4f2-667d7a8368cd

PR #3704, opened 2024-04-01
Fix planner header wrap and center planner settings button cogwheel

### LINKED ISSUE #3703: Course planner header text wrapped on wide viewports
### Describe the bug
Broken in #3701 
image

### Expected behavior
"Course Planner" should be in 1 line

jloh02:
## Context
Fix #3703 

## Implementation
Just CSS changes and removal of margin when "Settings" word removed

## Possible Improvements
There's a few UI minor details which I felt could've been done better but considering that we can release the planner then fix these minor things I'm going to leave these points here

- When the beta button is removed, the header text are not **visually** centered
- The module statistics portion ("0 Courses / 0 Units") should be right aligned especially at the smaller screen levels when the text is wrapped

## Screenshots
370px width with beta button:
image

500px width with beta button:
image

900px width with beta button:
image

1600px width with beta button:
image

370px width without beta button:
image

500px width without beta button:
image

900px width without beta button:
image

1600px width without beta button:
image

### DIFF
--- a/website/src/views/planner/PlannerContainer/PlannerContainer.scss
+++ b/website/src/views/planner/PlannerContainer/PlannerContainer.scss
@@ -17,8 +17,8 @@
 
   .headerLeft {
     display: flex;
+    flex: 1;
     align-items: center;
-    flex: 3;
     margin: 0;
 
     h1 {
@@ -37,14 +37,12 @@
 
   .headerRight {
     display: flex;
+    justify-content: flex-end;
     align-items: center;
-    justify-content: right;
-    flex: 2;
   }
 
   .moduleStats {
     display: flex;
-    flex-direction: row;
     flex-wrap: wrap;
     margin-right: 1rem;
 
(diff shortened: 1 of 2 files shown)


kokrui:
FYI we plan to keep the Beta - Send Feedback button for a sem or two still, so no worries about there for now

website/src/views/planner/PlannerContainer/PlannerContainer.scss
```diff
@@ -12,17 +12,17 @@
 
 .header {
   display: flex;
-  justify-content: space-between;
   align-items: center;
   width: 100%;
-  gap: 1rem;
```
kokrui, line 18: I think the PR is fine for now but let's keep the `gap` here so there's spacing between the "Beta - Send Feedback" and the "0 Courses / 0 Units" on a small viewport

kokrui (changes requested):
LGTM pending the quick change

kokrui (approved):
LGTM

PR #3712, opened 2024-04-09
Add Mobility page for ISBs, MRT info

kagiura:


zwliew:
Incredible work with this :) I'll clean this up over the holidays and hopefully bring it in by the end of the year.

website/src/views/routes/Routes.tsx
```diff
@@ -34,6 +35,9 @@ const Routes: React.FC = () => (
     {/* END LEGACY ROUTES */}
     <Route path="/archive/:moduleCode/:year/:slug?" component={ModuleArchiveContainer} />
     <Route path="/venues/:venue?" component={VenuesContainer} />
+    <Route path="/mobility/:type(service|stop)/:slug" component={MobilityContainer} />
+    <Route path="/mobility" component={MobilityContainer} />
```
jloh02, line 39: Perhaps instead of that, it'd somehow be better to integrate with current "venues". we have an excess of tabs and this could potentially be a problem on mobile?

website/src/apis/nextbus-new.ts
```diff
@@ -0,0 +1,26 @@
+const baseURL = 'https://nnextbus.nusmods.com'; // TODO: wait until we have an api proxy
```
jloh02, line 1: We should move this proxy to a serverless function of sorts? Or is something else better?

PR #3713, opened 2024-04-09
chore: fix typo in settings container

jsun1590:
## Context

Changes `survey to tells us` to `survey to tell us` in the settings container.

Alternatively, could consider changing it to `survey that tells us`. Thoughts?

## Implementation

N/A

## Other Information

N/A


kokrui:
Thanks for the catch @jsun1590! I do prefer `survey that tells us` -- let's switch it to that and I'll approve and merge

jsun1590:
@kokrui Thanks, I have changed it :)

kokrui:
😅

PR #3715, opened 2024-04-15
Add well-formed timetable validation check

### LINKED ISSUE #3673: Add well-formed timetable validation check

leslieyip02:
## Context
Resolves #3673 

## Implementation
Uses `joi` to check if export data conforms to `PageData`.

## Other Information
I chose to use this library because it was used in `nusmods/scrapers`.

### DIFF
--- a/export/package.json
+++ b/export/package.json
@@ -23,6 +23,7 @@
     "bunyan": "1.8.15",
     "fs-extra": "9.1.0",
     "http-graceful-shutdown": "2.4.0",
+    "joi": "^17.12.3",
     "koa": "2.14.2",
     "koa-router": "10.1.1",
     "koa-views": "6.3.1",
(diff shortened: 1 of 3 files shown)


ravern (approved):
LGTM! Thanks for the PR :)

Issue #3686, opened 2024-03-26
Include Room Name in Venue Search


jloh02:
### Is your feature request related to a problem? Please describe.
When searching for venues, a user may only be told the name of a venue e.g. SR15 and not the unit. Currently, the only way to figure out the location of the room is by google searching NUS websites. This feature request proposes to allow users to find the unit number by simply searching the venue name in NUSmods

### Describe the solution you'd like
Include searching of room name (including abbreviations such as SR for seminar rooms)

### Describe alternatives you've considered
Include room names into all venue codes (although this will make it look very long and hence not ideal)
- [COM3-01-21](https://nusmods.com/venues/COM3-01-21) should include the SR number for example `COM3-SR12-01-21`

mongj:
hi can I take this?

PR #3717, opened 2024-04-17
Include Room Name in Venue Search

mongj:
## Context

Implements #3686 

## Implementation

searchVenue function is modified to take in an optional VenueLocations argument and filter against both venue and roomName if the optional argument is provided. VenueLocations is made optional for backward compatibility.

venueContainters receive the venueLocations prop from withVenueLocations HOC. venueLocations is then passed to the searchVenue function in venueContainters's render method.

The venue buttons are also modified to show both venue and room names (if present).
Screenshot 2024-04-17 at 4 48 55 PM


kokrui (approved):
LGTM! Thanks for the thoughtful UI change to go along with the expanded search criteria, as well as writing sensible tests!

Issue #3720, opened 2024-04-17
Venue Search results should highlight matched substrings


mongj:
### Is your feature request related to a problem? Please describe.
Venue Search currently filters the list of venues according to what the user input into the search bar, but does not show which part of the result is actually being matched.

### Describe the solution you'd like
Matched substrings should be highlighted, similar to how it's implemented in the search bar in desktop view.
![Screenshot 2024-04-17 at 8 40 42 PM](https://github.com/nusmodifications/nusmods/assets/87565927/93d3afef-3c7d-418e-a65c-246d8d5467c9)

kokrui:
Resolved in #3722

PR #3722, opened 2024-04-18
feat(venues): add highlighting for matched substrings in Venue Search

mongj:
## Context

Implements #3720 

## Implementation

Implemented using existing [highlight](https://github.com/nusmodifications/nusmods/blob/9fdd1e806e201d48d47a7248c0ff5189af89cf00/website/src/utils/react.tsx#L46) function, which is already used for highlighting results in the main search bar in desktop view.
Screenshot 2024-04-18 at 4 23 10 PM

Venue search with highlighting
Screenshot 2024-04-17 at 8 46 20 PM

## Other Information

This PR is for the same commits as #3721. I closed #3721 because it was failing to upload the code coverage report which caused the build_and_test check to fail after merging from the master branch (I'm not sure why exactly though).


ravern (approved):
LGTM

PR #3729, opened 2024-05-17
Add menu button to add module from planner to timetable

### LINKED ISSUE #3726: Course Planner -> Timetable
Maybe the team can consider adding a feature that connects the Planner with the Timetable, where modules I add under my Planner can be automatically reflected into my Timetable with a button.

leslieyip02:
## Context

Resolves #3726 

## Implementation

- Add an additional option to the module menu dropdown to add the module to the timetable

https://github.com/nusmodifications/nusmods/assets/90888680/3bdb35b6-8614-45dd-aa7f-dcf95a44f652

## Other Information
- Would it be better if there was a single button to add all modules into the timetable/override the current timetable with the planner mods?


ravern:
This looks really good, and I've verified that it works :) Thank you for your contribution.

At a higher-level, I have two potential improvements/suggestions, that I hope can be included in this PR if not too complex to implement.

1. When I click _"Add to timetable"_, and the module is added to the timetable, the button is still present in menu, and now doesn't do anything when I click it. Is it possible to hide the button from the menu if the module has already been added to the timetable of the user? The button should re-appear if I remove the module from my timetable.

2. Perhaps this is more of a discussion, do you think we should only display the button for modules in the _current_ (highlighted) semester, or should we be able to add mods from non-current sems into the timetable as if they were in the current sem?

> Would it be better if there was a single button to add all modules into the timetable/override the current timetable with the planner mods?

I think this would be a good idea too, any ideas where this button can be placed? I think it might be better for this to go into a separate PR, but I'm okay for this to go into the current one too.

ravern (commented):
This looks really good, and I've verified that it works :) Thank you for your contribution.

At a higher-level, I have two potential improvements/suggestions, that I hope can be included in this PR if not too complex to implement.

1. When I click _"Add to timetable"_, and the module is added to the timetable, the button is still present in menu, and now doesn't do anything when I click it. Is it possible to hide the button from the menu if the module has already been added to the timetable of the user? The button should re-appear if I remove the module from my timetable.

2. Perhaps this is more of a discussion, do you think we should only display the button for modules in the _current_ (highlighted) semester, or should we be able to add mods from non-current sems into the timetable as if they were in the current sem?

> Would it be better if there was a single button to add all modules into the timetable/override the current timetable with the planner mods?

I think this would be a good idea too, any ideas where this button can be placed? I think it might be better for this to go into a separate PR, but I'm okay for this to go into the current one too.

leslieyip02:
>     2. Perhaps this is more of a discussion, do you think we should only display the button for modules in the _current_ (highlighted) semester, or should we be able to add mods from non-current sems into the timetable as if they were in the current sem?

I think it's ok for mods to be added to non-current sems because some people might want to plan both sem 1 and sem 2 together. You can also add the same mod to both sems and compare which schedule would be more favourable.

> I think this would be a good idea too, any ideas where this button can be placed? I think it might be better for this to go into a separate PR, but I'm okay for this to go into the current one too.

I think I can open a separate PR for this.

ravern (approved):
I think having the "View" and "Add" actions are a great idea! Let's stick with that.

> I think it's ok for mods to be added to non-current sems because some people might want to plan both sem 1 and sem 2 together. You can also add the same mod to both sems and compare which schedule would be more favourable.

Yep that makes sense.

I think what I meant to say was non-current _years_ and not sems. So for example, in this image, the button is available for CS3230 which is planned for next AY and not current one. I think it shouldn't be.

image

I think I'll leave this completely up to you though, I'm okay with either option.

leslieyip02:
> I think what I meant to say was non-current _years_ and not sems. So for example, in this image, the button is available for CS3230 which is planned for next AY and not current one. I'm just highlighting this consideration, but I think I'll leave this completely up to you though, I'm okay with either option.

Ah ok I see. I think I'll leave the button regardless of AY for now. It's possible that users may plan to take a module in a different year, but still want to add it in the current timetable for comparison.

ravern (approved):
LGTM

Issue #3399, opened 2022-07-03
I wish there is a "like" button or "saved" button for considered modules. If there is any, please forgive me, because I really cannot find it anywhere. 


tiif:
### Describe the solution you'd like

I wish there is a "like" button or "saved" button for considered modules. 

### Additional context

 If there is any, please forgive me, because I really cannot find it anywhere.

ravern:
@EssWhyy is taking this!

EssWhyy:
As discussed, I'll help to create a new button for each course description page to “Save Course” which adds the course to the new planner feature, under the section “Planning to take”. The planner feature will also be moved out of beta after this change (?)

Will update when I start to implement.

ravern:
Hey, we're doing some clean ups over the summer. Closing this for now because this issue is stale. Feel free to re-open this if you think otherwise!

PR #3731, opened 2024-05-31
website: Add button to save a course to planner from info page

EssWhyy:
## Context

Implementation for #3399 

## Implementation

Added a button to each module page to allow a user to save the module to the planner in the Plan to Take section, acting like a save button of sorts. 

If it is already added to Plan To Take, the button will remove the module from there instead. 

Test Run:

https://github.com/nusmodifications/nusmods/assets/39799639/bb7ccd98-f413-4400-9ca1-12440988319b


website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+type State = {
+  loading: Semester | null;
+};
+
+function isModuleInPlanToTake(module: Module, planToTakeModules: PlannerModuleInfo[]): boolean {
+  for (let i = 0; i < planToTakeModules.length; i++) {
+    if (planToTakeModules[i].moduleCode === module.moduleCode) {
+      return true;
+    }
+  }
```
ravern, line 35: Is it possible to use a `.some` call here instead?

e.g. `planToTakeModules.some(m => m.moduleCode === module.moduleCode)`

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return false;
+}
+
+function getModuleId(module: Module, planToTakeModules: PlannerModuleInfo[]): string {
+  // If duplicate ids, gets the first id.
+  for (let i = 0; i < planToTakeModules.length; i++) {
+    if (planToTakeModules[i].moduleCode === module.moduleCode) {
+      return planToTakeModules[i].id;
+    }
+  }
```
ravern, line 46: Is it possible to use a `.find` call here instead?

e.g. `planToTakeModules.find(m => m.moduleCode === module.moduleCode)?.id ?? '0'`

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  override state: State = {
+    loading: null,
+  };
+
+  onSelect(semester: Semester) {
+    const { module, planToTakeModules } = this.props;
+    const PLAN_TO_TAKE_YEAR = '3000';
+    const PLAN_TO_TAKE_SEMESTER = -2;
```
ravern, line 69: Is it possible for us to import these constants from `src/utils/planner.ts' instead?

Or better yet, create a new function that more accurately represents what this code is trying to do, i.e., add a module so far into the future that it will definitely be in the "plan to take" section (correct me if this _not_ what the code is trying to do)?

website/src/views/components/module-info/SaveModuleButton.scss
```diff
@@ -0,0 +1,34 @@
...
+  white-space: normal;
+
+  br {
+    display: none;
+  }
+
+  strong {
+    font-size: 1em;
+  }
+}
```
ravern, line 34: Seems to be unused?

website/src/views/components/module-info/SaveModuleButton.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  :global(.dropdown-toggle.dropdown-toggle-split) {
+    width: 2rem;
+    border-left: none;
+
+    @include media-breakpoint-down(sm) {
+      // Long selector for specificity
+      width: 3rem;
+    }
+  }
```
ravern, line 20: Global seems like quite a large scope, do you know if this styling is actually applied on the button?

ravern (changes requested):
Have verified that this works! Thanks for your contribution.

The code looks good, full props (pun intended) for following the current style of the surrounding components.

_This is completely optional_, but it would be a nice-to-have if we could create new components in the functional style instead of class style — I would like to slowly move the codebase towards this pattern.

The specific code comments are just nits, and I would still merge them in if you don't touch the code referenced in those comments.

I'm requesting changes only because of the **component name**, which I believe could be more specific than `SaveModuleButton`. Something that references the planner / plan-to-take action might be better.

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+  return '0';
+}
+
+export class SaveModuleButtonComponent extends PureComponent<Props, State> {
+  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+    const { planToTakeModules, module } = nextProps;
+    const { loading } = prevState;
+
+    if (loading != null && isModuleInPlanToTake(module, planToTakeModules)) {
```
zwliew, line 55: `loading ` doesn't seem used. I don't think a loading indicator is necessary so I think we can remove it entirely.

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+        Remove from <br />
+        <strong>Plan To Take</strong>
+      </>
+    ) : (
+      <>
+        Add to <br />
+        <strong>Plan To Take</strong>
+      </>
+    );
+  }
```
zwliew, line 96: This is simple enough to be inlined into the `render()` function.

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    if (isModuleInPlanToTake(module, planToTakeModules)) {
+      const id = getModuleId(module, planToTakeModules);
+      this.props.removeModule(id);
+    } else {
+      this.setState({ loading: semester });
+      this.props.addModule(PLAN_TO_TAKE_YEAR, PLAN_TO_TAKE_SEMESTER, {
+        type: 'module',
+        moduleCode: module.moduleCode,
+      });
```
zwliew, line 79: To add onto Ravern's comment [above](https://github.com/nusmodifications/nusmods/pull/3731/files#r1660190895) , the right thing to do is probably to add the following function definition to `actions/planner.ts`:

```js
export function addPlanToTakeModule(moduleCode: ModuleCode) {
// call addPlannerModule here with some defaults
}
```

website/src/views/components/module-info/SaveModuleButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    return null;
+  }
+
+  override state: State = {
+    loading: null,
+  };
+
+  onSelect(semester: Semester) {
```
zwliew, line 66: Similar to `loading`, we don't actually use `semester`, so we can remove it.

website/src/views/components/module-info/SaveModuleButton.tsx
zwliew: Going even further, I reckon we can replace `isModuleInPlanToTake` and `getModuleId` with:

```js
function findModule(
  module: Module,
  planToTakeModules: PlannerModuleInfo[],
): PlannerModuleInfo | undefined {
// use planToTakeModules.find(...)
```

and adjust the rest of the calling code accordingly.

website/src/views/modules/ModulePageContent.tsx
```diff
@@ -252,6 +253,12 @@ const ModulePageContent: React.FC<Props> = ({ module, archiveYear }) => {
                   </div>
                 )}
 
+                {!isArchive && offered && (
+                  <div className={styles.addToTimetable}>
+                    <SaveModuleButton module={module} className="btn-group-sm" block />
+                  </div>
```
zwliew, line 259: Please merge this into the block above (lines 251 to 253)

zwliew (changes requested):

zwliew:
I went ahead and made the requested clean up changes. Thanks for your contribution!

PR #3738, opened 2024-07-04
Add duplicate modules conflict

### LINKED ISSUE #3732: Duplicate Modules In Planner Columns
### Describe the bug

Duplicate Courses of the same course code are allowed on the same planner column.

### To Reproduce

1. Search any course in the dropdown for any of the 4 columns in planner (Semester 1, Semester 2, Exemptions, Plan to Take)
2. Add course to the column
3. Repeat steps 1-2 for the same column
4. Two identical course cards will be present for that column

### Expected behavior

There needs to be a function to check for duplicate courses for every column in the planner. 

### Screenshots

Planner Allows for Duplicate Entries 2
Duplicate Modules

### Environment
- Windows 10, Firefox v126.0

ChillinRage:
Duplicate conflict checks if there are modules with the same module code added to the same academic year and semester.

## Context

Proposal to fix #3732 which having same modules in the same year/semester shows no warnings.

## Implementation

Added a new conflict that checks if the module list for a given semester has more than one module of the same module code.

## Other Information

Added test case for the conflict, though I'm slightly confused as to what's the role/difference between `planner` and `moduleBank` (the variables in the `planner.test.ts`)?


website/src/types/planner.ts
```diff
@@ -30,7 +30,11 @@ export type NoInfo = {
   type: 'noInfo';
 };
 
-export type Conflict = PrereqConflict | ExamConflict | SemesterConflict | NoInfo;
+export type Duplicate = {
+  type: 'duplicate';
+};
```
ravern, line 35: nit: `Duplicate` -> `DuplicateConflict`

ravern (changes requested):
Thanks for this contribution! Been a bit busy, sorry for getting to this only now.

Awesome that you've added new tests for this, and I like the solution that you've come up with by adding a new conflict type!

I'm requesting changes because it would be great to have an error message indicating why this conflict occured. As can be seen in the screenshot, for pre-req conflicts there is a really specific reason, whereas for duplicate conflict the message is blank.

image

To render the error message, you can check out `website/src/views/planner/PlannerModule.tsx`.

ChillinRage:
Apologies for the missing message and late response too. I've added the section that should display a warning message. Is there a way in which I can run a live webpage that allows me to test my local changes?

ravern:
> Apologies for the missing message and late response too. I've added the section that should display a warning message. Is there a way in which I can run a live webpage that allows me to test my local changes?

Yes! Once you've cloned the repository, just run `cd website` to enter the website folder, `yarn install` to install all the dependencies, and then `yarn start` to start the web server.

Let me know if that works.

ChillinRage:
Yes I just tried `yarn start` and it works. Thank you!

As for the conflict message, it does appear after the changes.
image

ravern (approved):
LGTM

PR #3743, opened 2024-07-08
Remove Codecov on master and production branches

### LINKED ISSUE #3708: Remove Codecov as a CI step on master/production branches
As mentioned in https://github.com/nusmodifications/nusmods/issues/3688, I think we should remove Codecov as a CI step on the `master` and `production` branches.

I think we should keep it on `pull-request`s so we can still see how coverage changes each time a PR is made / while reviewing PRs, but once a commit has landed in master / prod it's no longer useful information.

itstrueitstrueitsrealitsreal:
## Context
Fixes #3708

## Implementation
Adds a conditional statement in `.circleci/config.yml` to check if the current branch is `master` or `production`, and does not run `yarn codecov --disable=gcov -f ./coverage/coverage-final.json`  if so. This reduces build times for both of these branches.


ravern:
Thank you for your contribution! Just wanted to let you know that we haven't forgotten about your PR, will get to it hopefully this weekend :)

zwliew (approved):
Thanks for contributing! LGTM 😊

PR #3744, opened 2024-07-09
Add new Packaged Laboratory lesson type

kokrui:
## Context

There is a new lesson type called Packaged Laboratory, with code Y. 

For now it looks like the only course with it is https://nusmods.com/courses/PHS2191/laboratory-techniques-in-pharmaceutical-science-i

It used to show:
image

Now it shows:
image

## Other Information

I've updated the prod scraper


ravern (approved):
LGTM

Issue #3205, opened 2021-01-27
Tooltipped buttons tend to require 2 taps to trigger onClick


taneliang:
### Describe the bug

Some of our buttons are wrapped with tooltips to provide additional info, e.g. the delete/hide buttons on the timetable page, as well as the nodes in the module page's prerequisite tree.

However, tooltipped buttons take 2 taps to trigger their `onClick` handler on iOS. Unfortunately, because the buttons still have working hover/focus/active styling, they still react visibly to users' taps, making them very misleading.

### To Reproduce

1. On mobile, go to or create a timetable with a selected module (or just import this timetable https://modsn.us/i3htm, but **note that importing will replace your current timetable!**)
2. Try tapping either the delete or hide buttons. They will likely just turn gray momentarily, but nothing else will happen. A second tap is necessary to trigger the delete/hide action.
	![image](https://user-images.githubusercontent.com/12784593/105927431-f192af00-607e-11eb-830d-85c3abafb099.png)

### Expected behavior

* Single-tapping the button should call the button's `onClick` handler.
* Long pressing the button should show the tooltip (which is the current behavior)

### Smartphone (please complete the following information):

iOS Safari/Chrome, all versions. Didn't try on Android.

### Additional context

We may want to see if upgrading @tippyjs (issue #3069) will fix this, otherwise if we can't find a solution we may want to find a replacement tooltip package.

With #3155 removing navigation tab titles on `sm`-width viewports, we may want to add new tooltips to nav tabs on `sm` and narrower in the future. Fixing this bug will unblock that.

zwliew:
Fixed in #3750

PR #3750, opened 2024-07-18
Fixed double clicking issue on touch devices

Incogdino:
## Context
 #3205 

## Implementation
 

image
Changed touch prop to trigger as "hold" if the button is held for more than 300ms, else it will be interpreted as a click and onClick() function is called instead. This only affects touch devices.

https://github.com/user-attachments/assets/e7ac8ce2-768e-41ae-a117-9358b27ea5c6

Running yarn test was successful. 
## Other Information

This is my first time contributing to nusmods. Pardon me if there are any mistakes :)


zwliew:
Hi, thanks for contributing! However, I don't understand what the issue is here. I get the right behaviour on Android: merely tapping a button clicks it, but holding for a few milliseconds doesn't click it and instead displays the tooltip. Of course, maybe there is an issue on other device/OS/browser combinations. May I know what issue you're facing and on what device+browser+OS?

Incogdino:
> Hi, thanks for contributing! However, I don't understand what the issue is here. I get the right behaviour on Android: merely tapping a button clicks it, but holding for a few milliseconds doesn't click it and instead displays the tooltip.
> 
> Of course, maybe there is an issue on other device/OS/browser combinations. May I know what issue you're facing and on what device+browser+OS?

Incogdino:
Hello sorry clicked on the wrong button 😓

https://github.com/user-attachments/assets/019a392c-e099-4798-aa1b-73426f4cf47c

Here is what I got when I tried in the video and the result : 
1. Single click on hide -> Button turns grey with no effect
2. Single click on delete -> Button turns grey with no effect
3. Hold delete button -> Tooltip pops up 
4. Click outside of the button -> Button loses grey color 
5. double click on delete button -> Module gets removed 
6. Click on undo button -> Module gets added back 

Device: ios safari

**Expected reaction** 
onClick function should trigger on a single click for both delete and hide

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
zwliew, line 1: Please remove this file. I believe this was mistakenly generated by running `yarn` in the root directory instead of the `website` directory.

website/src/views/timetable/TimetableModulesTable.tsx
```diff
@@ -55,7 +55,7 @@ export const TimetableModulesTableComponent: React.FC<Props> = (props) => {
     return (
       <div className={styles.moduleActionButtons}>
         <div className="btn-group">
-          <Tooltip content={removeBtnLabel} touch="hold">
+          <Tooltip content={removeBtnLabel} touch={['hold', 300]}>
```
zwliew, line 58: A 300ms delay is quite noticeable. Can we change it to something more shorter, say 50ms?

website/src/views/timetable/TimetableModulesTable.tsx
```diff
@@ -65,7 +65,7 @@ export const TimetableModulesTableComponent: React.FC<Props> = (props) => {
               <Trash className={styles.actionIcon} />
             </button>
           </Tooltip>
-          <Tooltip content={hideBtnLabel} touch="hold">
+          <Tooltip content={hideBtnLabel} touch={['hold', 300]}>
```
zwliew, line 68: Same thing here.

zwliew (changes requested):
I see. It mostly looks good to me, except for some minor issues.

website/src/views/timetable/TimetableModulesTable.tsx
Incogdino: Got ya :) I've made the necessary changes

zwliew (approved):

zwliew:
Thanks!

PR #3752, opened 2024-07-28
Disable planner conflicts in non-current years

### LINKED ISSUE #3751: Add button to silence planner module conflicts

I took UTC1702G in Sem 2 of AY23/24, but since it is only available in Sem 1 this year, there is a warning.

Screenshot 2024-07-26 181858

I think it could be helpful to have a button to silence warnings. I am also willing to work on this feature.

I don't know how the logic to fix the issue above would work unless module availability data from previous years is stored somewhere.

leslieyip02:
## Context
Resolves #3751. Disables planner conflicts for years other than the current year.

## Implementation

Screenshot 2024-07-29 at 00-11-22 Course Planner - NUSMods

## Other Information
I tried to write some tests too, but I'm not very familiar with writing them. Let me know if I should update anything.


website/src/views/planner/PlannerYear.tsx
```diff
@@ -70,6 +70,15 @@ export default class PlannerYear extends PureComponent<Props, State> {
...
     }
+    if (year !== config.academicYear) {
+      sortedSemesters = sortedSemesters.map(([semester, modules]) => [
+        semester,
+        modules.map((module: PlannerModuleInfo) => ({
+          ...module,
+          conflict: null,
+        })),
+      ]);
+    }
```
zwliew, line 81: Instead of modifying our view of `sortedSemesters` here, could we change the `renderModule` function in `PlannerSemester.tsx` to pass `null` as the `conflict` prop into `PlannerModule`? It's a simpler change. Besides, I don't like that we're calling `.map()` yet again here.

The line I'm referring to: https://github.com/nusmodifications/nusmods/blob/master/website/src/views/planner/PlannerSemester.tsx#L92

zwliew (changes requested):

zwliew:
Perfect. Thank you!

zwliew (approved):

PR #3755, opened 2024-07-30
Add button to add planner semester to timetable

leslieyip02:
## Context
Previously in #3729, I suggested a feature to add all modules from a semester in the planner into the timetable.

## Implementation
I placed a button under the "Add Courses" button, following the same style. This button will only appear for the current year.

https://github.com/user-attachments/assets/0f84d2a8-9f50-4dcc-a8a9-9cc12f74c512

## Other Information
I decided to add a notification to let the user know that the mods were added. Let me know if this makes sense?


kokrui (approved):
Thanks for the PR! We think it's a useful feature and we're going to merge it, but we'll swap out the icon of the button to something that's different from the Add Courses button

Issue #3740, opened 2024-07-06
Prerequisite check for AT LEAST n prefixed modules not working as intended


Wxy2003-xy:
### To Reproduce

1. Plan for a course with prereq tree contain 
     "nOf": [
              n,
              [
                "PREFIX%:GRADE"
              ]
            ]
    example: JS4207
2. fill in prerequisite modules accordingly 
    example: Japanese 1 - 5, any 3 JS coded modules
### Expected behavior
  
JS4207 should pass prerequisite check

### Screenshots
![Image 6-7-24 at 09 21](https://github.com/nusmodifications/nusmods/assets/134219782/0460b0c0-3f78-4b32-81d2-f31764bfabab)

### Desktop (please complete the following information):

- OS: [e.g. MacOS]
- Browser [e.g. Chrome]
- Version [e.g. Beta]

### Additional context

Found out this when testing our orbital project nusplanner

ravern:
I can reproduce this.

We just merged in a PR to update the pre-requisite check (#3737), but seems like it doesn't work specifically in the AT LEAST N prefixed modules case.

Thanks for filing this.

ChillinRage:
Hi can I check if there are modules that has prerequisites that contains both specific and prefix% codes? (i.e. at least 3 of CS3123/CS2% mods) ?

Because if the only combination is "at least N of " followed by only 1 prefix% code (like in JS4207), then I can try to come up with a solution for it without having to change the walkTree function too much.

Wxy2003-xy:
> Hi can I check if there are any modules that has prerequisites that contains both specific and prefix% codes in one nOf? (i.e. at least 3 of CS3123/CS2% mods) ?
> 
> 
> 
> Because if the only combination is "at least N of " followed by only 1 prefix% code (like in JS4207), then I can try to come up with a solution for it without having to change the walkTree function too much.



I'm not sure if there exists such case, but I'd like to propose a solution we used, which is to separate the match checking logic from the recursive check tree function, and handle wildcard case there.

ChillinRage:
Hm if you would like, you can try to create a PR for the proposed solution.

zwliew:
Fixed in #3758

PR #3758, opened 2024-07-31
planner: Fix 'nOf' prereq check with module code wildcards

Wxy2003-xy:
## Context
#3740 prerequisite check logic for nOf wildcard case not working as intended

Separately, we now also disambiguate conflict texts involving 'and', 'or', and 'nOf' prereq fragments.

## Implementation
handle wildcard case in nOf cases separately from normal course code


zwliew (approved):

zwliew:
Thank you. I refactored the code a bit to make it reduce code duplication.

PR #3770, opened 2024-08-13
planner: Display some types of conflicts in every semester

### LINKED ISSUE #3767: [Bug-Planner] Planning mods for future years do not show prerequisites on NUS Planner
### Describe the bug

Only mods that are put in the red highlighted zone (which I believe to represent 'current year') will be shown their prerequisites, but not anymore when you move to future years. 

### To Reproduce

1. Go to 'NUSMODS'
2. Click on 'NUS Planner'
3. Choose 'Settings' and set it to any years for 'Matriculated in' and any future years for 'Graduating' (For me I chose AY23/24 and In 3 years)
4. In my example, I selected module CS2109S for Year 2, prerequisites are shown and worked correctly. But when I moved this module to Year 3, prerequisites are not shown. 

### Expected behavior

Prerequisites should also be shown for future years apart from the current year, because users would want to be able to plan their modules for their entire tenure. If this is an intended feature, I think that restricting the module to have to be in the red highlighted area to show their prerequisites would not be practical for users e.g: What if they intend to take the module in Year 3 and want to see what prerequisites they need to clear and make adjustments accordingly? 

### Screenshots

Screenshot 2024-08-12 at 10 34 00 PM

Screenshot 2024-08-13 at 12 15 48 AM

- OS: MacBook Air
- Browser: Chrome

zwliew:
Fixes #3767, which is a regression caused by an oversight in #3752.

### DIFF
--- a/website/src/views/planner/PlannerSemester.tsx
+++ b/website/src/views/planner/PlannerSemester.tsx
@@ -79,6 +79,11 @@ const PlannerSemester: React.FC<Props> = ({
 
     const isInTimetable = moduleCode !== undefined && moduleCode in timetable;
 
+    const displayedConflict =
+      year === config.academicYear || (conflict && ['prereq', 'duplicate'].includes(conflict.type))
+        ? conflict
+        : null;
+
     return (
       <PlannerModule
         key={id}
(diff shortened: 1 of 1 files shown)


ravern (approved):
LGTM

PR #3771, opened 2024-08-17
website: timetable: Correctly display timetable in 15-min intervals

zwliew:
## Context
The timetable currently only supports 30-min blocks. As a result, a course like ETP5341 which starts at 2.15pm is shown to start at 2.30pm instead. Fix this by displaying the timetable in 15-min blocks instead.

**Original horizontal view**
Screenshot 2024-08-17 at 12 14 55 PM

**New horizontal view**
Screenshot 2024-08-17 at 12 32 32 PM

**Original vertical view**
Screenshot 2024-08-17 at 12 17 05 PM

**New vertical view**
Screenshot 2024-08-17 at 12 32 50 PM


kokrui (approved):
LGTM! thanks for making the code more maintainable too!

PR #3779, opened 2024-08-19
Add spacing between text in workload visualisation component

### LINKED ISSUE #3775: No spacing between text in workload visualisation component
On the course details and course search pages, we show the workload of each course using these colourful blocks:

Course Details page:
image

Course Search page:
image

When the hours for a specific workload type exceed some number, the number of hours is displayed. However, there is no spacing between the text for the number of hours and the label for the workload type directly to the right (if it exists). 

For example: https://nusmods.com/courses/EBA5004/practical-language-processing 
image

This looks odd, and we could consider adding a bit of spacing to the right of the "number of hours" label. We should make 

Relevant code: https://github.com/nusmodifications/nusmods/blob/master/website/src/views/components/module-info/ModuleWorkload.tsx

sarthakladhwe:
## Context

Closes #3775 

## Implementation

- Adjust the right margin of the workload label dynamically based on the value of 'hours'.
- If 'hours' is not a whole number, set a larger margin ('5%') for better spacing. Otherwise, use a smaller margin ('0.5rem').
image

### DIFF
--- a/website/src/views/components/module-info/ModuleWorkload.tsx
+++ b/website/src/views/components/module-info/ModuleWorkload.tsx
@@ -99,7 +99,12 @@ const ModuleWorkload = React.memo<Props>(({ workload }) => {
                 className={styles.moduleWorkloadComponent}
                 style={{ width: `${(100 / ROW_MAX) * Math.min(ROW_MAX, Math.ceil(hours))}%` }}
               >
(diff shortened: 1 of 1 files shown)


ravern (approved):
LGTM! Thank you for your contribution!

PR #3798, opened 2024-08-21
Replaced google groups email

### LINKED ISSUE #3777: Replace occurrences of Google Groups email
We used to use a Google Groups forum (`nusmods<at>googlegroups<dot>com`) to take public questions / requests, and have the email `mods<at>nusmods<dot>com`. However, we've since decided that we should prefer all communications take place via `mods<at>nusmods<dot>com` for various reasons, and that we should eventually deprecate the Google Group.

We should take the first step towards doing that by replacing all occurrences of the old email with the new email on the website!

shelialia:
## Context
Closes #3777 
Replacing all occurrences of the old email with the new email on the website. 

## Implementation
Under the opensearch.xml file, I replaced the old email with the new email according to the pull request. 
Under the FAQ tab, there is also an occurrence of the email and I replaced it. 
Screenshot 2024-08-21 at 3 07 28 PM


zwliew:
Hi, thanks so much for your contribution. However,

Issue #3776, opened 2024-08-19
AS3-0312 venue data looks incorrect


kokrui:
The venue data for AS3-0312 states that the name of the venue is `0314`. That feels incorrect to me -- we should figure it out and change it (or close this issue if it's correct!)

See: https://nusmods.com/venues/AS3-0312?q=AS3-0312

Relevant code: https://github.com/nusmodifications/nusmods/blob/59b78cc98376ed4ce3d3e20086614b9e92a46ad2/website/src/data/venues.json#L4246-L4253

PR #3801, opened 2024-08-21
fix venue data for AS3-0312

xinyuwang-nus:
## Context
Resolves https://github.com/nusmodifications/nusmods/issues/3776


zwliew:
Thank you!

PR #3810, opened 2024-08-24
website: Use local venue data for development and test

### LINKED ISSUE #3803: Use local venue data for development instead of prod data
### Describe the bug

The venue json data is updated locally, but the localhost is using prod data

### To Reproduce

1. Change local data for any venue, launch localhost and find venue data is not updated
2. Go to inspection on browser and see screenshot below

### Expected behavior

Use local venue data instead of prod data

### Screenshots

[
Pasted Graphic 1 copy
](url)

kavishsathia:
## Context

Closes #3803 

## Implementation

The solution checks if the user is on localhost, and if they are, it will opt to use the local venue data instead of the production data by default. To override this behavior, the user can add a query parameter `prodVenue` to their URL, which will enable them to use production data instead.


zwliew:
May I know what's the motivation behind changing this default?

website/src/utils/debug.ts
```diff
@@ -8,6 +8,10 @@ function getParams() {
   return qs.parse(window.location.search);
 }
 
+function isLocalhost() {
+  return window.location.hostname === 'localhost' || window.location.hostname === '127.0.0.1';
+}
+
```
ravern, line 14: Would it be make sense to use `NUSMODS_ENV` instead?

```typescript
// Snippet from src/types/global.d.ts

/**
 * NUSMods deployment environment.
 *
 * **Note:** although these environments may share the same name as
 * `NODE_ENV` and other envs, there are subtle differences. Definitions:
 *
 * - `production`: a production deployment of NUSMods, i.e. a deployment meant to
 *   be used by the general public.
 * - `staging`: a deployment of NUSMods's main development branch, which is
 *   not meant to be used by the general public but which may be promoted to
 *   production at any point.
 * - `preview`: a deployment of work-in-progress branches, e.g. PR deploy
 *   previews.
 * - `test`: we are in a test environment, e.g. a jest test run.
 * - `development`: all other situations.
 */
declare const NUSMODS_ENV: 'development' | 'production' | 'staging' | 'preview' | 'test';
```

website/src/utils/debug.ts
```diff
@@ -40,6 +44,11 @@ export function allowBusStopEditing() {
   return getParams().edit === '1';
 }
 
+// prodVenue query parameter allows the user to use production data instead of local data
+// By default, on localhost, local data will be used
 export function preferRepoVenues() {
-  return getParams().localVenue === '1';
+  // Don't use repo data in jest tests
+  if (process.env.JEST_WORKER_ID !== undefined) return false;
```
ravern, line 51: Using `NUSMODS_ENV` would allow checking for `NUSMODS_ENV === 'test'` as well.

ravern (approved):
Thank you for this contribution! Sorry it took me this long to get to it.

I like the idea of flipping the venue fetching such that it uses local by default and global in production.

Was there a specific reason you used a `window.location.hostname` check instead of the current environment string? If there wasn't, I'm thinking it might be better to use the existing config than the use the hostname.

ravern (changes requested):
Sorry, meant to request changes.

zwliew (approved):

zwliew:
I've made the following changes:
- Use NUSMODS_ENV to detect test and development environments
- Use localVenue and prodVenue flags to override the defaults in each case.

PR #3818, opened 2024-09-04
Fix privacy opt-out button not clickable

### LINKED ISSUE #3254: Privacy Opt-Out not clickable
### Describe the bug

Privacy Opt-Out not clickable

Accessed on Firefox 86.0 (64-bit)

yk-tuturu:
## Context
closes #3254 

## Implementation
 following instructions detailed in issue #3254, implemented checks to see if matomo is initialized and disables the privacy toggle if blocked by an ad-blocker


ravern (commented):
Thank you for your contribution! I am able to see your fix as intended — the toggle is set to "Opt-out" and disabled if `matomo` is not initialised.

While I think your solution fulfils the [original comment](https://github.com/nusmodifications/nusmods/issues/3254#issuecomment-868986124) made by an earlier maintainer, I think that adding a hook is unnecessary (and an incorrect usage of React hooks, see [this comment](https://github.com/nusmodifications/nusmods/pull/3818#discussion_r1764511588)).

I think the crux of this issue is if `matomo` is `undefined`, the opt-out toggle cannot be clicked, leading the user to believe that they cannot opt-out of the tracking, when in fact there is no tracking at all.

The current code that modifies the UI component actually already fixes this issue, and the addition of the hook is unnecessary. I think if **we reverse the changes to the hook portion of this PR**, it will LGTM.

website/src/bootstrapping/matomo.ts
```diff
@@ -77,3 +78,13 @@ export function trackPageView(history: History) {
...
+
+export function useMatomo() {
+  // need to use useState here or the matomo returned will always be undefined
+  const [matomoCopy, setMatomoCopy] = useState<Tracker | undefined>(undefined);
+  useEffect(() => {
+    setMatomoCopy(matomo);
+  }, [matomo]);
+
+  return matomoCopy;
+}
```
ravern, line 90: In the following code:
```typescript
useEffect(() => {
    setMatomoCopy(matomo);
  }, [matomo]);
```
This effect is only ever called once (on component mount) since the `matomo` variable exists outside of the React context. This is reflected by the linter as well: 

```
React Hook useEffect has an unnecessary dependency: 'matomo'. Either exclude it or remove the
dependency array. Outer scope values like 'matomo' aren't valid dependencies because mutating
them doesn't re-render the component.
```

ravern (changes requested):
Remove hook

website/src/bootstrapping/matomo.ts
ZhangYiJiang: I don't really understand this code at all. This in effect is forcing a re-render, but there are better ways to do that if that's the intention

ravern:

PR #3820, opened 2024-09-08
SASSDeprecationWarning

### LINKED ISSUE #3819: Fix SASS deprecation warnings during build
Currently, when building the web app with `yarn build`, the build process spews a lot of warnings, mostly regarding deprecated SASS features.

Any help with fixing these deprecation warnings is much appreciated!

Priyanshu1035:
## Context
Fixes #3819 

## Implementation
 Added `quietDeps : true` in saas option

## Other Information
  One can also deprecate the warnings by using the command `SASS_WARNINGS=false yarn build` in the terminal or command 
  line manually during yarn build.


zwliew:
Thanks for this! However, I would like to fix the issues causing the warnings as much as possible, and only rely on silencing the warnings as a last resort.

ravern:
I think we're in agreement that we're not going to silence the warnings, so I'm closing this PR.

PR #3823, opened 2024-09-10
Reduce SASS Deprecation Warnings

kaze-droid:
## Context
Partial Fix for Issue #3819 

## Implementation
Most SASS warnings were fixed by upgrading the packages:

```
yarn upgrade sass bootstrap @material/theme @material/ripple @material/fab
```

To temporarily fix the warning on passing percentage units to the global `abs()` function, I created a custom SCSS solution that defines an `abs()` function in `src/styles/scss/bootstrap/function-overrides.scss` that internally uses `math.abs()` and import it before the Bootstrap SCSS files

For the warning on using `/` for division, I used the automatic Sass migrator to refactor the code
```
npx sass-migrator division '**/*.scss'
```

However, this approach has limitations. It incorrectly uses `math.div` instead of `list.slash` in `node_modules/bootstrap/scss/mixins/_forms.scss`. Additionally, running the migrator with the `--pessimistic` flag resulted in numerous false negatives. For more details, [sass-migrator issues](https://github.com/sass/migrator/issues/205) 

## Other Information
Reduced SASS deprecation warnings
<ins>Image of reduced SASS deprecation warnings</ins>

A complete and permanent solution to these warnings will require migrating to BS5

### DIFF
--- a/website/.stylelintrc.js
+++ b/website/.stylelintrc.js
@@ -31,6 +31,7 @@ module.exports = {
           'return',
           'if',
           'else',
+          'use'
         ],
       },
     ],
--- a/website/src/styles/bootstrap/bootstrap.scss
+++ b/website/src/styles/bootstrap/bootstrap.scss
@@ -1,6 +1,7 @@
 // We duplicate the _bootstrap.scss so that we only include components we need
 
 // Our own variable overrides
+@import "function-overrides";
 @import "variable-overrides";
 
 // Core variables and mixins
--- a/website/src/styles/bootstrap/function-overrides.scss
+++ b/website/src/styles/bootstrap/function-overrides.scss
@@ -1,0 +1,5 @@
+@use 'sass:math';
+
+@function abs($value) {
+  @return math.abs($value);
+}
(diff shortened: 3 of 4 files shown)


ravern:
Thanks for your contribution!

I think the `sass-migrator` work isn't committed in this branch, can I confirm this is intentional, i.e., it doesn't work and won't be a part of this PR?

kaze-droid:
Oops I initially ran sass-migrator locally, which only updated the files in my local node_modules directory. I've just added a new script in package.json to run sass-migrate before the build process.

zwliew:
> Oops I initially ran sass-migrator locally, which only updated the files in my local node_modules directory. I've just added a new script in package.json to run sass-migrate before the build process.

I think there has been a mistake. sass-migrator should only need to be run once and should not be committed to the repo. Please revert your latest change!

kaze-droid:
Apologies for the confusion. I've reverted the changes as requested. Running sass-migrator only impacted files within node_modules, specifically with deprecation warnings originating from `@material/theme`. I attempted to update it, but the latest version that still supports BS4 doesn't resolve those warnings.

ravern (approved):
LGTM! Thank you very much for helping with this long-standing issue.

PR #3741, opened 2024-07-07
Migrate NUSMods to AY24/25

kokrui:
## Context
This PR migrates NUSMods scraper and the NUSMods frontend to AY24/25 (Sem 1).

Academic Calendar + 2024 Holidays and Well-Being Day from https://nus.edu.sg/registrar/docs/info/calendar/ay2024-2025.pdf

CourseReg Schedule from https://nus.edu.sg/coursereg/docs/UGCPE_CourseRegSchedule.pdf and https://www.nus.edu.sg/coursereg/docs/GD_CourseRegSchedule.pdf

## Other Information
Pending NUS API update before merge

PR #3832, opened 2024-10-11
Add CPEx notifications for AY24/25 S2

ravern:
Adds CPEx as a modreg schedule item, and fixed a small visual bug (extra whitespace).

image

Previous CPEx PRs:
* Added in #3665
* Removed in #3741


kokrui (approved):
LGTM, and nice catch on the extra whitespace!

PR #3836, opened 2024-10-15
Update error message for unexpected auth errors

jloh02:
## Context
"Unexpected error occurred" message occurs when login via NUS using email passes but no relayState given by NUS API

## Implementation
Return additional info by overriding defaultRescue handler


ravern (approved):
LGTM

PR #3291, opened 2021-06-26
Pr/reverse prereq tree

alvynben:
## Context

Fixes #3202 

This issue requests the reversal of the orientation of the prerequisite tree in the module page.

## Implementation

I have edited moduletree.tsx and moduletree.scss to reverse the order of the prerequisite tree. Then, I changed some of the words used to make the new format clearer. Finally, I updated the snapshots in moduletree.test.tsx to reflect the change in UI.

**For a single mod, with no prerequisite or dependant:**
image

**For mods with only dependants:**
image

**For mods with one layer of prerequisites and dependants:**
image

**For mods with two layers of prerequisites:**
image

**For mods with uneven layers of prerequisites:**
image

**For mods with uneven letters in prerequisites:**
image

## Other Information

Hello, I think the code works in most cases that I've managed to test, both on desktop and mobile. However, I feel that there should be a way to 'fix' the uneven nodes featured in the last picture. Do you have any suggestions?

This is my first issue and pull request by the way, so I do apologize for any mistake on my part. Thank you for taking the time to read through it.

PR #3840, opened 2024-10-21
Left to right prerequisites trees

zehata:
## Context

Issue #3202 Prerequisite tree right-to-left progression is confusing

## Implementation

I referenced the previous 2 PRs #3263 and #3291 to see what was good and why they were not merged.

Having noted https://github.com/nusmodifications/nusmods/pull/3263#discussion_r862281206, the implementation was changed to modify the DOM instead of using `flex-direction: row-reverse;` I **fully copied** @sciffany's copywriting since they are indeed easier to understand, so all credits to her.

Having noted the test cases used by @alvynben in PR #3291, I elected to use those cases (though with different modules).

I also noted that the last correspondence for both PRs were in 2021, so I believe opening a new PR should be appropriate.

### Options in user settings?
I noted that Issue #3202 specifically asked for left-to-right prerequisites trees, so this PR only targets that specification. However, I would favour an option in the settings that allow users to choose their preference. Having gotten used to the old prerequisites trees, users like myself might choose to stick to that. However, I also note that a left-to-right prerequisites tree is definitely more intuitive so it should not be counted out as well.

### UI tests

For a single mod, with no prerequisite or dependant:
image
For mods with only dependants:
image
For mods with one layer of prerequisites and dependants, also for mods with `requireNum` modules:
image
For mods with two layers of prerequisites:
image
For mods with uneven layers of prerequisites:
image
Highly complex prerequisites, prerequisites with uneven number of letters:
image

#### Any questions you may have
Is there a tele group specific for development?
#### Letting us know that you're new to this (so we know how much to help out)
Been almost 3 years since I touched any web dev. Sorry in advance if I had missed out anything.


zehata:
Fixed #3202

zehata:
Also, I wanted to make some other minor changes or refactor some things but if I get to that I will open a separate PR.

zehata:
Re: https://github.com/nusmodifications/nusmods/pull/3840/files#diff-c16227ad8088f0a3a8edf1ae53344bb0189198cb956157ec0a9083cf3e7069f9L106-L112 
The code originally had `prereqNode` class on conditional nodes. As far as I understand, modules don't conditionally unlock other modules, i.e. there are no situations like so:
```
B —┐
C —|— needs (isConditional == true) — A 
D —┘
```
so `isConditional` is never `true` and this style declaration is unused, 

https://github.com/nusmodifications/nusmods/pull/3840/files#diff-c16227ad8088f0a3a8edf1ae53344bb0189198cb956157ec0a9083cf3e7069f9R121 omits `{[styles.prereqNode]: isPreReq}` since there are never a situation whereby:
```
                    ┌— B
A — unlocks one of —|— C
                    └— D
```

website/src/views/modules/ModuleTree.tsx
```diff
@@ -103,18 +112,13 @@ const Tree: React.FC<TreeDisplay> = (props) => {
...
-        >
-          {formatConditional(node)}
-        </div>
         <Branch
           nodes={unwrapLayer(node)}
           layer={layer + 1}
+          isPrereq={isPrereq}
           getModuleCondensed={props.getModuleCondensed}
         />
+        <div className={classnames(styles.node, styles.conditional)}>{formatConditional(node)}</div>
```
zehata, line 121: `[styles.prereqNode]: isPrereq` is omitted
See: https://github.com/nusmodifications/nusmods/pull/3840#issuecomment-2428038029

zehata:
Update: I have a working copy of a version with a toggle in settings, but I am still refactoring. See preview at temporary branch https://github.com/zehata/nusmods/tree/zehata/website-module-page/3202/toggleable-left-to-right-prerequisites-trees-wip

zehata:
Reverted to draft because of
image

jloh02:
I think to allow this feature to have a toggle switch setting in future, it would be good to also have a boolean parameter to switch back to right-to-left.

zehata:
Toggleable implementation extending this PR is at #3842.

zwliew:
Since we've already merged #3842, this shouldn't be needed anymore.

Issue #3869, opened 2024-11-23
Reading week should only include weekdays but not weekends


ravern:
### Describe the bug

Currently, reading week is considered as the Mon to Sun after Week 13. However, according to the NUS academic calendar, it is technically exam week already.

https://nus.edu.sg/registrar/docs/info/calendar/ay2024-2025.pdf

### To Reproduce

1. Go to https://nusmods.com/today?date=2024-11-23 (explicit date)
2. Observe that it still counts as reading week

### Expected behavior

It should show as exam week instead.

### Screenshots

![image](https://github.com/user-attachments/assets/5818297c-4b2a-442c-8ee7-dfc8426a7eb6)

PR #3870, opened 2024-11-23
modify calculation of week type

kaze-droid:
## Context
Resolves Issue #3869 

## Implementation
Check if its a weekend on special acad weeks (recess, reading, examination) before returning the weekType


kokrui (approved):
Thanks for the PR! This looks like a reasonable implementation to me for this edge case. I'll add to the comments here and there and get this merged.

PR #3875, opened 2024-12-01
feat(website): Add TA mode

### LINKED ISSUE #3804: Allow TA's to hide non tutorial lessons

TA's would want to add timetable slots for tutorials that they are TA'ing but they can't only add a tutorial slot from a module.

A TA should be able to select a module and have an option to only show tutorial lessons on the timetable, additionally it should not count the units of the module in the total units

A more general feature can be to hide/show any lessons on the timetable, but the current code base only allows for all lessons in a module to be hidden so it might be annoying to refactor

leslieyip02:
## Context

Closes #3804.

Adds an option for TAs to hide non-TA lessons. The original issue requested an option to hide non-tutorial lessons, but this would hide labs and recitations as well, so I decided to implement a dropdown menu to toggle TA mode for different lesson types.

I have not been a TA before, so do let me know if this makes sense?

## Implementation

- Each module has a few different lesson types, so `TimetableState` was updated to have a `TaModuleConfig`, which maps each module to an array of lesson types.
- For modules with TA lessons, `TimetableContent` only renders lessons which are marked as TA and ignores other lesson types.
- TA modules will not be counted towards total units, and will also not be shown in the exam calendar.
- TA modules are suffixed with `(TA)` in the timetable
- The `ColorPicker` for TA modules has a half-filled color box

### Toggle TA mode

https://github.com/user-attachments/assets/7c8e890e-802d-44fa-bea6-77fd7e713ece

### No TA-able mods

When the mod has no TA-able mods (i.e. just lectures), the menu button is disabled.

image

### Export/Import

https://github.com/user-attachments/assets/965b2bf2-6415-4a59-a2fa-5b6d1ef6a0ca

### Exam clash

TA mods are not shown in the exam calendar, and do not contribute to exam clashes.

Left: Updated, Right: Current

Screenshot 2024-12-01 165844

I also made it so that hidden mods no longer creates a clash warning.

Screenshot 2024-12-01 165811

## Other Information

I had some trouble trying to get `Tooltip` to work with `Downshift`. When using `Tooltip` inside `DownShift`, I needed to call `getRootProps`, but it seems to be creating these warnings from `tippy.js`:

Screenshot 2024-12-01 170802

Since the versions of both `@tippy.js/react` and `downshift` are quite outdated, I wasn't able to find docs to resolve the warnings.


zwliew:
This looks quite good so far! However, the ability to add multiple tutorials/lab/recitations to the timetable in TA mode is missing. I suggest making each tutorial/lab/recitation slot toggleable in TA mode, instead of the usual behaviour that switches between slots.

zwliew (changes requested):
This looks quite good so far! Thanks for even considering timetable exports and adding some tests.

However, the ability to add multiple tutorials/lab/recitations slots to the timetable in TA mode is missing. This is useful for those TA-ing multiple slots. I suggest making each tutorial/lab/recitation slot toggleable in TA mode, instead of the usual behaviour that switches between slots.

leslieyip02:
I've added the ability to add multiple slots in TA mode. Let me know if the implementation makes sense.

https://github.com/user-attachments/assets/2495cedd-4cf6-43dc-a3ee-4f3b83e3de74

website/src/views/timetable/ModulesTableFooter.tsx
```diff
@@ -26,21 +27,38 @@ export const moduleOrders: { [moduleTableOrder: string]: ModuleOrder } = {
...
+): number {
+  return sumBy(
+    modules.filter(
+      (module) =>
+        !hiddenInTimetable.includes(module.moduleCode) &&
+        isEmpty(taInTimetable[module.moduleCode] ?? []),
+    ),
+    (module) => parseFloat(module.moduleCredit),
+  );
+}
```
zwliew, line 47: Refactoring this code into two other functions is quite unnecessary IMO, since the variables `totalMCs` and `shownMCs` already tell what the code is trying to do.

website/src/views/components/ColorPicker.scss
```diff
@@ -10,6 +10,18 @@ $arrow-height: 1.5rem;
 
 .container {
   position: relative;
+  border: 1px solid transparent;
+  border-radius: 0.25rem;
+
+  &.hidden {
+    border-style: dashed !important;
```
zwliew, line 17: Is `!important` necessary here?

website/src/types/timetables.ts
```diff
@@ -10,6 +10,16 @@ export type SemTimetableConfig = {
   [moduleCode: string]: ModuleLessonConfig;
 };
 
+// TaModulesConfig is a mapping of moduleCode to the TA's lesson types.
+export type TaModulesConfig = {
+  [moduleCode: ModuleCode]: TaLessonConfig;
```
zwliew, line 15: It would be simpler to map `ModuleCode` to an array of pairs`(lessonType, ClassNo)[]`. Do we need the extra indirection from the `TaLessonConfig` here?

website/src/utils/ical.ts
```diff
@@ -187,10 +189,15 @@ export default function iCalForTimetable(
 
     _.each(lessonConfig, (lessons) => {
       lessons.forEach((lesson) => {
+        if (moduleCode in taModules && !(lesson.lessonType in taModules[moduleCode])) {
+          return;
+        }
```
zwliew, line 194: I don't think this works correctly. The guard should simply return early if `moduleCode in taModules`, and we should have a separate loop afterwards to go through `taModules` and add them to the calendar.

zwliew (changes requested):
The implementation looks mostly fine! The only main issue is that there should be a single TA mode toggle for each course instead of per-lesson-type toggles. Besides that, there are a few minor questions and bugs to be resolved.
This PR should be ready for merging soon after these issues are resolved.

website/src/views/components/ColorPicker.scss
leslieyip02: I marked it with `!important` to handle the case where the mod is tagged as hidden and TA. But actually, I can just rearrange the styles for `&.ta` to be above `&.hidden` so that the dashed border will still show

website/src/views/timetable/ModulesTableFooter.tsx
leslieyip02: I initially created the functions so that it would be easier to write unit tests, but you are right that it's a bit redundant.

zwliew: Oh good point -- I didn't notice the tests.

export/src/render.ts
```diff
@@ -28,6 +28,7 @@ export async function launch() {
     headless: true,
     executablePath: config.chromeExecutable,
     devtools: !!process.env.DEVTOOLS,
+    args: ['--disable-gpu'],
```
leslieyip02, line 31: I'm not sure if this is absolutely necessary, but I couldn't get `browser.newPage()` to work without this.

docker-compose.yml
```diff
@@ -40,7 +40,7 @@ services:
       - NODE_ENV=development
       - PORT=8082
       - HOST=0.0.0.0
-      - PAGE=http://website:8081
+      - PAGE=http://website:8081/timetable-only/
```
leslieyip02, line 43: I'm not sure if this is necessary.

leslieyip02 (commented):
To test `export`, I had to change some configurations. I am not too sure if the changes are appropriate, but I couldn't get the `export` to work without the changes. Let me know if I missed anything?

leslieyip02:
The TA mode toggle is applied on a per-module basis now:

zwliew (changes requested):

export/src/render.ts
zwliew: I think this is fine.

docker-compose.yml
zwliew: We should make the same change for `docker-compose.prod.yml`.

zwliew: We should make the same change for `docker-compose.prod.yml`

export/src/types.ts
```diff
@@ -7,8 +7,12 @@ export interface PageData {
   readonly timetable: {
     [moduleCode: string]: ModuleLessonConfig;
   };
+  readonly hiddenInTimetable: string[];
+  readonly taInTimetable: {
+    [moduleCode: string]: [lessonType: string, classNo: string][];
+  };
```
zwliew, line 13: Should these be `hidden` and `ta` instead?
Also, it might be worth copy pasting the necessary type definitions (i.e. `TaModulesConfig`, `ModuleCode`) from `website/src/types` instead.

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -451,14 +529,18 @@ class TimetableContent extends React.Component<Props, State> {
...
   const hiddenInTimetable =
-    ownProps.hiddenImportedModules ?? (state.timetables.hidden[hiddenModulesKey] || []);
+    ownProps.hiddenImportedModules ?? (state.timetables.hidden[semester] || []);
+  const taInTimetable = ownProps.taImportedModules ?? (state.timetables.ta[semester] || {});
+  const timetableWithLessons = hydrateSemTimetableWithAllLessons(
+    timetable,
+    taInTimetable,
+    modules,
+    semester,
+  );
```
leslieyip02, line 543: For reference, `hydrateSemTimetableWithAllLessons` is used here.

website/src/utils/timetables.ts
```diff
@@ -107,8 +116,25 @@ export function randomModuleLessonConfig(lessons: readonly RawLesson[]): ModuleL
...
+  modules: ModulesMap,
+  semester: Semester,
+): SemTimetableConfigWithLessons {
+  // Add normal lessons without TA lessons
+  // Add TA lessons separately because TA lessons can map multiple classNos to a single lessonType
+  const taModuleCodes = Object.keys(filter(semTimetableConfig, (config) => !isEmpty(config)));
+  const timetableWithoutTaModules = omit(semTimetableConfig, taModuleCodes);
+  return {
+    ...hydrateSemTimetableWithNormalLessons(timetableWithoutTaModules, modules, semester),
+    ...hydrateSemTimetableWithTaLessons(taModules, modules, semester),
```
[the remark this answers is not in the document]
zwliew: Indeed. What I meant was that we should not mess with  the implementation of `hydrateSemTimetableWithLessons` in this PR -- the goal is to keep the PR concise and to keep the logic for hidden and TA-ed courses consistent.

How about this:
- Keep the original implementation of `hydrateSemTimetableWithLessons`. That is, the resulting `SemTimetableConfigWithLessons` should keep all lessons, even those for hidden and TA-ed courses.
- In places where we want to filter out lessons that the user doesn't TA for (i.e. image/PDF exports, Today page, timetable import), add some additional logic to filter them out at the call sites. This should be somewhat similar to following changes done for hidden courses:

Today page: https://github.com/nusmodifications/nusmods/pull/3927
Image/PDF exports: https://github.com/nusmodifications/nusmods/pull/3925

As for timetable imports, you'd need to slightly modify https://github.com/nusmodifications/nusmods/pull/3875/files#diff-0da48c065607eeaa76534e07e1141c7cda4f7da06d03b7fc69bee4dfd355840aR536-R543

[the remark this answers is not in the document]
leslieyip02: Ah ok, that makes sense.

zwliew (approved):
I merged in master and cleaned up the code slightly, but overall LGTM. Great job on this!

HollaG:
Just want to piggyback on this thread - it appears certain courses cannot be set to TA mode (example: CS2103T, CS3281). Should I file a bug report?

zwliew:
> Just want to piggyback on this thread - it appears certain courses cannot be set to TA mode (example: CS2103T, CS3281). Should I file a bug report?

Good catch - please go ahead and file one.

HollaG:
I really would like this feature to become useful to TAs, but the implementation of removing only the lectures is also a little bit restrictive, unfortunately - most CS courses that have lab + tutorials often also have separate TAs for them, and the current implementation doesn't allow hiding of either lab/tut individually. 

Nevertheless, thank you guys for your efforts!

PR #3921, opened 2024-12-30
Fix misnamed AsyncPlannerContainer

### LINKED ISSUE #3920: Misnamed AsyncTodayContainer
### Describe the bug

I think this should be `AsyncPlannerContainer` instead of `AsyncTodayContainer`.

https://github.com/nusmodifications/nusmods/blob/656b1447b9dbaa40befac5130d52bcad60b2341c/website/src/views/planner/PlannerContainer/index.tsx#L6-L20

It was likely copied from [here](https://github.com/nusmodifications/nusmods/blob/656b1447b9dbaa40befac5130d52bcad60b2341c/website/src/views/today/TodayContainer/index.tsx).

leslieyip02:
## Context

Fixes #3920.


kokrui (approved):
thanks for the catch!

PR #3930, opened 2025-01-03
Fix TA mode for lecture lessons

### LINKED ISSUE #3929: TA Mode doesn't work for some courses
### Describe the bug

Some courses cannot be set to "TA Mode". 
Example courses:
- CS2103T
- CS3281

### To Reproduce

1. Add either one of these courses to the timetable
2. Click on TA Mode Button
3. Nothing happens (no errors too, in console)

### Expected behavior

TA mode is enabled

- OS: Windows 10
- Browser: Chrome
- Version 131.0.6778.205

leslieyip02:
## Context

Fixes #3929. 

## Implementation

Previously, TA mode will filter out all lectures. However, some mods (e.g. CS2103T) only have lecture lessons, so this PR makes all lessons togglable.

`TaModulesConfig` now stores `[lessonType, classNo, startTime, day]` instead of `[lessonType, classNo]`.

```
export type TaModulesConfig = {
  [moduleCode: ModuleCode]: [
    lessonType: LessonType,
    classNo: ClassNo,
    startTime: StartTime,
    day: DayText,
  ][];
};
```

For the migration, I'm not sure on how `startTime` and `day` data can be pulled. Right now, I am wiping the existing TA config to avoid conflicts. Let me know if there's a better way to do this?

### DIFF
--- a/export/src/types.ts
+++ b/export/src/types.ts
@@ -12,9 +12,11 @@ export type ThemeState = Readonly<{
 }>;
 export type ColorScheme = 'LIGHT_COLOR_SCHEME' | 'DARK_COLOR_SCHEME';
 export type Semester = number;
-export type ClassNo = string; // E.g. "1", "A"
-export type LessonType = string; // E.g. "Lecture", "Tutorial"
 export type ModuleCode = string; // E.g. "CS3216"
+export type LessonType = string; // E.g. "Lecture", "Tutorial"
+export type ClassNo = string; // E.g. "1", "A"
+export type StartTime = string;
+export type DayText = string;
 export type SemTimetableConfig = {
   [moduleCode: ModuleCode]: ModuleLessonConfig;
 };
@@ -22,7 +24,12 @@ export interface ModuleLessonConfig {
   [lessonType: LessonType]: ClassNo;
 }
 export type TaModulesConfig = {
-  [moduleCode: ModuleCode]: [lessonType: LessonType, classNo: ClassNo][];
+  [moduleCode: ModuleCode]: [
+    lessonType: LessonType,
+    classNo: ClassNo,
+    startTime: StartTime,
+    day: DayText,
+  ][];
 };
 
 // `ExportData` is duplicated from `website/src/types/export.ts`.
--- a/website/src/reducers/timetables.ts
+++ b/website/src/reducers/timetables.ts
@@ -192,6 +192,13 @@ function semTaModules(state = DEFAULT_TA_STATE, action: Actions): TaModulesConfi
     case ADD_TA_LESSON_IN_TIMETABLE: {
       const { moduleCode, lessonType, classNo } = action.payload;
       if (!(moduleCode && lessonType && classNo)) return state;
+      // Prevent duplicate lessons
+      if (moduleCode in state) {
+        const isDuplicate = state[moduleCode].some((lesson) =>
+          isEqual(lesson, [lessonType, classNo]),
+        );
+        if (isDuplicate) return state;
+      }
       return {
         ...state,
         [moduleCode]: [...(state[moduleCode] ?? []), [lessonType, classNo]],
--- a/website/src/reducers/timetables.ts
+++ b/website/src/reducers/timetables.ts
@@ -49,9 +49,18 @@ export const persistConfig = {
       // eslint-disable-next-line no-underscore-dangle, @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-non-null-asserted-optional-chain
       _persist: state?._persist!,
     }),
+    3: (state) => ({
+      ...state,
+      ta: {},
+      // FIXME: Remove the next line when _persist is optional again.
+      // Cause: https://github.com/rt2zz/redux-persist/pull/919
+      // Issue: https://github.com/rt2zz/redux-persist/pull/1170
+      // eslint-disable-next-line no-underscore-dangle, @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-non-null-asserted-optional-chain
+      _persist: state?._persist!,
+    }),
   }),
   /* eslint-enable */
-  version: 2,
+  version: 3,
 
   // Our own state reconciler archives old timetables if the acad year is different,
   // otherwise use the persisted timetable state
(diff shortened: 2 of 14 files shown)


export/src/types.ts
```diff
@@ -12,17 +12,24 @@ export type ThemeState = Readonly<{
...
   [lessonType: LessonType]: ClassNo;
 }
 export type TaModulesConfig = {
-  [moduleCode: ModuleCode]: [lessonType: LessonType, classNo: ClassNo][];
+  [moduleCode: ModuleCode]: [
+    lessonType: LessonType,
+    classNo: ClassNo,
+    startTime: StartTime,
+    day: DayText,
+  ][];
```
zwliew, line 32: `TaModulesConfig` is starting to get quite big. I'm starting to think that it would be better to store this state differently. With this PR the state would look like:
```
timetable: {
	lessons: {
	  1: {
	     // note: the "CS2103T" keyed object is completely useless since the user is a TA for CS2103T.
	    CS2103T: {
	    	Lecture: "G01"
	   }
	},
	ta: {
		1: {
			CS2103T: [["Lecture", "G01", "0800", "Monday"]]
		}
	}
}
```
Instead, we could have `addTaLesson` actions overwrite the "lessons" keyed object like so:
```
timetable: {
	lessons: {
	  1: {
	    CS2103T: {
				// note: values are now arrays of arrays.
	    	Lecture: [["G01", "0800", "Monday"]]
	   }
	},
	ta: {
		1: ["CS2103T"]
	}
}
```
Note that we would have to change the type of the values for `ModuleLessonConfig` to an array.
Coincidentally (or not), this change causes the state schema for "ta" and "hidden" to match (they are both arrays of `ModuleCode`s.

For readability, we can go even further and replace `["G01", "0800", "Monday"]` with:
```
{
	classNo: "G01",
	startTime: "0800",
	day: "Monday"
}
```
Hence, the resulting state would look like:
```
timetable: {
	lessons: {
	  1: {
	    CS2103T: {
	    	Lecture: [
					{
						classNo: "G01",
						startTime: "0800",
						day: "Monday"
					}
				]
	   }
	},
	ta: {
		1: ["CS2103T"]
	}
}
```

As for the serialized state, it changes from `CS2103T=LEC:G01&ta=CS2103T(LEC:G01:0800:Monday)` to `CS2103T=(LEC:G01:0800:Monday)&ta=CS2103T`.

leslieyip02: Another thing I'm not sure how to handle is that there are some mods with lessons with identical startTime, endTime and day (e.g. HSI1000). The only thing differentiating the lessons are the weeks and venue. In this case, should we add weeks/venue to the schema as well? I feel that it would be introducing bloat for such a rare scenario though.

website/src/reducers/timetables.ts
```diff
@@ -190,20 +199,27 @@ function semTaModules(state = DEFAULT_TA_STATE, action: Actions): TaModulesConfi
...
+        const isDuplicate = state[moduleCode].some((existingConfig) =>
+          isEqual(existingConfig, [lessonType, classNo, startTime, day]),
+        );
+        if (isDuplicate) return state;
+      }
       return {
         ...state,
-        [moduleCode]: [...(state[moduleCode] ?? []), [lessonType, classNo]],
+        [moduleCode]: [...(state[moduleCode] ?? []), [lessonType, classNo, startTime, day]],
       };
```
zwliew, line 214: Can we merge this specific change in a separate PR first, without the addition of `startTime` and `day`? It is quite useful on its own even without the rest of the changes.

Also, we can simplify it to:
```javascript
const newLesson = [lessonType, classNo];
const curLessons = state[moduleCode] ?? [];
// Prevent duplicate lessons
if (curLessons.some((lesson) => isEqual(lesson, newLesson)) return state;
return {
	...state,
	[moduleCode]: [...curLessons, newLesson],
};
````

File: export/src/types.ts (hunk 1 above)
Comment (zwliew) [reply export/src/types.ts:32]: That's a problem. Can we hold off on the addition of `startTime` and `endTime` and any other distinguishing identifiers for now? Let's just make it possible to add lectures to TA, and figure out what we really need to make this feature possible.

Comment (leslieyip02) [reply export/src/types.ts:32]: Yep, that makes sense to me.

Comment (zehata) [reply export/src/types.ts:32]: @leslieyip02 Regarding the lessons with identical startTime, endTime and day in HSI1000, I am not sure which ones are colliding, it appears that they have different classNo?

Comment (leslieyip02) [reply export/src/types.ts:32]: image

The workshop lessons use the same `classNo`, but are offered in different weeks/venues.

So we would need to differentiate between different days, start/end times, weeks and venues.

Comment (zehata) [reply export/src/types.ts:32]: @leslieyip02 Ah I see I have just noticed that it disappears when modifying the schedule because the dedupe code uses classNo, which is a bug

Comment (leslieyip02) [reply export/src/types.ts:32]: One idea I had was to just sort all lessons of a particular type based on all of those fields, and then use the index of that lesson in the list as a key. This would be easier to serialize as well.

Feels a bit hacky though.

Comment (zehata) [reply export/src/types.ts:32]: Yeah... it feels quite hacky, using index as a key doesn't seem like a good idea to me.

Comment (zehata) [reply export/src/types.ts:32]: It looks like each module data already has individual lessons in a list inside semesterData, so if we are going the index-as-key route it might be possible to use that. But I am still worried about how stable the list is.

Comment (zehata) [reply export/src/types.ts:32]: https://github.com/user-attachments/assets/a3c1d5c5-5f56-4674-9cb4-2fd81376e431

I've gotten it working with the index-as-key method, using the list already in semesterData, and using the a structure adapted from https://github.com/nusmodifications/nusmods/pull/3930#discussion_r1902257170 :

```
timetable: {
	lessons: {
		1: {
			CS2103T: {
				Lecture: [15]
			}
		}
	},
	ta: ["CS2103T"]
}
```

Despite having adapted the validation mechanism over, I'm still concerned about the stability of the list. I think I can try to cook up a more sophisticated mechanism whereby it finds the minimum identifying information to save into state, and that should make it such that it doesn't depend on the order of the list, while minimizing the amount of distinguishing identifiers that need to be saved.

leslieyip02: Good work so far! This definitely isn't easy. I think as long as the list of `RawLesson[]` is sorted properly, the index-as-key method seems to be ok for now. I can help with adding unit tests as well.

The main weakness of this method is that if the module data is updated, then the order of lessons could change. But TBH if the lesson indices do changes, the timetable would need to be manually updated by the user anyways since new lesson options are available/existing lessons have changed/been removed. We would just need to handle the case where the index is out-of-bounds.

What do you think?

leslieyip02: In the meantime, I'll also try to look for the edge cases that aren't covered so we can try to figure out what the minimum identifying info is:

- Same classNo, different days: CG1111A
- Same classNo, same day, start time and end time, different weeks: HSI1000
- Same classNo, same lesson type at different times: CS2103T
- Same classNo, different lesson types at different times: FST2102B
    - There's only 1 lab and 1 tut slot, but because it's grouped together, you must have both
    - The current logic blocks selecting options if there's only 1 possible slot for that lesson type
    - This might be out-of-scope

So far, it looks like we need day, start time, end time, weeks and lesson type.

I think weeks might be a problem for serialization because it can either be an array or JSON.

zehata: > But TBH if the lesson indices do changes, the timetable would need to be manually updated by the user anyways since new lesson options are available/existing lessons have changed/been removed.

Yeah... Furthermore, currently things fail silently, so I think it would make sense to notify users when any discrepancies are found. It would be a separate issue though

> We would just need to handle the case where the index is out-of-bounds.

Yeah, that's part of the validation. currently I am handling it naively such that when it finds a non-valid lesson in normal mode it would also try to recover by finding lessons with the same classNo

zehata: @leslieyip02 For the 4th example, suppose there were 2 classes, 4 lessons

LEC:01 <- currently selected
TUT:01 <- currently selected
LEC:02
TUT:02

Do we need to make it such selecting LEC:02 will also switch the tutorial lesson type to TUT:02?

zehata: @zwliew @leslieyip02 From https://github.com/nusmodifications/nusmods/pull/3930#discussion_r1902257170
> When switching from TA to non-TA mode, we should also pick the first lesson for each type and discard the rest.

I think it would be better to take into account the lessons that the user currently have selected in TA mode, it's also not too difficult to implement

https://github.com/user-attachments/assets/acd0c96e-cb49-45a5-86f2-7cf3425082f9

zehata: @zwliew @leslieyip02 Since this discussion has veered from the original PR review, I have created a discussion which I think would be a more appropriate place to talk about the proposed changes and discuss the implementation
https://github.com/nusmodifications/nusmods/discussions/4098

leslieyip02:
Should be closed by #4431.

PR #3932, opened 2025-01-03
fix(website/ta): Hide TA button for non TA-able mods

### LINKED ISSUE #3931: TA mode button should be disabled for courses with no lessons
### Describe the bug

Some courses (e.g. internship courses) have no lessons. The TA button does nothing, but the tooltip still prompts users to "Enable TA".

### Expected behavior

The button should be disabled and should not show a tooltip to toggle TA for the mod.

### Screenshots

image

leslieyip02:
## Context

Fixes #3931.

## Implementation

Disables the TA button when the mod has no lessons.

| Before | After |
| - | - |
| Screenshot 2025-01-03 225029 | Screenshot 2025-01-03 230655 |


website/src/views/timetable/TimetableModulesTable.tsx
```diff
@@ -102,6 +102,7 @@ export const TimetableModulesTableComponent: React.FC<Props> = (props) => {
               type="button"
               className={classnames('btn btn-outline-secondary btn-svg', styles.moduleAction)}
               aria-label={taBtnLabel}
+              disabled={!module.canTa}
```
zwliew, line 105: I think it would be better to remove the button entirely.

zwliew (changes requested):

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -179,6 +179,14 @@ class TimetableContent extends React.Component<Props, State> {
 
   isTaInTimetable = (moduleCode: ModuleCode) => this.props.taInTimetable[moduleCode]?.length > 0;
 
+  canTa = (moduleCode: ModuleCode) => {
+    const { semester, modules } = this.props;
+
+    const module = modules[moduleCode];
+    const moduleTimetable = getModuleTimetable(module, semester);
+    return !isEmpty(moduleTimetable);
+  };
```
zwliew, line 188: Would this work in isolation for CS2103T? That is, it should return false for that module. I'd like to merge this fix before the other one, since there's a [comment](https://github.com/nusmodifications/nusmods/pull/3930/files/e56d83f6622c14dc53580f4f441b80883e5a8643#r1902257170) I'd like to discuss for that.

leslieyip02: Sure, I'll add logic to check if every lesson is a lecture.

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -179,6 +179,16 @@ class TimetableContent extends React.Component<Props, State> {
 
   isTaInTimetable = (moduleCode: ModuleCode) => this.props.taInTimetable[moduleCode]?.length > 0;
 
+  canTa = (moduleCode: ModuleCode) => {
+    const { semester, modules } = this.props;
+
+    const module = modules[moduleCode];
+    const moduleTimetable = getModuleTimetable(module, semester);
+    return !(
+      isEmpty(moduleTimetable) || moduleTimetable.every((lesson) => lesson.lessonType === 'Lecture')
```
zwliew, line 188: we can simplify this entire thing to just `moduleTimetable.some((lesson) => lesson.lessonType !== 'Lecture')`.

website/src/types/views.ts
```diff
@@ -55,6 +55,7 @@ export type ModuleWithColor = Module & {
   colorIndex: ColorIndex;
   isHiddenInTimetable: boolean;
   isTaInTimetable: boolean;
+  canTa?: boolean;
```
zwliew, line 58: Make this non-optional if possible.

website/src/views/timetable/TimetableContent.tsx
zwliew: Also, I'd like this function to be exported from `utils/modules.ts`, and we should add some tests (like for CS2103T) for it.

zwliew (approved):

PR #3942, opened 2025-01-04
fix(website/ta): Prevent duplicate TA lesson in timetable

### LINKED ISSUE #3941: Duplicate TA lessons in timetable
### Describe the bug

When modifying the timetable for a TA lesson, lessons that are already in the timetable are still given as options.

### To Reproduce

1. Select a TA mod and add 2 lessons
2. Try to modify the cell again

### Expected behavior

The existing lesson should no longer be an option.

### Screenshots

`TUT[02]` is already in the timetable, but is still given as an option.

Screenshot 2025-01-04 133102

leslieyip02:
## Context

Fixes #3941. This is caused because `render` in `TimetableContent` only filters out the `activeLesson`.

## Implementation

Skips existing lessons.

| Before | After |
| - | - |
| Screenshot 2025-01-04 133102 | image |


zwliew (approved):

zwliew:
Thanks for the quick fixes!

PR #3965, opened 2025-01-08
Fix Lecture Slots Rendering

### LINKED ISSUE #3964: Lecture Slots not showing up properly
### Describe the bug

Originally, clicking the lecture slot of a module will show all available lecture slots of all groups, i.e. L1 and L2. However, it now only shows the lecture slot of the other group and only of the other lecture timing as well.

Note: I have only tested with CS2040S as I am unsure what other mods have multiple lecture groups and timings.

### To Reproduce

1. Add CS2040S to the timetable
2. Click on any of the default lecture slot

### Expected behavior

I expect all lecture groups and lecture timings for all groups to appear in the timetable view.

### Screenshots

Default view:
image

Clicking on Lecture Group 1, lecture slot on Monday:
image

Clicking on Lecture Group 1, lecture slot on Wednesday:
image

### Desktop:

- OS: EndeavorOS
- Browser: Firefox

### Smartphone:

- Device: Samsung Note 21
- Browser: Chrome

### Additional context

As far as I remember, this was working before the TA feature and I have enabled the TA feature on both devices before.

kaze-droid:
## Context
Resolves #3964 of lecture slots not being shown properly

## Implementation
PR #3942 used `areLessonsSameClass` to prevent the duplicate TA lessons but this also caused different lecture timings for the same lecture group to be filtered out as duplicates since it identifies lessons from the same ClassNo but with different timings as same class.

To fix this, a new function `areLessonsDuplicate` is introduced with the additional check of day and startTime for filtering out the duplicate TA lessons while making sure lecture  slots are rendering correctly

Before:
Screenshot (1118)

After:
Screenshot (1117)

### DIFF
--- a/website/src/utils/modules.test.ts
+++ b/website/src/utils/modules.test.ts
@@ -17,6 +17,7 @@ import {
   renderExamDuration,
   getExamDuration,
   canTa,
+  areLessonsDuplicate,
 } from 'utils/modules';
 import { noBreak } from 'utils/react';
 
@@ -68,8 +69,9 @@ test('areLessonsSameClass should identify identity lessons as same class', () =>
   expect(areLessonsSameClass(mockLesson, deepClonedLesson)).toBe(true);
 });
 
-test('areLessonsSameClass should identify lessons from the same ClassNo but ' +
-  'with different timings as same class',
+test(
+  'areLessonsSameClass should identify lessons from the same ClassNo but ' +
+    'with different timings as same class',
   () => {
     const otherLesson: Lesson = lessonWithDifferentProperty(mockLesson, 'startTime', '0000');
     const otherLesson2: Lesson = lessonWithDifferentProperty(otherLesson, 'endTime', '2300');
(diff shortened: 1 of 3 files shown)


website/src/utils/modules.test.ts
```diff
@@ -93,6 +94,16 @@ test('areLessonsSameClass should identify lessons with different lessonType as d
   expect(areLessonsSameClass(mockLesson, otherLesson)).toBe(false);
 });
 
+test(
+  'areLessonsDuplicate should identify lessons from the same ClassNo but ' +
+    'with different timings as non duplicates',
+  () => {
+    const otherLesson: Lesson = lessonWithDifferentProperty(mockLesson, 'startTime', '0000');
+    const otherLesson2: Lesson = lessonWithDifferentProperty(otherLesson, 'endTime', '2300');
+    expect(areLessonsDuplicate(mockLesson, otherLesson2)).toBe(false);
```
zwliew, line 103: It would be more precise if this test checked that `mockLesson` is not a duplicate of `otherLesson`.

Also, could you add the following tests?
- `otherLesson` is duplicate of `otherLesson2`
- Different day implies not duplicates

zwliew (changes requested):

website/src/utils/modules.test.ts
kaze-droid: Sure, I'll make sure to update the tests

zwliew (approved):
Thank you!

PR #3970, opened 2025-01-12
Fix hover icon visible when no onClick action

### LINKED ISSUE #3969: TA lesson with no clickable action still shows a hover icon
### Describe the bug

For mods with only 1 lesson slot, they cannot be toggled off in TA mode. Hence, the minus icon should not appear when hovering over the cell.

### To Reproduce

1. Select a mod with only 1 lesson slot (e.g. UTS2709)
2. Hover over the cell

### Expected behavior

The cell should not have a hover icon.

### Screenshots

Screenshot 2025-01-12 172154

leslieyip02:
## Context

Fixes #3969,


kokrui (approved):
Thanks for the catch! LGTM

Issue #3785, opened 2024-08-20
Update faculty / department emails for Course Details error reporting


kokrui:
We have a "Report Errors" button on each Course Detail page. (See https://github.com/nusmodifications/nusmods/issues/3784)

However, some faculty emails listed there should be deprecated (e.g., GEQ1000, a course that no longer exists), and there are new emails which should be added.

We should incrementally improve this! If you notice data that should be added / removed / updated, please make a PR or issue to update https://github.com/nusmodifications/nusmods/blob/master/website/src/data/facultyEmail.json

dasha3412:
Hi, I'd like to work on this

itstrueitstrueitsrealitsreal:
Hi, wanted to ask about how we wanna go about doing this, do we wanna remove all the departments that no longer exist and all the emails for courses that no longer exist? Having a little trouble deciding which to remove in my PR.

kokrui:
I think eventually we'd want to remove all unused emails, but I'd be happy to accept PRs that make any sort of incremental progress (e.g. removing one or the other or any subset)!

itstrueitstrueitsrealitsreal:
Sounds good, I'll start working on it!

PR #4005, opened 2025-01-28
Update emails

itstrueitstrueitsrealitsreal:
## Context

Part of #3785
## Implementation

Updated the existing emails using the updated list of contacts for both [undergraduate](https://www.nus.edu.sg/coursereg/docs/UGFac_Contacts.pdf) and [graduate](https://www.nus.edu.sg/coursereg/docs/GDFac_Contacts.pdf) faculties.
## Other Information

Note that this does not remove deprecated courses, nor does it add any of the new ids for any new faculties, an example of which is adding a new id to differentiate DTK1234 from other CDE courses.

### DIFF
--- a/website/src/data/facultyEmail.json
+++ b/website/src/data/facultyEmail.json
@@ -62,7 +62,7 @@
   {
     "id": "chs_gd",
     "label": "Chinese Studies (Graduate)",
-    "email": "fasaslhl@nus.edu.sg",
+    "email": "chsma@nus.edu.sg",
     "match": {
       "type": "department",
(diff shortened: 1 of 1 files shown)


leslieyip02:
I've added some emails, mainly focusing on adding missing departments (e.g. Geography, Electrical Engineering, Pharmacy).

I gave up on updating the graduate emails because they can be quite specific:

Screenshot 2025-06-30 193557

I think it's probably fine though since the `facultyEmail.json` list serves to point users in the write direction for reporting errors, but they can always just look up the specific emails themselves.

PR #4008, opened 2025-02-03
allow choosing of no color

### LINKED ISSUE #3999: Allow choosing of "No Color" for color picker
### Feature Proposition
Allow choosing of no color to remove courses that one may not care about. This complements the incoming custom course feature in #3373 

Image

Can be done using `transparent` parameter in `TimetableCell.tsx`

kaze-droid:
## Context
Resolves #3999 

## Implementation
Provides an additional option of no color for the color picker by introducing an additional `colorIndex` for transparent cells

noColor


website/src/views/components/ColorPicker.tsx
```diff
@@ -51,7 +51,7 @@ const ColorPicker = memo<Props>((props) => {
           className={classnames(styles.palette, { [styles.isClosed]: !isOpen })}
           {...getMenuProps()}
         >
-          {_.range(NUM_DIFFERENT_COLORS).map((index: ColorIndex) => (
+          {_.range(TOTAL_COLORS).map((index: ColorIndex) => (
```
jloh02, line 54: Suggestion: Since the 8 colors currently form a 4x2 grid, the 9th cell makes it look kinda strange. Should we try to make the current color change to transparent color on click (e.g. if the current course color is blue, clicking on the originally blue cell should disable the color?)

website/src/utils/colors.ts
```diff
@@ -5,8 +5,12 @@ import { ColorMapping } from 'types/reducers';
 import { ModuleCode } from 'types/modules';
 
 export const NUM_DIFFERENT_COLORS = 8;
+export const TRANSPARENT_COLOR_INDEX = NUM_DIFFERENT_COLORS;
```
jloh02, line 8: Interestingly, `-1` as a color index also appears to force transparent colors. Would this complicate the code less? If so, is that a better alternative than to create the relevant additional scss code?

jloh02 (changes requested):
Thanks for the updates! Looks great just have a few comments for minor aesthetic and quality improvements

website/src/utils/colors.ts
kaze-droid: That makes sense. I've adjusted the color index to use -1 but kept the additional scss code (renamed to transparentColor) so that the button opening the colorPicker is not completely transparent

PR #4013, opened 2025-02-05
fix(tests): make ModuleArchiveContainer test more reliable

### LINKED ISSUE #3873: flaky `ModuleArchiveContainer` test
This test is flaky

https://github.com/nusmodifications/nusmods/blob/3c2418f9a76611e18b2cf0909be1e13785e1d4d5/website/src/views/modules/ModuleArchiveContainer.test.tsx#L82-L86

I believe it fails because the text is split up into multiple DOM elements / split up (e.g., if it renders like this I believe the test fails)

image

We should fix this!

itstrueitstrueitsrealitsreal:
## Context
Fixes #3873 

## Implementation
Updated  flaky `ModuleArchiveContainer` test by passing in `{ exact: false }` into `findByText`, following the approach that @zwliew suggested in the comments of #3873.

Verified that the tests were working by running the test repeatedly, without any failures.

### DIFF
--- a/website/src/views/modules/ModuleArchiveContainer.test.tsx
+++ b/website/src/views/modules/ModuleArchiveContainer.test.tsx
@@ -82,7 +82,9 @@ describe(ModuleArchiveContainerComponent, () => {
   test('should show 404 page when the course code does not exist', async () => {
     mockAxiosRequest.mockRejectedValue(notFoundError);
     make('/archive/CS1234/2017-2018');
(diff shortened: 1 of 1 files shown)


kokrui (approved):
LGTM, thanks for the change!

for posterity, I thinkkk something like this (https://stackoverflow.com/a/68429756) could be useful since we'll make sure the entire string as-expected is indeed renderred, but we could save that for a larger refactor down the line

Issue #4031, opened 2025-03-10
Updating `samlify` dependency breaks CPEx auth


ravern:
### Describe the bug

This update 278024035973b6ae842a34d8e9634367ba204e11 of our `samlify` dependency to `v2.8.11` broke our NUS auth.

### To Reproduce

1. Update the `samlify` dependency to `v2.8.11`
2. Observe (in Vercel logs, unfortunately maintainer-only) that NUS auth is now broken (error given below)

### Screenshots

![Image](https://github.com/user-attachments/assets/c08e7e97-5533-4f5c-b870-265bf41c901b)

ravern:
Could be related to this issue: https://github.com/tngan/samlify/issues/495

PR #4032, opened 2025-03-10
chore(deps): ignore updates to samlify

ravern:
## Context

An upgrade to `samlify` in 278024035973b6ae842a34d8e9634367ba204e11 broke our NUS auth. We reverted it in 0b8b091922dea151563079daede7fd703997be7b in order to prepare for CPEx this sem.

This config change will ensure the dependency doesn't get automatically updated again, until we properly investigated the cause.

I've filed an issue in https://github.com/nusmodifications/nusmods/issues/4031 to investigate the problem.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -65,7 +65,7 @@
       "groupName": "definitelyTyped"
     }
   ],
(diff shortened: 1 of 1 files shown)


jloh02 (approved):
LGTM. We will be upgrading the NUS Auth soon as well so can consider upgrading samlify then

PR #4039, opened 2025-03-20
website: Implement random course picker button

### LINKED ISSUE #3671: Feature Request: Random Course Picker
This was a feature requested via the NUSMods chat.

We could have a "random course" picker feature where users can see the course details page of a random course. It would help with basic discovering of U/Es, but I think it's mostly a fun feature to add.

Glenn-Chiang:
## Context

Closes #3671 random course picker button 

## Implementation
I added a "Random Course" button to the /courses page. When the button is clicked, the user will be redirected to the page of a random course. 

image

## Other Information
Hi, this is my first time contributing to an open source project. I hope that my changes are acceptable and look forward to making more substantial contributions as I learn more about the codebase.


leslieyip02:
Hi @Glenn-Chiang, thanks for the PR! Your code looks fine, and I think the course search page is a good place to put this feature. However, I think the feature would be more useful if it respected the search filters, so I've pushed some changes to support that as well.

Sorry for the late review!

leslieyip02 (approved):

ZhangYiJiang:
Not a blocker but I would suggest reviewing the UI/UX here a little. 

The button is quite large and prominent for a feature that I'm not sure many students would use. Also the tooltip goes over the search bar

leslieyip02:
@ZhangYiJiang  Maybe this would be better? I made the button more subtle and shifted the tooltip placement to bottom instead.

image

website/src/views/modules/ModuleFinderContainer/ModuleFinderContainer.tsx
```diff
@@ -16,17 +16,20 @@ import { ElasticSearchResult } from 'types/vendor/elastic-search';
...
 import ModuleFinderPager from 'views/modules/ModuleFinderPager';
 import ModuleFinderItem from 'views/modules/ModuleFinderItem';
 import Pagination from 'views/components/searchkit/Pagination';
+import RandomPicker from 'views/components/searchkit/RandomPicker';
 import Title from 'views/components/Title';
 
 import { forceElasticsearchHost } from 'utils/debug';
 import { HIGHLIGHT_OPTIONS } from 'utils/elasticSearch';
 import config from 'config';
+import { Book } from 'react-feather';
```
ZhangYiJiang: nit: Sort this up with the other non-internal imports

website/src/views/modules/ModuleRandomButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { modulePage } from 'views/routes/paths';
+
+import classNames from 'classnames';
+import styles from './ModuleRandomButton.scss';
+
+const ModuleRandomButton: React.FC<RandomPickerProps> = ({ getRandomModuleCode }) => {
+  const history = useHistory();
+
+  const handleClick = () => {
+    getRandomModuleCode().then((moduleCode) => history.push(modulePage(moduleCode)));
```
ZhangYiJiang: Note there's no loading and error handling state here. If this is fast it might be okay to not have a loading state but at least pop a toast or something telling the user the operation failed

ZhangYiJiang (commented):
The new placement is better, but you might still want to iterate on this. 

First it's best to avoid having interactive (the random course button) and non-interactive elements (the courses counter) share the exact same design, since that makes it hard to distinguish what you can interact with

Second I'm not sure adding the icons is a good idea since it doesn't really communicate anything that the text does not already, and kind of just adds noise, the page is already quite complex so you want to minimize these

leslieyip02:
Another option would be the bottom of the filter list? This would allow the button to not be as jarring, but still visuallly look like a button.

image

ZhangYiJiang:
I think that's a better placement yeah. The only change I would ask for is you're currently using a primary, which is typically reserved for the primary interactive element on the screen - think the submit button in a form for example. Here it's btn-secondary or even btn-link, since it's a navigational link of sorts, probably makes more sense

Issue #4050, opened 2025-04-07
Improve description for post-cpex period


jloh02:
After CPEx period, the disabled cpex stage still states "CPEx will open on" instead of something along the lines of CPEx has ended for this semester which is misleading.

![Image](https://github.com/user-attachments/assets/ce2733eb-6f8b-4439-ab81-2f1053b7372d)

PR #4056, opened 2025-04-12
website: Enable CPEx on staging and development branches

### LINKED ISSUE #4051: Enable CPEx on staging and development branches
CPEx should be enabled on branches that are not production to facilitate better testing. Currently this is manually enabled in `website/src/featureFlags.ts` but should be dynamically enabled using `process.env.VERCEL_ENV` as per `website/src/serverless/nus-auth.ts`

g0at1:
## Context

Closes #4051. CPEx should be enabled on branches that are not production to facilitate better testing.

## Implementation

I added new variable isProduction and use it to enable CPEx on non production branches.

### DIFF
--- a/website/src/featureFlags.ts
+++ b/website/src/featureFlags.ts
@@ -1,6 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 
 /** Enable Course Planning Exercise */
-export const enableCPEx = false;
+const isProduction = process.env.VERCEL_ENV === 'production';
+export const enableCPEx = !isProduction;
 
-export const showCPExTab = false;
+export const showCPExTab = !isProduction;


jloh02 (changes requested):
Thanks for this update. However there's 2 problems that don't quite seem right to me
1. Doing this in `featureFlags.ts` doesn't seem right as this no longer becomes a feature flag 
2. This implicitly breaks for production and the change away from a manual feature flag implies that it should be a fixed value

To solve this issue I would propose either of 2 alternatives:
1. Move this logic out featureFlags and convert the feature flag to something like "enableCPExForProd"; or
2. Create time based system for CPEx (which is difficult but preferred)

If we are keen on option 2 in this case (ie anyone deems the effort to reward ratio is worth to implement this), perhaps, opening a larger issue in place of #4050 and #4051 is better.

jloh02:
Since there were no updates. Closed in favor of #4063. Thanks though for the effort to contribute! Hope to see you on other PRs too

Issue #4026, opened 2025-03-06
TimetableAI


thejus03:
# TimetableAI

## Key Features

- Maximising Free Days: The algorithm prioritises creating schedules that align with users' preferences for free days
- Minimising Travel Time: It considers the physical distances between classrooms to reduce travel time between classes
- Lunch Time: Prioritises to give at least 1 hour for lunch between 12-2pm
- Clash-Free scheduling: Ensures all classes are conflict-free.

## High level integration

- We use the NUSMods API to preprocess venues data to clump different rooms in the same building together to get the average x and y coordinate of the building.
- From the user-selected modules, we use the NUSMods API to get all the slots for each module the user selected. We then remove all non-lecture slots that are on the free days the user selected. Then merge slots between the same modules based on those that are at the same timing, day and building location.
- Now that we have all the slots, we construct a graph where we make edges between slots that clash. 
- Then we use backtracking to create all possible timetables that do not clash, using the graph to prune immediately when there is a clash.
- Now that we have all the possible timetables, we score the timetables based on a scoring system that we elaborate on below. The top 5 ranking timetables are returned to the frontend.

## Scoring system
This system is an award-punish scoring system: 
- Awards for having at least an hour of time for lunch
- Punish timetables with lessons outside of the user-selected favourable timings
- Punish long distances travelled between lesson locations.

ravern:
This sounds cool!

Seems like you already have a working implementation of this algorithm at https://github.com/thejus03/hacknroll. Would this request be to integrate your existing work with the NUSMods website?

ravern:
Thank you for providing your reasoning @samthambad @thejus03!

> On the other hand, TimetableAI builds upon the functionality of NUS Optimizer by additionally optimising for an important aspect - travel distance.

This makes sense, sorry if I missed this point earlier. I think the ability to optimise by travel distance (in addition to everything else) is indeed quite a cool feature.

---

I see a few approaches suggested so far (correct me if I missed something out):

1. Using AWS Lambda or Google Cloud Run, for full set of features, running in the cloud
2. Integrate with existing Vercel serverless, without travel distance feature, running in the cloud
3. Compile Go code to WASM for reduced performance, for full set of features, running client-side
4. Rewrite into Rust and compile to WASM for better performance (it remains to be seen though), for full set of features, running client-side
5. Create a downloadable app

---

[...]

PR #4064, opened 2025-06-07
feat(timetable): add optimiser beta

thejus03:
## Context

#4026 Timetable Optimisation Feature

## Implementation

- Able to achieve significant improvement from initial optimisation of TimetableAI
- Works with Vercel Serverless Functions
- Takes about 5 to 40 seconds depending on complexity of task per request. In most cases taking 8-15 seconds.
- Returns an optimised timetable following user constraints
image


website/api/optimiser/README.md
```diff
@@ -0,0 +1,155 @@
...
+   cd nusmods/website/api/optimiser
+   ```
+
+2. **Install dependencies**
+   ```bash
+   go mod tidy
+   ```
+
+3. **Test the API**
+- Send a POST request following the request body format above to `http://localhost:3000/api/optimiser/optimise`
```
ravern: Can we update this section to be accurate? It doesn't include how to start the serverless function locally.

website/api/optimiser/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Limitations
+
+- **Venue Data Dependency**: Optimisation quality depends on accurate venue coordinate data from NUSMods
+- **Academic Year Coverage**: Limited to semesters with available NUSMods API data
+- **Lesson Type Support**: Optimises for standard NUS lesson types (may not handle special/custom lesson formats)
+
+## Potential Improvements
+
+- Once there is more concrete information on the building location for each venue, we can remove the current method of identifying building by taking the first few letters before the '-' in the venue name. This will improve the accuracy and reduce search space.
```
ravern, line 163: Let's bring this out into an issue once this PR is merged.

website/api/optimiser/README.md
```diff
@@ -0,0 +1,155 @@
...
+- Tweak the scoring function to prioritise more important constraints
```
ravern: What are the more important constraints?

website/api/optimiser/_models/models.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+)
+
+type OptimiserRequest struct {
+	Modules      []string `json:"modules"`      // Format: ["CS1010S", "CS2030S"]
+	Recordings   []string `json:"recordings"`   // Format: ["CS1010S Lecture", "CS2030S Laboratory"]
+	FreeDays     []string `json:"freeDays"`     // Format: ["Monday", "Tuesday"]
+	EarliestTime string   `json:"earliestTime"` // Format: "1504" (HHMM)
+	LatestTime   string   `json:"latestTime"`   // Format: "1504" (HHMM)
+	AcadYear     string   `json:"acadYear"`     // Format: "2024-2025" (YYYY/YYYY)
```
ravern: Nitpick: The format comment gives "-" for the example but a "/" for the format.

website/api/optimiser/_models/models.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Parsed fields
+	StartMin  int    // Minutes from 00:00 (e.g., 540 for 09:00)
+	EndMin    int    // Minutes from 00:00
+	DayIndex  int    // 0=Monday, 1=Tuesday, 2=Wednesday, 3=Thursday, 4=Friday
+	LessonKey string // "MODULE|LessonType"
+}
+
+// ParseModuleSlotFields parses and populates the computed fields in ModuleSlot for faster computation
+func (slot *ModuleSlot) ParseModuleSlotFields(lessonKey string) error {
```
ravern: Can we be consistent here, there's both "parsed" and "computed" fields in use now. Maybe "pre-computed" or "pre-parsed"? 

I think the method name can also be `PreParse` or `Precompute` instead of just `Parse`.

website/api/optimiser/_modules/modules.go
```diff
@@ -0,0 +1,205 @@
...
+		var moduleData struct {
+			SemesterData []struct {
+				Semester  int                 `json:"semester"`
+				Timetable []models.ModuleSlot `json:"timetable"`
+			} `json:"semesterData"`
+		}
+		err = json.Unmarshal(body, &moduleData)
+		if err != nil {
+			return nil, err
+		}
```
ravern: As previously mentioned,

* Can we pull HTTP requests into a single `client` file or package?
* Can we pull all URLs in use into a central `constants` file or package?

website/api/optimiser/_modules/modules.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	defer res.Body.Close()
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(body, &venues)
+	if err != nil {
+		return nil, err
+	}
```
ravern: * Can we pull HTTP requests into a single `client` file or package?
* Can we pull all URLs in use into a central `constants` file or package?

website/api/optimiser/optimise.go
```diff
@@ -0,0 +1,37 @@
...
+	"encoding/json"
+	"net/http"
+
+	models "github.com/nusmodifications/nusmods/website/api/optimiser/optimise/_models"
+	solver "github.com/nusmodifications/nusmods/website/api/optimiser/optimise/_solver"
+)
+
+func Handler(w http.ResponseWriter, r *http.Request) {
+	// CORS
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
ravern, line 14: Can we restrict this to only nusmods.com and vercel.app? Just to reduce the surface.

website/src/views/layout/Navtabs.tsx
```diff
@@ -44,6 +46,12 @@ const Navtabs: FC = () => {
         <Calendar />
         <span className={styles.title}>Timetable</span>
       </NavLink>
+      {beta && (
+        <NavLink {...tabProps} to="/optimiser">
+          <Cpu />
```
ravern, line 51: Haha nice

website/src/views/optimiser/OptimiserContent.tsx
```diff
@@ -0,0 +1,580 @@
...
+              <div className={styles.lunchTimeLabel}>:00</div>
+            </div>
+          </div>
+        </div>
+        <div className={styles.optimizeButtonSection}>
+          <button
+            type="button"
+            className={classnames(
+              'btn',
+              styles.optimizeButton,
```
ravern: Was there a reason not to use similar styles as the buttons for days of the week, e.g., Monday, Tuesday, Wednesday?

Is the intention here to make the "Optimise" button look different from the normal button styles of NUSMods (which is perfectly fine and can be discussed, just checking if it was intentional)?

website/src/views/optimiser/index.tsx
```diff
@@ -0,0 +1,20 @@
+import Loadable, { LoadingComponentProps } from 'react-loadable';
+
+import LoadingSpinner from 'views/components/LoadingSpinner';
+import ApiError from 'views/errors/ApiError';
+
+const AsyncPlannerContainer = Loadable({
```
ravern: Can we rename this to `Optimiser`?

website/src/views/optimiser/OptimiserContent.tsx
```diff
@@ -0,0 +1,580 @@
...
+                  <option value="13">13</option>
+                  <option value="14">14</option>
+                  <option value="15">15</option>
+                  <option value="16">16</option>
+                  <option value="17">17</option>
+                  <option value="18">18</option>
+                </select>
+                <div className={styles.timeLabel}>:00</div>
+              </div>
+            </div>
```
ravern: I don't really agree with how both the time inputs are being done here. Only selecting the hour of an HHMM value feels clunky.

Would it be better (and feel free to disagree) if we had the entire value in the selection box instead? i.e., `<option value="1900">19:00</option>`?

Also, is it technically possible to support 30-min intervals? So like 1600, 1630, etc.

website/src/views/optimiser/OptimiserContent.tsx
```diff
@@ -0,0 +1,580 @@
...
+        acadSem: activeSemester,
+        freeDays: Array.from(selectedFreeDays),
+        earliestTime: formatTime(earliestTime),
+        latestTime: formatTime(latestTime),
+        recordings,
+        lunchStart: formatTime(earliestLunchTime),
+        lunchEnd: formatTime(latestLunchTime),
+      }),
+    });
+    const data = await response.json();
```
ravern: Can we move the API fetching code into `website/src/apis/optimiser`? You can look at https://github.com/nusmodifications/nusmods/blob/master/website/src/apis/mpe.ts for an existing example.

website/api/optimiser/_modules/modules.go
```diff
@@ -0,0 +1,205 @@
...
+var E_Venues = map[string]bool{
+	"E-Learn_A":  true,
+	"E-Learn_B":  true,
+	"E-Learn_C":  true,
+	"E-Learn_D":  true,
+	"E-Hybrid_A": true,
+	"E-Hybrid_B": true,
+	"E-Hybrid_C": true,
+	"E-Hybrid_D": true,
+}
```
ravern: Is there a way not to hardcode this?

website/api/optimiser/_solver/nusmods_link.go
```diff
@@ -0,0 +1,89 @@
...
+	"PACKAGED LECTURE":              "PLEC",
+	"PACKAGED TUTORIAL":             "PTUT",
+	"RECITATION":                    "REC",
+	"SECTIONAL TEACHING":            "SEC",
+	"SEMINAR-STYLE MODULE TEACHING": "SEM",
+	"TUTORIAL":                      "TUT",
+	"TUTORIAL TYPE 2":               "TUT2",
+	"TUTORIAL TYPE 3":               "TUT3",
+	"WORKSHOP":                      "WS",
+}
```
ravern: Mentioned earlier as well, is there a way not to hardcode this? These things do change occassionally.

website/api/optimiser/README.md
thejus03: Sure

website/api/optimiser/README.md
thejus03: Ah okay... here what I meant was that in the future if we find out more constraints to let users further configure the optimiser requirements or improve the optimised timetables then we should update them. 

For instance, previously the scoring function was just for distance. But I later found out that there were large gaps (of 4-5 hours) in some timetables generated between lessons when there were other better alternatives. So now i also consider punishing gaps that are larger than 2 hours between lessons linearly resulting in more reasonable timetables.

website/api/optimiser/_models/models.go
thejus03: Fixed. Thanks!

website/src/views/layout/Navtabs.tsx
thejus03: 😂

website/src/views/optimiser/OptimiserContent.tsx
thejus03: Yea I wanted the optimise button to standout compared to normal button styles. I just felt that since it is the final button we press to "submit" it should be given some importance.

website/src/views/optimiser/OptimiserContent.tsx
thejus03: Hmmm... The reason I choose to put the : 00 outside is because I wanted to minimise using the input buttons simply because the select buttons do not look nice imo. Furthermore, it was to give users an indication that only hours can be changed since all slots are 1 hour slots and there are no 30 min slots iirc. So if the user selected 9:30 all the lessons would start earliest by 10:00 anyways. So I don't think there is any reason to let users choose minutes too?

website/src/views/optimiser/index.tsx
thejus03: Fixed. Thanks!

website/api/optimiser/_models/models.go
thejus03: image
Ah okay. I changed "computed fields" to "parsed fields". I am not sure why we should change to PreParse since this is the parsing function.

website/api/optimiser/_modules/modules.go
thejus03: Sure. That is much cleaner!

website/api/optimiser/_modules/modules.go
thejus03: Yep!

website/api/optimiser/_modules/modules.go
thejus03: Perhaps if we check if there is a prefix that starts with "E-Learn" or "E-Hybrid" then we take them to be E-Learning. But then again we are still hardcoding (but to a lesser extent?)
Not sure what to do here :(

website/api/optimiser/optimise.go
thejus03: image
image
Does this make sense? So any url that contains a substring matching allowedOrigin should be allowed?

website/api/optimiser/_solver/nusmods_link.go
thejus03: The map here is necessary to generate the url link based on the lesson type. Perhaps if there was some way to get the latest information?

website/src/views/optimiser/OptimiserContent.tsx
thejus03: Sure

website/api/optimiser/README.md
ravern: Cool! Feel free to add this (or a short reminder) to yourself in the README.

website/api/optimiser/optimise.go
ravern: Ah no, I meant more like using the `Access-Control-Allow-Origin` to do so, e.g., `https://nusmods.com` instead of `*`.

Can take a look at https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#examples.

website/src/views/optimiser/OptimiserContent.tsx
ravern: > do not look nice imo.

Valid! Definitely a subjective opinion.

> Furthermore, it was to give users an indication that only hours can be changed since all slots are 1 hour slots and there are no 30 min slots iirc.

This make sense. I'm happy for this to remain in hourly intervals.

website/api/optimiser/_constants/constants.go
```diff
@@ -0,0 +1,46 @@
...
+	"E-Hybrid_B": true,
+	"E-Hybrid_C": true,
+	"E-Hybrid_D": true,
+}
+
+var LessonTypeAbbrev = map[string]string{
+	"DESIGN LECTURE":                "DLEC",
+	"LABORATORY":                    "LAB",
+	"LECTURE":                       "LEC",
+	"PACKAGED LECTURE":              "PLEC",
```
ravern: Can you add a comment here to sync this with the constants here? And then add a comment in that file as well to sync that map with this map?
 https://github.com/nusmodifications/nusmods/blob/3c2c1e100b4a0017bc53b3f560cb207a6ebca140/website/src/utils/timetables.ts#L59-L61

```suggestion
# Ensure this is in sync with website/src/utils/timetables.ts
var LessonTypeAbbrev = map[string]string{
	"DESIGN LECTURE":                "DLEC",
	"LABORATORY":                    "LAB",
	"LECTURE":                       "LEC",
	"PACKAGED LECTURE":              "PLEC",
```

website/api/optimiser/_modules/modules.go
ravern: Agreed. I think I'm fine with keeping this as-is then.

ravern (changes requested):
This looks great! Most of the review comments have been addressed, including isolating constants, breaking down into components, creating an API client.

I like that the timetable doesn't just open automatically now, making it clearer to the user if it is a partial or complete timetable.

Just some final small changes to go:

- CORS header https://github.com/nusmodifications/nusmods/pull/4064#discussion_r2160210970
- "important constraints" (completely optional, more for yourself) https://github.com/nusmodifications/nusmods/pull/4064#discussion_r2160210753
- Other comments from this review
- Local development instructions (I can figure this one out, and send you a PR)

website/api/optimiser/_constants/constants.go
thejus03: yep sure!

ravern (approved):
LGTM 🔥 

You've been awesome to work with. Thank you for your patience in waiting for my (<1 per week) reviews.

I'll approve this now, and will do a final test on the preview deployment before merging.

ravern:
Preview deployment works!

PR #4077, opened 2025-06-27
Add planner import and export

### LINKED ISSUE #3843: [Course Planner]: Download/upload/share planner configuration

Similar problem to sharing timetable. I wish to replicate the course planner configuration by downloading the configuration and uploading the configuration json file.

1. A button that downloads the planner details on https://nusmods.com/planner
  a. Create a temporary blob from the redux store after stringification
  b. Create an object url for that blob
  c. Create a temporary anchor element with the create url
  d. Programmatically click on the anchor element
  e. Revoke the object url
2. A button that imports the planner details on https://nusmods.com/planner
  a. Drag or drop / file select for the planner config
  b. Maybe some sort of zod schema parsing
  c. Override the redux store with new state

URL sharing will be more convenient where URL contains the state of the course planner.

### Additional context

I tested my personal planner config and it is 5000+ characters after base64 encoding. It seems that URL should be kept below [2000 characters](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers). Nusmods cannot encode the config into the url without some sort of minification. Also, sharing a very long URL is pretty bad idea coz it might encounter character limits in texting applications

leslieyip02:
## Context

Closes #3843.

Since the new AY's modules are going to be added soon, the ability to import/export planners would be quite helpful for planning.

https://github.com/user-attachments/assets/af9eb7e3-4864-4c2a-b8e0-51ad9d5cabb7

The video looks a bit weird because my screen recorder hides my download window.

## Implementation
- Users can download a `.json` file of their current conifg
- Users can upload their `.json` config to write over the current config
- Uses `zod` for schema validation (since it's used in the export service as well)


ravern (changes requested):
Cool feature!

I'm mostly okay with the code, but I have some comments about the overall UX of the feature.

Feel free to disagree with any of these, and if I agree on all the disagreements I'm happy to approve this.

1. I think there should be a warning dialog for the "Clear" button, instead of a tooltip. This would make it similar to how the "Reset" button works on the timetable. Also, perhaps using the same icon and text ("Reset") as the timetable would make more sense.

2. We can afford to be more descriptive on errors. I don't think we should display the gory syntax error, but perhaps something like "Something went wrong when trying to import the file. Try to re-download the planner file from the source!" could help non-tech-savvy users.

Finally, I'm not sure if it's a typo but the "closes #3483" seems to be pointing to an unrelated PR? Wrong number?

leslieyip02:
Oops, it should be closing #3843.

ravern (approved):
LGTM! speedy af

Issue #4082, opened 2025-07-01
Ease multi-module selection in timetable


HilltopMeditator:
### Is your feature request related to a problem? Please describe.
When selecting modules within the timetable planner, I find it slightly troublesome to have to select multiple modules that start with the same prefix (eg: EE20xx, CS2xxx). I have to retype the module prefix in the search bar multiple times for each module I am taking.

### Describe the solution you'd like
Make shift-clicking add the module without clearing search bar details.

### Describe alternatives you've considered
1. Have buttons upon hovering that indicate what does the user want to do with the module? (scope potentially expandable into hiding module data for modules that the user is not interested in)
2. Delay adding execution, collecting modules that the user selects in a "shopping cart" until confirming the modules with a separate button. This can appear after the suggested modules that appear when user searches. The scroll bar would only apply to the suggestion list.

### Additional context
Alternate solution 2. example:
![Image](https://github.com/user-attachments/assets/61a44740-b65b-4c75-888b-b33f03a16f66)

PR #4084, opened 2025-07-03
allow multi-module selection in timetable

kaze-droid:
## Context
Resolves issue #4082 by adding shift click to module selection

## Implementation
Added isShiftClickRef to track when Shift key is held during mouse clicks
Modified stateReducer to handle clickItem events differently based on whether shift key was pressed

### DIFF
--- a/website/src/views/timetable/ModulesSelect.tsx
+++ b/website/src/views/timetable/ModulesSelect.tsx
@@ -1,4 +1,4 @@
-import { FC, useCallback, useState } from 'react';
+import { FC, useCallback, useState, useRef } from 'react';
 import { omit } from 'lodash';
 import Downshift, {
   ChildrenFunction,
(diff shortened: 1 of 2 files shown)


website/src/views/timetable/ModulesSelect.tsx
```diff
@@ -76,8 +78,19 @@ const ModulesSelect: FC<Props> = ({
           return changes;
 
         case Downshift.stateChangeTypes.keyDownEnter:
+          closeSelectAndEmptyInput();
+          return changes;
+
```
leslieyip02: I think `closeSelectAndEmptyInput` might not be needed here. When the user presses enter, they should still be able to continue typing.

Online version:

https://github.com/user-attachments/assets/52b233be-7a5f-41a5-bb62-769ad2df4626

Deployment preview:

https://github.com/user-attachments/assets/fa297488-22af-4f1e-9779-1ffc434319a1

kaze-droid: ive just updated the behaviour for enter (as well as regular click) by only clearing input instead of using closeSelectAndEmptyInput();

leslieyip02 (approved):
LGTM. Thanks!

PR #4087, opened 2025-07-04
Adds max consecutive hours feature in Optimiser

### LINKED ISSUE #4086: Add max consecutive hours of study in Optimiser

The timetable returned by optimiser may have lessons that are stacked back to back when there are other alternative timetables that have gaps which can be more reasonable for some.

Users get to choose max consecutive hours of study and optimiser returns a timetable that meets the constraint.

thejus03:
## Context

Closes #4086

## Implementation

Added in max consecutive hours of study as a heavily prioritised constraint like lunch timings
image
The optimiser now also punishes if there are back to back lessons that exceed user specified max hours. Therefore the timetable returned will take this into account and try to minimise this as much as possible.


ravern (changes requested):
Can we run `go fmt` on the Go code here?

PR #4094, opened 2025-07-08
Add Optimiser Unit Tests

### LINKED ISSUE #4092: Optimiser Unit Tests

I suggested a change to use `uniqueKey` instead of `displayText` in #4083, but it led to a regression bug in #4090. As more changes are made, we should make sure that they don't break existing functionality. We can introduce unit tests to improve the optimiser's stability.

We can start by refactoring the optimiser code to make a bit more testable. For example, the timetable functions for parsing, filtering and grouping are in `utils/timetables.ts`.

- [ ] Extract methods from `OptimiserContent` into `utils/optimiser`
- [ ] Implement unit tests

### LINKED ISSUE #4095: Optimiser free day conflict is not working properly
### Describe the bug
Optimiser thinks I am attending lessons live even if nothing is selected, raising conflict

### To Reproduce

Image

### Expected behaviour

No free day conflict should be thrown

### LINKED ISSUE #4102: Optimiser Saturday option not shown
### Describe the bug

Optimiser `hasSaturday` was broken by #4083. It should show Saturday as an option when there are Saturday classes.

Image

leslieyip02:
## Context

Closes #4092.

## Legacy Issues

Closes #4095.
Closes #4102.

## Implementation

- [x] Move types into `types` folder
- [x] Move util functions into `utils` folder + unit tests
- [x] Refactor `OptimiserForm` code
- [x] UI unit tests

### Replaced `uniqueKey` with `lessonKey`

Previously, we had:

- `uniqueKey` => `{moduleCode}-{lessonType}`
- `displayText` => `{moduleCode} {lessonType}`
- `lessonKey` => `{moduleCode}|{lessonType}`

I replaced `uniqueKey` because I think the deduplication makes the code a bit easier to follow.

### `useOptimiserForm` custom hook

We had a lot of `useState` calls in `OptimiserContent`, and they were being passed to `OptimiserForm` as well. I think having a custom hook makes it easier to pass around, and allowed to pass props down to individual form fields, e.g.:

```
const OptimiserLessonOptionSelect: React.FC<Props> = ({ lessonOptions, optimiserFormFields }) => {
  const { liveLessonOptions, setLiveLessonOptions } = optimiserFormFields;

  const toggleLiveLessonOption = useCallback(...);

  ...
}
```


jloh02 (approved):
LGTM good refactor and tests

jloh02 (changes requested):
Changes look good but i'm not sure if it's a refactor or latest merge changes. The tooltip (i) is broken and there are font changes. Not sure if these are intended. Just merged master in to this and hopefully some of these are fixed. If not, let's try to keep the fonts more uniform

image

jloh02:
Tooltips still broken :(

thejus03:
image

1. Not sure if this was intended but first select input looks too large. 
2. The second statement is wrong/copy of the first statement.

thejus03:
also very minor UI nitpick 😅:
image
the padding for the border bottom seems to be too thin. Something similar to the border bottom in timetable page would look nice!

leslieyip02:
> Tooltips still broken :(

The weird thing is that if you navigate to another page and then come back to the optimiser, the tooltips suddenly start working.

## Update: 

Fixed by importing `Tooltip` directly instead of using this async version:

image

leslieyip02:
>  1. Not sure if this was intended but first select input looks too large.

@thejus03 I standardized the dropdown sizes so that the dropdowns look more uniform. However, if we add more dropdowns in the future, we can always customize their widths.

thejus03:
image

@leslieyip02 Just a heads up the info tooltip is not wrapped around in mobile viewport size. Not sure if that was intended because it looks okay to me.
I think we can either remove the info tooltip for mobiles or we need to fix this problem since it would persist for any sentence that wraps.

Besides that the UI looks good!

website/src/views/optimiser/OptimiserForm/OptimiserFreeDaySelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const days = hasSaturday ? [...WorkingDays] : dropRight([...WorkingDays], 1);
+
+  const toggleDay = useCallback(
+    (day: DayText) => {
+      const isSelected = freeDays.has(day);
+      setFreeDays(
+        (prev) =>
+          new Set(isSelected ? [...prev].filter((existing) => existing !== day) : [...prev, day]),
+      );
+    },
```
ZhangYiJiang: Using the prev state as a dep into a state update fn is generally an anti-pattern since you can use the prev state directly 
```suggestion
      setFreeDays(
        (prev) =>
          new Set(freeDays.has(day) ? [...prev].filter((existing) => existing !== day) : [...prev, day]),
      );
    },
```

website/src/views/optimiser/OptimiserForm/OptimiserTimeRangeSelect.test.tsx
```diff
@@ -0,0 +1,79 @@
+import { mount } from 'enzyme';
```
ZhangYiJiang: enzyme is pretty outdated. If you can I would try using react testing library and hook testing library which is what most people have converged on as the successor  https://testing-library.com/docs/react-testing-library/intro/. For the test cases here you can convert them pretty easily I think

website/src/utils/optimiser.ts
```diff
@@ -0,0 +1,181 @@
...
+  return `${moduleCode} ${lessonType}`;
+}
+
+export function getOptimiserAcadYear(acadYear: AcadYear): string {
+  const [from, to] = acadYear.split('/');
+  return `${from}-${to}`;
+}
+
+export function getOptimiserTime(time: LessonTime): string {
+  const hh = padStart(`${getLessonTimeHours(time)}`, 2, '0');
```
ZhangYiJiang: nit: `String(getLessonTimeHours(time))` communicates intent to cast better than template strings, even if it is a bit longer

website/src/views/hooks/useOptimiserForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+  setLessonTimeRange: Dispatch<SetStateAction<TimeRange>>;
+  lunchTimeRange: TimeRange;
+  setLunchTimeRange: Dispatch<SetStateAction<TimeRange>>;
+  maxConsecutiveHours: number;
+  setMaxConsecutiveHours: Dispatch<SetStateAction<number>>;
+};
+
+export default function useOptimiserForm(): OptimiserFormFields {
+  const [liveLessonOptions, setLiveLessonOptions] = useState<LessonOption[]>([]);
+  const [freeDays, setFreeDays] = useState<Set<DayText>>(new Set());
```
ZhangYiJiang:
```suggestion
  const [freeDays, setFreeDays] = useState(new Set<DayText>());
```

website/src/views/optimiser/OptimiserButton.tsx
```diff
@@ -16,41 +16,41 @@ const OptimiserButton: React.FC<OptimiserButtonProps> = ({
...
-        </span>
-      )}
-      {isOptimising ? 'Searching and optimising...' : 'Optimise Timetable'}
-    </button>
-    <div className={styles.estimateTime}>
-      <div>estimated time:</div>
-      <div className={styles.estimateTimeValue}>5s - 40s</div>
+}) => {
+  const isDisabled = isOptimising || isEmpty(lessonOptions) || !isEmpty(freeDayConflicts);
+  const zapFill = isEmpty(lessonOptions) || !isEmpty(freeDayConflicts) ? '#69707a' : '#ff5138';
```
ZhangYiJiang: `fill` is also available as a CSS property

website/src/views/optimiser/OptimiserForm/OptimiserMaxConsecutiveHoursSelect.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <OptimiserFormTooltip content="Prioritises having less than this number of consecutive hours of live lessons" />
+      </h4>
+
+      <select
+        value={maxConsecutiveHours}
+        onChange={(e) => setMaxConsecutiveHours(parseInt(e.target.value, 10))}
+        className={styles.optimiserDropdown}
+      >
+        {values.map((value) => (
+          <option value={value}>{value}</option>
```
ZhangYiJiang: There's a missing `key` prop here

website/src/views/optimiser/OptimiserForm/OptimiserFreeDaySelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <section className={styles.freeDaysSection}>
+      <h4 className={styles.optimiserDescription}>
+        Select days you would like to be free
+        <OptimiserFormTooltip content="Chosen days will have no physical classes" />
+      </h4>
+
+      <div className={styles.freeDaysButtons}>
+        {days.map((day) => (
+          <button
```
ZhangYiJiang: In general using a button for a stateful toggle is not ideal, because the design-wise the active state looks exactly the same as hover, so clicking on it with a mouse doesn't seem to do anything until I move my mouse away. Accessibility-wise there's also no indication the button is now active, although there are workarounds. 

Ideally you would use the appropriate element (an input type="checkbox") but if you really want the button style interaction maybe consider alternative styling for selected buttons or adding a check icon (although that will shift layout which causes its own problems). 

For accessibility you can try using role="checkbox" and aria-checked like in this example https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/

website/src/views/hooks/useOptimiserForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+const defaultLunchTimeRange: TimeRange = {
+  earliest: '1200',
+  latest: '1400',
+};
+
+const defaultMaxConsecutiveHours = 4;
+
+export type OptimiserFormFields = {
```
ZhangYiJiang, line 17: Unfortunately while the idea is good, using an object like this and passing it around as a single prop is a React anti-pattern. React allows using memo to optimize rendering by only re-rendering the child if any of its prop changes. Before, a form element would just take its own value, and so would only re-render when it updates, but now since the object is updated when any of its properties updates, _every_ form element updates when _any_ of them changes. 

In practice for a component of this size the difference is negligible, and memo isn't used either, but its not a good habit to get into. 

If you need to manage large complex forms, maybe consider the useReducer hook or a library instead like https://react-hook-form.com/

website/src/views/optimiser/OptimiserForm/OptimiserForm.scss
```diff
@@ -0,0 +1,62 @@
...
+  margin: 0;
+  margin-bottom: 0.5rem;
+  font-size: medium;
+
+  @include media-breakpoint-down(xs) {
+    font-size: small;
+  }
+}
+
+.optimiserDropdown {
```
ZhangYiJiang, line 41: Can't we use Bootstrap's default select styling here? https://getbootstrap.com/docs/4.6/components/forms/#select-menu

website/src/views/optimiser/OptimiserForm/OptimiserFormTooltip.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Tooltip from 'views/components/Tooltip/Tooltip';
+
+import styles from './OptimiserFormTooltip.scss';
+
+type Props = {
+  content: string;
+};
+
+const OptimiserFormTooltip: React.FC<Props> = ({ content }) => (
+  <Tooltip content={content} placement="right">
```
ZhangYiJiang, line 11: Putting a tooltip around a non-interactive element is not ideal for accessibility since the non-interactive element cannot gain focus so it can never been seen by a keyboard user. Try adding `tabIndex={0}` to the icon

website/src/views/optimiser/OptimiserForm/OptimiserLessonOptionSelect.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <section>
+      <h4 className={styles.optimiserDescription}>
+        Select lessons you plan to attend live (in person/online)
+        <OptimiserFormTooltip content="Chosen lessons will only be allocated on your school days" />
+      </h4>
+
+      {isEmpty(lessonOptions) ? (
+        <div className={styles.noLessonsWarning}>
```
ZhangYiJiang: The warning divs should use Bootstrap `alert` styles https://getbootstrap.com/docs/4.6/components/alerts/. The current styling does not correctly work in light mode

ZhangYiJiang (commented):
Not a complete review, just some notes 

Also I've noticed while the code splitting logic exists in optimizer/index.tsx, it's not actually used since the router directly connects to OptimizerContent, bypassing the split chunk

website/src/views/hooks/useOptimiserForm.tsx
leslieyip02: I think I'll add a // TODO for this? I think for now it's ok since the optimiser is still in beta.

website/src/views/optimiser/OptimiserForm/OptimiserFreeDaySelect.tsx
leslieyip02: Yea I agree that the indication could be a bit better, maybe something like this?

image

website/src/views/hooks/useOptimiserForm.tsx
ZhangYiJiang: Yeah sure, just remember to get to it eventually

website/src/views/optimiser/OptimiserForm/OptimiserFreeDaySelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <div className={styles.freeDaysButtons}>
+        {days.map((day) => {
+          const checked = freeDays.has(day);
+
+          return (
+            <button
+              type="button"
+              role="checkbox"
+              aria-checked={checked}
+              tabIndex={0}
```
ZhangYiJiang: I don't think this is necessary, buttons already have this by default

website/src/views/optimiser/OptimiserForm/OptimiserFreeDaySelect.tsx
ZhangYiJiang: That's okay but you might consider using a proper checkbox instead https://getbootstrap.com/docs/4.6/components/forms/#custom-forms

website/src/views/optimiser/OptimiserResults.tsx
```diff
@@ -1,122 +1,122 @@
...
+                >
+                  {lesson.displayText}
+                </div>
+              ))}
+            </div>
+
+            <div className={styles.unassignedExplanation}>
+              <div className={styles.unassignedExplanationHeader}>Why did this happen?</div>
+              <div className={styles.unassignedExplanationItem}>
+                • <strong>Venue constraints:</strong> NUSMods may not have complete or accurate
```
ZhangYiJiang: Should these be some heading and a `ul` + `li`?

website/src/views/optimiser/OptimiserForm/OptimiserTimeRangeSelect.tsx
```diff
@@ -0,0 +1,158 @@
...
+  setTime: (lessonTime: LessonTime) => void;
+};
+
+const OptimiserTimeRangeSelect: React.FC<TimeRangeSelectProps> = ({
+  currentValue,
+  timeValues,
+  setTime,
+}) => (
+  <select
+    data-testid="optimiserTimeRangeSelect"
```
ZhangYiJiang: Form elements generally don't need this because they should be labeled https://testing-library.com/docs/queries/about#priority. 

Here you should be associating the select with a `<label>` instead https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#defining_an_explicit_label_with_the_for_attribute

website/src/views/optimiser/OptimiserContainer/OptimiserContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setLiveLessonOptions: setPhysicalLessonOptions,
+    freeDays,
+    lessonTimeRange,
+    lunchTimeRange,
+    maxConsecutiveHours,
+  } = optimiserFormFields;
+
+  const [hasSaturday, setHasSaturday] = useState(false);
+  const [unassignedLessons, setUnassignedLessons] = useState<LessonOption[]>([]);
+  const [shareableLink, setShareableLink] = useState<string>('');
```
ZhangYiJiang: Using empty string as a default state is pretty iffy, you're usually better off being explicit about it and using `null` to explicitly signal this is optional and empty by default

website/src/views/optimiser/OptimiserResults.tsx
```diff
@@ -1,122 +1,122 @@
...
 const OptimiserResults: React.FC<OptimiserResultsProps> = ({
   shareableLink,
-  unAssignedLessons,
-  openOptimisedTimetable,
+  unassignedLessons,
 }) => {
   const optimiserResultsRef = useRef<HTMLDivElement>(null);
 
+  useEffect(() => {
+    scrollToBottom();
```
ZhangYiJiang: This will also trigger on first render. The div is not rendered so nothing would happen but you should be explicit about this

website/src/views/optimiser/OptimiserResults.tsx
```diff
@@ -1,122 +1,122 @@
...
+
   const scrollToBottom = () => {
     const element = optimiserResultsRef.current;
     if (element) {
       element.scrollIntoView({ behavior: 'smooth' });
     }
   };
 
-  useEffect(() => {
+  const openOptimisedTimetable = () => {
```
ZhangYiJiang: This should be an anchor with button styling instead, no need for a callback

website/src/views/optimiser/OptimiserResults.tsx
```diff
@@ -1,122 +1,122 @@
...
   return (
-    <div ref={optimiserResultsRef}>
-      {/* Partially optimised timetable */}
-      {unAssignedLessons.length > 0 && (
-        <div className={styles.unassignedWarning}>
-          <div className={styles.unassignedHeader}>
-            <AlertTriangle size={24} />
-            Optimiser Warning : Unassigned Lessons
-          </div>
+    !!shareableLink && (
```
ZhangYiJiang: This is just a style thing but usually for rendering precondition you'd want to call it out a bit more explicitly so `if (!shareableLink) return null;` is preferable. It also lets you remove one layer of nesting

website/src/views/optimiser/OptimiserContainer/OptimiserContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+    [lessonOptions, physicalLessonOptions],
+  );
+
+  useEffect(() => {
+    const availableKeys = new Set(lessonOptions.map((option) => option.lessonKey));
+    setPhysicalLessonOptions((prev) =>
+      prev.filter((lesson) => availableKeys.has(lesson.lessonKey)),
+    );
+    setHasSaturday(isSaturdayInOptions(lessonOptions));
+  }, [lessonOptions, setHasSaturday, setPhysicalLessonOptions]);
```
ZhangYiJiang: This looks like it should either be an initial state, if this is meant to only be set once when the component is mounted, a derived value, or a state that should be updated together, in which case it is actually one large state instead

website/src/views/optimiser/OptimiserContainer/OptimiserContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+      .then(parseData)
+      .catch((e) => setError(e))
+      .finally(() => setIsOptimising(false));
+  };
+
+  const parseData = async (data: OptimiseResponse | null) => {
+    const link = data?.shareableLink;
+    if (!link) {
+      return;
+    }
```
ZhangYiJiang, line 107: This feels like an error case

ravern (commented):
Can I check if the design of the components should be changing? Since this is posed as a refactor rather than a design update.

On the branch preview:

image

On production:

image

leslieyip02:
The UI changes were made to address @ZhangYiJiang's comments on accessibility. The free days select should be OK, but I think I'll push a commit to make the dropdown fit in better with the rest of the form (I hadn't checked Safari, but it looks a bit jarring).

**Update:**

I've updated the dropdown to use `form-control` to be more in line with the other dropdowns.

| Firefox | Safari |
| - | - |
| Screenshot 2025-08-18 at 4 15 46 PM |  Screenshot 2025-08-18 at 4 14 20 PM |

PR #4096, opened 2025-07-09
fix(free days conflict): bug fix in key

### LINKED ISSUE #4095: Optimiser free day conflict is not working properly

thejus03:
## Context
Fixes #4095 

## Implementation
Minor bug in changing the key from uniqueKey to displayText for another line too. Previously in #4091 I only updated one line but there was another line that needed to be changed to fix this too.

### DIFF
--- a/website/src/views/optimiser/OptimiserContent.tsx
+++ b/website/src/views/optimiser/OptimiserContent.tsx
@@ -94,7 +94,7 @@ const OptimiserContent: React.FC = () => {
           }
         }
       });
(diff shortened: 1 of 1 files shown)


leslieyip02:
Please hold on for a bit. I'm refactoring the optimiser functions into a utils file (#4094), so I've moved and modified the free day conflict function, so it should already be fixed in that version.

thejus03:
sounds good!

PR #4099, opened 2025-07-10
fix: layout overflow in Course Not Found page

### LINKED ISSUE #4089: Course Not Found page layout overflow
### Describe the bug

In the "Course Not Found" page, there is a list of previous AY archives. The list does not wrap properly

Image

### To Reproduce

1. Go to https://nusmods.com/courses/PF1101/fundamentals-of-project-management

### Expected behavior

The archives should wrap properly, or be displayed as a vertical list like in the mobile view:

Image

### Additional Context

If you are interested in helping out with this issue, you can start looking at `ModuleNotFoundPageComponent`.

zamora-carlos:
## Context

In the "Course Not Found" page, the list of AY archives did not wrap and overflowed its container.

Closes #4089

## Implementation

Added `flex-wrap: wrap;` to allow the list to wrap properly and `gap` to space the elements instead of margins.

Screenshot 2025-07-09 at 8 07 48 p m

The `.buttons` class is used in the `ModuleFinderNoHits` component, so I verified that the styles still render correctly there as well:

Screenshot 2025-07-09 at 8 06 39 p m

I also set a `min-width` on the elements to prevent layout issues where rows would end up with different numbers of buttons. The following image shows how it looked **before** adding the `min-width` property:

image

### DIFF
--- a/website/src/views/errors/ErrorPage.scss
+++ b/website/src/views/errors/ErrorPage.scss
@@ -40,8 +40,8 @@
   display: flex;
   flex-direction: column;
   flex-wrap: wrap;
-  gap: 0.8rem 1rem;
   margin-bottom: 2rem;
+  gap: 0.8rem 1rem;
 
   @include media-breakpoint-up(sm) {
     flex-direction: row;


jloh02:
Love the fix thanks! Looks great!

PR #4101, opened 2025-07-10
fix(optimiser): account for non-overlapping same-timeslot courses

### LINKED ISSUE #4100: Optimiser does not take into account weeks of the semester
### Describe the bug

The optimiser thinks there are clashes in time when there are lessons at the same time and day but weeks that do NOT clash.

### To Reproduce
These are the optimiser preferences:
Image

Optimiser thinks there is a conflict in this:

Image
if you notice they are both at the same time but the weeks do not clash.

### Expected behavior

A timetable must be returned by optimiser that takes into account the weeks of the lesson

thejus03:
## Context
fixes #4100 
fixes another problem i only realised when looking at code which was:
- it did not merge duplicate slots (slots happening at same time, day, building and weeks) for lessons that were not selected to be live lessons...which does not make sense.
fixes problem with creating partial timetables:
- Previously it did not really create partial timetables. It just showed partial timetable when all the slots after filtering out those outside of earliest-latest timings and missing venues was 0. 

I also made other optimisations that are language-specific that caused unnecessary copies to be made 

## Implementation
* Added Weeks field to ModuleSlot
* Introduced WeeksSet and WeeksString for efficient week handling in module timetable processing.
* Updated merge and filter logic to incorporate weeks in combination keys and conflict checks.
* Updated beam search to allow for better partial timetable creations.

## Regards to fix for week support
While most courses have weeks to be a []int, some have weeks to be a map with keys like `start` and `end`. For these courses, there is no week conflict detection being done.


website/api/optimiser/_models/models.go
```diff
@@ -34,12 +34,15 @@ type ModuleSlot struct {
 	StartTime   string      `json:"startTime"`
 	Venue       string      `json:"venue"`
 	Coordinates Coordinates `json:"coordinates"`
+	Weeks       []int       `json:"weeks"`
```
leslieyip02: I think you may need to do more parsing/put error handling because `Weeks` might not be an array of numers:

image

leslieyip02: If it's just a bunch of weird mods that have non-standard week ranges, we can consider rejecting them. I'll see what I can find.

leslieyip02 (changes requested):

website/api/optimiser/_models/models.go
leslieyip02: I've dug around a bit and there are actually quite a few mods that have non-standard week ranges:

image

One more thing to note is that a lot of the special term mods will have these weird week ranges, so if we want to support them in the future, we would need to handle the weeks properly.

However, that sounds like a lot of additional complexity for an unclear benefit, so I think it would be easier to just skip lessons that have these ranges. What do you think?

thejus03: just to clarify, i think it will work for all slots that have weeks to be an []int. So it will work for all the lessons you showed in the picture except for `CDE4501A` and `LL4359Z` where weeks are not []int but look like this:
image

Also can I ask what do you mean by "skip lessons"? Like remove lessons from being considered by optimiser or not care about the weeks for these lessons?

jloh02: A good example of common mods to support are HSI which has lots of such "odd weeks" but many piece together these nicely. LSM mods should also have this feature to somehow "piece together" the courses from non-conflictin weeks so this is a common use case

thejus03:
@leslieyip02 @jloh02 I have made the changes requested and also fixed a problem with partial timetable creation. 
For instance for this mod, previously it would return empty timetable and not a partial one
image
But now,
image
It returns a non-empty partial timetable

Ideally this should allow for better partial timetable creation from now!

website/api/optimiser/_models/models.go
thejus03: yea so I think we can ignore week conflicts for non-[]int weeks for now

website/api/optimiser/_modules/modules.go
```diff
@@ -47,8 +48,31 @@ func GetAllModuleSlots(optimiserRequest models.OptimiserRequest) (map[string]map
 			}
 		}
 
+		// Parse the weeks
+		for i := range moduleTimetable {
+
+			// Note: if weeks is not a []int, then skip parsing
+			// Currently we are not handling week conflict for non-[]int weeks
+			if _, ok := moduleTimetable[i].Weeks.([]any); !ok {
+				// fmt.Println("weeks is not a []int", moduleTimetable[i].Weeks)
```
leslieyip02: Here.

website/api/optimiser/_modules/modules.go
```diff
@@ -47,8 +48,31 @@ func GetAllModuleSlots(optimiserRequest models.OptimiserRequest) (map[string]map
...
+				weekInt := int(week.(float64))
+				moduleTimetable[i].WeeksSet[weekInt] = true
+				weeksStrings[j] = strconv.Itoa(weekInt)
+			}
+			moduleTimetable[i].WeeksString = strings.Join(weeksStrings, ",")
+		}
+
 		// Store the module slots for the module
 		moduleSlots[module] = mergeAndFilterModuleSlots(moduleTimetable, venues, optimiserRequest, module)
+		// fmt.Println("moduleSlots", moduleSlots[module])
```
leslieyip02: And here.

website/api/optimiser/_modules/modules.go
thejus03: oh yes

ravern (approved):
LGTM

PR #4105, opened 2025-07-11
Add Optimiser Beta toaster

thejus03:
## Context

image

- User can click the "Turn on BETA" button to easily turn on NUSMods BETA and get access to optimiser tab.
- User can close the toaster by either pressing the "Turn on BETA" button or pressing the close button.

## Implementation

- Introduced a new BetaAnnouncement component with associated styles.
- The component conditionally renders based on the user's beta testing status.
- Integrated the BetaAnnouncement into the TimetableContent view for visibility.


website/src/views/components/notfications/BetaAnnouncement.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const beta = useSelector(({ settings }: State) => settings.beta);
+  const [isVisible, setIsVisible] = useState(true);
+
+  // Don't show if user is already a beta tester or manually dismissed
+  if (beta || !isVisible) {
+    return null;
+  }
+
+  const handleDismiss = () => {
+    setIsVisible(false);
```
leslieyip02: When the notification is dismissed, I think it should stay dismissed. It could get annoying for users to close the notification, navigate to another page (e.g. courses, planner), and have the notification appear again when they go back to the timetable page.

You can take a look at how the course announcement does it:

https://github.com/nusmodifications/nusmods/blob/2959785d063768aa7cd8eb93785f75f54a2114c4/website/src/views/components/notfications/Announcements.tsx#L45-L55

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -442,6 +443,7 @@ class TimetableContent extends React.Component<Props, State> {
         <Title>Timetable</Title>
 
         <Announcements />
+        <BetaAnnouncement />
```
leslieyip02: Bit of a nitpick but I think it would be better to rename this to `BetaBanner` or something similar to how `HacktoberfestBanner` is named. I think it would be confusing to have `Announcements` followed by `BetaAnnouncement`.

website/src/views/components/notfications/BetaAnnouncement.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <p className={styles.bodyElement}>
+          Turn on NUSMods BETA to try out the new timetable optimiser to help you find your optimal
+          timetable based on your preferences.
+        </p>
+        <p className={styles.bodyElement}>Share your feedback to help us make it even better!</p>
+        <div className={styles.Betabutton}>
+          <button
+            className="btn btn-info"
+            type="button"
+            onClick={() => dispatch(toggleBetaTesting())}
```
leslieyip02: Just a comment on the UX, but maybe it would be better if the button would instead redirect to the settings page? This allows users to make a more conscious choice, and also lets them know where they can turn the setting off.

image

So the new flow would be:

1. Click button
2. Redirect to settings page and beta setting (i.e. https://nusmods.com/settings#beta)
3. Don't show banner again

What do you think?

leslieyip02: If you end up doing the settings page redirect, please also help to convert "Current tests" to title case, and maybe we can also change the description to "Build a timetable that suits you" or something more neutral instead because I think "perfect" might be overpromising.

thejus03: > So the new flow would be:
> 
> 1. Click button
> 2. Redirect to settings page and beta setting (i.e. https://nusmods.com/settings#beta)
> 3. Don't show banner again
> 
> What do you think?

Yep this makes sense to me

leslieyip02 (approved):
Thanks for the quick fixes! LGTM

PR #4107, opened 2025-07-12
feat(planner): add setting to include exempted module credits in total calculation

FrancisRalph:
## Context
I was planning my mods but had quite a few exemptions and realised the credit calculation didn't include exempted modules. So I thought it would be useful to have it as a setting

## Implementation
- Added setting to include exempted module credits in the calculation
- Updated total MC calculation to conditionally include exempted modules based on the setting

### DIFF
--- a/website/src/actions/planner.ts
+++ b/website/src/actions/planner.ts
@@ -35,6 +35,14 @@ export function setIgnorePrerequisitesCheck(prereqsCheck: boolean) {
   };
 }
 
+export const SET_INCLUDE_EXEMPTED_MODULE_CREDITS = 'SET_INCLUDE_EXEMPTED_MODULE_CREDITS' as const;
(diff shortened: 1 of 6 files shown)


FrancisRalph:
Thoughts on enabling this setting by default? Currently have it disabled by default so existing users do not see changes in the unit calculation, but to me it makes more sense to enable it

FrancisRalph:
Nvm, I realise there is no real need for this feature as I can just calculate the total MC i need excluding the exempted mods

And it affects the pre-req warning if I have a module I took a placement test for e.g. LAC2201 requires LAC1201. I can take LAC2201 if I took the placement test, so to silence the warning I can add LAC1201 to the exemptions. But adding LAC1201 in the exemptions counts towards the total MC which is not accurate as it doesnt actually get included in your transcript

Ideally the solution to that should be to ignore the pre-req warning for a singular module, but that's too much effort lol

PR #4108, opened 2025-07-13
fix(moduletree): remove leading line for inactive root node

### LINKED ISSUE #4103: Prerequisite Tree leading branch on unavailable mods
### Describe the bug

`ModuleTree` node has a leading branch when

1. The class is not currently offered and
2. The class is not a prerequisite for another mod

Image

### To Reproduce

Reproducible on Chrome and Firefox.

1. Go to https://nusmods.com/courses/CS6202/advanced-topics-in-programming-languages or https://nusmods.com/courses/CS5271/performance-analysis-of-embedded-systems

### Expected behavior

Should not have a leading branch:

Image

Punpun1643:
## Context
Fixes #4103 

### Issue
- Leading horizontal line always appear when the node is *inactive* (i.e. `moduleActive` is `undefined`).

### Fixes:
- This PR updates the logic to **exclude the horizontal line** when:
    - The module is *inactive* (`moduleActive` is `undefined`), **and**
    - The node is a *root* node (i.e. `layer = 1`) 
- Update snapshot as a new css selector is introduced and added to a `classname`

Before: 

as per issue description

After:
1CF83590-96F1-44FA-8BD4-BE712F5F7BAB

## Other Information
Alternative implementation of PR #4110.

It preserves the original line length even when the root node is inactive and has `fulfillRequirements`. However, that PR introduces slightly more complex logic.

In this PR, the appearance of this specific case where the node is inactive but has `fulfillRequirements` is not preserved where the left line before the node appears shorter.

Before:

6EE708D7-F1F4-4FB9-8767-7CEF62D3B826

After:

582B5BEA-47E0-4891-AD01-1C293865B19C


leslieyip02 (changes requested):
Good work on the fix!

However, I think it introduces some inconsistencies in the lengths of the line:

| Right prereqs | Left prereqs |
| - | - |
| Screenshot 2025-07-14 195555 | Screenshot 2025-07-14 195609 |

If I'm not wrong, we can change this line:

https://github.com/nusmodifications/nusmods/blob/2959785d063768aa7cd8eb93785f75f54a2114c4/website/src/views/modules/ModuleTree.tsx#L171-L178

so that `prereqTreeOnLeft={prereqTreeOnLeft}` instead of `prereqTreeOnLeft={false}`, this should fix the spacing. This is based on my own testing so far. Let me know if this makes sense/works?

Punpun1643:
Hi @leslieyip02, thanks for the review!

I'm not sure if I'm testing correctly as I'm quite new to the codebase. How I produce the example in your screenshot is:

Change the prop below to `reverse={prereqTreeOnLeft}` (line 166) to invert `Tree` and `Branch` positions
https://github.com/nusmodifications/nusmods/blob/c713b35e0869e3ec35bea9eb5271e4d1e97fe920/website/src/views/modules/ModuleTree.tsx#L166-L180

Then, change the prop below to `prereqTreeOnLeft={!prereqTreeOnLeft}` (line 171) to invert the `Branch`

https://github.com/nusmodifications/nusmods/blob/c713b35e0869e3ec35bea9eb5271e4d1e97fe920/website/src/views/modules/ModuleTree.tsx#L168-L173

I then try your suggestion to change the below to `prereqTreeOnLeft={prereqTreeOnLeft}` 

https://github.com/nusmodifications/nusmods/blob/c713b35e0869e3ec35bea9eb5271e4d1e97fe920/website/src/views/modules/ModuleTree.tsx#L175-L180

And I still get the result below:
96B27675-CF95-4874-8024-7D3817E45AB6_4_5005_c

The code change suggestion make sense to me that the padding in `.leftNode` selector should follow the state in `prereqTreeOnLeft` so I'm wondering if I'm missing out anything.

leslieyip02:
Sorry that I wasn't very clear, but I meant that the inconsistency happens when we toggle the prereq tree direction in the settings:

https://github.com/user-attachments/assets/e003aa15-1198-4033-a8cc-9bf3afc81fd2

I think after we change the `prereqTreeOnLeft={false}` line then it should be fixed. Could you push that change? Thanks in advance!

Punpun1643:
Oh I see, thanks for the explanation! I have made the fix.

leslieyip02 (approved):
LGTM. Thanks for the PR!

PR #4110, opened 2025-07-14
fix(moduletree): remove leading line for inactive root node (alternative of #4108)

### LINKED ISSUE #4103: Prerequisite Tree leading branch on unavailable mods

Punpun1643:
## Context

Fixes #4103

### Issue
- Leading horizontal line always appear when the node is *inactive* (i.e. `moduleActive` is `undefined`).

### Fixes:
- This PR updates the logic to **exclude the horizontal line** when:
    - The module is *inactive* (`moduleActive` is `undefined`), **and**
    - The node is a *root* node (i.e. `layer = 1`) , **and** 
    - The node doesn't have `fulfillRequirements`

Before: 

as per issue description

After:
1CF83590-96F1-44FA-8BD4-BE712F5F7BAB

## Other Information

Alternative implementation of PR #4108

### DIFF
--- a/website/src/views/modules/ModuleTree.scss
+++ b/website/src/views/modules/ModuleTree.scss
@@ -126,7 +126,7 @@ $connector-size: 0.75rem;
 
 .leftBranch {
   justify-content: flex-end;
-
+  
   &::before,
   &::after {
     right: 0;


leslieyip02:
I think I prefer #4108 because it modifies the style on the node instead of the entire `ModuleTreeComponent`. Let's continue over there  instead.

PR #4111, opened 2025-07-17
implement clearer UX for transparent color in color picker

### LINKED ISSUE #4081: Clearer UX required for transparent color in color picker
With the updates in #4008, we should ensure that the feature is more obvious. My recommendation is to change the currently chosen color to an indicator for transparent color

junnjiee:
## Context

Fixes #4081 

## Implementation

image
image

Instead of an `X` symbol, a transparent color icon is used to hint that the button removes the color.


website/src/views/components/ColorPicker.scss
```diff
@@ -139,3 +138,12 @@ $arrow-height: 1.5rem;
...
   }
 }
+.transparentIcon {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: 1rem;
+  height: 1rem;
+  transform: translate(-50%, -50%);
+  fill: var(--gray-lighter);
```
leslieyip02: I think that the light gray color could be a bit hard to see on some colors, e.g.:

Screenshot 2025-08-03 at 8 48 21 PM

Could you remove fill here? I think the `color-${index}` class automatically adjusts the stroke color of the icon. This is how the text color is handled for timetable cells:

Screenshot 2025-08-03 at 9 02 57 PM

junnjiee:
Hi @leslieyip02, thanks for reviewing!

Committed changes in 4564f4fce81ff289bfc5820b3d8658be5f61955d which solves the issue your brought up. The icon now inherits its color based on the background using `fill="currentColor"`.

Examples:
image
image
image

leslieyip02 (approved):
LGTM!

PR #4174, opened 2025-08-17
feat(website): Sync state changes across tabs

TheMythologist:
## Context
Previously, changes to a redux state were not reflected in other tabs. This causes 2 main issues:
1. **Major** issue: Scenario where 2 tabs are opened to different modules. Clicking "Add to Semester ..." button in 1 tab, followed by the "Add to Semester ..." button in the other tab, will only cause the 2nd module to be added. (The change in the second tab overrode the change in the first tab).
2. _Minor_ issue: Scenario where 2 tabs are opened, 1 to the timetable and another tab to a module. After clicking "Add to Semester..." button in the module tab, you have to manually refresh the timetable tab for this module to appear.

## Implementation
Uses [`redux-state-sync`](https://github.com/AOHUA/redux-state-sync) to sync redux states across tabs, which uses a polyfilled [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) under the hood.

## Other Information
Another option I looked at was [`redux-persist-crosstab`](https://github.com/rt2zz/redux-persist-crosstab), which is unmaintained and incompatible with the current version of `redux-persist`.

Additionally, I had to forcefully resolve `broadcast-channel` to version `5.3.0`.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -63,6 +63,7 @@
     "@types/react-router-dom": "5.3.3",
     "@types/react-scrollspy": "3.3.9",
     "@types/redux-mock-store": "1.5.0",
+    "@types/redux-state-sync": "^3.1.10",
     "@types/use-subscription": "1.0.2",
     "@types/webpack-env": "1.18.5",
     "@typescript-eslint/eslint-plugin": "4.33.0",
@@ -168,6 +169,7 @@
     "react-scrollspy": "3.4.3",
     "redux": "4.2.1",
     "redux-persist": "6.0.0",
+    "redux-state-sync": "^3.1.4",
     "redux-thunk": "2.4.2",
     "reselect": "4.1.8",
     "samlify": "2.7.7",
@@ -179,6 +181,7 @@
     "extends browserslist-config-nusmods"
   ],
   "resolutions": {
+    "broadcast-channel": "5.3.0",
     "cheerio": "1.0.0-rc.12"
   }
 }
--- a/website/src/middlewares/state-sync-middleware.ts
+++ b/website/src/middlewares/state-sync-middleware.ts
@@ -1,0 +1,27 @@
+import type { AnyAction } from 'redux';
+import { PERSIST, PURGE, REHYDRATE } from 'redux-persist';
+import { createStateSyncMiddleware } from 'redux-state-sync';
+
+const reduxStateSyncConfig = {
+  predicate: (action: AnyAction) => {
+    // Reference: https://github.com/aohua/redux-state-sync/issues/53
+    const blacklist = [PERSIST, PURGE, REHYDRATE];
+
+    // redux-state-sync relies on BroadcastChannel, which only supports
+    // objects that are clonable by `structuredClone`
+    if (typeof action === 'function') {
+      return false;
+    }
+
+    // Request actions starting with `FETCH_` are excluded as well
+    if (action.type.toString().startsWith('FETCH_')) {
+      return false;
+    }
+
+    return !blacklist.includes(action.type);
+  },
+};
+
+const stateSyncMiddleware = createStateSyncMiddleware(reduxStateSyncConfig);
+
+export default stateSyncMiddleware;
--- a/website/src/bootstrapping/configure-store.ts
+++ b/website/src/bootstrapping/configure-store.ts
@@ -6,6 +6,7 @@ import { setAutoFreeze } from 'immer';
 import rootReducer from 'reducers';
 import requestsMiddleware from 'middlewares/requests-middleware';
 import ravenMiddleware from 'middlewares/raven-middleware';
+import stateSyncMiddleware from 'middlewares/state-sync-middleware';
 import getLocalStorage from 'storage/localStorage';
 
 import type { GetState } from 'types/redux';
@@ -25,7 +26,7 @@ export default function configureStore(defaultState?: State) {
   // to reduce the amount of data NUSMods is using
   getLocalStorage().removeItem('reduxState');
 
-  const middlewares = [ravenMiddleware, thunk, requestsMiddleware];
+  const middlewares = [ravenMiddleware, thunk, requestsMiddleware, stateSyncMiddleware];
 
   if (NUSMODS_ENV === 'development') {
     // eslint-disable-next-line @typescript-eslint/no-var-requires, global-require, import/no-extraneous-dependencies
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -1083,6 +1083,13 @@
     "@babel/plugin-transform-modules-commonjs" "^7.25.9"
     "@babel/plugin-transform-typescript" "^7.25.9"
 
+"@babel/runtime@7.22.10":
+  version "7.22.10"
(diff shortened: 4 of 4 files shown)


website/src/middlewares/state-sync-middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+    // Reference: https://github.com/aohua/redux-state-sync/issues/53
+    const blacklist = [PERSIST, PURGE, REHYDRATE];
+
+    // redux-state-sync relies on BroadcastChannel, which only supports
+    // objects that are clonable by `structuredClone`
+    if (typeof action === 'function') {
+      return false;
+    }
+
+    // Request actions starting with `FETCH_` are excluded as well
```
zwliew: Instead of stating what the code does, can you explain why you exclude these actions?

zwliew (approved):
Thanks for looking into this! The change looks good to me with a minor nit.

Do let one of the other active maintainers @ravern or @jloh02 review it too before merging.

website/package.json
```diff
@@ -63,6 +63,7 @@
     "@types/react-router-dom": "5.3.3",
     "@types/react-scrollspy": "3.3.9",
     "@types/redux-mock-store": "1.5.0",
+    "@types/redux-state-sync": "^3.1.10",
```
zwliew: We use renovate to update our dependencies (though we haven't been very diligent in doing so)

```suggestion
    "@types/redux-state-sync": "3.1.10",
```

Could you also update the yarn.lock file?

website/package.json
```diff
@@ -168,6 +169,7 @@
     "react-scrollspy": "3.4.3",
     "redux": "4.2.1",
     "redux-persist": "6.0.0",
+    "redux-state-sync": "^3.1.4",
```
zwliew:
```suggestion
    "redux-state-sync": "3.1.4",
```

website/package.json
```diff
@@ -179,6 +181,7 @@
     "extends browserslist-config-nusmods"
   ],
   "resolutions": {
+    "broadcast-channel": "5.3.0",
```
zwliew: Is this an incompatibility between versions of broadcast-channel after 5.3.0 and redux-state-sync? If so, is there an issue for this? I wonder what the cause is.

TheMythologist: At first I was getting some error regarding a type error in the broadcastchannel's constructor method, but after sticking the `redux-state-sync` package and its typescript definitions to `3.1.4`, the error is resolved. Still not sure exactly what the error is 🤷🏻

zwliew: Oh, I see that the CircleCI test job now fails with your recent change. That's probably the reason. I'm fine with pinning broadcast-channel, but could you also add the failure log to your PR description?

TheMythologist: I think I managed to resolve the test issues now without fixing the `broadcast-channel` npm package 👍🏻

zwliew: Nice! Unfortunately, it seems like the test job is still failing. Could you look into it?

TheMythologist: Yup, it's because the broadcast channel handle is left open so the jest isn't closing. I think I'll be mocking redux-state-sync in the tests so it won't actually do the syncing, but I'm a bit busy at the moment. I'll get to it when I have more time 👍

TheMythologist:
Hi @zwliew I finally got to working on this again, could you help me take a look at the changes and review this PR once more? Thanks!

jloh02 (changes requested):
hi i noticed a bug. steps to reproduce
1. Open 2 tabs on https://nusmods-website-mmnouxt0t-modsbots-projects.vercel.app/timetable/sem-1
2. Add a course on one tab
3. Other tab has the screenshot as below

image

website/src/middlewares/state-sync-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+    const blacklist = [PERSIST, PURGE, REHYDRATE];
+
+    // redux-state-sync relies on BroadcastChannel, which only supports
+    // objects that are clonable by `structuredClone`
+    if (typeof action === 'function') {
+      return false;
+    }
+
+    // `FETCH_` request actions should not be synced to other tabs
+    if (action.type.toString().startsWith('FETCH_')) {
```
zwliew:
````
hook.js:608 TypeError: Cannot read properties of undefined (reading 'find')
    at u (modules.ts:30:30)
    at d (modules.ts:35:16)
    at timetables.ts:168:21
    at mapValues.js:38:34
    at _createBaseFor.js:17:11
    at e.exports (_baseForOwn.js:13:20)
    at e.exports (mapValues.js:37:3)
    at ue (timetables.ts:167:10)
    at timetables.ts:127:14
````

The problem is that in `getModuleSemesterData()` of `modules.ts`, the `semesterData` field of `module` is undefined for a module that was synced over from another tab.

Perhaps a fix for this is to whitelist the `FETCH_MODULE_REQUEST` and `FETCH_MODULE_SUCCESS` actions, which are there to add module info (like `semesterData`) for newly added modules. 

As an aside, I would be more comfortable with switching over to an action whitelist rather than a blacklist.

TheMythologist:
@jloh02 @zwliew Finally got to fixing the issue. The reason for the weird filtering by requests starting with "FETCH_" was ping-pong responses between tabs whenever any fetch request was started, resulting in actions with type `FETCH_MODULE_REQUEST_REQUEST_REQUEST_REQUEST_REQUEST_REQUEST_SUCCESS`. The fix was in the requests middleware, to prevent ongoing requests from being propagated again.

ravern:
Thanks for the recent fix!

Just as a first pass, I'm going to enable an AI code review tool. But we (humans) will still be the ones reviewing and approving later on.

@greptile

TheMythologist:
I can't recall why I initially locked the version of `broadcast-channel` package to 5.3.0, and that commit was too long ago that I forgot as well 😅 

That being said, the current version that is installed (3.7.0) is pretty old. I did some local testing by fixing broadcast-channel to the latest version (7.3.0) and it works fine, looking at the changelog nothing that is used by broadcast-channel is affected. Should I fix it to the latest version? @ravern

ravern (commented):
LGTM overall :D

Thank you for implementing this, and apologies for the insanely long review time.

Regarding the `broadcast-channel`, since it is a dep of a dep, I'd rather just leave it at whatever version `redux-state-sync` specifies. Unless there's a security flaw that warrants a version bump.

Fix the build errors, and I'll be happy to approve and merge this in!

TheMythologist:
@ravern thanks for the review! No worries, it took me just as long to work on it as well 😅 

Regarding the build failures, what specifically is failing? If you're referring to the typecheck failures, those files are not touched by me so I'm not sure what you mean. I tried `yarn build` locally and it seems to be working fine for me. I also can't exactly see the vercel build logs so I'm not sure what went wrong, perhaps you could copy paste the logs from there so I can see what needs to be fixed?

ravern:
For the build failure, I was referring to the CircleCI workflow, which is publicly-accessible. This is the one that's failing: https://app.circleci.com/pipelines/github/nusmodifications/nusmods/16305/workflows/f5019a8b-cee9-49c4-9d6e-efb647bc40ba/jobs/96459

ravern:
Yeah it doesn't look like any files you've touched.

I will take over this PR to fix the unrelated issues + get it merged!

ravern (approved):
LGTM since build passes now.

Issue #4168, opened 2025-08-14
Only one class shows up when selecting Lab with multiple classes with same class number


undefined-func:
### Describe the bug

When a class number is linked to multiple classes during the same timeslot, only one class will show up when selecting classes.

### To Reproduce

1. Add ME2143 to timetable
2. Click on lab timeslot to select other lab class number
3. For each class number, only 1 class is shown. However, each class number has 2 labs: either weeks 6 & 12 or weeks 12 & 13.

### Expected behavior

Both labs should show up for each class number.

### Screenshots

Actual timetable:
<img width="608" height="439" alt="Image" src="https://github.com/user-attachments/assets/3be939ff-35ce-42bb-8259-13df5bc7bcb3" />

What shows up when selecting lab:
<img width="812" height="406" alt="Image" src="https://github.com/user-attachments/assets/fc156446-e2c5-4a09-8551-53d5572ebefc" />
* Lab 1 is missing Week 6
* Lab 2 is missing Week 12, etc.

### Additional context

This issue also occurs for other classes such as ME2134, ME2121, etc.

PR #4214, opened 2025-08-28
website: timetables: Fix deserialization of TA courses in share link

zwliew:
## Context

Currently, the `deserializeTa` function does not correctly parse share
links with multiple TA-ed courses.

Fix it and add some unit tests for the function.

Fixes #4171

PR #4225, opened 2025-09-05
Change representation of lesson configs from classNo to lesson indices

zehata:
## Implementation
- The lesson index of a lesson is its index in the timetable in the API response.
- Lesson configs are now saved as mapping of lesson types to arrays of lesson indices. Example:
```TS
{
    CS1010S: {
        Lecture: [53],
        Tutorial: [19],
        Recitation: [32],
    },
}
```
- TA configs have been removed. TA modules are now saved as an array of module codes. `["CS1010S"]`
- TA lessons will use the same lesson config as non-TA modules. TA lessons can now be added and removed individually. Example:
```TS
{
    CS1010S: {
        Lecture: [53],
        Tutorial: [19],
        Recitation: [32, 33],
    },
}
```
- When a TA module is changed back to non-TA, it will use the closest matching valid non-TA lesson config.
- Current format `classNo` lesson configs and TA lesson configs will be migrated to the new `lessonGroup` format
- Serialization format has been changed

|Change|Before|After|
|---|---|---|
|`(..., ...)` denotes arrays of lesson indices|`?CG1111A=LAB:03`|`?CG1111A=LAB:(5,6)`|
|`;` separates different lesson types|`?CS1010S=LEC:1,TUT:1`|`?CS1010S=LEC:(53);TUT:(41)`|
|TA modules serialized similar to hidden|`?CG1111A=LAB:03&ta=CG1111A(LAB:03)`|`?CG1111A=LAB:(5,6)&ta=CG1111A`|
- Current share link serialization format will be deserialized into the new format
- Optimiser has been updated to share links in the new `lessonGroup` format.

## Resolves
All modules can now be set as TA modules.
Resolves https://github.com/nusmodifications/nusmods/issues/3929 and https://github.com/nusmodifications/nusmods/pull/3930 , from which this implementation came from
|Before|After|
|---|---|
|||

Issue with deserialization of TA module lesson configs with multiple lessons in a module, from https://github.com/nusmodifications/nusmods/pull/4214#issuecomment-3257159652.
|Before|After|
|---|---|
|||

All lessons of a single classNo now show up, and are highlighted when one of them is hovered over. Provided that the module is not set as a TA mod. Resolves https://github.com/nusmodifications/nusmods/issues/4168.
|Before|After|
|---|---|
|||

TA module lessons can now be added and removed individually.
User request from: https://t.me/NUSMods/12508 (No issue opened, at least I can't find one)
|Before|After|
|---|---|
|||

### Array Index Stability
Using the index of an array comes with the inherent issue of stability. Furthermore, it may **fail silently**, since lessons of the same lesson type are typically placed next to each other in the timetable list in the API response. So the module lesson config may remain valid even after the module timetable had been updated.

An alternative is to serialize additional info as proposed in https://github.com/nusmodifications/nusmods/pull/3930 . It will make share links very long, because of the serialization of weeks information.

Personally I think the lessons provided by the faculties are stable enough, and they will most likely not add an additional class in the middle of the semester, so I think it will be better idea to create UI to inform users that the module info has changed if it changes, which can be done simply by hashing the module info during scraping.

### DIFF
--- a/website/src/views/timetable/ShareTimetable.scss
+++ b/website/src/views/timetable/ShareTimetable.scss
@@ -60,6 +60,6 @@
 
   svg {
     display: block;
-    margin: 0;
+    margin: 0 auto;
   }
 }


website/api/optimiser/_solver/nusmods_link.go
```diff
@@ -23,39 +23,48 @@ func CreateConfig(assignments map[string]string) map[string]map[string]string {
...
 
 	for moduleCode, lessons := range config {
 		var lessonParams []string
-		for lessonType, classNo := range lessons {
+		for lessonType, lessonIndex := range lessons {
 			// Get abbreviation for lesson type
 			abbrev := constants.LessonTypeAbbrev[strings.ToUpper(lessonType)]
 
-			lessonParams = append(lessonParams, fmt.Sprintf("%s:%s", abbrev, classNo))
+			lessonParams = append(lessonParams, fmt.Sprintf("%s:%s", abbrev, "("+strings.Trim(strings.Join(strings.Fields(fmt.Sprint(lessonIndex)), ","), "[]"))+")")
```
leslieyip02: Could we make a function for this instead? The nesting makes it harder to read.

website/api/optimiser/_solver/nusmods_link.go
```diff
@@ -23,39 +23,48 @@ func CreateConfig(assignments map[string]string) map[string]map[string]string {
...
 		}
 		if len(lessonParams) > 0 {
-			moduleParams = append(moduleParams, fmt.Sprintf("%s=%s", moduleCode, strings.Join(lessonParams, ",")))
+			moduleParams = append(moduleParams, fmt.Sprintf("%s=%s", moduleCode, strings.Join(lessonParams, ";")))
```
leslieyip02: I think we can define a constant for the separator.

zehata:
Changes from #4226 is only a bugfix. The delta from this PR will apply to the fixed master too.

zehata:
@leslieyip02 Any updates?

leslieyip02:
Sorry about the delay, I'm a bit busy at the moment but I will try to review once I have the time. Thanks for your patience!

website/src/actions/timetables.test.ts
```diff
@@ -49,24 +60,72 @@ test('select module color should dispatch a select of module color', () => {
...
+    const state: any = {
+      timetables: timetablesState(ta),
+      moduleBank: { modules: { CS1010S, CS3216 } },
+    };
+    const dispatch = jest.fn();
+    const action = actions.disableTaModule(semester, 'CS1010S');
+
+    action(dispatch, () => state);
+
+    expect(dispatch).toHaveBeenCalled();
```
leslieyip02: Other than just checking that `dispatch` is called, I think we could also add a `toMatchSnapshot()` check in this test.

website/src/types/timetables.ts
```diff
@@ -1,17 +1,48 @@
...
+export type TaModulesConfig = ModuleCode[];
+
+/**
+ * ClassNoTaModulesConfig is the v1 representation of TA modules\
+ * It is a mapping of moduleCode to the TA's lesson types\
+ * It is only used for type annotations in the migration logic
+ */
+export type ClassNoTaModulesConfig = {
   [moduleCode: ModuleCode]: [lessonType: LessonType, classNo: ClassNo][];
 };
```
leslieyip02, line 45: I think it could be clearer to just name these `ModuleLessonConfigV1`/`ModuleLessonConfigV2` and `TaModulesConfigV1`/`TaModulesConfigV2`. It would make it more explicit that the `V1` versions are no longer used and only kept for compatibility.

Also, I personally think `ClassNoTaModulesConfig` is a bit too verbose and the `NoTa` substring in the name could be misleading.

website/src/actions/timetables.ts
```diff
@@ -115,21 +128,62 @@ export function setLesson(
...
+}
+
+export function changeLesson(
+  semester: Semester,
+  moduleCode: ModuleCode,
+  lessonType: LessonType,
+  lessonIndices: LessonIndex[],
+) {
+  return setLesson(semester, moduleCode, lessonType, lessonIndices);
 }
```
leslieyip02: Maybe we can just rename `setLesson` to `changeLesson` instead of adding this function. `setLesson` doesn't seem to be used anywhere else.

website/src/utils/timetables.ts
```diff
@@ -670,17 +1001,259 @@ export function isSameLesson(l1: Lesson, l2: Lesson) {
...
+      return {
+        config: {
+          ...accumulated.config,
+          [semester]: migrated.migratedSemTimetableConfig,
+        },
+        ta: {
+          ...accumulated.ta,
+          [semester]: migrated.migratedTaModulesConfig,
+        },
+        alreadyMigrated: migrated.alreadyMigrated || accumulated.alreadyMigrated,
```
leslieyip02: Should this be `&&` instead of `||`? If any of the configs are somehow not migrated, it would probably be safer to dispatch `Internal.setTimetables()` just in case.

website/src/actions/timetables.ts
```diff
@@ -273,36 +346,62 @@ export function showLessonInTimetable(semester: Semester, moduleCode: ModuleCode
...
+/**
+ * A helper function for disableTaModule\
+ * Removes a module from the semester's timetable TA modules list\
+ * Does not check for the lessonConfig validity as a non-TA lesson config. Use disableTaModule instead.
+ * @param semester
+ * @param moduleCode
+ * @param lessonConfig
+ * @returns
+ */
+export function removeTaModule(
```
leslieyip02, line 364: Since it's a helper function (and not used in tests), we probably don't need the `export`. This lowers the risk of accidentally calling `removeTaModule` without the proper checks.

website/src/reducers/timetables.ts
```diff
@@ -267,19 +285,22 @@ function timetables(
...
       const { semester } = action.payload;
 
       return produce(state, (draft) => {
         draft.lessons[semester] = semTimetable(draft.lessons[semester], action);
         draft.colors[semester] = semColors(state.colors[semester], action);
         draft.hidden[semester] = semHiddenModules(state.hidden[semester], action);
-        draft.ta[semester] = semTaModules(state.ta[semester], action);
+        const taModules = state.ta[semester];
+        const taModulesList = taModules;
+        draft.ta[semester] = semTaModules(taModulesList, action);
```
leslieyip02: Not sure if this reassignment is necessary?

website/src/utils/timetables.test.ts
```diff
@@ -81,51 +89,29 @@ test('randomModuleLessonConfig should return a random lesson config', () => {
...
   const configWithLessons: SemTimetableConfigWithLessons = hydrateSemTimetableWithLessons(
     config,
-    modules,
+    modulesMap,
     sem,
   );
   expect(configWithLessons[moduleCode].Tutorial[0].classNo).toBe('8');
   expect(configWithLessons[moduleCode].Recitation[0].classNo).toBe('4');
   expect(configWithLessons[moduleCode].Lecture[0].classNo).toBe('1');
 });
```
leslieyip02: We could add another test case with multiple lesson indices:

```
  const config: SemTimetableConfig = {
    [moduleCode]: {
      Tutorial: [42, 43, ...],
    },
  };
```

Even though TA config isn't an argument, we should still test that it handles multiple lesson indices.

website/src/utils/timetables.test.ts
```diff
@@ -413,44 +400,232 @@ test('findExamClashes should return non-empty object if exams starting at differ
...
-});
 
-test('deserializing edge cases', () => {
-  // Duplicate module code
-  expect(deserializeTimetable('CS1010S=LEC:01&CS1010S=REC:11')).toEqual({
-    CS1010S: {
-      Lecture: '01',
-      Recitation: '11',
-    },
+  test('deserializing ta modules', () => {
```
leslieyip02: We cover TA modules, but let's add a test for deserializing hidden modules too.

website/src/utils/timetables.test.ts
```diff
@@ -529,16 +704,26 @@ describe(validateTimetableModules, () => {
...
 });
 
-describe('validateModuleLessons', () => {
+// TODO: validate module lessons
+// - either normal non-TA or TA module
+//   - remove lesson group if there are lessons of other lesson types
+// - if module is a normal non TA module:
+//   - remove lesson group if any lesson is missing
+//   - remove lesson group if there are extra lessons
+//
```
leslieyip02, line 720: Currently, the `validateModuleLessons()` test cases only covers the `validateNonTaModuleLesson()` path. We should probably cover the `validateTaModuleLesson()` path as well since the logic is slightly different.

website/src/utils/timetables.ts
```diff
@@ -261,9 +252,9 @@ export function arrangeLessonsWithinDay(lessons: ColoredLesson[]): TimetableDayA
 //    ],
 //    ...
 //  }
-export function arrangeLessonsForWeek(lessons: ColoredLesson[]): TimetableArrangement {
+export function arrangeLessonsForWeek<T extends RawLesson>(lessons: T[]): { [x: string]: T[][] } {
```
leslieyip02: I think the return type can be `TimetableArragement<T>` instead.

website/src/utils/timetables.ts
```diff
@@ -439,6 +430,127 @@ export function validateTimetableModules(
...
+  return {
+    validatedLessonConfig,
+    valid,
+  };
+}
+
+export function getFirstClassNoLessonIndices(
+  lessonsWithLessonType: RawLessonWithIndex[],
+): LessonIndex[] {
+  const { classNo } = lessonsWithLessonType[0];
```
leslieyip02: I think it would be safer to guard this with an empty array check (not sure if it's even possible though).

website/src/utils/timetables.ts
```diff
@@ -439,6 +430,127 @@ export function validateTimetableModules(
...
+  const lessonTypesInLessonConfig = keys(lessonConfig);
+  const { config: validatedLessonConfig, valid: configValid } = reduce(
+    lessonsByType,
+    (accumulatedValidationResult, lessonsWithLessonType, lessonType) => {
+      const lessonTypeInLessonConfig = lessonTypesInLessonConfig.includes(lessonType);
+      const configLessonIndices = lessonConfig[lessonType];
+
+      if (!lessonTypeInLessonConfig || !configLessonIndices.length) {
+        // TODO: Open an issue to make recovery use random lessons instead
+        const validLessonIndices = getFirstClassNoLessonIndices(lessonsWithLessonType);
```
leslieyip02: If it's just for recovery, I don't think randomness matters. Also, maybe we could rename `getFirstClassNoLessonIndices()` to `getRecoveryLessonIndices()` or something else since the underlying logic for choosing the recovery indices shouldn't matter.

website/src/utils/timetables.ts
```diff
@@ -568,94 +653,340 @@ export function formatNumericWeeks(unprocessedWeeks: NumericWeeks): string | nul
...
+
+      const moduleLessonConfig = lessons
+        .split(LESSON_SEP)
+        .reduce((accumulatedModuleLessonConfig, lesson) => {
+          // TUT:2
+          const [lessonTypeAbbr, classNo] = lesson.split(LESSON_TYPE_KEY_VALUE_SEP);
+          // ["TUT", "2"]
+          const lessonType = LESSON_ABBREV_TYPE[lessonTypeAbbr];
+          if (!lessonType) return accumulatedModuleLessonConfig;
+          const lessonsByLessonTypeByClassNo = groupLessonsByLessonTypeByClassNo(timetable);
```
leslieyip02: I think this can be moved out of the `reduce()` block since it doesn't need to be re-computed every time.

website/src/utils/timetables.ts
```diff
@@ -568,94 +653,340 @@ export function formatNumericWeeks(unprocessedWeeks: NumericWeeks): string | nul
...
+): ModuleLessonConfig {
+  // LEC:1,TUT:1,REC:1
+  return reduce(
+    serializedModuleLessonConfig.split(LESSON_SEP),
+    (accumulatedModuleLessonConfig, lessonTypeSerialized) => {
+      // LEC:1
+      const [lessonTypeAbbr, classNo] = lessonTypeSerialized.split(LESSON_TYPE_KEY_VALUE_SEP);
+      // ["LEC", "1"]
+      const lessonType = LESSON_ABBREV_TYPE[lessonTypeAbbr];
+      const lessonsByLessonTypeByClassNo = groupLessonsByLessonTypeByClassNo(timetable);
```
leslieyip02: I think this can be moved out of the `reduce()` block since it doesn't need to be re-computed every time.

website/src/utils/timetables.ts
```diff
@@ -451,37 +563,34 @@ export function validateModuleLessons(
...
+): LessonsByLessonTypeByClassNo => {
+  const lessonsByLessonType = groupBy(lessonsWithIndex, 'lessonType');
+  return mapValues(lessonsByLessonType, (lessonsWithLessonType) => {
+    const lessonsByClassNo = groupBy(lessonsWithLessonType, 'classNo');
+    return mapValues(lessonsByClassNo, (lessonsWithClassNo) =>
+      map(lessonsWithClassNo, 'lessonIndex'),
+    );
+  });
+};
+
```
leslieyip02: This is grouping lesson indices, and not lessons, so we might want to rename this.

`LessonsByLessonTypeByClassNo` and `groupLessonsByLessonTypeByClassNo` are both quite verbose, so one possibility is to rename them to `LessonsIndicesMap` and `makeLessonIndicesMap()`. Then we define a `getLessonIndices(lessonIndicesMap, lessonType, classNo)` helper function to get the lesson indices. That way, the exact order of grouping doesn't matter since it will handled by the helper.

website/src/utils/timetables.ts
```diff
@@ -568,94 +653,340 @@ export function formatNumericWeeks(unprocessedWeeks: NumericWeeks): string | nul
...
+ */
 export function serializeTimetable(timetable: SemTimetableConfig): string {
   // We are using query string safe characters, so this encoding is unnecessary
   return qs.stringify(mapValues(timetable, serializeModuleConfig), { encode: false });
 }
 
-export function deserializeTimetable(serialized: string): SemTimetableConfig {
-  const params = qs.parse(serialized);
-  return mapValues(omit(params, ['hidden', 'ta']), parseModuleConfig);
+// TODO merge logic for TA modules and hidden modules
```
leslieyip02: https://github.com/nusmodifications/nusmods/blob/3f774f365ea9788ae6160aa1bc3443e2234e376f/website/src/views/routes/paths.ts#L25-L34

Are you referring to abstracting the serialization + empty list logic into a common helper function?

```
export function serializeModuleListWithKey(modules: ModuleCode[], key: string): string {
  return isEmpty(modules) ? '' : `&${key}=${modules.join(LESSON_SEP)}`;
}
```

If so, I think that would be worth doing.

website/src/utils/timetables.ts
```diff
@@ -568,94 +653,340 @@ export function formatNumericWeeks(unprocessedWeeks: NumericWeeks): string | nul
...
+  getModuleSemesterTimetable: (moduleCode: ModuleCode) => readonly RawLessonWithIndex[],
+): SemTimetableConfig {
+  // CS2100(TUT:2,TUT:3,LAB:1),CS2107(TUT:8)
+  const trimmedSerializedTaModulesConfig = taSerialized.slice(0, -1);
+  // CS2100(TUT:2,TUT:3,LAB:1),CS2107(TUT:8
+  return reduce(
+    trimmedSerializedTaModulesConfig.split(`)${LESSON_SEP}`),
+    (accumulatedTaTimetableConfig, moduleConfig) => {
+      // CS2100(TUT:2,TUT:3,LAB:1
+      // CS2107(TUT:8
```
leslieyip02: Truncating the last character feels a bit hacky to me. Instead, maybe we could use a regex:

```
taSerialized.split(/(?<=\)),/); // ['CS2100(TUT:2,TUT:3,LAB:1)', 'CS2107(TUT:8)']
```

website/src/views/components/module-info/AddModuleDropdown.test.tsx
```diff
@@ -76,7 +76,7 @@ describe(AddModuleDropdownComponent, () => {
 
   test('should show remove button when the module is in timetable', () => {
     // eslint-disable-next-line no-useless-computed-key
-    const container = make(CS3216, { [1]: { CS3216: { Lecture: '1' } } });
+    const container = make(CS3216, { [1]: { CS3216: { Lecture: [0] } } });
```
leslieyip02, line 79: I think the test at line 62 needs to be edited too.

https://github.com/nusmodifications/nusmods/blob/3f774f365ea9788ae6160aa1bc3443e2234e376f/website/src/views/components/module-info/AddModuleDropdown.test.tsx#L62-L65

website/src/views/timetable/TimetableContent.tsx
```diff
@@ -353,79 +460,19 @@ class TimetableContent extends React.Component<Props, State> {
...
+    ).filter((lesson) => !this.isHiddenInTimetable(lesson.moduleCode));
 
-    // Inject color into module
-    const coloredTimetableLessons = timetableLessons.map(
-      (lesson: Lesson): ColoredLesson => ({
-        ...lesson,
-        colorIndex: colors[lesson.moduleCode],
-        isTaInTimetable: this.isTaInTimetable(lesson.moduleCode),
-      }),
+    const coloredTimetableLessons: InteractableLesson[] = this.getInteractableLessons(
```
leslieyip02: Let's rename this to `interactableLessons` to reflect the change.

zehata:
@leslieyip02 

> Also, just in case we missed something in the migration logic, I think we should wait until the end of the semester before merging this PR (lest we nuke everyone's active timetables).

Same thought actually haha

Thanks for reviewing, I was thinking it might be worth for us to ask other people to take a look at the implementation too, since it is definitely quite disruptive

website/src/utils/timetables.ts
zehata: Haha even I got confused reading this back.

I think when I was writing this I was making that assumption that configs are either fully migrated or not migrated at all, but yea it makes a lot more sense to not make this assumption.

website/src/actions/timetables.ts
zehata: I receive this error if the dispatch action is not exported

```
ts: No overload matches this call.
  The last overload gave the following error.
    Argument of type '{ type: "REMOVE_TA_MODULE"; payload: { semester: number; moduleCode: string; lessonConfig: ModuleLessonConfig; }; }' is not assignable to parameter of type 'DispatchRequestAction<RequestType>'.
      Property 'meta' is missing in type '{ type: "REMOVE_TA_MODULE"; payload: { semester: number; moduleCode: string; lessonConfig: ModuleLessonConfig; }; }' but required in type 'DispatchRequestAction<RequestType>'.
```

leslieyip02: Oh I see. We can just leave it as is.

website/src/utils/timetables.ts
zehata: Yeah, `groupBy<T>` returns `Record<T[]>` even though it should be impossible to have 0 member in any group, the correct one should have been `Record<[T, T[]]>`. I thought of overriding it with a custom implementation but eh it's fine, not worth it.

website/src/utils/timetables.ts
zehata: hmm, I agree with the implementation but I think the name still needs work, I am struggling to think of one too, perhaps `LessonGroup` so it becomes `groupLesson` and `getLessonGroupIndices`?

website/src/utils/timetables.test.ts
zehata: lol accidentally left that todo in, sorry about that :sweat_smile:

zehata: Yeah we should test multiple lesson indices, though I think it's probably fine to just test multiple lesson indices? I don't see why multiple would work but not a single lesson index.

website/src/utils/timetables.test.ts
zehata: I was thinking that the migration tests would be sufficient but you are right, I will add it

website/src/utils/timetables.test.ts
zehata: haha i wanted to do that in a separate PR because it's not atomic to this PR

website/api/optimiser/_solver/nusmods_link.go
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/04319c954b0006c713bfbb66b3d6ffdcf96e98d1

website/api/optimiser/_solver/nusmods_link.go
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/04319c954b0006c713bfbb66b3d6ffdcf96e98d1

website/src/actions/timetables.test.ts
zehata: I added equality tests so that it's easier to see the action called https://github.com/nusmodifications/nusmods/pull/4225/commits/b05656f264570163a0aacb1a76e7ff82dbd9f87b

website/src/types/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/8137c0a628d94d5fab5b4a0ea4831f25d8b6f3c4

website/src/reducers/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/670ff1d8a32d96a3948188c98bfa6dd26a7e5880

website/src/utils/timetables.test.ts
zehata: Do you think this is sufficient? https://github.com/nusmodifications/nusmods/pull/4225/commits/49f3b599cee984f22126eb26a32d21f64540249d

website/src/utils/timetables.test.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/3da612389a77c5e715f09bff753a6691998dd9e9

website/src/utils/timetables.ts
zehata: Fix: 
Unnecessary randomness: https://github.com/nusmodifications/nusmods/pull/4225/commits/4ce7d74c593740a5c7807f236a107ae099073607
Renaming: https://github.com/nusmodifications/nusmods/pull/4225/commits/d205e80a5289cb2b4e1d206090acb51456cbb5fc

website/src/utils/timetables.test.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/9200762871f26335d5f745fcde51601c23ab52d6
Thanks for calling out this one: I noticed that I had missed the recovery logic on the TA mods validation

website/src/utils/timetables.ts
zehata: https://github.com/nusmodifications/nusmods/pull/4225/commits/d24b5af6f3de6aa047b1d7fc5db4247b8ee368b3

website/src/utils/timetables.ts
zehata: Fixed in conjunction with https://github.com/nusmodifications/nusmods/pull/4225#discussion_r2442669087 in commit https://github.com/nusmodifications/nusmods/pull/4225/commits/d24b5af6f3de6aa047b1d7fc5db4247b8ee368b3

website/src/utils/timetables.ts
zehata: Fixed in conjunction with https://github.com/nusmodifications/nusmods/pull/4225#discussion_r2442669087 in commit https://github.com/nusmodifications/nusmods/pull/4225/commits/d24b5af6f3de6aa047b1d7fc5db4247b8ee368b3

website/src/utils/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/6106f4a04fcf85972067eb9e6f8b4fd10bc506db

website/src/utils/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/f47622134c017d2ab144e812dd5ccad218eacf12

website/src/utils/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/8a20993da3b2b5ba43ae4256656c97103a137a01

website/src/actions/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/670ff1d8a32d96a3948188c98bfa6dd26a7e5880

website/src/utils/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/9a9f2c47b31707596d4903a343ce8dfdea2fc585

Note that I moved the params key logic, as I think the abstraction for adding the params key is unnecessary

website/src/utils/timetables.ts
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/6c43b1f22f3445374b2e3274a61c893860e35482 Let me know if I missed any in the same file

website/src/views/components/module-info/AddModuleDropdown.test.tsx
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/14df7f4b60ad4042583a47e608fd83cbe0a0ba30

website/src/views/timetable/TimetableContent.tsx
zehata: Fix: https://github.com/nusmodifications/nusmods/pull/4225/commits/705984050982c06503d3ea16b4b0839ccbf91a15

website/src/utils/timetables.test.ts
```diff
@@ -499,14 +499,14 @@ describe('timetable serialization/deserialization', () => {
     test('deserialize v1', () => {
       expect(
         deserializeTimetable(
-          'CS1010S=LEC:1,TUT:8&CS3216=LEC:1&ta=CS3216(LEC:1),CS1010S(LEC:1,TUT:2)&hidden=CS3216',
+          'CS1010S=LEC:1,TUT:8&CS3216=LEC:1&ta=CS3216(LEC:1),CS1010S(LEC:1,TUT:2,TUT:3)&hidden=CS3216',
```
zehata, line 502: I noticed that I hadn't add a test case to test multiple lessons for a particular lesson type

website/src/utils/timetables.test.ts
```diff
@@ -90,7 +90,6 @@ test('randomModuleLessonConfig should return a random lesson config', () => {
   });
 });
 
-// TODO: how to test TA config...
```
zehata: I wrote this to remind myself to write the test cases to deserialize TA configs. I had forgotten to remove this.

zehata:
Blocked by #4249 because some classes are removed in the optimiser logic, which causes lesson indices reported to be invalid lesson index combinations.

zehata:
Thanks for reviewing again, let me know if there's anything you would like me to change, I just finished some stuff so I should be mostly free this week.

PR #4232, opened 2025-09-13
[Planner] Allow all conflict checks so that future mods correctly flag Prereq check failure 

### LINKED ISSUE #4231: [Planner] Prerequisite check false-positive for alternate-semester mods if placed in the wrong semester of a future year
### Describe the bug

If a module (eg `LAC4202`) that is held on alternative semesters (ie only offered **Semester 2**) is placed in the wrong semester (ie **Semester 1**) of any future year, it is incorrectly marked green (meeting all prerequisites + available for semester). 

- Placing on the *correct semester* (ie **Semester 2**) of **any year** will correctly trigger prerequisite check (either turn green or yellow depending on whether the prerequisites are met)
- Placing on the *wrong semester* of any year **less than or equal** to the current year will turn the module yellow with the warning that it is not offered for the semester.

year|correct sem|wrong sem
----|-----------|---------
up to current year | prereq check works correctly | not-offered check triggers 
\> current year     | prereq check works correctly | **Green** (no prereq check happens)

### To Reproduce

1. Go to planner
2. Go to settings
4. Set *Graduating* to any future year (eg *In 2 years (AY27/28)*)
5. Exit settings
6. Add `LAC4202` to any valid semester (ie Semester 2 of current year)
7. Drag it to the wrong semester of a future year (eg *2026/2027 Semester 1*)
8. Module turns green

### Expected behavior

Module turns yellow with either:
- incorrect semester warning (ie `Module may only be offered in Sem 2`)
- unfulfilled prerequisites warning (ie `These courses may need to be taken first`)

### Screenshots

- ✅  `LAC4202` on the wrong sem = wrong sem warning
- ✅ `CS4221` on the correc sem = unfulfilled prerequisites warning
Image

Bug:
- ❌ **`LAC4202` on the wrong sem = Green** 
- ✅ `CS4221` on the correc sem = unfulfilled prerequisites warning
- ✅ `LAC3201` / `LAC2201` (offered on both semesters) = unfulfilled prerequisites warning
- ignore the banana cursor
Image

- OS: NixOS 25.11.20250706.1fd8bad (Xantusia) x86_64
- DE: niri (Wayland) 
- Browser: Brave 1.80.113 (Official Build) (64-bit) Chromium: 138.0.7204.49
- NUSMods R version: [20250907-d81459f](https://github.com/nusmodifications/nusmods/commit/d81459f68b2f3d54ba07aa6ee86b7528f8559117).

yadobler:
Calculate all conflicts so that prereq and duplicate checks for future sems do not get shadowed.

## Context

Fixes #4231 

image

If a module in the future fails `semesterCheck`, it will not ignore `prereqCheck` or `duplicateCheck` warnings. This allows for the intended case of only showing Prerequisites / Duplicates warnings regardless of whether Module is currently offered in that semester. 

Correct Future Sem? | Prereqs fail? | Duplicates? | result
--|--|--|--
No | No | No | ✅ Green Light
No | Yes | No | ⚠️ Prereq Error 
No | Yes | Yes | ⚠️ Prereq Error 
No | No | Yes | ⚠️ Duplicate Error
Yes | No | No | ✅  Green Light
Yes | Yes | No | ⚠️ Prereq Error
Yes | Yes | Yes | ⚠️ Prereq Error
Yes | No | Yes | ⚠️ Duplicate Error

## Implementation

- `ModuleInfo` now takes an array of `Conflict` instead of just a single one
- All conflict checks are done, instead of terminating at the first non-null conflict
- The `displayedConflict` will either be the first conflict in the `Conflicts` array if current year, or else either Prerequisite or Duplicate conflict for non-current years

## Assumptions:
- conflict-checking functions are added into the conflictChecks array in sequence of priority
- prerequisites takes precedence over duplicates
- future semesters do not include `Semester` conflicts due to possible changes in future timetable

## Other Information

- Not sure if I need to edit any test files
- please ignore banana cursor in screenshot

### DIFF
--- a/website/src/types/planner.ts
+++ b/website/src/types/planner.ts
@@ -55,7 +55,7 @@ export type PlannerModuleInfo = {
   // Custom info added by the student to override our data or to fill in the blanks
   // This is a separate field for easier typing
   customInfo?: CustomModule | null;
-  conflict?: Conflict | null;
+  conflicts?: Conflict[] | null;
   moduleCode?: ModuleCode;
   placeholder?: PlannerPlaceholder;
 };
--- a/website/src/views/planner/PlannerSemester.tsx
+++ b/website/src/views/planner/PlannerSemester.tsx
@@ -4,7 +4,7 @@ import AddCalendarIcon from 'img/icons/add-calendar.svg';
 import classnames from 'classnames';
 
 import { Semester, ModuleCode } from 'types/modules';
-import { AddModuleData, PlannerModuleInfo } from 'types/planner';
+import { AddModuleData, PlannerModuleInfo, Conflict } from 'types/planner';
 import { Dispatch } from 'types/redux';
 import config from 'config';
 import { getExamDate, renderMCs } from 'utils/modules';
@@ -78,16 +78,22 @@ const PlannerSemester: React.FC<Props> = ({
   const dispatch = useDispatch<Dispatch>();
 
   const renderModule = (plannerModule: PlannerModuleInfo, index: number) => {
-    const { id, moduleCode, moduleInfo, conflict, placeholder } = plannerModule;
+    const { id, moduleCode, moduleInfo, conflicts, placeholder } = plannerModule;
 
     const showExamDate = showModuleMeta && config.academicYear === year;
 
     const isModuleInTimetable = moduleCode !== undefined && moduleCode in timetable;
 
-    const displayedConflict =
-      year === config.academicYear || (conflict && ['prereq', 'duplicate'].includes(conflict.type))
-        ? conflict
-        : null;
+    let displayedConflict: Conflict | null = null;
+
+    if (conflicts?.length) {
+      displayedConflict =
+        year === config.academicYear
+          ? conflicts[0] // Topmost conflict if correc year
+          : conflicts.find((c) => c.type === 'prereq') ||
+            conflicts.find((c) => c.type === 'duplicate') ||
+            null; // either prereq or duplicate or none if not same year
+    }
 
     return (
       <PlannerModule
--- a/website/src/selectors/planner.ts
+++ b/website/src/selectors/planner.ts
@@ -136,12 +136,9 @@ function mapModuleToInfo(
   }
 
   if (moduleCode) {
-    // Only continue checking until the first conflict is found
-    let index = 0;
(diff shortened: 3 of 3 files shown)


website/src/types/planner.ts
```diff
@@ -55,7 +55,7 @@ export type PlannerModuleInfo = {
   // Custom info added by the student to override our data or to fill in the blanks
   // This is a separate field for easier typing
   customInfo?: CustomModule | null;
-  conflict?: Conflict | null;
+  conflicts?: Conflict[] | null;
```
leslieyip02: I think replacing `conflict` with an array of `conflicts` is a good choice, and can give us flexibility in the future if we want to add more conflict-related logic.

However, I think it's safer if we made `conflicts` non-null. The default value could just be an empty array.

website/src/views/planner/PlannerSemester.tsx
```diff
@@ -78,16 +78,22 @@ const PlannerSemester: React.FC<Props> = ({
...
+    let displayedConflict: Conflict | null = null;
+
+    if (conflicts?.length) {
+      displayedConflict =
+        year === config.academicYear
+          ? conflicts[0] // Topmost conflict if correc year
+          : conflicts.find((c) => c.type === 'prereq') ||
+            conflicts.find((c) => c.type === 'duplicate') ||
+            null; // either prereq or duplicate or none if not same year
+    }
```
leslieyip02: Maybe we could put this inside `utils/planner.ts` instead. Something like:

```
export function getDisplayedConflict(
  conflicts: Conflict[],
  isCurrentYear: boolean,
): Conflict | null {
  if (isEmpty(conflicts)) {
    return null;
  }

  if (isCurrentYear) {
    return conflicts[0];
  }

  ...

  return null;
}
```

I think this abstraction could help if we ever want to re-order the conflicts in the future, and makes testing easier since we can just call `getDisplayedConflict()`.

website/src/types/planner.ts
yadobler: good idea!

website/src/views/planner/PlannerSemester.tsx
yadobler: Makes sense, I have moved to `utils` and also updated the test cases and also cut down on the single-line shortcircuiting

Issue #4145, opened 2025-08-05
Optimiser partial timetable does not show unallocated lessons


thejus03:
### Is your feature request related to a problem? Please describe.

<img width="479" height="385" alt="Image" src="https://github.com/user-attachments/assets/f17f0297-5d58-41fa-84d9-fdbdb31c0af0" />

<img width="1063" height="691" alt="Image" src="https://github.com/user-attachments/assets/ce7efcfa-82c8-4b9a-9c03-7f9c4dd8fa48" />

When the optimiser returns a partial timetable, it does not show the unallocated timetable in it. So users would have to re-add the module to get the lesson options for the unallocated lesson.

### Describe the solution you'd like

One solution is for the optimiser to just return a random slot for unallocated lessons. This way users can manually look at other lesson timings for it and update their timetable instead of having to re-add the module.

PR #4250, opened 2025-10-31
Removed deduplication logic. See #4249

### LINKED ISSUE #4249: Optimiser Result Lesson Collision
### Describe the bug
Collision between lessons A and B where:
- A is a lesson where there are other lessons in the same module with the same
  - lesson type
  - day
  - start time
  - building name
  - weeks
- B is a lesson from a different module

### To Reproduce
1. Select BN1112 and CS2103T
2. Run optimiser with BN1112 Tutorial and CS2103T Lecture selected

### Expected behavior
No collision between classes

### Screenshots
Image

### Additional context
From #4101 https://github.com/nusmodifications/nusmods/pull/4101/files#diff-a09b3d8eb9d207a92b619e76f23a1c7ccbdfe8bd4483b7f9ff64c750037eb1f5R168

@thejus03 If I am not mistaken, we might have to account for `classNo` as well? It combines slots with different `classNo` that share these details, which makes it possible for slots from other modules to collide with the removed slots.

Image

zehata:
## Context
Resolves #4249 

## Implementation

If I am not mistaken, the previous deduplication logic is no longer needed if we are going to be accounting for classNo within the logic like this:
```diff
-				combinationKey := slot.LessonType + "|" + slot.Day + "|" + slot.StartTime + "|" + buildingName + "|" + slot.WeeksString
+				combinationKey := slot.LessonType + "|" + slot.classNo + "|" + slot.Day + "|" + slot.StartTime + "|" + buildingName + "|" + slot.WeeksString
```

### DIFF
--- a/website/api/optimiser/_modules/modules.go
+++ b/website/api/optimiser/_modules/modules.go
@@ -151,49 +151,21 @@ func mergeAndFilterModuleSlots(timetable []models.ModuleSlot, venues map[string]
 		}
 	}
 
-	/*
-		Now merge all slots of the same lessonType, slot, startTime, weeks and building
-		We are doing this to avoid unnecessary calculations & reduce search space
-	*/
-
-	mergedTimetable := make(map[string]map[string][]models.ModuleSlot) // Lesson Type -> Class No -> []ModuleSlot
-	seenCombinations := make(map[string]bool)
-
+	lessonsMap := make(map[string]map[string][]models.ModuleSlot) // Lesson Type -> Class No -> []ModuleSlot
 	for _, slots := range validClassGroups {
 		for _, slot := range slots {
-
-			if !constants.E_Venues[slot.Venue] {
-				buildingName := extractBuildingName(slot.Venue)
-
-				combinationKey := slot.LessonType + "|" + slot.Day + "|" + slot.StartTime + "|" + buildingName + "|" + slot.WeeksString
-
-				if seenCombinations[combinationKey] {
-					continue
-				}
-				seenCombinations[combinationKey] = true
+			if lessonsMap[slot.LessonType] == nil {
+				lessonsMap[slot.LessonType] = make(map[string][]models.ModuleSlot)
 			}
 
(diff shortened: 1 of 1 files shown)


thejus03:
@zehata Thanks for finding this bug! 

However, I believe that the solution was not to include classNo. For that section of code, I was trying to merge all slots with the lessonType, Day, StartTime and building name to reduce the search space. If we did include classNo, then we would effectively never merge anything because we would simply never encounter another grouping with the same classNo because they are all unique.

I believe the problem that I oversaw was that each grouping (eg G07) has 3 slots right? I was previously merging them as individual entities without looking at them as a whole grouping. So the solution was to find a better way to create combinationKey for the WHOLE grouping

I will send my PR and please take a look and see if you agree!

zehata:
Let me know of a potential alternative to my proposed change, since mine will incur quite the cost.

As you said, adding the `classNo` would be effectively removing the logic, and so the PR I proposed was removing the logic for combining. But of course my change would be negatively impacting performance.

You may want to take a look at #4225 as well, since I actually found this bug while diagnosing an issue whereby some lessons are missing from the resultant share link, and I don't want us to spend too much time on a solution that would not work in conjunction with #4225.

thejus03:
> Let me know of a potential alternative to my proposed change, since mine will incur quite the cost.
> 
> As you said, adding the `classNo` would be effectively removing the logic, and so the PR I proposed was removing the logic for combining. But of course my change would be negatively impacting performance.

Yea for an alternative solution you can refer to #4251. I think it works but please try and test it out, if you have time.

> You may want to take a look at #4225 as well, since I actually found this bug while diagnosing an issue whereby some lessons are missing from the resultant share link, and I don't want us to spend too much time on a solution that would not work in conjunction with #4225.

Ah I see. Just to clarify the missing lessons in the resultant share link are for the partial timetables only right? So something similar to #4145?

zehata:
@thejus03 I am closing this as it's superseded by 4251. let's move any further discussion to #4251

PR #4251, opened 2025-10-31
refactor(optimiser): Enhance module slot merging and filtering logic

### LINKED ISSUE #4249: Optimiser Result Lesson Collision

thejus03:
## Context

Fixes #4249 
## Implementation
Explained in some detail in #4250

### DIFF
--- a/website/api/optimiser/_modules/modules.go
+++ b/website/api/optimiser/_modules/modules.go
@@ -2,6 +2,7 @@ package modules
 
 import (
 	"encoding/json"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


zehata:
Sorry, was busy with something else, let me try this out locally.

zehata:
@thejus03 Thanks for the PR, I have tried it locally, it fixed the issue relating to collision, and I have found that it has no conflict with #4225 and has also fixed the issue with some slots not being serialized, so it looks good to me :+1: 

cc: @leslieyip02

Issue #4256, opened 2025-11-19
Share/Sync Timetable QR Code is not centered


leslieyip02:
### Describe the bug

The QR Code generated for in the Share/Sync Timetable modal is not centered:

<img width="711" height="554" alt="Image" src="https://github.com/user-attachments/assets/6d316d60-b1b6-40e4-bb19-e67326c81382" />

### Expected behavior

It should be centered.

Harshith-Teja:
Hey @leslieyip02, can I work on this?

leslieyip02:
Sure, go ahead!

Harshith-Teja:
Hi @leslieyip02, thank you for this opportunity. I've submitted a [PR](https://github.com/nusmodifications/nusmods/pull/4257#issue-3641529212) for this

Harshith-Teja:
Hi @leslieyip02, can you please review this pr and merge it if it is alright?

leslieyip02:
Thanks for the fix! (and sorry for the delay). Also, as a tip, for future PRs you can use a [keyword to link an issue to the PR](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) which would auto-close the issue when the PR gets merged.

Closed by #4257.

Harshith-Teja:
No problem! Sure, will keep that in mind.

PR #4257, opened 2025-11-19
Fix QR code centering issue

Harshith-Teja:
## Context

This PR solves the issue #4256 of QR not centering

## Implementation

Updated the margin property from '0' to '0 auto' for the svg element inside the .qrCode class to center the QR code

image

## Other Information

I'm new to open source, and this is my first PR. I came across this repository as I was searching for 'good first issues'. I'm very thankful to the team for this opportunity!!

### DIFF
--- a/website/src/views/timetable/ShareTimetable.scss
+++ b/website/src/views/timetable/ShareTimetable.scss
@@ -60,6 +60,6 @@
 
   svg {
     display: block;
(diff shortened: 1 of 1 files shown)


leslieyip02 (approved):
LGTM. Thanks for the fix! (and sorry for the delay)

PR #4263, opened 2025-12-02
Improve dev experience and CI for optimiser

### LINKED ISSUE #4247: Improve development experience for optimiser
Currently testing optimiser for development requires manually changing the endpoint. Furthermore, there is also no testing workflow set up, requiring manual testing, and running the entire workflow even if we are just testing a small function.

thejus03:
## Context

- Fixes #4247

## Implementation

- Added a new global variable called `OPTIMISER_API_URL`
- Depending on the script that is run it assigns either `http://localhost:8020/optimise` or `api/optimiser/optimise`
- Added a new script called `yarn start:local` for running the localhost api endpoint on the website
- Updated the `yarn start` script to include the real endpoint for production
- Updated the optimiser README to reflect the above changes

## Updated on 21 Dec 2025
- Add linting and formatting using `golangci-lint` library.
- Update README for developers to lint and format themselves locally
- Update CI to run lint and formatting and run the tests

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -27,21 +27,20 @@ $ yarn
 This will install all of the dependencies you need.
 
 ### Development
-- To run the development build with optimiser, refer to the [optimiser README](./api/optimiser/README.md).
 
-- To run the development build without optimiser, simply run:
+To run the development build, simply run:
 
-  ```sh
-  $ yarn start
-  ```
+```sh
+$ yarn start
+```
 
-  This will start Webpack dev server, which will automatically rebuild and reload any code and components that you have changed.
+This will start Webpack dev server, which will automatically rebuild and reload any code and components that you have changed. If you need to use the optimiser, please refer to [the optimiser's README](./api/optimiser/README.md).
 
-  We recommend the following development tools to help speed up your work
+We recommend the following development tools to help speed up your work
 
-  - React Developer Tools ([Chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi), [Firefox](https://addons.mozilla.org/firefox/addon/react-devtools/))
-  - [Redux DevTools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd)
-  - [Firefox Developer Edition](https://www.mozilla.org/en-US/firefox/developer/)
+- React Developer Tools ([Chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi), [Firefox](https://addons.mozilla.org/firefox/addon/react-devtools/))
+- [Redux DevTools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd)
+- [Firefox Developer Edition](https://www.mozilla.org/en-US/firefox/developer/)
 
 ### Writing styles
 


.circleci/config.yml
```diff
@@ -110,6 +114,37 @@ jobs:
...
+              exit 1
+            fi
+      - run:
+          name: Run tests
+          command: |
+            go run ./_test/server/main.go -port 8020 &
+            SERVER_PID=$!
+            sleep 5
+            go test ./_test/. -v
+            kill $SERVER_PID
```
leslieyip02: 1. I'm not sure if `sleep 5` is guaranteed to work. I think we can check if the port is open before running the tests by doing something like this:

```
for i in {1..10}; do
  nc -z localhost 8020 && break
  sleep 1
done
```

2. Instead of `kill $SERVER_PID`, maybe we can do

```

go run ./_test/server/main.go -port 8020 &
SERVER_PID=$!

trap "kill $SERVER_PID" EXIT

for i in {1..10}; do
  nc -z localhost 8020 && break
  sleep 1
done

go test ./_test/. -v

```

This ensures the server is always killed.

website/api/optimiser/_solver/solver.go
```diff
@@ -269,41 +325,78 @@ func calculateLunchGap(physicalSlots []models.ModuleSlot, optimiserRequest model
...
 }
 
+// scoreConsecutiveHoursofStudy calculates a penalty for having too many consecutive hours
+// of classes without a break. It tracks blocks of back-to-back classes (where one class
+// ends exactly when the next begins) and penalizes any block exceeding maxConsecutiveHours.
+//
+// Algorithm:
+//  1. Accumulate duration of consecutive classes (no gap between them)
+//  2. When a gap is detected, check if accumulated time exceeds limit and penalize if so
+//  3. After the loop, check the final consecutive block
```
leslieyip02: Nitpick but step 3 isn't accurate since the final check isn't "after the loop".

website/api/optimiser/_test/api_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+func makeRequest(t *testing.T, req models.OptimiserRequest) (*http.Response, []byte) {
+	t.Helper()
+
+	body, err := json.Marshal(req)
+	if err != nil {
+		t.Fatalf("Failed to marshal request: %v", err)
+	}
+
+	client := &http.Client{Timeout: 60 * time.Second}
```
leslieyip02: Should we reuse the client instead of recreating one for each request?

website/api/optimiser/_test/api_test.go
```diff
@@ -0,0 +1,235 @@
...
+	resp, err := client.Post(baseURL, "application/json", bytes.NewReader(body))
+	if err != nil {
+		t.Fatalf("Failed to send request: %v", err)
+	}
+
+	respBody, err := io.ReadAll(resp.Body)
+	resp.Body.Close()
```
leslieyip02: I think it could be better to do

```
	defer resp.Body.Close()
	respBody, err := io.ReadAll(resp.Body)
```

so that the body is closed even if `io.ReadAll` panics.

leslieyip02:
@greptile

website/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "3.0.0-alpha",
   "description": "NUSMods v3",
   "scripts": {
-    "start": "cross-env NODE_ENV=development webpack serve --config webpack/webpack.config.dev.js",
+    "start": "cross-env NODE_ENV=development OPTIMISER_API_URL=api/optimiser/optimise webpack serve --config webpack/webpack.config.dev.js",
```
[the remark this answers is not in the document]
leslieyip02: Also, we may want to update the [`README`](https://github.com/nusmodifications/nusmods/blob/master/website/README.md) for `/website` since the optimiser doesn't work if we just run `yarn start`.

Or we could also hard code `https://nusmods.com/api/optimiser/optimise` as `OPTIMISER_API_URL` if we want it to work?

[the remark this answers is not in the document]
thejus03: Thats a good point. I think it makes more sense for all developers to use just one script which would now be `yarn start:local`. Meanwhile, `yarn start` would be used for production purposes. So I will update the README.

Besides that hardcoding could have random issues like between production version of optimiser and the one at latest.nusmods.com when they have not synced yet (which I am actually facing now while testing the latter approach)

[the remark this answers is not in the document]
leslieyip02: On second thought, I think leaving `yarn start` as the default start command is fine. If the user doesn't run `yarn start:optimiser`, then `yarn start:local` also wouldn't work. We can just add a note that if they want the optimiser to work then they should refer to its README (since not everyone needs the optimiser for local development).

leslieyip02 (approved):
LGTM!

PR #4264, opened 2025-12-03
fix(optimiser): Add default slots to unallocated lessons

### LINKED ISSUE #4145: Optimiser partial timetable does not show unallocated lessons

thejus03:
## Context

- Fixes #4145 
## Implementation

- Added default slots for unassigned lessons


leslieyip02 (commented):
The code is LGTM, but I'm not convinced that assigning a default slot is necessarily better than doing nothing. I feel that omitting unassigned modules makes it more explicit to the user that they need to tinker a bit with those modules (and/or adjust their preferences).

However, that's just my opinion so maybe a better way is to add an opt-in for this? So some sort of checkbox for "Assign default lessons if needed".

|Warning|Result Timetable|
|--|--|
|Screenshot 2025-12-14 at 5 23 58 PM|Screenshot 2025-12-14 at 5 24 08 PM|

thejus03:
Sure, I think adding an opt-in would make sense for some people. However, the reason why I would think most would want to have the default slot is because, let’s say, from the above example, CS2100 Lecture was unassigned. Now, if no default slots are added, he needs to go and remove the mod and re-add CS2100. Doing this, however, resets the other slots chosen for the other CS2100 lessons by the optimiser, like tutorial, lab, etc.

So I thought adding a default slot was better because now users can just modify the slot for CS2100 Lec without having the need to re-add and lose the other lessons chosen by the optimiser. 

Hope that makes sense. I am okay with anything. Maybe what we can do is put the default state for the opt-in as 'assign default slot' ?

jloh02:
alternatively, what if we do a flow where we show the warning, then have a **button** to auto-assign to default slots. i reckon this would not require a new Vercel function call, and allows for both behaviours. if we feel this is too convoluted, i'm more in favor of a warning due to the above reasons @leslieyip02 mentioned

website/api/optimiser/_modules/modules.go
```diff
@@ -15,18 +15,22 @@ import (
 - Get all module slots that pass conditions in optimiserRequest for all modules.
 - Reduces search space by merging slots of the same lesson type happening at the same day and time and building.
 */
-func GetAllModuleSlots(optimiserRequest models.OptimiserRequest) (map[string]map[string]map[string][]models.ModuleSlot, error) {
+func GetAllModuleSlots(optimiserRequest models.OptimiserRequest) (map[string]map[string]map[string][]models.ModuleSlot, map[string]map[string][]models.ModuleSlot, error) {
```
leslieyip02: Could we can add some type aliases for this? It's getting a bit hard to read.

Maybe something like

```
type ModuleSlotsMap = map[string]map[string][]models.ModuleSlot
```

leslieyip02 (approved):

PR #4272, opened 2025-12-17
Patch for #4225, with additional checks to prevent overwriting user config if error encountered and allow recovery for malformed configs

zehata:
This is an extension to #4271, details of which I have copied over.
@leslieyip02 @jloh02 Patch for #4225, for the issue reported on the morning of 16 Dec by LH (https://t.me/NUSMods/13136)

## Post-mortem
1. When the app fetches the timetable modules, it populates the lesson indices which allows the migration of the module config to the new schema
https://github.com/nusmodifications/nusmods/blob/d31261e060e45bcd79b07d251f0ffe2403905dee/website/src/reducers/moduleBank.ts#L47-L60

2. The issue was caused by
https://github.com/nusmodifications/nusmods/blob/d31261e060e45bcd79b07d251f0ffe2403905dee/website/src/views/AppShell.tsx#L64-L66
calling
https://github.com/nusmodifications/nusmods/blob/d31261e060e45bcd79b07d251f0ffe2403905dee/website/src/actions/timetables.ts#L231-L239

When `validateTimetable(1)` is called when the app successfully fetches the module data for semester 1, it calls `migrateTimetableConfigs` which migrates the configs for _all_ semesters, including semester 2. Since the request for semester 2 has not returned, the lesson indices has not been populated for semester 2 (see point 1). Trying to migrate the config for semester 2 will thus fail.

## Fix
This patch fixes this issue by removing the `migrateTimetableConfigs` function and instead migrates each semester's timetable config as data is fetched for each of the semester.

It adds checks to prevent overwriting the user's config if any error occurs during the migration process.

It also adds code to create a random config to overwrite malformed configs caused by the migration failure.

### DIFF
--- a/website/src/utils/timetables.ts
+++ b/website/src/utils/timetables.ts
@@ -1194,7 +1194,6 @@ export function migrateModuleLessonConfig(
 }
 
 /**
- * A helper function for migrateTimetableConfigs\
  * Migrates a semester's timetable config
  * @param semTimetableConfig the semester timetable config to migrate
  * @param taModulesConfig the TA lesson configs overrides the semester timetable config


zehata:
Also, while doing testing, I realized that the error occurred before writing to state, so users' configs were actually not overwritten. Nonetheless, I think it would make more sense to add checks to explicitly ensure that users' configs are not overwritten when errors are encountered.

website/src/actions/timetables.ts
```diff
@@ -229,29 +221,47 @@ export function setTimetable(
...
+        ? getModuleTimetable(moduleBank.modules[moduleCode], semester)
+        : [];
+
+    const {
+      migratedSemTimetableConfig: timetable,
+      migratedTaModulesConfig: ta,
+      alreadyMigrated,
+    } = await Promise.resolve(
+      migrateSemTimetableConfig(semTimetableConfig, taModulesConfig, getModuleSemesterTimetable),
+    );
```
leslieyip02, line 244: Is the `Promise.resolve` needed? I don't think `migrateSemTimetableConfig` is async.

zehata: No, I am using it to end validation early if an error occurs during migration

zehata: Of course there are other ways of achieving the same effect

Issue #3926, opened 2025-01-02
Make testing the export service locally less painful


zwliew:

zehata:
Currently,
1. The docker instructions are outdated and uses the docker compose v1 API
2. The example `.env` does not work correctly if only configured with https://github.com/nusmodifications/nusmods/blob/master/export/README.md?plain=1#L10-L11 because it is missing the `CHROME_EXECUTABLE`, `ACADEMIC_YEAR` and `PAGE` environment variables.
3. There is no reference to the http://localhost:8081/timetable-only/index.html for setting the `PAGE` environment variable
4. The base url for development https://github.com/nusmodifications/nusmods/blob/master/website/src/apis/export.ts#L13-L14 should be http://localhost:3000 instead

Since sunsetting docker (#3098), the docker configuration have not been updated, but I have managed to update it to a working config.

While #3098 proposes to fully remove docker from the app, I think it has some merits in the convenience of being able to spin up the export service for development, without having to manually setup the env vars and spinning up each of the 3 services (website, timetable-only and export).

I would like to propose that we keep docker for development convenience, especially since optimiser is being introduced and that would add another service that needs to be started. However, since most of the time we aren't testing both the export service and optimiser, it's probably not necessary.

PR #4277, opened 2025-12-30
Fix export service error with the new timetable config schema

zehata:
Updated the validation logic in the export service

## Context

Resolves https://t.me/NUSMods/13264
This was caused by the validation logic in https://github.com/nusmodifications/nusmods/blob/f7227585ad3301adcae7ad700ea0237716c31fb3/export/src/data.ts#L64-L71 not updated for the updated in timetable config schema.

The systemic problem is that there is currently no test coverage for the export service. It also took a long time to fix because the [readme](https://github.com/nusmodifications/nusmods/tree/f7227585ad3301adcae7ad700ea0237716c31fb3/export) is really outdated.

## Other Information

Sample image and pdf output of the export service, showing no regression from before #4225:
[My Timetable.pdf](https://github.com/user-attachments/files/24380987/My.Timetable.pdf)
My Timetable

### DIFF
--- a/website/src/views/timetable/TimetableContainer.test.tsx
+++ b/website/src/views/timetable/TimetableContainer.test.tsx
@@ -125,6 +125,20 @@ describe(TimetableContainerComponent, () => {
     await expectRedirectToHomepageFrom('/timetable/2017-2018/v1');
   });
 
+  test('should display blank timetable if share string to import is empty', async () => {
+    make('/timetable/sem-1/share?');
+
+    // Expect no spinner when share string is empty
+    expect(screen.queryByText(/Loading/)).not.toBeInTheDocument();
+
+    // Expect import header to be present
+    expect(await screen.findByRole('button', { name: 'Import' })).toBeInTheDocument();
+
+    // Expect page to load
+    expect(screen.getByText(/Semester 1/)).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


leslieyip02:
Yeah the export service is quite old and the testing issue was raised in #3926. We should probably do something about it.

leslieyip02 (approved):
LGTM

leslieyip02:
For future reference, we missed 3 issues:

1. `.positive()` expects a lesson index greater than 0, so it flags lesson indices of 0 as invalid
2. `.ordered()` will restrict the lesson indices array's length, so we should use `.items()` instead
3. Default value of the `ta` key in the export state should be `[]` instead of `{}`

Fixed by these commits: (this is pretty sloppy but I just wanted to get it fixed)

- https://github.com/nusmodifications/nusmods/commit/e58d5c5c958294cf930a656b45e8ab7e72aec411
- https://github.com/nusmodifications/nusmods/commit/0eef0ba19259337c2dffd3a5f6e4ca46048acaed
- https://github.com/nusmodifications/nusmods/commit/1091598b05a3b0e6507d68dae20538a279893ef2
- https://github.com/nusmodifications/nusmods/commit/9c69174c8213b838e598e035a7b4ac415b02e928

PR #4282, opened 2026-01-02
Attempt recovery of TA module config if migration fails

zehata:
## Context

When the migration script fails, because a module was removed by the faculty, the migration script refuses to overwrite old v1 config because it will cause data loss.

However, the user is left in an unrecoverable state because the TA reducer can only access a migrated TA config.

## Implementation

This implementation **_will_ cause data loss**. If the user requests a module to be removed, the reducer will delete the old TA config and only keeps the module codes of the TA modules. This will allow the user to recover to a usable state, but **all TA configs for that semester will be deleted**.

## Other Information

There are several other ways that a recovery can be attempted
1. This implementation is the simplest method, and will be less riskiest because it is the least involved. However, it **necessitates that the TA lesson configs be removed** for recovery to take place.
2. We allow the reducer to produce old TA configs, which will prevent data from non-affected modules from being overwritten. This **incurs a lot of technical debt** from various downstream changes and guards that have to be added, and we will have to revert this behavior in the future for maintainability.
3. We allow the migration script to recover the state for other non-affected modules, but this **runs the risk of deleting module data invisibly when it encounters an unrelated issue**, for example, a network error.
4. We make no changes, and **have the user clear data manually to recover to a blank slate, which is the most destructive**, but depending on how widespread the issue is, this might be the least risky and it doesn't introduce further changes.

I would like to ask for advice on what is the most prudent course of action.

### DIFF
--- a/export/src/data.ts
+++ b/export/src/data.ts
@@ -78,7 +78,7 @@ export function validateExportData(data: ExportData) {
    */
   const timetableSchema = Joi.object().pattern(Joi.string(), moduleLessonConfigSchema);
 
-  const taModulesConfigSchema = Joi.array().ordered(Joi.string());
+  const taModulesConfigSchema = Joi.array().items(Joi.string());
   const themeSchema = Joi.object({
     id: Joi.string(),
     timetableOrientation: Joi.string().valid('HORIZONTAL', 'VERTICAL'),
--- a/website/src/utils/timetables.test.ts
+++ b/website/src/utils/timetables.test.ts
@@ -1,5 +1,5 @@
 import NUSModerator from 'nusmoderator';
-import _, { get } from 'lodash';
+import _, { filter, find, get, map, mapValues, some } from 'lodash';
 import { parseISO } from 'date-fns';
 import {
   TaModulesConfigV1,
@@ -11,8 +11,11 @@ import {
   TimetableDayArrangement,
   TimetableDayFormat,
   ModuleLessonConfigV1,
+  LessonWithIndex,
+  InteractableLesson,
 } from 'types/timetables';
 import {
+  LessonIndex,
   LessonType,
   ModuleCode,
   RawLesson,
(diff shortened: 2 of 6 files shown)


website/src/reducers/timetables.ts
```diff
@@ -208,7 +208,16 @@ function semHiddenModules(state = DEFAULT_HIDDEN_STATE, action: Actions) {
 
 // Map of semester to list of TA modules
 const DEFAULT_TA_STATE: ModuleCode[] = [];
-function semTaModules(state = DEFAULT_TA_STATE, action: Actions): ModuleCode[] {
+function semTaModules(
+  state: ModuleCode[] | TaModulesConfigV1 = DEFAULT_TA_STATE,
+  action: Actions,
+): ModuleCode[] {
+  // Should TA module migration fail, attempt to recover to a usable state
```
zehata, line 215: This is a destructive action, because it only takes the module codes of the old TA config, and we lose the lesson configs.

leslieyip02: While not ideal, I think that's preferable to not being able to recover at all.

website/src/reducers/timetables.ts
```diff
@@ -208,7 +208,16 @@ function semHiddenModules(state = DEFAULT_HIDDEN_STATE, action: Actions) {
...
+  if (!isArray(state)) {
+    const recoveryAttemptState = keys(state);
+    return semTaModules(recoveryAttemptState, action);
```
leslieyip02, line 218: I think this is ok, just that maybe we can add a notification to tell the user that the migration may have dropped some of the lessons.

Not sure what the message should be though. Maybe "Lessons have been modified. Please review your timetable."?

zehata: Given that migration is mostly complete, I think introducing this change might be technical debt. I would rather close this.

zehata: Given that migration is mostly complete, I think introducing this change might be technical debt. I would rather close this.

leslieyip02: That makes sense.

PR #4284, opened 2026-01-03
Fix export service validation for multiple TA modules

zehata:
## Context
As reported by Michael: https://t.me/NUSMods/13410

## Implementation
In my previous fix #4277 , I confused `ordered` with `items` and the array validation logic fails when the number of TA modules is more than 1.


zehata:
This is similar to the issue that was fixed by https://github.com/nusmodifications/nusmods/commit/0eef0ba19259337c2dffd3a5f6e4ca46048acaed

leslieyip02 (approved):
LGTM!

PR #4298, opened 2026-02-10
fix(build): copy venues.json in pre-commit hook instead of before build

ravern:
## Summary
- Vercel compiles Go serverless functions **before** running `installCommand` or `buildCommand`, so the `prepare:optimiser` script (which copies `venues.json` into `api/optimiser/_constants/`) never runs in time for `//go:embed`
- Fix: commit `venues.json` directly to git so it's present at clone time when Go compiles
- Add a husky pre-commit hook to automatically keep the committed copy in sync with `src/data/venues.json`
- Remove the `installCommand` and Go `includeFiles` from `vercel.json` (neither works for Go functions on Vercel)

## Test plan
- [x] Verify Vercel preview build succeeds on this PR
- [x] Verify optimiser API still works on the preview deployment
- [x] Verify pre-commit hook copies venues.json on commit


jloh02:
Does this mean our repo now needs to maintain 2 versions of venues.json? is there no better way for this?

ravern:
Yes, unfortunately for Vercel Go serverless, I couldn't find a way to insert the copy command before the Go gets compiled.

jloh02 (approved):
Approving this for now as it is a blocker and created side issue #4299

PR #4310, opened 2026-03-01
fix(scraper): recognize old-format NUS API attribute values

ravern:
## Summary

- Add old-format uppercase values (`YES`, `NO`, `HT`) to `truthyValues`/`falsyValues` so boolean attributes from modules still using the old API format are correctly parsed
- Add old-format short codes (`S1`, `S2`, `S1&S2`) to `mpeValueMap` for MPE semester matching
- Add `YES` to `sfsTruthyValues` for SkillsFuture Series detection
- Add tests for `mapAttributes` covering both old and new format values

The NUS API migrated its `CourseAttributes` value format (e.g. `YES`→`Yes`, `HT`→`HT - Honours Thesis/Rsh Project`, `S1`→`S1 - Sem 1`) but the migration is incomplete — roughly half of modules still return old-format values, causing ~16,470 attribute entries to be silently dropped.

## Test plan

- [x] Add unit tests for `mapAttributes` covering old-format and new-format values
- [x] Verify scraper logs no longer emit warnings for unrecognized attribute values like `YES`, `NO`, `HT`, `S1`, `S2`, `S1&S2`
- [x] Confirm attribute counts increase after the fix (previously ~16,470 entries were dropped)

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
@@ -1,10 +1,13 @@
 import { RawLesson } from '../types/modules';
+import { ModuleAttributeEntry } from '../types/api';
+import { Logger } from '../services/logger';
 import departments from './fixtures/departments.json';
 import faculties from './fixtures/faculties.json';
 import {
   cleanModuleInfo,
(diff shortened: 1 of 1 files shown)


ravern:
@greptile review

PR #4311, opened 2026-03-01
fix(scraper): exclude propagated timetables from alias detection

ravern:
## Summary

The scraper's data pipeline has two relevant stages:

1. **`GetSemesterData`** — fetches module info, timetables, and exams. For "shadow" modules that lack their own timetable (e.g. dual-coded courses where only one code has lesson data in the API), `findEquivalentModules()` propagates timetable data from a matching module (matched by title, credits, and description). This is correct and necessary for timetable display.

2. **`CollateVenues`** — processes all module timetables to build venue availability maps and detect co-listed (aliased) modules by finding lessons that occupy the same venue and timeslot.

The bug: propagated timetables are *copies* of another module's lessons. When `CollateVenues` processes both the original and the copy, it "discovers" that they share the same venue+timeslot — and incorrectly flags them as aliases. This produces ~3,110 false alias pairs (e.g. GESS1000T, a SCALE Academy course, incorrectly cross-listed with GES1002/GESS1000).

### Fix

- Add a `timetablePropagated?: boolean` flag to `WritableSemesterModuleData`
- Set it to `true` in `GetSemesterData` when timetable data is propagated from another module
- Skip modules with `timetablePropagated` in `CollateVenues` so they don't participate in alias detection

Timetable propagation itself is unchanged — the propagated data still flows through to the frontend for display. Only alias detection is affected.

## Test plan

- [x] All 138 existing tests pass across 18 test suites

### DIFF
--- a/scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
+++ b/scrapers/nus-v2/src/tasks/GetSemesterData.test.ts
@@ -1,10 +1,13 @@
 import { RawLesson } from '../types/modules';
+import { ModuleAttributeEntry } from '../types/api';
+import { Logger } from '../services/logger';
 import departments from './fixtures/departments.json';
 import faculties from './fixtures/faculties.json';
 import {
   cleanModuleInfo,
   getDepartmentCodeMap,
   getFacultyCodeMap,
+  mapAttributes,
   parseWorkload,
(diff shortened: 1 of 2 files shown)


ravern:
@greptile review

PR #4313, opened 2026-03-02
fix(scrapers): filter out inactive faculties and departments

ravern:
## Summary
- The new API returns all faculties and departments regardless of their active status, unlike the old API which accepted an `eff_status: 'A'` parameter for server-side filtering
- Added client-side filtering by `EffectiveStatus === 'A'` for both faculties and departments after fetching (and after the 099 fallback), before the `cleanFacultyDepartment` mapping

## Test plan
- [x] Verify that inactive faculties/departments are excluded from the output
- [x] Verify that the manually-added 099 faculty (which has `EffectiveStatus: 'A'`) is retained
- [x] Verify that `facultyDepartments.json` output only contains active entries


ravern:
This is actually a no-op in terms of data produced, since it looks like there are no inactive faculties/departments.

So the intention here is to just preserve functionality from previous version of scraper.

jloh02 (approved):
Thanks Claude (and Ravern)!

PR #4322, opened 2026-03-04
chore: shift oxfmt to root of the repo

yangshun:
## Summary

- Consolidates the identical `.oxfmtrc.json` config from all 6 sub-packages into a single root-level `.oxfmtrc.json`
- Moves the `oxfmt` devDependency from each package to the root `package.json`
- Adds root-level `format` and `format:check` scripts

## Test plan

- [x] Run `pnpm install` to update lockfile
- [x] Verify `oxfmt` runs correctly from each sub-package (it will resolve the config from the repo root)

### DIFF
--- a/scrapers/cpex-scraper/src/index.ts
+++ b/scrapers/cpex-scraper/src/index.ts
@@ -50,15 +50,15 @@ function getTimestampForFilename(): string {
 /** Strip HTML tags, decode common HTML entities, remove NBSPs, collapse whitespace. */
 function cleanString(s: string): string {
   return s
-    .replace(/<[^>]*>?/gm, ' ')        // strip HTML tags
-    .replace(/&nbsp;/gi, ' ')           // decode &nbsp;
-    .replace(/&amp;/gi, '&')            // decode &amp;
-    .replace(/&lt;/gi, '<')             // decode &lt;
-    .replace(/&gt;/gi, '>')             // decode &gt;
-    .replace(/&quot;/gi, '"')           // decode &quot;
-    .replace(/&#39;/gi, "'")            // decode &#39;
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -62,7 +62,6 @@ jobs:
           at: ~/nusmods
       - run: corepack enable && corepack prepare pnpm@10.30.3 --activate
       - run: pnpm install --frozen-lockfile
-      - run: pnpm format:check
```
yangshun, line 65: I don't think CI should check for formatting. 

Instead, we should add formatting as a precommit hook. It's so fast (few ms for changed files, few seconds for the entire repo) it won't be much of a delay.

package.json
```diff
@@ -5,6 +5,13 @@
   "engines": {
     "node": ">=22"
   },
+  "scripts": {
```
yangshun, line 4: This is the only file that contains the real changes that need to be reviewed. The rest are formatting changes resulting from running oxfmt.

.circleci/config.yml
ravern: Do you think we can add the pre-commit hook in this PR as well? Any reason not to 🤔

ravern:
Will re-review once merge conflicts + comments are addressed!

.circleci/config.yml
yangshun: In the first place CI isn't the best place to be catching formatting issues, which can be super wasteful for large repos (https://news.ycombinator.com/item?id=38248119):

I didn't add it in this PR since:

1. It's not crucial — it's ok for the repo to temporarily be in a state that doesn't have a precommit formatting step
2. This PR already contains many changes, modifying precommit hooks will make it even harder to review

Followed up in #4327

PR #4334, opened 2026-03-06
fix(website): increase async util timeout to fix flaky tests

yangshun:
## Summary

- `ModulePageContainer` and `ModuleArchiveContainer` tests intermittently fail when run via `pnpm -r test` but always pass when run standalone with `pnpm test` in `website/`
- Root cause: all three workspace packages (`nusmoderator`, `nus-v2`, `website`) run Vitest simultaneously, causing CPU/memory contention. The two tests rely on dynamic `import()` + mocked axios responses resolving before Testing Library's default 1000ms `findByText` timeout, which isn't enough under load
- Fix: increase `asyncUtilTimeout` to 5000ms globally in `vitest-setup.ts`

## Test plan

- [x] `pnpm -r test` passes consistently from repo root (previously failed ~30% of runs)
- [x] `pnpm test` in `website/` still passes

### DIFF
--- a/export/.env.example
+++ b/export/.env.example
@@ -1,6 +1,6 @@
 # For explanation about these values, refer to config.js
-# ACADEMIC_YEAR=2020-2021
-PAGE=https://nusmods.vercel.app/timetable-only
+ACADEMIC_YEAR=2025-2026
+PAGE=https://nusmods.com/timetable-only
 SENTRY_DSN=https://37215374a3fd42a198c7b144c8886335@sentry.io/265295
 # CHROME_EXECUTABLE=
 # MODULE_DATA=
--- a/export/README.md
+++ b/export/README.md
@@ -73,6 +73,13 @@ Download PNG image of the timetable.
 
 Returns the HTML content of the page that Puppeteer renders.
 
+## Testing
+
+After running locally, you can test with the following URLs on `localhost:3000`:
+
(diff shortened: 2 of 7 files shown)


yangshun:
@ravern can you have a look at this? Tests are now flaky and sometimes fails on resource-constrained envs 😭

leslieyip02:
@yangshun I ran into more flaky tests caused by `downshift` while running CI. It's trying to access `document` after the tests end:

Screenshot 2026-03-08 at 3 53 00 PM

I saw the fake timer fix in `website/src/views/timetable/ModulesSelect.test.tsx`, but I think ensuring the DOM is properly reset between tests could be a more permanent fix (it's also easier than tracking down every component that uses `downshift`).

What do you think?

yangshun:
> @yangshun I ran into more flaky tests caused by `downshift` trying to access `document` after the tests end:
> 
> Screenshot 2026-03-08 at 3 53 00 PM
> 
> I saw the fake timer fix in `website/src/views/timetable/ModulesSelect.test.tsx`, but I think ensuring the DOM is properly reset between tests could be a more permanent fix (it's also easier than tracking down every component that uses `downshift`).
> 
> What do you think?

I haven't been able to reproduce this well. I'll work on fixing it properly.

yangshun:
@leslieyip02 Oh I didn't realize you pushed to this branch. Yeah I think it's good, let's ship this. Thanks so much!

yangshun:
Let's hold off on merging first. I tried running the test 50 times to detect flakiness and saw other problems.

If you're interested, here's the command I added to `package.json`

```json
"test:flaky": "node -e \"const { spawnSync } = require('node:child_process'); const command = process.platform === 'win32' ? 'pnpm.cmd' : 'pnpm'; for (let seed = 1; seed <= 50; seed += 1) { console.log('\\n[flaky] seed=' + seed); const result = spawnSync(command, ['exec', 'vitest', 'run', '--coverage', '--sequence.shuffle', '--sequence.seed=' + seed], { stdio: 'inherit', env: { ...process.env, CI: '1' } }); if (result.status !== 0) process.exit(result.status ?? 1); }\""
 ```

yangshun:
Ok fixed, I made Codex run `test:flaky` (50 runs) and fix the tests until the command succeeded. All 50 test runs were successful on the 3rd attempt.

leslieyip02:
It seems the tests for components using downshift still fail from time to time, so I've moved the original fake timer fix into a util and added it to the affected tests.

I also migrated the `ModulesSelect` (MPE) tests to React Testing Library as it previously used a timeout hack to wait for the component to paint, and this was incompatible with the fake timers.

yangshun:
@leslieyip02 Thank you! Let's merge this.

Issue #4339, opened 2026-03-06
Optimiser - No longer filter out slots without venue information


thejus03:
# Context
Currently the optimiser removes all slots without venue information and works with the remaining slots so that we can also optimise timetables using venue locations.

However, there are still many lessons with missing venue locations causing many users to receive a partial timetable. 

# Suggested Solution
Venue optimisation is an additional feature that is nice to have, but it is not strictly necessary. Therefore
we can avoid filtering out slots without venue information. Instead during the venue optimisation phase, we can simply ignore those slots.

Ideally this would reduce the number of unnecessary partial timetables (due to missing venues) generated but at the cost of some venue optimisation.

PR #4340, opened 2026-03-06
refactor(optimiser): include slots with unknown venue locations and several code quality improvements

thejus03:
## Context
#4339

### Logic Changes
- Stop filtering out module slots that lack venue coordinates, so classes in venues not listed in venues.json are still included in timetable optimisation
- Add NO_VENUE_PENALTY = 100.0 constant — scores worse than the furthest known on-campus distance (~80), discouraging unknown venues without making them unviable
- Update beam width from 2500 -> 5000 for deeper searches.
### Bug Fixes
- **Unsafe type assertions on `Weeks` field** (`_modules/modules.go`, `_solver/solver.go`): Added comma-ok checks before `week.(float64)` assertions to prevent panics on malformed NUSMods API data
- **HTTP status check in `GetModuleData`** (`_client/client.go`): Non-200 responses from the NUSMods API now return a descriptive error instead of silently passing bad data downstream
- **Raw error message leaked to client** (`optimise.go`): JSON decode errors now return a generic `"Invalid request format"` message instead of exposing internal error details

### Input Validation
- **Empty modules guard** (`_models/models.go`): `ParseOptimiserRequestFields` now returns an error if the modules list is empty
- **JSON tag hygiene** (`_models/models.go`): Parsed/computed fields (`EarliestMin`, `LatestMin`, etc.) annotated with `json:"-"` to exclude them from JSON decode/encode

### Code Quality
- **Magic numbers replaced with named constants** (`_constants/constants.go`): `BeamWidth = 2500`, `BranchingFactor = 100`, `DaysPerWeek = 6` — all usages in `_solver/solver.go` updated
- **Variable shadowing fixed** (`_solver/solver.go`): Inner loop variable `i` renamed to `slotIdx` to avoid shadowing the outer loop index
- **`serializeLessonIndices` simplified** (`_solver/nusmods_link.go`): Replaced a convoluted `fmt.Sprint` + `strings.Fields` + `strings.Trim` pipeline with a straightforward `strconv.Itoa` loop; also adds `strconv` import
- **`Handler` doc comment fixed** (`optimise.go`): Blank line between comment blocks caused `revive` to report a malformed doc comment — merged into a single contiguous block
- **Pre-existing style fixes** from earlier in the branch: snake_case variables, function name casing (`scoreConsecutiveHoursOfStudy`), block-to-line comment conversion, duplicate map construction removed, recordings format standardised to `MODULE|LessonType` pipe separator

### Tests
- `TestOptimiser_EmptyModules` — validates the new empty-modules guard returns a non-200 response
- `TestOptimiser_InvalidTimeFormat` — validates malformed time strings are rejected
- `TestOptimiser_NonExistentModule` — validates unknown module codes are rejected (exercises the HTTP status check in `GetModuleData`)
- `TestOptimiser_MethodNotAllowed` — validates GET requests return 405
- `TestOptimiser_ShareableLinks` — verifies both shareable links are well-formed URLs containing all requested module codes
- `TestOptimiser_AllSlotsHaveAssignments` — internal consistency check that every slot in `DaySlots` has a corresponding entry in `Assignments`

### Documentation
- **README** comprehensively updated:
  - Explains `_` directory prefix convention (Vercel serverless requirement)
  - End-to-end data flow walkthrough (handler → module fetch → beam search → link generation)
  - Hard vs soft constraints clearly distinguished, with guidance on where to add new constraints
  - MRV heuristic documented in algorithm section
  - All scoring constants documented with rationale and priority order
  - Response fields table (`shareableLink` vs `defaultShareableLink` explained)
  - Venue data maintenance instructions (`venues.json` embed, deduplication logic)

## Test plan
- [ ] Run existing optimiser tests: start test server with `pnpm start:optimiser`, then `go test ./_test/... -v`
- [ ] Verify modules with unknown venues appear in optimiser results
- [ ] Confirm timetables with known-venue slots are still preferred over unknown-venue ones
- [ ] Test error cases: empty modules, bad time format, non-existent module — all should return non-200


website/api/optimiser/_constants/constants.go
jloh02: constants in go should be using the `const` keyword. in addition, go style guide recommends constants to be in lowerCamelCase for non-exported, and UpperCamelCase for exported constants. In this case, UpperCamelCase is fine

website/api/optimiser/_modules/modules.go
```diff
@@ -122,15 +122,7 @@ func mergeAndFilterModuleSlots(
...
-		if !constants.E_Venues[slot.Venue] {
-			venueLocation := venues[slot.Venue].Location
-			if venueLocation.X == 0 && venueLocation.Y == 0 {
-				continue
-			}
-		}
-
-		// Add coordinates to slot
+		// Add coordinates to slot (will be (0,0) for venues not in venues.json)
 		slot.Coordinates = venues[slot.Venue].Location
```
jloh02: go idiomatics recommend using `value, ok := myMap[key]` to evaluate if the value is not in the given map. We should then assign an error of sorts rather than (0,0) as that is still "valid" to some extent

website/api/optimiser/_constants/constants.go
```diff
@@ -40,6 +40,7 @@ var ModulesURL = "https://api.nusmods.com/v2/%s/modules/%s.json"
 var NUSModsTimetableBaseURL = "https://nusmods.com/timetable"
 
 var MAX_WALK_DISTANCE = 0.250 // 250 meters
+var NO_VENUE_PENALTY = 100.0
```
jloh02: we should document what this penalty exactly means, along with the NO_LUNCH_PENALTY and GAP_PENALTY_RATE. is the value simply arbitrary?

website/api/optimiser/_solver/solver.go
```diff
@@ -176,14 +176,22 @@ func calculateDayDistanceScore(daySlots []models.ModuleSlot, recordings map[stri
...
-			curr.Coordinates.X == 0 || curr.Coordinates.Y == 0 ||
-			isLessonRecorded(prev.LessonKey, recordings) ||
+		// Skip if either lesson is recorded
+		if isLessonRecorded(prev.LessonKey, recordings) ||
 			isLessonRecorded(curr.LessonKey, recordings) {
 			continue
 		}
 
+		prevNoCoords := prev.Coordinates.X == 0 || prev.Coordinates.Y == 0
+		currNoCoords := curr.Coordinates.X == 0 || curr.Coordinates.Y == 0
```
jloh02: as per previous comment, this should be:
1. abstracted to a isInvalidCoordinate method
2. remove the idea of (0,0) being the "invalid coordinate"

website/api/optimiser/_constants/constants.go
jloh02: What's the significance of all of these being `true`? Most practices in go for a set would use a struct{} set as the "true" value has no real significance. And then use the `_, exist := set[key]` to assess if value is in set

website/api/optimiser/_constants/constants.go
thejus03: Should we document this in the code or in the README? or both?

jloh02: in code constants make sense to me so we don't need to update both

thejus03:
@jloh02 Since we have decided to refactor the code now, I have made several changes in the code and abstracted it out. I also made few performance improvements due to unnecessary copies that were being made. I have updated the PR description to match the changes.

Beside including slots without venue locations (the original PR) there has been no other functional changes.

jloh02:
@greptileai

website/api/optimiser/_modules/modules.go
```diff
@@ -67,13 +79,17 @@ func GetAllModuleSlots(
...
-				moduleTimetable[i].WeeksSet[weekInt] = true
+				weekFloat, ok := week.(float64)
+				if !ok {
+					continue
+				}
+				weekInt := int(weekFloat)
+				moduleTimetable[i].WeeksSet[weekInt] = struct{}{}
 				weeksStrings[j] = strconv.Itoa(weekInt)
 			}
 			moduleTimetable[i].WeeksString = strings.Join(weeksStrings, ",")
```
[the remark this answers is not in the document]
thejus03: nice

website/api/optimiser/_constants/constants.go
```diff
@@ -35,17 +37,31 @@ var LessonTypeAbbrev = map[string]string{
...
+	ConsecutiveHoursPenaltyRate = 100
+)
 
 // This is used by [nusmods_link.SerializeLessonIndices] to serialize the result of optimiser into a timetable share link to return to the client
-var MODULE_CODE_SEPARATOR = ";"
+const ModuleCodeSeparator = ";"
+
+// Beam search parameters
+const (
+	BeamWidth       = 5000
```
thejus03, line 61: @jloh02 I doubled the beam width from 2500  to 5000 now for deeper and more accurate searches. This is the only other logical change in the code besides adding back the slots with no venue info.

thejus03:
@jloh02 wondering if the vercel preview for the website still works? I wanted to see if there were any performance improvements for optimising large module count (10 mods) timetables from the one in prod.

jloh02:
> @jloh02 wondering if the vercel preview for the website still works? I wanted to see if there were any performance improvements for optimising large module count (10 mods) timetables from the one in prod.

@thejus03 mybad i broke it partially in #4350 #4328. will try to push a fix out in the next 1-2 days

website/api/optimiser/_constants/constants.go
jloh02: Sorry I'm not super well versed. Does this potentially lead to longer runtimes?

thejus03:
My bad I should have expanded on that. Yes it does increase runtime because we are searching through double the amount of timetables as before. But overall I think latency should have decreased from the one in prod now because I reduced latency by more than half from before. So we can afford to search deeper?

We can test out how fast it is compared to prod using the vercel preview link.

Also the reason why I doubled it is because for one of my friend's 10 mod timetable it was not returning a completely optimised timetable (sometimes) because we didn't search deep enough. But after modifying, it was able to find it every time.

image


### Prod
image

### Current PR
image

thejus03:
> sorry facing a financial crunch a lil

Just confirming but we are using Vercel's free tier right?

jloh02:
> > sorry facing a financial crunch a lil
> 
> Just confirming but we are using Vercel's free tier right?

Unfortunately not, we have moved out of it, but it seems like our vercel function invocations are still fine for now will allow this change and monitor

Issue #4314, opened 2026-03-02
NUSMods v4


yangshun:
This is an umbrella issue for the next generation of NUSMods — NUSMods v4. 

I'm inspired to start this due to advancements in AI such that we can use ~vibe coding~ agentic engineering to upgrade the current iteration of NUSMods, which is almost ten years old as of this year.

Conventional wisdom tells you software rewrites never succeed. There are good reasons to that but with AI, things have changed and it's a good time to challenge that advice and test the limits of what technology in this age.

## Goals

- Modernize tech stack
- Improve code base structure with better modularization
- Pay off tech debt
- Improve UX
- Pave the way for new features

## Plan

That said, big projects always pose risk, and there's a non-zero chance that this initiative falls through. That's fine, at least we tried. To mitigate risk, we'll break down this project into a few phases and have clear steps and milestones. Even if we don't complete the project, the work done in the process will still improve maintainability of the project and benefit future generations.

We'll start with the obvious, easier changes that nobody will disagree with, such as upgrading non-user facing aspects such as tooling. Then slowly move up the stack and improve the product, changing the UX and adding/killing features.

### Phase 1 – Tooling upgrades

Since we want to use AI to do the heavy-lifting, we need to make things easier for AI to do its thing. The first step is to upgrade tooling so that AI can easily verify correctness (somewhat). Tooling improvements are also easier for humans to verify. Development experiemce improvements that good for humans tend to also be good for AI.

Issue: https://github.com/nusmodifications/nusmods/issues/4319

### Phase 2 — Documentation, Modules & abstractions

We eventually want to rewrite the user interface, which means both versions of the website will co-exist for a bit (potentially for a long time depending on how long the rewrite takes).

Issue: https://github.com/nusmodifications/nusmods/issues/4318

#### Documentation

Before we do that, we (I) should be familiar with what exists, and that's where documentation comes in helpful. AI makes documentation trivial, so we can firstly generate some engineering docs, feature list, etc. for the current repo.

#### Modules & abstractions

Next, it'll be good to extract out the business logic (stuff are independent of UI and doesn't change often) into standalone packages/modules (not referring to the timetable ones), so that both current and new interfaces can use them.

The `nusmoderator` package was a first step (albeit a thin one), but the idea is there. I suspect some parts of the timetable reducer can also be extracted out. While we do this, we can also add more tests and design good module boundaries / APIs.

[...]

Issue #4290, opened 2026-01-14
Exporting calendar as .png leads to 504 error


somebodyelsedidthis:
### Describe the bug

Exporting my calendar led to nusmods.com timing out and leading to a 504 GATEWAY_TIMEOUT code "FUNCTION_INVOCATION_TIMEOUT", id "sin1::mf8qd-1768390909781-20d7020ef866"

### To Reproduce

1. Click on 'Download > Image (.png)'
2. Wait, then see error

### Expected behavior

Calendar to download as a .png file

### Screenshots

<img width="1280" height="719" alt="Image" src="https://github.com/user-attachments/assets/2ebb0f72-0a05-4fbc-a8f9-e04a38dd4ea7" />

### Desktop (please complete the following information):

- OS: Windows 11 Pro 25H2
- Chrome
- Version 143.0.7499.193

leslieyip02:
Hi, what does your timetable config look like? Could you share a link?

From the looks of it, the export service timed out (i.e. took more than 15s).
Previously, we had increased the timeout from the default of 10s to 15s in #4255.

@jloh02 Perhaps we could just increase this timeout again?

PR #4343, opened 2026-03-07
refactor(export): migrate from puppeteer to satori

darentanrw:
## Context
The current export service uses Puppeteer (via `puppeteer-core` and `@sparticuz/chromium`) to render timetable images and PDFs. 

This approach has several drawbacks, the biggest of which being slow performance due to cold starts and chromium load time. This has resulted in issues such as 504 Errors (#4290) where function time outs, and fixes to increase timeout to 15 seconds such as (#4255)

## Implementation
Instead, I propose to replace Puppeteer with [Satori](https://github.com/vercel/satori) + [resvg-js](https://github.com/thx/resvg-js) for server-side timetable rendering:

- Satori converts React JSX → SVG entirely in Node.js
- resvg-js converts SVG → PNG
- PDFKit composites the rendered PNG into a properly formatted A4 PDF

ExportData plus modules are parsed into buildRenderableTimetable, which produces a RenderableTimetable view model. 

That model is fed into TimetableImage in view.tsx, which is then rendered from JSX → SVG → PNG → PDF

### Drawbacks and Limitations 
Because Satori does not run a real browser, It only accepts JSX elements that are pure and stateless, and a subset of layout and styling (inline styles, no CSS classes, no full DOM). The NUSMods timetable uses SCSS, class-based styling, so it can’t be rendered directly by Satori. `view.tsx` hence is a separate React tree that mirrors the timetable layout using only Satori-compatible inline styles (thanks Claude).

However: 
1. This is not currently pixel-perfect with the website (scroll to bottom to see UI changes). In particular, the font is not the same as NUSMods currently uses system font?
2. There is duplicated logic and data: 
- `export/src/types.ts`- ExportData, Module, SemTimetableConfig, etc. (from `website/src/types/export.ts`, `types/timetables.ts`, `types/modules.ts`)
- `export/src/satori/theme.ts` - Theme color palettes (from `website/src/styles/constants.scss $nusmods-theme-colors`)
- `export/src/satori/render-model.ts` — Time conversion, lesson arrangement, week formatting, color mapping (from `website/src/utils/timify.ts`, `utils/timetables.ts`, `utils/colors.ts`)

This has potential future painful implication - as any change to the timetable UI (layout, themes, week formatting, etc.) must be reflected in both the website components and the export view.tsx / render-model.ts, increasing the chance of inconsistency.

Note: This was also done from my assumption of that you are deploying your monorepo separately onto different vercel instances - there are ways we can reduce this data duplication with packages and etc, but it will require a more significant rewrite?

Future work: Should v4 migration (#4314) be implemented and Next.js is chosen, this feature also can be implemented via [Next.js ImageResponse](https://nextjs.org/docs/app/api-reference/functions/image-response) as well.

## Performance Improvements
Performance improvements of 60~70% 🥳! 

Wrote a shell script, at `export/utils/endpoint-performance-test.sh` which ran 10 simulated timetables against both the old (Puppeteer) and new (Satori) endpoints. I deployed the new export service at https://nusmods-export.vercel.app, and latency was measured on an M3 MacBook Pro over Ethernet under typical usage. 

CleanShot 2026-03-08 at 00 56 20@2x
Full log here: <a href="https://github.com/user-attachments/files/25816185/080326-1255-performance-test-output.txt">080326-1255-performance-test-output.txt</a>

For more precise numbers, we should inspect Vercel function logs to compare cold starts and execution time between the two implementations.

## Visual Differences
<table>
<tr>
  <th> New Export (Satori)
  <th> Old Export (Puppeteer)
<tr>
<tr>
 <td> 1-new-image
 <td> 1-old-image
<tr>
 <td> 2-new-image
 <td> 2-old-image
<tr>
 <td> 4-new-image <br> <a href="https://github.com/user-attachments/files/25815965/4-new-pdf.pdf">PDF Version</a>
 <td> 4-old-image <br>
<a href="https://github.com/user-attachments/files/25815967/4-old-pdf.pdf">PDF Version</a>
<tr>
 <td> 8-new-image <br>
<a href="https://github.com/user-attachments/files/25815971/8-new-pdf.pdf">PDF Version</a>
 <td> 8-old-image
<a href="https://github.com/user-attachments/files/25815974/8-old-pdf.pdf.error.txt">PDF Failed to Render (Error Here)</a>
</table>

## Misc
First PR to nusmods, any feedback will be appreciated! 

I found many edge cases, but I'm sure that there are more that I have yet to consider, so would love your inputs and edits on this PR.

Thanks a lot! Especially to @ravern and @leslieyip02 for planting this idea in my head.


export/package.json
```diff
@@ -10,17 +10,19 @@
     "start": "pm2 start ecosystem.config.js",
     "build": "tsc",
     "nodemon": "nodemon --no-update-notifier -r dotenv/config ./build/src/index.js",
+    "test": "tsx --test src/**/*.test.ts",
```
yangshun: Use Vitest

export/package.json
```diff
@@ -10,17 +10,19 @@
...
     "watch": "tsc --watch",
     "dev": "run-p nodemon watch",
-    "devtools": "cross-env DEVTOOLS=1 pnpm dev",
+    "check": "run-s lint typecheck",
```
yangshun, line 17: Change it to `run-s lint typecheck test`

yangshun:
> Note: This was also done from my assumption of that you are deploying your monorepo separately onto different vercel instances - there are ways we can reduce this data duplication with packages and etc, but it will require a more significant rewrite?

Sounds like a great idea. I'll be starting on Phase 2 of #4314 soon and we can tackle this extraction.

export/src/types.ts
```diff
@@ -1,9 +1,8 @@
-import type { Page } from 'puppeteer-core';
+// These types are duplicated from `website/src/types`.
```
yangshun, line 3: Nice one. You could also add a comment there so that anyone changing there will know to change here as well (this will not be necessary after we extract into a core package)

export/package.json
```diff
@@ -31,8 +33,10 @@
     "koa-views": "6.3.1",
     "lodash": "4.17.23",
     "nodemon": "2.0.22",
+    "pdfkit": "^0.17.2",
     "pug": "3.0.3",
-    "puppeteer-core": "22.15.0"
+    "react": "18.3.1",
```
yangshun, line 41: Why not use latest React?

export/package.json
```diff
@@ -43,15 +47,17 @@
     "@types/koa-views": "2.0.4",
     "@types/lodash": "4.17.14",
     "@types/node": "22.13.5",
+    "@types/pdfkit": "^0.17.5",
     "@types/pug": "2.0.10",
+    "@types/react": "18.3.18",
```
yangshun, line 56: Ditto

export/package.json
```diff
@@ -43,15 +47,17 @@
...
     "@vercel/node": "1.15.4",
-    "cross-env": "7.0.3",
     "dotenv": "8.6.0",
     "eslint-plugin-no-only-tests": "^3.3.0",
     "eslint-plugin-perfectionist": "^5.6.0",
     "eslint-plugin-unused-imports": "^4.4.1",
     "npm-run-all": "4.1.5",
     "oxlint": "^1.51.0",
+    "tsx": "^4.21.0",
```
yangshun: Prefer Vitest since the entire repo is moving to Vitest

export/package.json
yangshun: Add test step to `.circle/config.yml` for `export`

yangshun:
Some thoughts:

1. The current export service acts like a general screenshotting service, which is in some sense better because it doesn't need to contain any rendering or layout logic. The export service will work regardless of how the website looks, which may or may not be desired. 
2. If we're going to share the data model between `website` and `export`, we will need to ensure that either data model changes are backwards-compatible, or both apps are deployed at the same time. The current `export` already is coupled to the data model, but not as much as the new way.

Lastly, can we also explore if https://www.renoun.dev/screenshot works? I think it runs on the client and maybe we can use it for images and avoid a server round-trip.

I'm torn between the export approaches. Decoupling the rendering has good benefits too, so I guess the question is how much the visual consistency between website and export matters.

darentanrw:
Thanks for the review! I've fixed the low hanging fruits that you commented on.

I will take a further look into [https://www.renoun.dev/screenshot](https://www.renoun.dev/screenshot), however a potential problem that I foresee is that it might not be easy to select only the timetable and module info as a PNG export. 

The current workflow with Puppeteer screenshots [https://nusmods.vercel.app/timetable-only](https://nusmods.vercel.app/timetable-only) after injecting it with the timetable data. However, what is currently rendered to the user also includes the `TimetableActions` and `ModuleSelectContainer` components. Will need to figure out a way to exclude them from the local screenshot, or to redirect to timetable-only, then screenshot, which is not an ideal user experience.

In this case of using Serverless functions for export, I believe that a cleaner approach to couple the data models is by shifting the export serverless handlers into the main website APIs (to `website/api/export/*`). 

Personally, I don't think that the visual consistency between website and export matters that much (See Visual Differences in original PR), and I think that we can make it more similar (except for fonts) - but ultimately this is a decision for the maintainers to make.

yangshun:
> Thanks for the review! I've fixed the low hanging fruits that you commented on.
> 
> I will take a further look into https://www.renoun.dev/screenshot, however a potential problem that I foresee is that it might not be easy to select only the timetable and module info as a PNG export.
> 
> The current workflow with Puppeteer screenshots https://nusmods.vercel.app/timetable-only after injecting it with the timetable data. However, what is currently rendered to the user also includes the `TimetableActions` and `ModuleSelectContainer` components. Will need to figure out a way to exclude them from the local screenshot, or to redirect to timetable-only, then screenshot, which is not an ideal user experience.
> 
> In this case of using Serverless functions for export, I believe that a cleaner approach to couple the data models is by shifting the export serverless handlers into the main website APIs (to `website/api/export/*`).
> 
> Personally, I don't think that the visual consistency between website and export matters that much (See Visual Differences in original PR), and I think that we can make it more similar (except for fonts) - but ultimately this is a decision for the maintainers to make.

Those are great points! I'm convinced this is a good approach. Like you said, we can also co-locate the exporting functions into the web app, which is a good idea for the future.

Btw I've started extracting the types in https://github.com/nusmodifications/nusmods/pull/4346, depending on whose PR merges first, someone will have to resolve the conflicts, but it shouldn't be hard.

jloh02:
> Personally, I don't think that the visual consistency between website and export matters that much (See Visual Differences in original PR), and I think that we can make it more similar (except for fonts) - but ultimately this is a decision for the maintainers to make.

Agree on this. I think this is fine. Ultimately, my biggest barrier is still the configuration/duplicated logic issues which have been lightly addressed for now.

> In this case of using Serverless functions for export, I believe that a cleaner approach to couple the data models is by shifting the export serverless handlers into the main website APIs (to website/api/export/*).

I am agreeable to this since technically having 2 separate Vercel projects doesn't yield much benefit if that can ultimately reduce duplicated code. @ravern @leslieyip02 can weigh in on this too.

ravern:
> Btw I've started extracting the types in https://github.com/nusmodifications/nusmods/pull/4346, depending on whose PR merges first, someone will have to resolve the conflicts, but it shouldn't be hard.

Since we've stopped the effort to extract/share types, let's not include that into the scope of this PR.

I also personally don't think that the visual differences don't matter too much.

I do like the effort here in bringing in a new approach to exporting timetables. And I do think there are pros and cons, so why not we merge this in to sit _alongside_ the current export behind a **beta**, and then see what users think?

Export is pure-function-ish and we can just have both current and beta endpoints without much hassle.

I can make the changes to support both.

ravern:
@greptileai review

ravern:
@greptileai review

PR #4346, opened 2026-03-08
refactor(types): share export types with export package

yangshun:
## Summary
- add shared export-facing types to @nusmods/types
- switch export/ to import ExportData from the shared package
- point website/src/types/export.ts at the same shared source of truth

## Verification
- pnpm --dir packages/types typecheck
- pnpm --dir export typecheck
- pnpm --dir website typecheck


ravern:
@yangshun did u mean to commit FRONTEND_REWRITE_PLAN.md into this PR?

If u did, maybe we should re-scope this PR to include adding this plan? Not sure if we just want to dump the plan into the repo root either.

yangshun:
> @yangshun did u mean to commit FRONTEND_REWRITE_PLAN.md into this PR?
> 
> If u did, maybe we should re-scope this PR to include adding this plan? Not sure if we just want to dump the plan into the repo root either.

Yes, so that we can just update the plan and track the progress in the Markdown file. But I understand that seem noisy.

Since we're not going to start on Phase 2 anytime soon I'll just close this out first.

PR #4355, opened 2026-03-09
fix(export): align local API routes and config cleanup

yangshun:
## Summary
- align the local export server routes with the deployed `/api/export/*` paths used by the website client and docs
- add a simple `export/public/test.html` page for manually exercising local image and PDF exports
- clean up export config/docs by removing the stale `CHROME_EXECUTABLE` config entry, simplifying `.env.example`, and updating the local export comments
- remove the unused PM2-specific deploy path by deleting `export/ecosystem.config.js`, dropping the `deploy` script, and making `pnpm start` run the built server directly

## Test plan
- `pnpm --dir export typecheck`
- `pnpm --dir export build`

### DIFF
--- a/export/src/app.ts
+++ b/export/src/app.ts
@@ -1,4 +1,4 @@
-import path from 'path';
+import path from 'node:path';
 import Koa from 'koa';
 import Router from 'koa-router';
 import serve from 'koa-static';


export/public/test.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
```
yangshun, line 1: This file will make it easier to test the deployment, just go to `/test.html`.

I'll look into adding some unit/E2E tests, but Vercel has a different deployment flow so this is still needed for testing on Vercel.

export/src/app.ts
```diff
@@ -15,7 +15,7 @@ const app = new Koa<State>();
 const router = new Router();
 
 router
-  .get('/image', async (ctx) => {
+  .get('/api/export/image', async (ctx) => {
```
yangshun, line 20: Align the routes on Koa vs Vercel.

export/ecosystem.config.js
yangshun, line 1: I think this file is no longer used.

ravern (approved):
LGTM

PR #4357, opened 2026-03-10
fix(website): use dynamic import for lodash-es in serverless function

ravern:
## Summary
- Inline a small typed `mapValues` helper in `nus-auth.ts` to replace the `lodash` import, avoiding the `ERR_REQUIRE_ESM` error on Vercel serverless (Node runs CJS by default and can't `require()` the ESM-only `lodash-es`)
- Switch `config/index.ts` from `lodash` to `lodash-es` since it's bundled by webpack

## Test plan
- [x] `pnpm --dir website typecheck` passes
- [x] Verify Vercel serverless auth endpoints work without `ERR_REQUIRE_ESM`

🤖 Generated with [Claude Code](https://claude.ai/code)


website/src/serverless/nus-auth.ts
```diff
@@ -99,6 +98,7 @@ export const authenticate = async (req: Request) => {
     extract: { attributes },
   } = await sp.parseLoginResponse(idp, 'post', requestToProcess);
 
+  const { mapValues } = await import('lodash-es');
```
yangshun, line 105: I think it'd be helpful to leave an explanation as to why this has to be done, it's super unobvious.

A comment at the top explaining that these files don't work with ESM is gonna be helpful too.

website/src/config/index.ts
```diff
@@ -1,4 +1,4 @@
-import { mapValues } from 'lodash';
+import { mapValues } from 'lodash-es';
```
yangshun, line 1: `mpe.ts` loads this file and `mpe.ts` also runs on Vercel functions.

```
import config from '../config';
```

Is it gonna be an issue?

I think better test this on staging. Jon should have reverted this part for a reason.

ravern:
https://nusmods-export-git-fix-serverless-lodash-esm-modsbots-projects.vercel.app/test.html

I'm able to download Image and PDF.

ravern:
I think I'm trolling - the reversion was fixing CPEx endpoints, not export. I will merge this once CPEx is over.

website/src/config/index.ts
jloh02: Yep this was exactly the reason. We should test this on preview environment as far as possible. As well as on staging

PR #4358, opened 2026-03-11
fix(build): Allow fork PR builds on vercel

jloh02:
The initial logic did not account for "forks". It made it such that the vercel builds run on forks, instead of including PRs from forks


leslieyip02:
#4359

Oops it's still broken:

image

PR #4360, opened 2026-03-11
fix(build): install jq on forks in vercel build test

jloh02:


jloh02:
image
image


@leslieyip02 for reference i logged the request at the start of the file. please verify this is fine. will remove these lines now

leslieyip02:
I've verified the script works locally, and the log seems correct, so I think the only issue was just `jq` not being installed.

ravern (approved):
LGTM

PR #4361, opened 2026-03-11
fix(build): fork detection without jq

leslieyip02:
- Compared to #4360, this implementation doesn't require a `jq` install each time the script is run
- Add `GITHUB_API_TOKEN` so we don't get rate-limited (the rate limit for unauthenticated requests is 60 per hour)


leslieyip02:
Probably too hacky.

PR #4363, opened 2026-03-12
chore(deps): use pnpm catalog for shared devDependencies

yangshun:
## Motivation

Make it easier to update package, we just have to update on the root level. I've only migrated the tooling packages as they are less risky and don't affect the build.

## Summary
- Add `catalog:` section to `pnpm-workspace.yaml` with 12 shared devDependencies (typescript, vitest, oxlint, eslint plugins, cross-env, npm-run-all, @types/node)
- Replace hardcoded versions with `catalog:` in all 5 workspace packages (website, export, nusmoderator, cpex-scraper, nus-v2)
- Runtime dependencies and version-coupled type packages (`@types/lodash`, `@types/fs-extra`, `@vercel/node`) are intentionally left in their respective packages

## Test plan
- [x] `pnpm install` succeeds without errors
- [x] `pnpm -r check` succeeds without errors
- [x] CI passes


yangshun:
Have regenerated the lockfile.

PR #4366, opened 2026-03-13
fix(scraper): fix build output path and yargs compatibility with pnpm

ravern:
## Summary
- Exclude `vitest.config.ts` from `tsconfig.prod.json` to fix build output path (`build/src/index.js` → `build/index.js`). With the file included, TypeScript inferred `rootDir` as the project root instead of `src/`.
- Upgrade yargs from v16 to v17 since `parseAsync()` is only available in v17. Previously resolved via yarn's hoisting, but pnpm's stricter resolution exposed the mismatch.

## Test plan
- [x] Run `pnpm build` in `scrapers/nus-v2` and verify output is at `build/index.js`
- [x] Run `pnpm scrape` and verify the scraper starts without `parseAsync` errors

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -277,8 +277,8 @@ importers:
         specifier: 1.0.0
         version: 1.0.0
       yargs:
-        specifier: 16.2.0
-        version: 16.2.0
+        specifier: 17.7.2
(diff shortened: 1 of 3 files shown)


jloh02 (approved):
lgtm thanks

PR #4372, opened 2026-03-21
refactor(optimiser): code quality improvements

thejus03:
## Summary
**No logic changes**
- Decouples `Solve` from `http.ResponseWriter` — solver now returns `(SolveResponse, error)` and the HTTP handler owns response writing. Introduces `SolveError` to carry HTTP status codes back to the handler without coupling the solver to `net/http`
- Renames `GenerateNUSModsShareableLink` to `FillDefaultsAndGenerateShareableLinks` to make the in-place mutation of `assignments` and `lessonToSlots` explicit
- Adds a 10-second timeout to the HTTP client to prevent hung upstream requests from consuming the entire serverless budget
- Replaces repeated `Weeks.([]any)` type assertions and `float64` casts in `hasConflict` with direct `WeeksSet` map iteration — cleaner and avoids redundant work in the beam search hot path
- Fixes week-unaware collision check in `validateTimetable` — two slots overlapping in time but on different weeks are not a conflict; test now mirrors the solver's `WeeksSet` intersection logic
- Changes `Coordinates` fields from `float32` to `float64`, removing the lossy widening cast when passing to the haversine function
- Fixes incorrect `DayDistance` comment ("Squared" → accurate description)
- Replaces two `strings.Split` calls on the same key with a single `strings.SplitN`


jloh02:
@greptileai

leslieyip02:
@greptileai

leslieyip02 (approved):
LGTM

PR #4374, opened 2026-03-24
fix: do not allow builds for prs

howenyap:
## Context

Build environment secrets can be leaked via malicious code on PRs

## Implementation

Update [`vercel_ignore_build.sh`](https://github.com/nusmodifications/nusmods/blob/master/vercel_ignore_build.sh) to not build for PRs by default

## Other Information

Aligned with @ravern


jloh02:
@howenyap @ravern  cmiiw but when we deploy with changes to the .sh, vercel warns us and let's us know of changes? how else might these secrets be leaked? if we ignore these by default, then we can merge PRs from forks without any preview which is unintended?

image

@nusmodifications/nusmods-developers maintainers please lmk your thoughts too, but if we merge this means CI passes without any previous by default

howenyap:
> @howenyap @ravern cmiiw but when we deploy with changes to the .sh, vercel warns us and let's us know of changes? how else might these secrets be leaked? if we ignore these by default, then we can merge PRs from forks without any preview which is unintended?

cmiiw but secrets can be leaked via changes in any code, not just in `.sh`

there is also the second gate where you have to authorise vercel for the build to actually happen, but the concern was that it could be a large PR and the review might've missed something

but at the end of the day you guys are the maintainers so I'll just leave it to you to decide, I raised this to ravern and sent a patch just for disclosure

jloh02:
hmm that's a good point and it's true. that's a natural drawback of the preview environment. will discuss with the other maintainers on this. thanks for bringing it up!

jloh02:
I am personally against disabling for PRs

- if u can leak secrets through PRs, it means you can leak secrets through production too
- it is technically our responsiblity to ensure this doesn't happen in either env
- it is better we leak a preview env secret than a prod one

PR #4376, opened 2026-03-24
feat(cpex-scraper): add file logger for better debugging and tracing

faizkhairi:
## Summary
- Add a lightweight file logger (`src/logger.ts`) that writes timestamped lines to both stdout and a log file
- Log files are written to `scrapers/cpex-scraper/logs/` with format `cpex-YYYY-MM-DD.HH-mm-ss.log`
- The logs directory is created automatically at runtime (already covered by root `.gitignore`)
- Update `src/index.ts` to use the file logger instead of the default `console`
- Zero new dependencies -- uses only Node.js built-in `fs` module
- Satisfies the existing `Pick<Console, 'log'>` interface, so `scraper.ts` is unchanged

## Related Issue
Closes #4371

## Test Plan
- [ ] Run `pnpm --filter nusmods-cpex-scraper build` and verify compilation succeeds
- [ ] Run the scraper and verify log file is created in `logs/` directory
- [ ] Verify console output still shows timestamped log lines
- [ ] Verify existing unit tests still pass (`pnpm --filter nusmods-cpex-scraper test`)


faizkhairi:
**Note on testing**: I was unable to run the full CI suite locally (`pnpm run ci`) as I used a clone. Happy to add unit tests for `logger.ts` if desired, the main cases would be:

1. `log()` writes timestamped lines to both stdout and the file stream
2. `close()` resolves after the stream is fully flushed
3. Constructor creates the logs directory if it doesn't exist
4. Stream error handler prevents unhandled crashes

The existing `scraper.test.ts` tests are unaffected since `FileLogger` is only instantiated in `index.ts`, not in the testable `scraper.ts`.

Let me know if you'd like me to add these tests.

jloh02:
@greptile

PR #4388, opened 2026-04-17
Refactor timetable utils into smaller files

zehata:
## Context
- The timetable utils file was originally 1400+ locs while its tests is another 1400. Refactoring them into smaller logical files makes it significantly easier to read, writes and test
- I tried re-exporting symbols from utils/timetables but realized that it causes circular references which causes ReferenceError: can't access lexical declaration X before initialization

## Implementation
I split up `utils/timetables` into what I consider are logical blocks
- exams
- interactabilityHydration
- lessonHydration
- lessonsArrangement
- migration
- shareLinks
- validation

## Other Information
This is part of the effort to fix the lesson drifting issue described at #4387 
I think refactoring is necessary and is better done before other PRs are pushed out

Since this is just splitting up code, and all the tests are green, I don't think this is a high risk PR.


zehata:
I don't know why I hadn't thought of it, but I just had to split up the offending circular reference into its own file.
I also realized it's probably a good idea to rename `timetable.ts` to `index.ts` and nest it within `timetable`

PR #4407, opened 2026-05-23
fix(website): regressions from cross-tab state sync

### LINKED ISSUE #4383: Non-clonable Actions are Broken
### Describe the bug

Actions with payloads containing functions are broken by https://github.com/nusmodifications/nusmods/pull/4174 because

https://github.com/nusmodifications/nusmods/blob/7b7d91cba015bce2c6e0206c807747a6a07b7198/website/src/middlewares/state-sync-middleware.ts#L14-L18

is an unreliable check for clonability. Children that are not clonable will also result in [`DataCloneError`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone)

In fact, this is a meaningless check, since `Action`s are necessarily objects

https://github.com/reduxjs/redux/blob/aaa04ae8402ba2caba55a9c75bfa8d3df6c78f8c/src/types/actions.ts#L41
```ts
export interface AnyAction extends Action {
  // Allows any extra properties to be defined in an action.
  [extraProps: string]: any
}
```

### To Reproduce

Use the `openNotification` action with the action handler being a function, the most prominent example being

https://github.com/nusmodifications/nusmods/blob/7b7d91cba015bce2c6e0206c807747a6a07b7198/website/src/views/AppShell.tsx#L69-L76

TheMythologist:
## Context

Follow-up to #4174, which introduced `redux-state-sync` for cross-tab state synchronization. Three regressions surfaced after that landed - each rooted in the fact that `redux-state-sync` broadcasts actions via `BroadcastChannel.postMessage`, which uses `structuredClone` and (a) refuses to copy certain values, and (b) silently swallows any `DataCloneError` it raises.

1. **Notifications were leaking across tabs.** `OPEN_NOTIFICATION` payloads can contain `action.handler` / `willClose` function references; `structuredClone` throws on functions, and the broadcast was lost. Even if it weren't, notifications are inherently per-tab UI (one tab's "Undo" toast shouldn't appear in another). Closes #4383.

2. **Undo no longer worked across tabs.** `undoHistory` is a non-persisted slice built up by listening to live actions. `redux-state-sync` never syncs state at tab-open time, so a tab opened after another already had history would start with empty `undoHistory.past` - and a synced `UNDO` action would silently bail in `computeUndoStacks`, leaving the two tabs' timetables diverged.

3. **Request lifecycle actions carried non-cloneable values.** `*_FAILURE` actions wrap `AxiosError` (which holds an `XMLHttpRequest` reference); `*_SUCCESS` carries `AxiosHeaders` and the full `AxiosRequestConfig`. `postMessage` either threw outright or degraded shapes silently - observer tabs got stuck "loading" states on every API failure.

## Implementation

**1. Blacklist `OPEN_NOTIFICATION`.** Added to the state-sync predicate so notifications stay local to the originating tab.

**2. Sync the full store on tab open via `initStateWithPrevTab`.**

- Wrap the root reducer with `withReduxStateSync` so it intercepts `RECEIVE_INIT_STATE` and adopts the peer tab's state -including `undoHistory`, so cross-tab undo behaves correctly.
- Call `initStateWithPrevTab(store)` after `persistStore` in `configure-store`.
- The broadcast goes through `structuredClone`, so add a `prepareState` that strips `app.notifications` (function handlers) and `requests` (transient + may contain `AxiosError`). A matching `receiveState` keeps the adopting tab's own in-flight requests / notification queue intact rather than the (empty) stripped versions.

**3. Make request lifecycle actions cloneable.** Each problem field is handled based on whether it's needed downstream:

| Field | Treatment | Why |
|---|---|---|
| `*_FAILURE` actions | Dropped from broadcast via `meta.requestStatus === FAILURE` predicate | `AxiosError.request` is an `XMLHttpRequest`, blocked by `structuredClone`. Originating tab still applies it locally; the `requests` slice is already per-tab (also stripped in `prepareState`). |
| `meta.request` on SUCCESS/FAILURE | Dropped at source | Re-attached the full `AxiosRequestConfig` for no consumer. Wasn't in the declared `SuccessAction` / `FailureAction` types either — removing aligns runtime with the contract. |
| `meta.responseHeaders` on SUCCESS | Normalized to plain object via `{ ...response.headers }` | Read by `moduleBank` for `last-modified`. Axios 1.x returns an `AxiosHeaders` class instance whose prototype `structuredClone` strips — normalizing at dispatch guarantees identical shape on sender and receiver. |
| `payload` on REQUEST | Narrowed to `{ method, url, headers, params, data }` | Full `AxiosRequestConfig` permits function fields (`transformRequest`, `paramsSerializer`, etc.). No reducer reads this payload — the safe subset is sufficient and keeps the action FSA-shaped. |

Updated the test mock for `redux-state-sync` to stub the additional exports (`withReduxStateSync`, `initStateWithPrevTab`, `RECEIVE_INIT_STATE`).

## Other Information

The three commits are deliberately kept granular so each fix can be reviewed (or reverted) on its own.


ravern:
@greptileai review

website/src/middlewares/requests-middleware.ts
```diff
@@ -74,10 +74,23 @@ const requestMiddleware: Middleware<RequestsDispatchExt, State, Dispatch> =
...
+    // `transformRequest` and `paramsSerializer`; if a caller ever adds one,
+    // structuredClone would throw silently in state-sync. None of the request
+    // status reducers read this payload, so a narrowed shape is sufficient.
+    const safePayload: AxiosRequestConfig = {
+      method: payload.method,
+      url: payload.url,
+      headers: payload.headers,
+      params: payload.params,
+      data: payload.data,
+    };
```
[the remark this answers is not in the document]
ravern: Agree this is worth fixing. It is low impact today because the REQUEST payload is not currently read, but spreading headers keeps sender/receiver shapes consistent and avoids a future AxiosHeaders footgun.

website/src/__mocks__/redux-state-sync.ts
```diff
@@ -1,6 +1,15 @@
...
+import type { Middleware, Reducer, Store } from 'redux';
 
-// eslint-disable-next-line import/prefer-default-export
 export function createStateSyncMiddleware(): Middleware {
   return () => (next) => (action) => next(action);
 }
+
+export function withReduxStateSync<T extends Reducer>(reducer: T): T {
+  return reducer;
+}
```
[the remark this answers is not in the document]
ravern: Agree with the signature gap, but this suggested mock still drops receiveState. To actually cover the new behavior, the mock should call receiveState when it receives RECEIVE_INIT_STATE.

website/src/middlewares/state-sync-middleware.ts
```diff
@@ -9,18 +37,28 @@ const reduxStateSyncConfig = {
   channel: 'redux_state_sync',
   predicate: (action: AnyAction) => {
     // Reference: https://github.com/aohua/redux-state-sync/issues/53
-    const blacklist = [PERSIST, PURGE, REHYDRATE];
+    const blacklist = [PERSIST, PURGE, REHYDRATE, OPEN_NOTIFICATION, RECEIVE_INIT_STATE];
```
ravern: If notifications are per-tab, `POP_NOTIFICATION` should be blacklisted too. Otherwise one tab's timeout/close can pop a different local notification in another tab.

ravern (commented):
A few state-sync concerns. Typecheck passes locally (`pnpm --dir website typecheck`).

website/src/middlewares/state-sync-middleware.ts
```diff
@@ -9,18 +37,28 @@ const reduxStateSyncConfig = {
...
     // objects that are clonable by `structuredClone`
     if (typeof action === 'function') {
       return false;
     }
 
+    // *_FAILURE actions carry an AxiosError as payload, which holds a
+    // reference to the underlying XMLHttpRequest -- structuredClone throws
+    // DataCloneError on those. Action types are dynamic (e.g.
+    // `FETCH_MODULE_LIST_FAILURE`), so discriminate on the lifecycle meta.
+    if (action.meta?.requestStatus === FAILURE) {
```
ravern: Dropping FAILURE while still syncing REQUEST can still leave observer tabs stuck loading: they receive `*_REQUEST` but never the matching `*_FAILURE`. If requests are per-tab, REQUEST should probably be filtered too.

website/src/bootstrapping/configure-store.ts
```diff
@@ -58,5 +58,10 @@ export default function configureStore(defaultState?: State) {
   }
 
   const persistor = persistStore(store);
+
+  // Ask any already-open tab to hand over its current state (including
+  // non-persisted slices like undoHistory) so cross-tab undo stays in sync.
+  initStateWithPrevTab(store);
```
ravern: `persistStore` rehydrates asynchronously, so this can import peer state before local `REHYDRATE` later overwrites persisted slices. That can pair peer `undoHistory` with local timetable/theme state; consider calling this from the persist callback after bootstrapping.

ravern:
@greptileai review

website/src/bootstrapping/configure-store.ts
ravern: The undo-history fix does address the original issue (new tabs previously started with an empty `undoHistory`, so a later synced `UNDO` could diverge). The remaining problem is just ordering: `persistStore(store)` kicks off async `REHYDRATE`, but `initStateWithPrevTab(store)` currently runs immediately after it. That means a new tab can adopt the peer tab’s `undoHistory` first, then later have persisted slices like `timetables` / `theme` overwritten by local rehydration, leaving the peer undo stack paired with local current state.

I think the minimal fix is to keep the author’s approach, but move the init sync into the redux-persist bootstrapped callback:

```ts
const persistor = persistStore(store, undefined, () => {
  initStateWithPrevTab(store);
});
```

That preserves the startup undo sync, but waits until local persisted slices have settled before importing peer state.

ravern (changes requested):
One last comment: https://github.com/nusmodifications/nusmods/pull/4407#discussion_r3410734550

ravern:
Verified the latest fix locally.

Checks run:
- `pnpm --dir website typecheck`
- targeted configure-store-adjacent Vitest files: 7 files / 29 tests passed
- `pnpm --dir website test`: 103 files / 669 tests passed
- `pnpm --dir website test:integration`: 103 files / 669 tests passed
- `pnpm --dir website lint:code`: 0 warnings / 0 errors
- `pnpm --dir website build`: compiled successfully

Manual Chrome E2E on `localhost:8080`:
1. Added `CS1010S` in tab 1.
2. Opened tab 2 and confirmed it booted with `CS1010S` / 4 Units.
3. Removed `CS1010S` in tab 1.
4. Opened tab 3 after the removal and confirmed it booted with no courses.
5. Clicked Undo in tab 1.
6. Confirmed tabs 1, 2, and 3 all restored `CS1010S` / 4 Units.

That covers the important startup-sync path: a fresh tab opened after an undoable timetable change still receives enough peer `undoHistory` to apply the later synced Undo correctly. No state-sync / clone / rehydration errors showed up in browser logs. Only caveat: local shell was on Node v24.7.0 while the project asks for Node 22.x, so pnpm emitted engine warnings, but all checks passed.

ravern (approved):
Approved. The remaining rehydration/undo-history ordering issue has been addressed, automated checks are passing, and I manually verified the cross-tab undo startup flow in Chrome.

website/src/middlewares/state-sync-middleware.ts
```diff
@@ -9,18 +37,42 @@ const reduxStateSyncConfig = {
...
+      OPEN_NOTIFICATION,
+      POP_NOTIFICATION,
+      RECEIVE_INIT_STATE,
+    ];
 
     // redux-state-sync relies on BroadcastChannel, which only supports
     // objects that are clonable by `structuredClone`
     if (typeof action === 'function') {
       return false;
     }
```
zehata, line 54: @ravern @TheMythologist  I believe all actions are objects and never a function?

PR #4418, opened 2026-06-10
Display relevant weeks on exam timetable

### LINKED ISSUE #1717: Exams on Saturday should show full week

If exams fall on Saturday, the exam view does not show Sundays.
e.g.: http://modsn.us/QbDtf

Exam views should show Sundays on larger screens if possible.

Alternative designs for mobile view should be considered.

### Additional context

photo6192635155231910127

photo6190264625342294204

not-a-scam:
## Context
Initially just created to close #1717

Additionally updated exam timetable to only display the week when exams to be taken are happening. Previously, the first week will always be displayed even if there are no exams to be taken.
timetable with extra weeks

### Full Week Display
timetable with full week displayed

- Check if there is a Saturday exam and display full week if there is
- Change cell width to be equal across days with the addition of Sunday

### Display relevant weeks
1. Filter out examinable mods from visible mods 
2. Map visible and examinable mods to their exam dates
3. Reduce by comparison to get first and last exam dates
4. Normalise result to always be a Monday 
5. Calculate number of weeks for display
 
timetable with only relevant weeks

## Other Information

First PR here! Feedback is appreciated!

- Created it so that if there are no exams, the getExamCalendar would return the current date as the first date instead for simplicity, but not sure if there are better return value? Considered using the first day of exams from the academic calendar but I thought the additional call would be unnecessary since the value isn't used.

- I think the addition of Sunday looks fine on mobile. Seemed to be a concern in #1717 but included a gif for reference
<p align="center">
  mobile view of timetable
</p>


leslieyip02 (approved):
LGTM 👍 

I've pushed some minor changes to use `date-fns` instead of custom logic, but other than that, the PR looks great!

Thanks for your contribution, and sorry for taking so long to approve this!

leslieyip02:
@greptileai

leslieyip02:
@greptileai

PR #4419, opened 2026-06-11
Cleanup dead code remaining that was missed in #4225

zehata:
I realized that this function is no longer used, so it would be a waste of time to update it for config format V3

Note: This is part of a series of PRs related to #4387, please refer to the discussion for context


zehata:
Sorry, wrong branch

PR #4427, opened 2026-06-19
Introduce `LessonId`s

### LINKED ISSUE #4283: Lessons Configs are Not Preserved When Faculty Modifies Timetable

zehata:
- Generate `ModuleLessonMap` after module data is fetched, which stores a map of `RawLesson` to their `LessonId` so that serializeLessonDetails only needs to be called once.
- Generate `ModuleLessonMap` for sample modules as well.
- Generate `ModuleLessonMap` in the export service so lessons can be rendered.
- Currently these are inactive, and would only be used in a future PR.

## Context
This is part of the changes meant to fix #4283 
Please refer to #4387 for additional info

## Implementation
Other than adding the types, it also generates the `ModuleLessonMap` when modules are downloaded. This memoizes the `LessonId` so that they do not have to be generated every time they are used. This also applies to the export service. This also includes a couple of helper functions for serializing and deserializing `LessonId`s.

`LessonId`s are serialization of all a lesson details, except its module info and lesson type, please refer to https://github.com/nusmodifications/nusmods/pull/4292#issuecomment-3788288704 for why this was chosen. 
```ts
        {
          "classNo": "22",
          "lessonType": "Tutorial", // not serialized
          "weeks": [2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13],
          "day": "Wednesday",
          "startTime": "0900",
          "endTime": "1000",
          "venue": "S16-0518"
        },
```
would serialize to
`22|TUE|0900|1000|S16-0518|2_3_4_5_6_7_8_9_10_11_12_13`.

Serialization and deserialization can be quite expensive. So we calculate and store a map of the `LessonType` to `LessonId`s to lessons in `SemesterData`.
```ts
// Semester-specific information of a module.
export type SemesterData = {
  // ...
  readonly lessonMap: LessonMap<RawLesson>;
};
```

## Other Information
Currently these changes are inactive, they will be used in the a future PR. The delta to switch over to use `LessonId`s is already large enough as is.

### DIFF
--- a/export/src/makeModuleLessonMap.ts
+++ b/export/src/makeModuleLessonMap.ts
@@ -1,16 +1,4 @@
-import {
-  fromPairs,
-  get,
-  groupBy,
-  invert,
-  isNaN,
-  isUndefined,
-  map,
-  mapValues,
-  omitBy,
-  some,
-  split,
-} from 'lodash';
+import { fromPairs, groupBy, invert, isUndefined, map, mapValues } from 'lodash';
 import { DayText, Lesson, ModuleLessonMap, RawLesson, WeekRange, Weeks } from './types';
 
 type lessonTypeAbbrev = { [lessonType: string]: string };
@@ -128,56 +116,11 @@ export const serializeLessonDetails = <T extends RawLesson>(lesson: T): string =
 /**
  * Group lessons by lesson types
  * @param lessons lessons to group
- * @returns map of `LessonId`s, not lessons
+ * @returns a {@link ModuleLessonMap|ModuleLessonMap} from the `RawLesson`s provided
  */
 export const makeModuleLessonMap = (lessons: Readonly<Array<RawLesson>>): ModuleLessonMap => {
   const lessonsByLessonType = groupBy(lessons, 'lessonType');
   return mapValues(lessonsByLessonType, (lessonsWithLessonType) =>
     fromPairs(map(lessonsWithLessonType, (lesson) => [serializeLessonDetails(lesson), lesson])),
   );
-};
-
-const deserializeWeekNumbers = (serializedWeekNumbers: string): Array<number> => {
-  const weeks = map(split(serializedWeekNumbers, WEEKS_SEP), (week) => Number.parseInt(week, 10));
-
-  if (some(weeks, isNaN)) {
-    throw 'Serialized weeks is malformed';
-  }
-
-  return weeks;
-};
-
-/**
- * Parses serialized weeks by first attempting to deserialize it as a week range\
- * if that fails, attempt to deserialize it as week numbers\
- * if that also fails, the string is malformed
- * @param serializedWeeks
- * @returns
- */
-export const parseWeeks = (serializedWeeks: string): Weeks => {
-  const parsedRegex =
-    /(?<start>[0-9]{4}-[0-9]{2}-[0-9]{2})_(?<end>[0-9]{4}-[0-9]{2}-[0-9]{2})_(?<weekInterval>[0-9])_?(?<weeks>(?:_*[0-9])*)/.exec(
-      serializedWeeks,
-    );
-  const regexGroup = parsedRegex?.groups;
-  if (!regexGroup) {
-    return deserializeWeekNumbers(serializedWeeks);
-  }
-
-  const start = get(regexGroup, 'start');
-  const end = get(regexGroup, 'end');
-  const weekIntervalString = get(regexGroup, 'weekInterval');
-
-  const weekInterval = Number.parseInt(weekIntervalString, 10);
-  const weeks = get(regexGroup, 'weeks', undefined);
-
-  return omitBy(
-    {
-      end,
-      start,
-      weekInterval: weekInterval === 0 ? undefined : weekInterval,
-      weeks: weeks ? deserializeWeekNumbers(weeks) : undefined,
-    },
-    isUndefined,
-  ) as WeekRange;
 };
--- a/export/src/makeModuleLessonMap.ts
+++ b/export/src/makeModuleLessonMap.ts
@@ -1,0 +1,125 @@
+import { fromPairs, groupBy, invert, isUndefined, map, mapValues } from 'lodash';
+
+type lessonTypeAbbrev = { [lessonType: string]: string };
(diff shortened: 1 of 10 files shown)


leslieyip02:
@greptileai

zehata:
@leslieyip02 I have made some changes, and you might want to look at the corresponding changes I made to address the issues that Greptile generated, or retrigger the review.

leslieyip02:
@greptileai

website/src/utils/timetables/lessonId.test.ts
```diff
@@ -0,0 +1,169 @@
...
+import { ModuleLessonMap, RawLesson, WeekRange } from 'types/modules';
+import { CS1010S } from '__mocks__/modules';
+import { getModuleTimetable } from 'utils/modules';
+
+const semester = 1;
+const sampleModuleTimetable = getModuleTimetable(CS1010S, semester);
+
+describe('serialize/parse weeks', () => {
+  test('serialized weekRange should be deserialized to the same weekRange', () => {
+    const weekRanges: WeekRange[] = [
```
leslieyip02, line 20: I think we should add 1 more test case where with only `start` and `end`. ([BME5802](https://api.nusmods.com/v2/2025-2026/modules/BME5802.json) is an example with just `start` and `end` in its week range).

website/src/utils/timetables/lessonId.ts
```diff
@@ -24,9 +52,252 @@ export const LESSON_TYPE_ABBREV: lessonTypeAbbrev = {
...
+ * @returns a {@link ModuleLessonMap|ModuleLessonMap} from the `RawLesson`s provided
+ */
+export const makeModuleLessonMap = (lessons: readonly RawLesson[]): ModuleLessonMap<RawLesson> => {
+  const lessonsByLessonType = groupBy(lessons, 'lessonType');
+  return mapValues(lessonsByLessonType, (lessonsWithLessonType) =>
+    fromPairs(map(lessonsWithLessonType, (lesson) => [serializeLessonDetails(lesson), lesson])),
+  );
+};
+
+const deserializeWeekNumbers = async (serializedWeekNumbers: string): Promise<number[]> => {
```
leslieyip02: Does this need to be async?

website/src/utils/timetables/lessonId.ts
```diff
@@ -24,9 +52,252 @@ export const LESSON_TYPE_ABBREV: lessonTypeAbbrev = {
...
+};
+
+/**
+ * Parses serialized weeks by first attempting to deserialize it as a week range\
+ * if that fails, attempt to deserialize it as week numbers\
+ * if that also fails, the string is malformed
+ * @param serializedWeeks
+ * @returns
+ */
+export const parseWeeks = async (serializedWeeks: string): Promise<Weeks> => {
```
leslieyip02: Does this need to be async?

website/src/utils/timetables/lessonId.ts
```diff
@@ -24,9 +52,252 @@ export const LESSON_TYPE_ABBREV: lessonTypeAbbrev = {
...
+    weeks,
+  };
+};
+
+/**
+ * Used to recover from the config of a lesson type of a non-TA module with invalid lessons
+ * @param lessonsWithLessonType lessons with the same lesson type to generate a valid lesson config from
+ * @returns a `ClassNo`. The current implementation generates a config containing the first `ClassNo`
+ */
+export function getRecoveryClassNo(lessonsWithLessonType: Record<LessonId, RawLesson>): LessonId[] {
```
leslieyip02: Nit: I think the return type is supposed to be `ClassNo`.

Also, is there a reason to return `LessonId[]` instead of a single nullable `LessonId`? The function name `getRecoveryClassNo` implies a single return value.

export/src/makeModuleLessonMap.ts
```diff
@@ -0,0 +1,126 @@
...
+ *   end: "2025-02-14",
+ *   weekInterval: 1,
+ *   weeks: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13],
+ * }
+ * ```
+ * will be serialized to `2025-01-13_2025-02-14_1_1_2_3_4_5_6_7_8_9_10_11_12_13`\
+ *
+ * If `weekInterval` is undefined, it is serialized as `0`, _despite the fact that a 1 week interval is assumed in the logic_\
+ * If weeks is an empty array, the serialized string would be `2025-01-13_2025-02-14_`\
+ * If weeks is undefined, the serialized string would be `2025-01-13_2025-02-14`, without the dangling `_`\
```
leslieyip02: Minor nit: weeks should be enclosed by backticks (i.e. \`weeks\`).

leslieyip02 (commented):
Mostly LGTM, just some minor nits.

website/src/utils/timetables/lessonId.ts
zehata: No, I don't mind rewriting with try catch

website/src/utils/timetables/lessonId.ts
zehata: Ah right, fair point there

website/src/utils/timetables/lessonId.ts
```diff
@@ -24,9 +53,258 @@ export const LESSON_TYPE_ABBREV: lessonTypeAbbrev = {
...
+    configLessonTypeLessonIdsSet.has(lesson.classNo),
+  );
+  if (size(recoveryLessons) > 0) return keys(recoveryLessons);
+
+  const firstLessonId = first(keys(lessonsWithLessonType));
+  if (!firstLessonId) {
+    return [];
+  }
+  return [firstLessonId];
+}
```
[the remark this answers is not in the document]
leslieyip02: @zehata Maybe using [`intersection`](https://lodash.com/docs/4.17.23#intersection) is a better fit here?

```ts
const commonLessonIds = intersection(keys(lessonsWithLessonType), configLessonTypeLessonIds);
...
```

leslieyip02 (approved):
LGTM 👍

PR #4429, opened 2026-06-21
feat: parse and apply cohort-year conditions in prerequisite trees

### LINKED ISSUE #4258: Add additional information into prerequisite tree
Some courses have additional information in the prereq text e.g. [BT4103](https://nusmods.com/courses/BT4103/business-analytics-capstone-project)

> If undertaking an Undergraduate Degree THEN (**if in one of the cohorts prior to 2021 inclusive** THEN must have completed BT3102 at a grade of at least D AND must have completed BT3103 at a grade of at least D AND must have completed 1 of ES1601/ES2002/IS2101 at a grade of at least D) AND (**if in one of the cohorts from 2022 inclusive** THEN must have completed BT2101/BT3103 at a grade of at least D AND must have completed 1 of ES1601/ES2002/IS2101/UWC2101% at a grade of at least D)

These are essential texts that should be added, perhaps in the branching logic of sorts. In addition, it seems the AND path compression in this feels wrong but imo OR seems more fitting

TheMythologist:
Closes #4258

## Problem

The scraper's ANTLR grammar already recognises `COHORT_YEARS` predicates, but `ReqTreeVisitor` had no visitor for them, so the predicate fell through to the default `visitChildren` and was silently discarded. Cohort-specific requirements were therefore flattened into unconditional ones. For example, [BT4103](https://nusmods.com/courses/BT4103/business-analytics-capstone-project) has differing prerequiste logic, based on studens' matriculation cohorts.

## What this does

This carries the cohort condition through the whole pipeline instead of dropping it. The scraper now emits a `{ cohort, then }` node, and the website both evaluates it (in the planner, where the student's matriculation year is known) and displays it (on the module page, which has no user context).

### Scraper (`scrapers/nus-v2`)

- Added `CohortRule` / `CohortCondition` types and a `{ cohort: CohortCondition; then: PrereqTree }` variant to `PrereqTree`. The `years` field holds the raw `YEARS` tokens (`S:` = from-inclusive, `E:` = until-inclusive; the year part may be a calendar year like `2022` or an academic year like `2019/20`; two tokens form a closed range).
- Fixed the grammar so the cohort consequence binds correctly. Previously the `THEN` clause bound only to the immediately-following clause, so a trailing `AND COURSES …` escaped the gate and became an unconditional requirement (for NGT2001E this wrongly required `NSW%` of every cohort). The cohort consequence is now lifted to a new `compound`-level rule `cohort_conditional: cohort_years THEN compound`, placed above `binop`, so the consequence greedily absorbs trailing `AND`/`OR` clauses. `cohort_years` is now a bare condition, the `THEN` handling lives in the new `visitCohort_conditional`, and a bare cohort (no `THEN`) still contributes nothing. `program_types` / `programs` / `plan_types` are deliberately left untouched to avoid regressions. The ANTLR parser/visitor were regenerated from the updated grammar.
- Taught `flattenTree` to descend only into `then` so the condition metadata never leaks into `fulfillRequirements`, and to bail on non-object values — `flattenTree` walks nodes via `Object.values`, which recurses into the `nOf` tuple and passes the count (a number) back in, and the new `in` check only works on objects.

### Website

- Mirrored the type on the website copy of `PrereqTree`.
- `checkPrerequisite` gained an optional `cohortYear` parameter and a new `cohortConditionApplies` helper that compares the matriculation year against the `S:`/`E:` bounds and inverts the match for `IF_NOT_IN` / `MUST_NOT_BE_IN`. When the cohort year is unknown it conservatively assumes the requirement applies, preserving the previous behaviour.
- The planner selector threads `planner.minYear` (the "Matriculated in" setting) into the check, so a requirement is only enforced for matching cohorts. `conflictToText` renders the gated requirement.
- `ModuleTree` renders the condition as a label such as "For cohort 2022 onwards" since the module page has no user context.

## Testing

The full scraper suite (140 tests) and the affected website suites pass, lint is clean, and typecheck introduces zero new errors. New scraper cases cover the greedy trailing-`AND` grouping (the NGT2001E shape), `E:` bounds, academic-year ranges, `IF_NOT_IN` inversion, the bare-predicate drop, and `nOf` flattening (standalone and nested under a cohort gate). New website cases cover the planner evaluation across cohort bounds and the unknown-cohort fallback, plus a `ModuleTree` snapshot for the cohort-gated render. Regenerating NGT2001E with these changes produces `{ cohort: { rule: "IF_IN", years: ["S:2022"] }, then: { and: ["NTW%:D", "LC1016:D", "NSW%:D"] } }`, with all three modules correctly gated by cohort 2022 onwards.

Before:
image

After:
image

## Notes for reviewers

- `planner.minYear` defaults to the current academic year when the user has never opened planner settings, so gates evaluate as "current cohort" until a matriculation year is chosen.
- This is a data-shape change, so cohort nodes only appear in the API once the scraper's collation step re-runs (`combine <year>` re-collates from cached semester data without re-fetching from NUS, or the next scheduled full scrape) and the output is redeployed. The website code is forward-compatible and inert until then.
- The grammar change regenerates `NusModsParser.ts` / `NusModsVisitor.ts`, so those files show large diffs, mostly formatting related — the only hand-written grammar change is in `NusMods.g4`.


jloh02:
@greptileai

leslieyip02:
@greptileai

scrapers/nus-v2/src/services/requisite-tree/antlr4/NusMods.g4
```diff
@@ -56,7 +62,7 @@ cohort_years:
 		| IF_NOT_IN
 		| must_be_in
 		| must_not_be_in
```
leslieyip02, line 64: Nit: Can we standardise the casing? We could introduce a parser rule for `IF_IN` and `IF_NOT_IN` like this:

```
if_in: IF_IN;
if_not_in: IF_NOT_IN;
```

then it would be more consistent with:

https://github.com/nusmodifications/nusmods/blob/d424d32f45c9610901facbe2371ab00adfb4bb5f/scrapers/nus-v2/src/services/requisite-tree/antlr4/NusMods.g4#L85-L86

scrapers/nus-v2/src/services/requisite-tree/antlr4/NusMods.g4
```diff
@@ -10,7 +10,13 @@ program_types:
...
 	)*;
 
-compound: | '(' compound ')' | binop | op;
+compound: | '(' compound ')' | cohort_conditional | binop | op;
+
+// A cohort-gated requirement. The consequence is a full compound (placed above
+// binop) so that trailing AND/OR clauses are absorbed into the gate, i.e.
+// `COHORT_YEARS ... THEN a AND b` parses as the gate over `(a AND b)` rather
+// than `(gate over a) AND b`.
+cohort_conditional: cohort_years THEN compound;
```
leslieyip02, line 19: I'm not sure if this is handled, but it is possible for `COHORT_YEARS` to appear as a requirement without a following `THEN`.

[DBA3702](https://nusmods.com/courses/DBA3702/descriptive-analytics-with-r):

> PROGRAM_TYPES IF_IN Undergraduate Degree THEN (COURSES (1) DAO1704%:D, DSC1007:D AND COHORT_YEARS MUST_BE_IN S:2017)

[MNO2707A](https://nusmods.com/courses/MNO2707A/business-ethics) (more recent cohort year):

> PROGRAM_TYPES IF_IN Undergraduate Degree THEN (COURSES (1) MNO1706%:D, PL3239:D AND COHORT_YEARS MUST_BE_IN S:2023/24)

For what it's worth, most of the courses that have a cohort year requirement are archived courses, or have cohort years that are really old (i.e. 2017). However I think we should try to support this syntax as well (if feasible)? What do you think?

Disclaimer: I'm not familiar with ANTLR4.

leslieyip02 (changes requested):

scrapers/nus-v2/src/services/requisite-tree/antlr4/NusMods.g4
TheMythologist: Fixed in 2f8dd4f47ff72e2052926f1e9cbd7ad6a69aa9b7, that's a good nit!

scrapers/nus-v2/src/services/requisite-tree/antlr4/NusMods.g4
TheMythologist: Yup it's possible, though it's not a ANTLR4 change. Fixed in 6a5c4c61a8ec431f0f8885e2be32efa6e0f14586!

It now looks like this:
DBA3702:
image

MNO2707A:
image

There a was small display issue that was fixed as well, where childless nodes had dangling connector lines.

leslieyip02 (approved):
LGTM, thanks for the quick fixes!

PR #4431, opened 2026-06-23
Switch over to use `LessonId`

### LINKED ISSUE #4283: Lessons Configs are Not Preserved When Faculty Modifies Timetable

zehata:
- Changes validation and serialization/deserialization logic over use V3 format
- Reverted changes to optimiser service for it to return V1 serialization strings
- Reverted timetable export service to accept `ClassNo` or `LessonId`

## Context
This activates the `LessonId` introduced in #4427 to resolve #4283

## Implementation
As discussed in https://github.com/nusmodifications/nusmods/discussions/4387#discussioncomment-17071656
Module lesson configs are now stored as
```ts
{
  [lessonType: LessonType]: ClassNo[],
}
```
for non-TA modules, and as
```ts
{
  [lessonType: LessonType]: LessonId[],
}
```
for TA modules
Because `ClassNo` does not change when the faculty modifies timetables, changes made to lessons will not cause configs to become invalid unless the `ClassNo` was deleted.

Migration and validation logic has been updated to migrate v1 and v2 configs.

Tested that:
- optimiser
`https://nusmods.com/timetable/sem-2/share?CS1010E=TUT:09,SEC:2&CS1010S=REC:18,LEC:1,TUT:33`
- export service
My Timetable
are working.

## Other Information
Just pushing this to get Greptile to review it. Not ready for review yet.


website/src/reducers/timetables.ts
```diff
@@ -94,33 +94,27 @@ function moduleLessonConfig(
...
       };
     }
     case ADD_LESSON: {
-      const { lessonIndices, lessonType } = action.payload;
-      if (!(lessonIndices && lessonType)) return state;
+      const { lessonIds, lessonType } = action.payload;
       return {
         ...state,
-        [lessonType]: uniq([...lessonIndices, ...state[lessonType]]),
+        [lessonType]: uniq([...lessonIds, ...get(state, lessonType)]),
```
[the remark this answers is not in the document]
zehata: This should never happen but I guess it doesn't hurt to add a guard.

zehata:
Linter didn't catch the debug `console.log` that was left behind...

website/src/actions/timetables.ts
```diff
@@ -348,27 +354,71 @@ export function showLessonInTimetable(semester: Semester, moduleCode: ModuleCode
...
+          ? keys(pickBy(lessonsWithLessonType, (lesson) => lesson.classNo === lessonIds[0]))
+          : lessonIds;
+      },
+    );
+
+    dispatch(addTaModule(semester, moduleCode, taModuleLessonConfig));
   };
 }
 
 export const REMOVE_TA_MODULE = 'REMOVE_TA_MODULE' as const;
```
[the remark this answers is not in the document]
zehata: I don't think semester key and module key can ever be absent . In any case, if `moduleLessonConfig` is `undefined`, there wouldn't be any data loss since it is undefined.

website/src/utils/timetables/migration.ts
```diff
@@ -39,12 +115,22 @@ export function migrateSemTimetableConfig(
         ? taModulesConfig.includes(moduleCode)
         : moduleCode in taModulesConfig;
 
-      const timetable = getModuleSemesterTimetable(moduleCode);
+      const module: Module | undefined = get(modules, moduleCode, undefined);
+      if (!module) return accumulatedSemTimetableConfig;
```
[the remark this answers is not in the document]
zehata: The cached `moduleBank` must contain the module code if the user had previously added it to their timetable.

zehata:
I ran the tests again on my machine, and couldn't reproduce the test failure. Can we re-run the CircleCI test?

website/src/actions/timetables.test.ts
leslieyip02: While we're at it, let's update the `addModule` to use snapshots as well. The test isn't useful in it's current state, the GitHub link is broken, and I don't know what the TODO is for).

Something similar to the `"disabling ta module"` tests, like this:

```ts
describe(actions.addModule, () => {
  const semester: Semester = 1;
  const moduleCode: ModuleCode = "CS1010";

  beforeEach(() => {
    vi.spyOn(moduleBankActions, "fetchModule")
      .mockImplementation(() => () => Promise.resolve({} as Module));

    vi.spyOn(modulesUtils, "getModuleLessonMap")
      .mockReturnValue({} as ModuleLessonMap<RawLesson>);

    vi.spyOn(timetablesUtils, "randomModuleLessonConfig")
      .mockReturnValue({} as ModuleLessonConfig);
  });

  afterEach(() => {
    vi.restoreAllMocks();
  });

  test("dispatches addModule when the module exists", async () => {
    const state: any = {
      moduleBank: {
        modules: {
          [moduleCode]: { moduleCode },
        },
      },
    };

    const dispatch = vi.fn().mockResolvedValue(undefined);
    await actions.addModule(semester, moduleCode)(dispatch, () => state);
    expect(dispatch).toHaveBeenCalledTimes(2);
    const [, [secondAction]] = dispatch.mock.calls;

    expect(secondAction).toEqual(
      actions.Internal.addModule(semester, moduleCode, expect.any(Object))
    );
  });

  test("dispatches a notification when the module cannot be loaded", async () => {
    const state: any = {
      moduleBank: {
        modules: {},
      },
    };

    const dispatch = vi.fn().mockResolvedValue(undefined);
    await actions.addModule(semester, moduleCode)(dispatch, () => state);
    expect(dispatch).toHaveBeenCalledTimes(2);
    const [, [secondAction]] = dispatch.mock.calls;

    expect(secondAction).toEqual({
      type: OPEN_NOTIFICATION,
      payload: {
        message: `Cannot load ${moduleCode}`,
        action: expect.objectContaining({
          text: "Retry",
        }),
      },
    });
  });
});
```

website/src/actions/timetables.ts
```diff
@@ -348,27 +354,71 @@ export function showLessonInTimetable(semester: Semester, moduleCode: ModuleCode
...
+}
+
+/**
+ * While the non-TA modules use `ClassNo`s, the TA modules use `LessonId`s\
+ * Thus, the corresponding lessons need to be obtained from the `ClassNo` of the non-TA config to create the TA config
+ * @param semester
+ * @param moduleCode
+ * @returns
+ */
+export function enableTaModule(semester: Semester, moduleCode: ModuleCode) {
```
leslieyip02, line 388: I think we should test `enableTaModule`. Minimally, we should cover these 3 cases:

- should dispatch action to add the module
- should dispatch action to add the module when config consists of `ClassNo`
- should dispatch action to add the module even with missing `semesterData`

website/src/types/timetables.ts
```diff
@@ -1,21 +1,31 @@
...
   ModuleCode,
+  ModuleLessonMap,
   ModuleTitle,
   RawLesson,
   Semester,
 } from './modules';
 
 export type ModuleLessonConfig = {
+  [lessonType: LessonType]: ClassNo[] | LessonId[];
+};
```
leslieyip02, line 15: One suggestion for better type safety would be making `ClassNo` and `LessonId` [branded types](https://www.learningtypescript.com/articles/branded-types) instead of just type aliases for `string`:

```
export type ClassNo = string & { __brand?: 'ClassNo' };
export type LessonId = string & { __brand?: 'LessonId' };
```

We could also enforce the following invariants (correct me if I'm wrong):

```
export type ModuleLessonConfig = {
  [lessonType: LessonType]:
    | [ClassNo]                 // If `lessonType` is `ClassNo[]`, length = 1
    | [LessonId, ...LessonId[]] // If `lessonType` is `LessonId[]`, length >= 1
};
```

Granted, this will be quite annoying, but I think it's worth the additional safety?

website/src/utils/timetables/lessonHydration.ts
```diff
@@ -36,14 +44,16 @@ function hydrateModuleConfigWithLessons(
...
-    const newLessons = lessonsWithIndices.filter((lesson: RawLessonWithIndex) =>
-      lessonIndices.includes(lesson.lessonIndex),
-    );
-    return newLessons.map((lesson: RawLessonWithIndex) => ({
+): ModuleLessonMap<Lesson> {
+  const lessonMap: Readonly<ModuleLessonMap<RawLesson>> = getModuleLessonMap(module, semester);
+  return mapValues(moduleLessonConfig, (lessonIds: LessonId[], lessonType: LessonType) => {
+    const lessonsWithLessonType: Record<LessonId, RawLesson> = get(lessonMap, lessonType, {});
+
+    const isClassNo: boolean = lessonIds.length === 1 && !includes(lessonIds[0], '|');
```
leslieyip02: Let's extract this into a helper function (perhaps in `utils/timetables/validation.ts`)? It seems this `lessonIds.length === 1 && !includes(lessonIds[0], '|')` check is duplicated 3 times across the codebase.

website/src/utils/timetables/index.ts
```diff
@@ -43,29 +32,16 @@ export function isValidSemester(semester: Semester): boolean {
...
-      return map(randomlySelectedLessons, 'lessonIndex');
-    },
-  );
+export function randomModuleLessonConfig(
+  lessonMap: ModuleLessonMap<RawLesson>,
+): ModuleLessonConfig {
+  return mapValues(lessonMap, (lessons) => {
+    const lessonsByClassNo = groupBy(lessons, 'classNo');
+    const randomClassNo = sample(keys(lessonsByClassNo));
+    return randomClassNo ? [randomClassNo] : [];
```
leslieyip02: I'm not sure if this should return an empty array if we can't get a `randomClassNo`. Would an empty array in `ModuleLessonConfig` be a problem?

This function is called by `addModule`, which dispatches `ADD_MODULE`:

https://github.com/nusmodifications/nusmods/blob/9b93e5aeab6f84ef79f80825236cc2965ce2b6e9/website/src/actions/timetables.ts#L59-L84

I'm worried this would bring the user into an invalid state. Would it be safer to omit the config if `randomClassNo` is unavailable instead?

leslieyip02 (changes requested):
This is gonna take a while. Will continue reviewing tomorrow. Mostly seems good so far though!

website/src/utils/timetables/index.ts
zehata: Actually you raise a good point, in `utils/timetables/validation`, `validateNonTaModuleLesson` also defaults to `[]` for lesson types in a `ModuleLessonMap` with no lessons.

~~Removing the lesson type from the `ModuleLessonConfig` will not work though because the validation process iterates through all lesson types in the `ModuleLessonMap` which will attempt to add back the lesson type with no lessons. I'm not too sure how to approach this problem.~~

~~Lesson type configs that are empty arrays are also allowed because TA modules do not need to have at least one lesson for each lesson type.~~

Actually I think your method is fine too, I don't think it makes too much of a difference anyhow we deal with it.

I don't think this would realistically happen though? `ModuleLessonMap` is created by call `groupBy` on a module's lessons, so each lesson types has to have at least one lesson. Even if the user makes the change themselves, `ModuleLessonMap` is refreshed each time the module data is downloaded.

website/src/types/timetables.ts
zehata: TIL about branded types actually. Good idea though, I'll work on it.

zehata: The initial idea that I had was that all `ClassNo`s are treated as valid `LessonId`s. I have made some changes to make it such that only `[ClassNo]` are valid configs.

website/src/actions/timetables.test.ts
zehata: Yea it's definitely something to deal with. There were actually a couple of other stuff I actually cut out because this PR is really long, So long as you're ok with it I can try writing something out.

zehata:
I haven't actually reviewed this back yet. I'm just pushing it because I have something else to do tonight.

website/src/utils/timetables/index.ts
leslieyip02: I guess I'm being pessimistic, but it's not impossible for the loaded module data to be invalid. I think it's better to be defensive in this case.

website/src/utils/timetables/interactabilityHydration.ts
```diff
@@ -27,122 +41,83 @@ export function areOtherClassesAvailable(
...
 ): lesson is InteractableLesson {
-  return 'lessonIndex' in lesson;
+  return 'canBeSelectedAsActiveLesson' in lesson;
 }
 
 /**
  * Hydrate timetable lessons with interactability info\
  * See type defintion of {@link InteractableLesson} for properties added
  */
 export function getInteractableLessons(
```
leslieyip02, line 259: Can we refactor this function for better readability?

- Extract some of the more complex boolean expressions into helpers (e.g. `canBeSelectedAsActiveLesson`, `canBeAddedToLessonConfig`, etc).
- I'm not a fan of the triply-nested `mapValues`. Maybe we can extract each level to a helper (e.g. `getInteractableLesson(lesson, lessonId, ...)`, `groupInteractableLessonsByLessonId(lessonMap, ...)`, etc)?

website/src/utils/timetables/migration.ts
```diff
@@ -81,56 +167,67 @@ export function migrateSemTimetableConfig(
...
 } {
-  const lessonIndicesMap = makeLessonIndicesMap(timetable);
   return reduce(
     moduleLessonConfig,
     (accumulatedModuleLessonConfig, lessonsIdentifier, lessonType) => {
-      if (isArray(lessonsIdentifier)) {
+      if (!isV1(lessonsIdentifier)) {
+        const configIsV2 = isV2(lessonsIdentifier);
+        const migratedLessonConfig: [ClassNo] | LessonId[] = configIsV2
+          ? migrateLessonTypeLessonsFromLessonIndicesToLessonIds(
```
leslieyip02: I think we should split the logic in `migrateModuleLessonConfig` into `migrateLessonConfigV1`, `migrateLessonConfigV2` and `migrateLessonConfig` for better readability.

leslieyip02 (commented):
I've finished looking at most files, and the implementations seems fine. I'm just left with sharing/serialization/deserialization. I'll give it another pass tomorrow.

zehata:
~~@leslieyip02 Sorry I am making some refactors to deserialization right now~~ Done as below

website/src/utils/timetables/shareLinks.ts
```diff
@@ -150,49 +201,132 @@ function deserializeTaModulesConfigV1(
...
+  lessonType: LessonType;
+  serializedLessons: string;
+} | null {
+  // LAB:(1)
+  const [lessonTypeAbbr, serializedLessonTypeConfig] =
+    lessonTypeSerialized.split(LESSON_TYPE_KEY_VALUE_SEP);
+  const lessonType: LessonType = get(LESSON_ABBREV_TYPE, lessonTypeAbbr);
+  if (!lessonType) return null;
+
+  const unwrappedLessonType: RegExpMatchArray | null =
```
zehata, line 226: I realize a few of these could be rewritten as named groups regex. However, my rationale is that, we know that this regex works considering we have been using for 6 months. I would prefer to suggest this change in a future PR.

zehata:
Thanks for covering for me for some of the changes.

I am still going to be doing a final manual check later, just running everything manually as a final check.

leslieyip02:
I've updated the export deployment to point at the website deployment. This should simplify testing the export service: https://nusmods-export-iu6oqsuwj-modsbots-projects.vercel.app/.

zehata:
@leslieyip02 I have done a manual check (which is by no means thorough) as a sanity check. I tested migration from v2 config with multiple populated semesters, validating a malformed config, running optimiser and the export service. I did not notice anything out of the blue, so I think it should be ok.

leslieyip02 (approved):
LGTM. Thanks for the massive effort on this issue! Let's keep monitoring this.

zehata:
Thanks a lot for reviewing and helping with the changes!

PR #4435, opened 2026-06-26
fix: fetch module details when importing a planner file

TheMythologist:
## Problem

When adding a module to the planner through the UI, its details are fetched automatically. But uploading a planner JSON file via the **Upload** button did not fetch the details of the imported modules — they rendered without their info until a page refresh.

## Cause

The import flow only dispatched `importPlanner(importedState)` to update the Redux state. Two other flows already fetch module details, which is why this gap wasn't obvious:

- `onAddModule` calls `fetchModule` after adding each module.
- `componentDidMount` fetches details for every module already in the planner on page load — this is why refreshing the page "fixed" the imported modules.

The upload path went through neither, so the details were never fetched.

## Fix

Added an `onImportPlanner` handler in `PlannerContainer` that dispatches the import and then fetches each imported module's details, mirroring `onAddModule`. Placeholder entries without a `moduleCode` are skipped, and fetch errors are swallowed to match the existing `componentDidMount` behavior. The import button now uses this handler instead of dispatching the raw action.

## Testing

- Manually: on a completely new session (with no module cache), navigate directly to /planner and upload a planner JSON and confirm module titles/credits appear immediately without a refresh.

Note that the TODO comment is also copied from the existing `onAddModule` logic


website/src/views/planner/PlannerContainer/PlannerContainer.tsx
```diff
@@ -146,6 +146,20 @@ export class PlannerContainerComponent extends PureComponent<Props, State> {
...
 
+  onImportPlanner = (importedState: PlannerState) => {
+    this.props.importPlanner(importedState);
+
+    // Fetch the details of each imported module so they display correctly
+    // without requiring a page refresh (componentDidMount only fetches on load)
+    values(importedState.modules).forEach((module) => {
+      if (module.moduleCode) {
+        this.props.fetchModule(module.moduleCode).catch(() => {
+          // TODO: Handle error
```
leslieyip02: While we're at it, shall we implement the error handling in this PR? It seems `componentDidMount` also has the same TODO, so we could write a function to wrap the call to `this.props.fetchModule` and dispatch a call to `openNotification` with a message to indicate the module wasn't found (if it's not too annoying, add a unit test since idk how else to make `fetchModule` fail otherwise).

`openNotification` reference for convenience:

https://github.com/nusmodifications/nusmods/blob/550b4ce177f4186ad35ba9468925067d411a082f/website/src/views/planner/PlannerImportButton.tsx#L16-L23

leslieyip02 (changes requested):

website/src/views/planner/PlannerContainer/PlannerContainer.tsx
TheMythologist: Fixed in b30fd273080614e189003f3166859dfcd9a00357

I was lazy to fix the TODO comments in this PR, but you're right, someone has to fix it eventually HAHA

leslieyip02 (approved):
LGTM!

PR #4437, opened 2026-06-27
feat(prereq): model program-type-gated prerequisites

TheMythologist:
## What & why

Prerequisites that branch by program type — e.g. `(PROGRAM_TYPES IF_IN Graduate Degree Coursework THEN X) OR (PROGRAM_TYPES IF_IN Undergraduate Degree THEN Y)` — previously failed to parse and produced no prerequisite tree at all. The grammar only allowed `program_types THEN compound` at the top level, so any program-type conditional that was parenthesised, combined with `AND`/`OR`, or gated on a disjunction of program types raised a "no viable alternative" error. This silently dropped the tree for ~65 modules (ESE5608, MA5202, IT5003, EE5113, the honours `*HM` modules, and others).

This PR parses those prerequisites and carries the program type as a first-class node in the tree (`{ programType, then }`), mirroring the existing cohort modelling. It is display-only: the planner has no program-type input to evaluate against, so these gates are shown in the tree but are never treated as unmet requirements.

## Scraper

- Grammar: `program_types_conditional` now takes a `program_types_gate` — `'(' program_types_gate ')' | program_types (OR program_types)*` — so the gate may be a single program type, a disjunction of them, or that disjunction parenthesised. `compound` gained `program_types_conditional`, and `overall` was simplified to `compound EOF`. The parser was regenerated with `pnpm antlr4ts`.
- Visitor: builds `{ programType, then }`; a gate that is a disjunction of `IF_IN` program types is collapsed into one condition over the union of types (since `IF_IN X OR IF_IN Y` is just `IF_IN X, Y`); bare program types are dropped; and the old "too many program types" bail-out is removed. The result is then normalised — redundant same-type repeats are flattened (the greedy `THEN` makes `pt THEN A OR pt THEN B` parse as `pt THEN (A OR (pt THEN B))`, and flattening recovers the symmetric `A OR B`), and the single outermost wrapper is dropped only when it is the lone "Undergraduate Degree" gate (the ubiquitous assumed audience). Any gate naming more is a real restriction and is kept.
- `flattenTree` walks only the gated requirement.

## Website

- Mirrored the `ProgramTypeRule` / `ProgramTypeCondition` types and the new `PrereqTree` variant.
- Planner: `checkPrerequisite` skips program-type nodes (never unmet) and `conflictToText` handles them defensively.
- `ModuleTree` renders a program-type gate as "For <program type>" (joining a disjunction's types with " or "). A disjunction whose branches are all program-type gates is a per-degree split rather than a choice, so it drops the misleading "one of" label and collapses the now-empty junction so the connector line runs straight through.

## Behaviour by case

- A normal module's lone `Undergraduate Degree THEN …` wrapper is dropped, so every existing tree is unchanged.
- Same-type branches (the honours modules) flatten to a plain `A OR B`.
- Differing-type branches are modelled and labelled — e.g. IT5003 shows `IT5001` under both "For Graduate Degree Coursework" and "For CPE (Certificate)", and MA5202 shows three branches.
- An OR-of-program-types gate is collapsed and labelled with the union of its types — EE5113 (`Undergraduate OR Graduate Coursework`) shows "For Undergraduate Degree or Graduate Degree Coursework → EE5934", and EE6438 / BL5232D show "For Graduate Degree Coursework or Graduate Degree Research".
- A graduate/CPE-only wrapper is kept and labelled, so ESE5608 — whose undergraduate branch is entirely `PROGRAMS`/`SPECIAL` constraints the tree cannot represent — shows "For Graduate Degree Coursework → ESE5003" instead of a bare, unexplained `ESE5003`.

## Testing

- Scraper `parseString`: added cases for differing-type modelling, same-type flattening, the ESE5608 degenerate shape, parenthesised and unparenthesised OR-of-program-types gates, and multi-type gates; updated the two previously-`null` "cannot parse" tests. Full scraper suite passes (151).
- Website: added planner cases (program-type nodes are never unmet, plus `conflictToText`) and a `ModuleTree` snapshot for the per-degree split. Planner + `ModuleTree` suites pass.
- Typecheck and lint clean on touched files.

Before:
image

with error `["no viable alternative at input '(PROGRAM_TYPESIF_INGraduate Degree CourseworkTHEN'"]`

After:
image

with no errors


leslieyip02 (approved):
LGTM 👍

Issue #3338, opened 2021-08-10
Make all date and time use local timezone


RezwanArefin01:
### Problem Description

Due to COVID-19, many students are currently stranded overseas. NUSMods have some issues with timezones. For example: 
- The time shown above the timetable is hard-coded in SGT. 
- But the marker that runs on the timetable is showing current time takes local time. 
- Also the time till next class shows time in only SGT. 

It is very inconvenient to always convert the time to local time in our mind. 

### Describe the solution you'd like

Make all times shown in the website use local time / have a settings for choosing timezone.

VVinter-melon:
After downloading the .ics file and importing to google calendar, the timezone is set to (GMT+00:00) Coordinated Universal Time. When editing for example a lecture event at 10am SG time, google calendar shows 2am at (GMT+00:00).  
![Screenshot 2021-08-11 101251](https://user-images.githubusercontent.com/39219189/128959192-4c23b718-6151-4379-920b-b0ade56d1f6f.jpg)

yaofeng-wang:
Hi, I'm interested in taking this. 

Also, which solution should we choose? Implicitly use local time or have a setting for timezone?

li-kai:
@yaofeng-wang Apologies for the late reply, but if you are still interested in taking this on, I think it's easier to use the implicit local time.

These would be the following sub tasks:
- [x] Convert local time to sgt for the marker.
- [ ] Convert local time to sgt for the time till next class
- [x] Fix ics timezone from UTC to Singapore

PR #4439, opened 2026-06-29
fix: show current time indicator in Singapore time

TheMythologist:
## First commit (377d520e48d2c0e959670d0a0e3f44da422d93b9)

The timetable's current time indicator derived its position from the viewer's local timezone via the deprecated `getCurrentHours`/`getCurrentMinutes` helpers. Use the `currentTime` injected by `withTimer` and convert it through `toSingaporeTime`, so the indicator's day and time always reflect SGT regardless of the viewer's timezone.

Part of #3338

## Second commit (ab4fcdd6336364725522f1ff6b4db26985be8167)

Removes/Replaces deprecated functions in timify.ts


TheMythologist:
> The one open item is that AvailabilitySearch.defaultSearchOptions still seeds its day and hour from local time rather than SGT — a pre-existing inconsistency that is now more visible since the timetable next to it is SGT-aware, but it does not regress anything this PR touches.
>
> AvailabilitySearch.tsx — defaultSearchOptions derives both the default day and hour from the viewer's local time rather than SGT, which is a pre-existing gap worth addressing in a follow-up.

Out of scope of this PR

leslieyip02 (approved):
LGTM!

PR #4440, opened 2026-06-30
fix: default tracking boolean to `false`

### LINKED ISSUE #3254: Privacy Opt-Out not clickable

TheMythologist:
Closes #3254

When matomo fails to load (such as due to adblocker), `allowTracking` remains defaulted to `true` in the `useState` declaration. Swap the defaults to `false` instead.

When matomo does load successfully, `setAllowTracking` will be called with the correct value.


jloh02 (approved):
Simple fix with low implications. LGTM!

PR #4442, opened 2026-07-01
fix(today): base "time till next class" on Singapore time

TheMythologist:
The Today view computed the next-lesson countdown by placing the lesson start time on the viewer's local calendar day while comparing against the true current instant. Since lesson times are in Singapore time, overseas students saw a countdown that was off by their timezone offset.

Derive "now" from the SGT wall-clock (via `toSingaporeTime`) when reasoning about today, and anchor the lesson start time to that same reference so the displayed duration reflects the real time until the next class. No-op for viewers already on Singapore time.

Part of https://github.com/nusmodifications/nusmods/issues/3338


jloh02:
We should probably fix greptile's comments in this samePR while we're at it?

TheMythologist:
> We should probably fix greptile's comments in this samePR while we're at it?

You're right - fixed in 361be69a2eeee3eec52f3195cb938c48970b29c6

jloh02 (approved):

PR #4445, opened 2026-07-02
fix(scraper): keep current-year module info during special-term AY fallback

TheMythologist:
## Problem

During an academic-year migration, modules offered in the **current** year were showing **stale module info from the previous year**. For example, in the AY2026/2027 output `GE4204` came out with its AY2025/2026 title *"New Geographies of Urban Theory"* and description, and `"acadYear": "2025/2026"`, even though the NUS API correctly returns the new title *"'Slumdog' Cities: Urban Theory from the Margins"* for 2026/2027. The module's actual 2026/2027 Semester 1 timetable was correct — only the module info (title/description/acadYear/prereqs) was stale. This affected essentially every module offered in Sem 1/2 that also existed in the previous year's catalog but isn't offered in a special term.

## Root cause

Introduced in #4402, which added the special-term fallback (Sem 3 & 4 sourced from the previous academic year during a migration) together with the `preserveModuleInfoSemesters` guard in `combineModules`.

`GetSemesterData` emits an entry for **every** module in the catalog it's given, offered or not — modules without a timetable that semester are pushed with `semesterData: undefined` but still carry that academic year's module info. For special terms, that catalog is the **previous** year's, so those timetable-less entries carry stale titles.

`combineModules` merges semesters in order 1 → 2 → 3 → 4, and its preserve guard was gated on the incoming entry having a timetable:

```js
if (semesterData && preserveModuleInfoSemesters?.has(semesterData.semester)) { ... }
```

A module not offered in the special term has `semesterData === undefined`, so `semesterData && …` is falsy, the guard is skipped, and execution falls through to the "always use the latest semester's data" branch — overwriting the current-year module info with the previous-year record. Because special terms are processed last, the stale info wins.

The regression test added in #4402 only covered the *offered* case (preserved semester **with** a timetable), so this path was never exercised.

## Fix

Determine whether a batch is a preserved (previous-AY) semester **per batch**, from any offered module's semester, rather than per-entry from `semesterData.semester`. All modules in a batch come from the same semester's fetch, so this correctly classifies timetable-less entries too. For preserved semesters we now keep the existing current-year module info and only merge in a timetable when the module is actually offered.

## Testing

- Added a regression test reproducing the exact scenario: a module offered in the current year (Sem 1) that is present-but-not-offered (no timetable) in a previous-AY special term must keep its current-year title. It fails on `master` and passes with this change.
- Existing `combineModules` tests still pass (including the #4402 offered-in-preserved-semester case).
- `tsc --noEmit` and `oxlint` are clean.

### DIFF
--- a/scrapers/nus-v2/src/index.ts
+++ b/scrapers/nus-v2/src/index.ts
@@ -123,7 +123,7 @@ yargs
             logger.warn(error, `No semester data available for ${semester}`);
             return [];
           });
-        semesterData.push(modules);
+        semesterData.push({ modules, semester });
       }
 
       await new CollateModules().run({ aliases, semesterData });
--- a/scrapers/nus-v2/src/tasks/CollateModules.test.ts
+++ b/scrapers/nus-v2/src/tasks/CollateModules.test.ts
@@ -182,6 +182,103 @@ describe(combineModules, () => {
       },
     ]);
   });
+
+  test('should keep current-year module info for modules not offered in a preserved semester', () => {
+    // Reproduces the AY-migration bug: a module offered in the current year
+    // (Semester 1) that also exists in the previous year's catalog but is NOT
+    // offered in the previous-AY special term. Its timetable-less special-term
+    // entry must not overwrite the current year's module info.
+    const currentYearModule = {
+      acadYear: '2026/2027',
+      department: 'Geography',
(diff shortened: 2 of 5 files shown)


ravern:
TREX

jloh02 (approved):

PR #4449, opened 2026-07-05
fix: update wording for attendance requirement in free day conflicts

thejus03:
## Summary
- Update the free day conflicts message in the optimiser to say lessons require **live** attendance instead of **physical** attendance, better reflecting that these lessons must be attended at their scheduled time.

## Test plan
- No functional change; copy-only update to `OptimiserFreeDayConflicts`.


jloh02:
Is this because the mentioned "live" lessons may be remotely assessed too?

thejus03:
> Is this because the mentioned "live" lessons may be remotely assessed too?

Yes because "live" can be either in person or online as seen in the first option

image

jloh02 (approved):

Issue #4455, opened 2026-07-09
Can you add different set-up slots for the timetable?


graizecad:
### Is your feature request related to a problem? Please describe.

I'm frustrated when i want to compare the arrangement of timetables in a semester but i cannot do it unless i screenshot the current one, then move things around on the page.

### Describe the solution you'd like

I wish there was way to directly compare the arrangements of timetable on the website to make everything easier. something like how in games, you have different save slots/team-line ups so i can always refer to the different arrangements ive made.

### Describe alternatives you've considered

display of saved timetables/screenshots directly on the page. even if we cant go back to access it or modify it, at least the direct comparison is there. I know i can js windows key and split screen but lowkey would be nice if the website already has the function itself...

eugenewong22:
i would like to work on this

jloh02:
Could you clarify the use of this? This is probably a once a semester usage by approximately 20% of the user base, when comparing multiple (infinitely many) links is probably more useful? It also has major implications on the redux setup?

leslieyip02:
Hi @eugenewong22, appreciate the effort in raising PRs for this feature request. However, on a first glance it seems to be introducing a lot of risk for an unclear benefit. We need to change a lot of the code in order to support storing multiple timetable configs, which may break certain assumptions in the codebase, impact existing behaviors and cause 2nd order effects.

As a baseline, an alternative to taking screenshots (as mentioned in the original feature request) would be using sharing links:

1. Configure timetable
2. Share "Share/Sync", copy the link and open it in a new tab
3. Go back to the original tab, modify the timetable and repeat step 2 as needed
4. Compare timetables between tabs
5. Click "Import" on the tab that you want to keep to set your timetable

https://github.com/user-attachments/assets/689474be-c0a4-4b64-9da0-1ce7b2645bc5

This approach is already supported, and also lets users compare timetables. The main difference is that this approach uses the browser's tabs, while the approach in #4460 and #4461 shows a comparison view within NUSMods:

<img width="1255" height="208" alt="Image" src="https://github.com/user-attachments/assets/3236f3b2-4d68-4da9-aa36-b785eef277a4" />

<img width="1261" height="648" alt="Image" src="https://github.com/user-attachments/assets/0f814475-12f6-4e52-bab1-d9bcec0c6a78" />

The new approach would be:

1. Configure a timetable (Timetable 1)
2. Create a new timetable and repopulate it (Timetable 2)
3. Open the menu and click "Compare"
[...]

PR #4460, opened 2026-07-10
feat(website): support multiple saved timetables per semester

eugenewong22:
## Context

Closes the "save slot" half of #4455: users want to keep multiple timetable arrangements per semester and switch between them without screenshotting before every change. (A follow-up PR adds the side-by-side comparison view — see the issue for the full plan.)

## What this does

- **Save slots per semester**: a pill-tab row under the semester switcher shows each saved arrangement. Switching tabs swaps the entire arrangement — lessons, colours, hidden modules and TA modules — instantly. Edits always go to the active slot.
- **Create / duplicate**: a `+` menu adds a new blank timetable or duplicates the current one.
- **Rename / delete**: a `⋯` menu with modal dialogs (following `ResetTimetable`'s pattern). Deleting is undoable via the existing undo notification; the last remaining slot can't be deleted.
- **Import as new timetable**: shared timetable links now offer importing into a fresh slot, so importing no longer has to overwrite the saved timetable. The existing overwrite Import is unchanged.

## Implementation

- **Swap-on-switch**: the active timetable stays in the existing `TimetablesState.{lessons,colors,hidden,ta}` maps, so the ~15 existing timetable actions, all selectors, undo, share and export code are untouched. A new `slots: { [semester]: TimetableSlot[] }` array stores arrangements; `SWITCH_TIMETABLE_SLOT` snapshots the live maps into the outgoing slot and loads the target's data (`reducers/timetables.ts`).
- **Slot `data` is authoritative for inactive slots only** — the active slot's data lives in the live maps and its snapshot is refreshed on switch-away. All reads go through the new `getSlotTimetableData` selector.
- **Lazy initialization + v3 persist migration**: semesters with no `slots` entry implicitly have one active slot ("Timetable 1"). The redux-persist migration (`migrateV2toV3`, following the planner's exported-migration pattern) only adds two empty maps — no user data is rewritten.
- **Undo**: `DELETE_TIMETABLE_SLOT` joins `actionsToWatch`. Switching/adding is deliberately not watched — switching back is lossless, and with `limit: 1` it would clobber the single undo step.
- **Module bank**: slot modules are pinned against LRU eviction (`getPinnedLessonConfigs`), and the switch/delete thunks fetch the incoming slot's modules *before* loading it into the live timetable.
- **Year rollover**: unchanged — only the active lessons are archived; slots are reset with the rest of the year's state (documented in the reconciler; archiving all slots would need an archive shape change and is left as a follow-up).

## Testing

- Reducer: slot CRUD, switch round-trip preserves both timetables exactly, delete-active activates the neighbour, no-ops return the same state reference, migration unit tests (`reducers/timetables.test.ts`).
- Selectors, thunk ordering (fetch before switch), LRU pinning, undo restore, switcher component behaviour, and an integration test that import-as-new-slot doesn't clobber the saved timetable.
- `pnpm run ci` passes (lint, typecheck, 800+ tests, build).
- Manually verified in Chromium: create/rename/duplicate/delete/switch, share-import both paths, reload persistence, **v2→v3 migration with real persisted v2 state**, and cross-tab sync with two tabs.

switcher

### DIFF
--- a/website/src/actions/moduleBank-lru.ts
+++ b/website/src/actions/moduleBank-lru.ts
@@ -1,20 +1,28 @@
-import { flatMap, sortBy, get } from 'lodash-es';
+import { flatMap, sortBy, get, values } from 'lodash-es';
 
-import { ModulesMap } from 'types/reducers';
-import { TimetableConfig } from 'types/timetables';
+import { ModulesMap, TimetablesState } from 'types/reducers';
+import { SemTimetableConfig, TimetableConfig } from 'types/timetables';
 import { ModuleCode } from 'types/modules';
 
 // Module bank utils - exported separately so this does not become exported as an action creator
 
+// All lesson configs that pin their modules in the module bank: the live
+// timetable of every semester, plus every saved slot's snapshot
+export function getPinnedLessonConfigs(timetables: TimetablesState): SemTimetableConfig[] {
+  return [
+    ...values(timetables.lessons),
+    ...flatMap(values(timetables.slots), (slots) => slots.map((slot) => slot.data.lessons)),
+  ];
+}
+
 // Export for testing
-// eslint-disable-next-line import/prefer-default-export
 export function getLRUModules(
   modules: ModulesMap,
-  lessons: TimetableConfig,
+  lessons: TimetableConfig | SemTimetableConfig[],
   currentModule: string,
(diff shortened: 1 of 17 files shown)


leslieyip02:
@greptile

PR #4467, opened 2026-07-15
feat(optimiser): allow pinning specific class slots to optimise around

### LINKED ISSUE #4341: Optimiser: Allow users to choose certain slots and optimise the remaining

Currently, the optimiser optimises for all the lessons in the timetable. But many students have specific slots they would like to attend for some lessons eg with friends. So they would want the optimiser to only optimise the remaining lessons around these explicitly chosen slots.

There is a way for students to enter the specific slot they want for some lessons. After entering all the other preferences and clicking the Optimise button, an optimal timetable is generated around the chosen slots.

eugenewong22:
## Context

Closes #4341

The optimiser currently optimises every lesson in the timetable. Many students already know exactly which class they want for some lessons (e.g. a tutorial with friends) and want the optimiser to keep those and only optimise the rest.

## Implementation

Users can now pin a specific class per lesson type in a new **"Pin specific classes (optional)"** section of the optimiser form. Each lesson gets a dropdown that defaults to *Any (optimise for me)* and lists every class with its schedule (paired slots of the same class number are shown together, e.g. `1 — Tue 08:00-10:00, Fri 08:00-10:00`).

**API (Go):**
- New `pinnedSlots` request field with entries `"MODULE|LessonType|ClassNo"`, mirroring the existing `recordings` convention.
- Pins are validated against the raw module timetable (nonexistent class / module not in request / malformed / duplicate entries → 400).
- In `mergeAndFilterModuleSlots`, a pinned lesson's candidates are reduced to just the pinned class **before** the duplicate-schedule merge, so the beam search is forced to assign it — no solver changes needed.
- Pinned classes are exempt from the free-day and time-range hard filters (same as recorded lessons): an explicit pin wins over general preferences. The pinned class also becomes the lesson's default/backup slot so `defaultShareableLink` respects the pin.

**Website:**
- `useOptimiserForm` gains `pinnedSlots` (`Record<lessonKey, classNo>`), serialised into the request payload on optimise.
- New `OptimiserPinnedSlotSelect` (dropdown per lesson) and `OptimiserPinnedSlotConflicts` (non-blocking warning when a pinned class falls on a selected free day or outside the selected times — the pin is kept anyway, the warning just explains the override).
- Pinned lessons are excluded from `getFreeDayConflicts` so a pin can *resolve* a free-day conflict that would otherwise disable the Optimise button.
- Stale pins are pruned when modules are removed from the timetable.
- Pinning is orthogonal to the live/recorded toggle; conflict warnings are only computed for live lessons since recorded ones are exempt from physical constraints anyway.

Verified end-to-end against a local optimiser server: pinning CS2040S Laboratory 08 (Mon 14:00–16:00) with Monday selected as a free day shows the warning, keeps the button enabled, and the returned links contain `CS2040S=LAB:08` with the lab scheduled on Monday.

Tests:
- `api/optimiser`: 6 new integration tests (pin assigned, pin wins over free day, and the four 400 cases); `validateTimetable` updated to exempt pinned lessons like recorded ones. All 15 tests pass; `golangci-lint` clean.
- `website`: unit tests for `getPinnedSlotOptions` / `getAllPinnedSlotOptions` / `getPinnedSlotConflicts` / `getPinnedSlotsPayload` and both new components. `pnpm test`, `pnpm typecheck`, `pnpm lint` all pass.


thejus03:
@greptile

website/src/utils/optimiser.ts
```diff
@@ -153,6 +162,117 @@ export function getFreeDayConflicts(
...
+    .map(([classNo, classLessons]) => {
+      const sortedLessons = [...classLessons].sort(
+        (a, b) =>
+          WorkingDays.indexOf(a.day as Day) - WorkingDays.indexOf(b.day as Day) ||
+          a.startTime.localeCompare(b.startTime),
+      );
+      const schedules = uniq(
+        sortedLessons.map(
+          (lesson) =>
+            `${lesson.day.slice(0, 3)} ${getOptimiserTime(lesson.startTime)}-${getOptimiserTime(
```
thejus03: Image

This is the current UI. I think we would rather want it same as buttons as we see for free days here:

Image

How this would work is that selected lessons would be pinned according to the slot chosen in the timetable tab.

website/src/utils/optimiser.ts
```diff
@@ -153,6 +162,117 @@ export function getFreeDayConflicts(
...
+          reasons.push(`falls on your free day(s): ${freeDayClashes.join(', ')}`);
+        }
+        if (isOutsideTimeRange) {
+          reasons.push(
+            `is outside your selected lesson times (${getOptimiserTime(
+              lessonTimeRange.earliest,
+            )} - ${getOptimiserTime(lessonTimeRange.latest)})`,
+          );
+        }
+        return isEmpty(reasons)
```
thejus03: Correct me if I am wrong but we are not using this in your current solution right?

website/api/optimiser/README.md
```diff
@@ -42,7 +42,7 @@ A single request moves through the following stages:
 
 1. **`optimise.go` — HTTP handler**: Decodes the JSON request body into `OptimiserRequest` and calls `solver.Solve`.
 
-2. **`_modules/GetAllModuleSlots`**: For each requested module, fetches timetable data from the NUSMods API (`_client`). Slots are then filtered (removing those outside the time window or on free days) and deduplicated — two class numbers that share the same day, start time, and building are treated as equivalent and merged to reduce the search space. Returns a map of `Module → LessonType → ClassNo → []Slot`.
+2. **`_modules/GetAllModuleSlots`**: For each requested module, fetches timetable data from the NUSMods API (`_client`). Slots are then filtered (removing those outside the time window or on free days) and deduplicated — two class numbers that share the same day, start time, and building are treated as equivalent and merged to reduce the search space. Lessons pinned via `pinnedSlots` (format `"MODULE|LessonType|ClassNo"`) are reduced to just the pinned class, which is kept even if it falls outside the time window or on a free day — an explicit pin wins over general preferences. Returns a map of `Module → LessonType → ClassNo → []Slot`.
```
thejus03: pinned lessons should be considered for free day conflicts and start/end time conflicts

thejus03 (changes requested):
In terms of logic, pinned lessons MUST not be in conflict with free days and start/end time. 

If a **live** pinned lessons is in conflict with a free day then just include that to the already existing free Day conflict UI component.

In addition, now we need to have a Start/End time conflict UI component as well for **live** pinned lessons which ideally looks and acts similar to free day conflict UI component. 

All this logic can happen in frontend because we have the information already. 


In terms of backend code for the optimiser, I think we only need minimal changes in the logic using the above set of rules. All non-live lessons (regardless of pinned or not) should be considered for scoring heuristic as well.

website/api/optimiser/README.md
eugenewong22: alright made this change in the modified PR

website/src/utils/optimiser.ts
eugenewong22: changed the UI in the modified PR

website/src/utils/optimiser.ts
eugenewong22: yup removed this component

website/api/optimiser/_solver/solver.go
```diff
@@ -316,65 +322,46 @@ func isInvalidCoordinates(coord models.Coordinates) bool {
...
-	physicalSlots := make([]models.ModuleSlot, 0, len(daySlots))
-	for i := range daySlots {
-		slot := &daySlots[i]
-		if !isLessonRecorded(slot.LessonKey, recordings) {
-			physicalSlots = append(physicalSlots, *slot)
-		}
-	}
-
-	return physicalSlots
-}
```
thejus03, line 376: The deleted `getPhysicalSlots` was used to strip recorded lessons because only those slots run through the lunch-gap and consecutive hours scoring heuristic. Right now, you have a regression allowing all slots to be passed sent for this scoring heuristic. 

This filter is necessary because we don't care where non-live lessons are because students are not attending them. 

I believe you should bring this back.

website/src/utils/optimiser.ts
```diff
@@ -153,6 +173,109 @@ export function getFreeDayConflicts(
...
+  return modules.flatMap((module) => {
+    const { moduleCode } = module;
+    const lessons = getModuleTimetable(module, semester);
+    const lessonOptions = get(groupedPhysicalLessonOptions, moduleCode, []);
+    return compact(
+      lessonOptions.map((lessonOption) => {
+        const { lessonType, lessonKey, displayText } = lessonOption;
+        const classNo = pinnedClassNos[lessonKey];
+        if (!classNo) return null;
+
```
thejus03, line 233: If two pinned lessons have time conflicts between them then it should raise a warning as well and disable sending to optimiser. No additional changes needed in backend here.

thejus03 (changes requested):
Great work! The UI looks good so far. I made a couple minor changes. Take a look and tell us if you are fine with it.

There were a couple bugs/issues with the backend. I have also sent a commit to fix one issue to make sure that pinned lessons were always allocated first during beamSearch.

website/src/utils/optimiser.ts
eugenewong22: alright updated this component

website/api/optimiser/_solver/solver.go
eugenewong22: ok i included this in the new commit

eugenewong22: also updated UI looks better

thejus03 (changes requested):
Currently, only LIVE pinned classes with time conflict show warning. Even if they are not live, but pinned it should show warning.

thejus03:
Had to revert the UI for the pinned lessons to remove the venues because I realised that courses like CS2103T have grouped slots with multiple venues within each slot.

thejus03:
@greptile

thejus03 (approved):
LGTM

PR #4478, opened 2026-07-23
Remove Disqus identifier for comment migration

leslieyip02:
Apparently, [`disqus_identifier` needs to be removed](https://eaglesky.github.io/2016/01/02/MigrateDisqus/) in order for URL mapper comment migration to work.

I have re-submitted the URL mapping:

```csv
http://nusmods.com/courses/GEH1043/reviews,http://nusmods.com/courses/GEC1011/reviews
http://nusmods.com/modules/GEH1043/reviews,http://nusmods.com/courses/GEC1011/reviews
```

DO NOT MERGE.

I'll be back in 24 hours.

### DIFF
--- a/website/src/types/views.ts
+++ b/website/src/types/views.ts
@@ -32,7 +32,6 @@ export type RefinementItem = { key: string; doc_count?: number; missing?: boolea
 export type RefinementDisplayItem = RefinementItem & { selected: boolean };
 
 export type DisqusConfig = {
-  readonly identifier: string;
   readonly url: string;
   readonly title: string;
 };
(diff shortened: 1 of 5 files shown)


leslieyip02:
Supposedly, we can use the [URL Mapper](https://help.disqus.com/en/articles/1717129-url-mapper) in Disqus to move comments from one thread to another. However, this does not work.

This could be because we use both `/courses` and `/modules` paths (`nusmods.com/modules` redirects to `nusmods.com/modules`, which confuses Disqus. A full migration of all reviews might be necessary in order to fix this, but we also don't know if there are issues on Disqus's end because the only feedback we get from Disqus is this:

Screenshot 2026-09-02 at 10 47 19 AM

Since the use case of this is just moving comments from 1 mod to another, I don't think it's worth the risk of doing such a migration. I have copied the comments over manually using the NUSMods Disqus account instead.

PR #4525, opened 2026-08-18
fix: update map to current ISB network

ThinkerPal:
## Context

Currently the ISB network overlay on the map displays an outdated version of the Internal Shuttle Bus. This PR updates it with the new services.

## Implementation

Removed E/BTC/L Services from network
Removed EA and SDE3 Bus Stops no longer served by ISB 
Added R1, R2 and P Services


jloh02:
@greptileai

jloh02 (approved):

PR #4541, opened 2026-08-25
chore: archive AY2025/2026 after AY2026/2027 starts

jloh02:
## Summary

- Add `2025/2026` to `website/src/config/app-config.json` `archiveYears` for PR3.
- Keep `specialTermAcademicYear` as `null` so the completed AY overlap configuration remains automatic.

## Validation

- `python3 -m json.tool website/src/config/app-config.json`
- PR3 config assertion via Node
- `git diff --check`

`oxfmt --check` was unavailable because the local dependencies are not installed.


ravern (approved):
LGTM

